PR #23, opened 2015-03-22
Added handling to remove unused definitions.

halindrome:
Simple approach but seems to work.  Note that to use this you need to change the include of the glossary to use the new method restrictReferences

That method does the same thing as fixIncludes, but also adds a callback to clean up the unused definitions.  Please take a look at this and merge if you think it is okay @michael-n-cooper @cookiecrook


halindrome:
Any comments on this?  I think it is pretty stable and I would like to merge it in.

michael-n-cooper:
No technical comments from me, I'm not really qualified to review the script. But the approach looks clean and simple to me. 

One comment is it seems this should be in a shared script, rather than just the one for the ARIA spec. Somewhere in common/script, either a dedicated file ("restrictReferences.js" perhaps) or a "ARIA respec additions file" like "aria-respec.js" or something. I lean towards the latter. Since the existing resoveReference.js already is a respec extension I think, the cleanest thing might be to add the new code to that file, then rename that file to be more generic, and update the references to it from the various specs.

PR #49, opened 2015-04-27
Required properties missing

halindrome:
Thanks @Joanmarie for running this down.  I think it is fine to merge.  I will need to update ariaChild.js too, but can do that separately.

### DIFF
--- a/aria/script/aria.js
+++ b/aria/script/aria.js
@@ -272,21 +272,22 @@ respecEvents.sub("end", function( msg ) {
                 }
                 // are there supported states / properties in this role?  
                 var attrs = [];
-                node = parentNode.querySelector(".role-properties");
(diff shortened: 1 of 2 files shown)


halindrome:
I don't want to perturb anything but I think this is an important fix.  @cookiecrook can you take a look and see if you are okay with it?

halindrome:
My checkin message was unclear there.  I incorporated her changes into ariaChild.js

halindrome:
As per PFWG meeting today, merging this.  I will then update roleInfo.js from the latest aria spec in the trunk and push that so it is available for DPUB.

cookiecrook:
@halindrome wrote:

> I don't want to perturb anything but I think this is an important fix. @cookiecrook can you take a look and see if you are okay with it?

Sorry, I've been to busy to review this. I have complete faith in the two of you. ;-)

PR #68, opened 2015-06-22
Master

chaals:
spelling, choose the right process document, comment the dummy test for
changelogs and point to stuff that is real, remove the reference to
funders who didn't fund this work.


LjWatson:
@Chaals Merged with thanks.

PR #71, opened 2015-07-07
Add div button example, fix document outline

MichielBijl:
The 'JavaScript' heading was a level two, while the 'html' heading is a level three (because it's beneath a level two with a label of 'Code'); hence JavaScript should be a level three, too. This fixes that.


MichielBijl:
It seems GitHub merged my changes into one request.

**Additional notes:**

* Add a clearer description for the button example.
* Add `<div>` element based button.
* Add comments to JavaScript for button example.

MichielBijl:
Linked the button and link example to a common css-file to prevent duplicates, and inconsistency.

MichielBijl:
Changed “do no repurpose…” to “Authors should not repurpose…”.

klown:
Hi @MichielBijl 
> Changed “do no repurpose…” to “Authors should not repurpose…”.

Long ago, PF decided that the APG was a non-normative document, and that all RFC2119 keywords be removed from the document (must, should, may, may not, etc.).  As editor, I replaced all (most?) of those normative phrases with simple statements (e.g., "do X", "do not do Y"), and sometimes, "it is recommended that ...".

Unless that decision has been reversed, and the APG is now normative, you need to avoid normative keywords.

MichielBijl:
@klown I wasn't aware of that. So would something like this be more appropriate?

“It is recommended that authors do not re-purpose other elements to create buttons.”

klown:
@MichielBijl ,
> So would something like this be more appropriate?
> 
> “It is recommended that authors do not re-purpose other elements to create buttons.”

Yes.  However, I think "Do not re-purpose... " is clearer.  If you stick with the "It is recommended ..." version, use "It is recommended that authors not re-purpose other elements to create buttons."

MichielBijl:
The original version had “Do not re-purpose…” and received comments from the group at yesterdays telecon; hence the change. So following your advice, I have changed it to “It is recommended that authors not re-purpose…”.

PR #89, opened 2015-09-14
revised link example with template

a11ydoer:
Template has 4 sections : 1)Examples, 2)keyboard support, 3)ARIA Roles,
Properties and States, and 4)Source code


MichielBijl:
This all looks very good, but why is there a `role="banner"` on the `<header>`, that is an implied role; no need for it right?

mcking65:
Putting role banner on header will still work in browser versions that do not support the HTML 5 mapping.

stevefaulkner:
recommend fixing HTML conformance errors
https://validator.w3.org/nu/?doc=http%3A%2F%2Frawgit.com%2Fa11ydoer%2Faria%2F2e887c7ad1c597f422d93059c6705fdc4fd19aca%2Fpractices%2Fexamples%2Flink%2Flink.html

MichielBijl:
It is my understanding that these examples should focus on conformance and not so much legacy support.

mcking65:
Best practice in conformance, yes. My understanding is that it is still considered best practice to not fully rely on the HTML 5 mapping of landmark regions by adding the landmark role to html5 elements. I guess we should learn more about compatibility and support to be clear on this. For example, I think browsers only started supporting the main element last year.

 

We have agreed that we do not want our examples to perform gymnastics to support very old browsers, e.g., IE8 and 9 … that was my understanding of legacy support.

MichielBijl:
I agree. It seems IE11 doesn't support either `footer`, `header`, or `main`: http://html5accessibility.com/. Support in the Edge browser is unknown at this point. 

I'm all for adding support for current browsers as long as it's not through proprietary code.

I'm working on a list of things to take into account when one works on the examples. Shall I add this to the list?

“It's okay to add support for current browsers as long as that support can be added with conforming code. You can check both html5accessibilit.com and caniuse.com for support tables.”

stevefaulkner:
@MichielBijl my comment here http://www.sitepoint.com/avoiding-redundancy-wai-aria-html-pages/#comment-2258207796 may be helpful

MichielBijl:
What I take away from that is that we should look at AT support rather than browser support, is that correct?

—Michiel

stevefaulkner:
@MichielBijl if it's interoperably implemented in browsers that is enough me thinks

MichielBijl:
Cool. Do we have any sources for this?

stevefaulkner:
@MichielBijl sources for?

mcking65:
>if it's interoperably implemented in browsers that is enough me thinks

 

Previous example was IE11 does not interoperably implement header, footer, and main. So, should we consider it a best practice to add redundant ARIA roles? Or, should we look at how many AT implement support based on their own DOM parsing before deciding to recommend adding extra code to compensate for a browser deficiency?

MichielBijl:
@stevefaulkner to decide what to include / which browsers to support. I don't know wether to include `role`'s or not; I don't have any data to base my decision on.

PR #93, opened 2015-09-19
Aria graphics module: editorial changes & new/improved examples

AmeliaBR:
I've also re-introduced the text for the figure role, as an editorial note, so that we keep a reference copy while deciding on the final text for the main ARIA 1.1 spec.

Suggestions welcome for how to improve formatting of the complex examples (code + embed + introductory paragraph text).


fredesch:
merged

PR #94, opened 2015-10-06
Fixed wrong escape char used for aria-kbdshortcuts

mlewand:
Seems like there's a wrong escape character used for `&quot;` entity.


Joanmarie:
Hey, thanks for the patch! You're correct, but you need to escape the ampersand char. I went ahead and made that change.

mlewand:
Wasn't sure about this one, as you're a little inconsistent in terms of escaping within the documents, so took a wrong shot. :wink: Anyway, it's great to see that it's fixed, cool!

Issue #106, opened 2015-11-13
RFE: Add means to programmatically control link creation for rref, pref, sref, and glossary references


Joanmarie:
Our README.md states:

> * The first reference in a section to a role or state should be linked to the role or state with the appropriate class. Subsequent references needn't be referenced, but may be in certain circumstances.
> * repeated links to aria features should just be in `<code>`, not in cross-reference links (just link first time)

I have no objections to this style. However, it has some maintenance implications: If we insert a new reference before any other references in that section, we then need to modify the spec text which follows the insertion -- text whose content wouldn't need to be modified were it not for the tag change. This in and of itself is no big deal. But it introduces noise into the patch-review process (especially given our convention of making paragraphs a single line), as well as into any subsequent forensics we wish to do when researching when/why an earlier change was made.

Then there are questions regarding the "needn't be... but may in certain circumstances". What are those circumstances? Without them being clearly defined, we're likely not going to be consistent. Moreover, what if we change our minds later on regarding when a reference should be a link versus stylized text? Then we're once again (functionally) changing spec _content_ when all we really want to change is spec _style_. Again, not a big deal, but it's also not ideal.

I think the above issues could be addressed by always using `rref`, `pref`, `sref`, and `a` (for glossary) and using some JavaScript magic to turn each into either a link or stylized text.

Joanmarie:
> Note that a is within the scope of ReSpec.

And hence beyond the scope of our spec and this github issue? If so, please close this as FIXED. And thank you (thank you, thank you) for doing this!

PR #108, opened 2015-11-22
Changed reference conversion logic

halindrome:
Now only the first [spr]ref in a section will be a link to the
base definition.  The others will just be spans with the same class
so that they style the same (except the link version will have an
underline of course).

This should support the feature requested in #106

### DIFF
--- a/common/script/resolveReferences.js
+++ b/common/script/resolveReferences.js
@@ -79,7 +79,10 @@ function updateReferences(base) {
         var content = $item.text();
         var ref = $item.attr("title");
         if (!ref) {
-            ref = content;
+            ref = $item.attr("data-lt");
+            if (!ref) {
+                ref = content;
+            }
         }
 
         // what sort of reference are we?
--- a/common/script/resolveReferences.js
+++ b/common/script/resolveReferences.js
@@ -77,12 +77,17 @@ function updateReferences(base) {
 
         // what are we referencing?
         var content = $item.text();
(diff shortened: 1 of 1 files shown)


halindrome:
@Joanmarie any thoughts on this change?

Joanmarie:
Sorry (very, very sorry!) for the delay in responding.

Some observations using the 'grid' role to compare the [before](https://rawgit.com/w3c/aria/master/aria/aria.html#grid) and [after](https://rawgit.com/w3c/aria/shane-p106/aria/aria.html#grid):

The 'row' role is referenced twice in the paragraph which begins with

> Authors must ensure that elements with role treegrid are owned by elements with role row

In master, both are links; in the shane-p106 branch neither are. I think that the first instance should remain a link because it's the first time that role is referenced within the 'grid' text. I think that the second instance should (probably) be styled in the same fashion as we style code elements.

The 'gridcell', 'rowheader', and 'columnheader' roles are similarly not linked after the change. But 'treegrid' seems to work partially correctly: The first instance is a link and the second is not. The second instance, again, should (probably) be styled in the same fashion as we style code elements.

The 'aria-selected' attribute is only referenced once in the 'grid' text, but that reference loses its link. But 'aria-multiselectable', which is also only referenced once in the 'grid' text, keeps its link.

Joanmarie:
This seems to be doing everything I want and expect. Merging. Thanks!!!!

PR #110, opened 2015-11-30
Added a check to see if a ref is in termlist

### LINKED ISSUE #109: Removal of glossary entries failing when the only reference is within the glossary itself.
@halindrome:

The bug is inside restrictReferences() within the handler for the "core/link-to-dfn".  See:
https://github.com/w3c/aria/blob/master/common/script/resolveReferences.js#L119

The rationale is that $("a.internalDFN") will produce a list of all the reference links to glossary terms within the main content of the document.  The problem is that some of those links occur only within the glossary itself.

An example is the glossary entry "Ontology".  There is a reference to it ("ontology") from the glossary entry for "Taxonomy".  Because of that, the handler leaves the entry for "Ontology" in the document.  However, If "Taxonomy" is not referenced, its entry is removed.  The result is a glossary entry for "Ontology" where it is not referenced from anywhere within the document.

You can see an example by loading the rawgit url of accname-aam.html.  Both "Ontology" and "Keyboard Accessible" are listed in the important terms section, even though neither is linked to from anywhere within the document:
http://rawgit.com/w3c/aria/master/accname-aam/accname-aam.html#glossary

(note:  "Keyboard Accessible" is referenced from the glossary entry for "Operable").

I've been puzzling over this for a while, but haven't thought of an elegant way to get around it.

halindrome:
If terms reference one another, they were being left in the
glossary section EVEN IF they were not referenced elsewhere.

This should fix #109 @klown

### DIFF
--- a/common/script/resolveReferences.js
+++ b/common/script/resolveReferences.js
@@ -54,7 +54,7 @@ function linkCrossReferences() {
     setHrefs ('a.html-mapping', htmlMappingURL);
   }
   else {
-    console.log ("linkCrossReferences():  Note -- htmleMappingURL is not defined.");
+    console.log ("linkCrossReferences():  Note -- htmlMappingURL is not defined.");
   }
 
 }
@@ -112,13 +112,25 @@ function restrictReferences(utils, content) {
     });
 
     // add a handler to come in after all the definitions are resolved
+    //
+    // New logic: If the reference is within a 'dl' element of
+    // class 'termlist', and if the target of that reference is
+    // also within a 'dl' element of class 'termlist', then
+    // consider it an internal reference and ignore it.
 
     respecEvents.sub('end', function(message) {
         if (message == 'core/link-to-dfn') {
             // all definitions are linked
             $("a.internalDFN").each(function () {
                 var $item = $(this) ;
-                var r = $item.attr('href').replace(/^#/,"") ;
+                var t = $item.attr('href');
+                if ( $item.closest('dl.termlist').length ) {
+                    if ( $(t).closest('dl.termlist').length ) {
+                        // do nothing
+                        return;
+                    }
+                }
+                var r = t.replace(/^#/,"") ;
                 if (termNames[r]) {
                     delete termNames[r] ;
                 }


michael-n-cooper:
I think this might be creating a problem. I pulled this commit into my publication branch, and immediately got broken link fragment errors. If a term A links to another term B, but this processing excludes term B from being included in the doc because its only reference in the doc is from term A, then the reference link is broken, and I can't publish. I haven't followed this issue closely enough to know what problem it is meant to solve, but it's creating a new one. I'm not sure if that's an issue of implementation or conflicting requirements.

halindrome:
Hmm.  Or rather, Yikes.  This was meant to fix the situation where term B is only referenced by the definition of term A.  In that case, if term B and term A are in the glossary, them remove term B from the document.  I will dig into this right away.

halindrome:
Uggh!  Okay - consider this situation: Somewhere in the document, outside of the glossary, there is a reference to term A. Nowhere in the document outside of the glossary are there references to term B nor term C.  Term A references term B.  Term B references term C.  When trying to decide if term C should be in the document or not, we look at who references it.  We see ONLY references from within the glossary.  The terms in the glossary that reference term C are themselves ONLY referenced from within the glossary.  Ad infinitum...

There is also the possibility that something more fundamental would break.  Consider a loop.  C references B, which references A, which references C.  A link into that circle from outside would lead to madness.

Anyone got a brilliant idea beyond reverting this change?

klown:
@michael-n-cooper 
I'm just catching up on this.  You wrote:
> I haven't followed this issue closely enough to know what problem it is meant to solve, ...

The issue is here, for your reading pleasure:
https://github.com/w3c/aria/issues/109

klown:
@halindrome wrote:
> Anyone got a brilliant idea beyond reverting this change?

I doubt it's brilliant, and I'm not sure it would work, but here's an outline:
1. Before including the glossary, make a list of all the terms used in the document.
1. Remove duplicates from this term list, i.e. each term is unique within the list.
1. For each item in the term list, cherry pick the corresponding glossary item.
1. For each glossary item picked, check if it references others terms in the glossary, and append them term to the term list, but only if that term is not already in the list.
1. When the term list has been exhausted, you're done.

halindrome:
Might work... except that ReSpec processing order means included content comes in pretty early.  A similar approach could work at the end of processing:

1. Upon load develop a complete list of terms defined in the glossary.
2. Find each internal reference that is a) not from within the glossary and b) references something within the glossary. Now you have a list of glossary terms referenced.
3. For each term referenced, check its definition (dd element) and see if any other terms are referenced.  If so, and if those terms are not in the list, add them and restart this loop.
4. Once complete you have a full list of required referenced terms.  Remove these from the list of terms defined in the glossary.
5. Remove any terms from the glossary that remain in the list.

I think that would work.  I hate restarting the loop.  Expensive.

PR #111, opened 2015-12-02
Revert "Added a check to see if a ref is in termlist"

halindrome:
Reverts w3c/aria#110 because situations where items reference one another were not adequately considered.


halindrome:
We do not need to revert this.

PR #112, opened 2015-12-02
Changed logic to preserve referenced items.

halindrome:
New logic that is smart about terms that reference other terms.  This resolves the broken links that were introduced via the merge of pull request #110

### DIFF
--- a/common/script/resolveReferences.js
+++ b/common/script/resolveReferences.js
@@ -94,24 +94,35 @@ function updateReferences(base) {
 
 // We should be able to remove terms that are not actually
 // referenced from the common definitions
+//
+// the termlist is in a block of class "termlist", so make sure that
+// has an ID and put that ID into the termLists array so we can 
+// interrogate all of the included termlists later.
 var termNames = [] ;
+var termLists = [] ;
+var termsReferencedByTerms = [] ;
 
 function restrictReferences(utils, content) {
     var base = document.createElement("div");
-    var termContainer = $(base).makeID("terms") ;
     base.innerHTML = content;
-    updateReferences(base);
 
-    // strategy: Traverse the content finding all of the terms defined
+    // New new logic:
+    //
+    // 1. build a list of all term-internal references
+    // 2. When ready to process, for each reference INTO the terms, 
+    // remove any terms they reference from the termNames array too.
     $.each(base.querySelectorAll("dfn"), function(i, item) {
         var $t = $(item) ;
-        var titles = $t.getDfnTitles() ;
-        var n = $t.makeID("dfn", titles[0]) ;
+        var titles = $t.getDfnTitles();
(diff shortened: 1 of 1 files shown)


Joanmarie:
Since I hadn't been following the #109 + #110 discussion, I started with a diff of glossary terms that are present as a result of this fix:

    --- before	2015-12-16 20:34:51.909841211 +0100
    +++ after	2015-12-16 20:39:13.606577300 +0100
    @@ -2,8 +2,11 @@
     
     accessibility api
      accessibility apis
     +accessibility tree
     +accessible description
      accessible name
      accessible names
     +accessible object
      accessible objects
      assistive technologies
      assistive technology
     @@ -28,6 +31,7 @@
      managed state
      managed states
      nemeth braille
     +node
      normative
      object
      objects
     @@ -51,6 +55,7 @@
      state
      states
      taxonomy
     +text node
      understandable
      user agent
      user agents

Seems harmless to me, and hopefully fixes the issue @michael-n-cooper reported.

I'm guessing node or text node are referenced in other specs we manage? (If not, I'm not convinced it's a big deal -- read "worth the apparent hassle")

Joanmarie:
Ok, so one verifiable example of something that is broken in master but fixed in the branch associated with this issue is the inclusion of Accessible Description in the glossary. Accessible Description is referenced by Accessible Name's glossary entry, and Accessible Name is referenced in the spec. Thus both Accessible Name and Accessible Description belong in the spec.

@halindrome Since @michael-n-cooper is the one who experienced the breakage, I think it would make sense for him to verify if all of his issues have been addressed and nothing else regressed as a result of your proposed change.

halindrome:
Works for me!

On Wed, Dec 16, 2015 at 3:56 PM, Joanmarie <notifications@github.com> wrote:

> Ok, so one verifiable example of something that is broken in master but
> fixed in the branch associated with this issue is the inclusion of
> Accessible Description in the glossary. Accessible Description is
> referenced by Accessible Name's glossary entry, and Accessible Name is
> referenced in the spec. Thus both Accessible Name and Accessible
> Description belong in the spec.
>
> @halindrome <https://github.com/halindrome> Since @michael-n-cooper
> <https://github.com/michael-n-cooper> is the one who experienced the
> breakage, I think it would make sense for him to verify if all of his
> issues have been addressed and nothing else regressed as a result of your
> proposed change.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/w3c/aria/pull/112#issuecomment-165256534>.
>



-- 
Shane McCarron
halindrome@gmail.com

halindrome:
Uggh - I apparently broke the merge by checking in the fixed stuff.  I will attempt to resolve the conflicts ON THE BRANCH.  Fingers crossed.

halindrome:
I am going to throw this away and start over.  Merge conflicts are not cool.

Issue #117, opened 2015-12-12
WAI-ARIA 1.1: Links to CSS2


momdo:
In <a href="http://www.w3.org/TR/2015/WD-wai-aria-1.1-20151119/#aria-hidden">aria-hidden (state)</a> of section 6.4,

>CSS 2 provides a way to select on attribute values ([CSS2]).
This link (http://www.w3.org/TR/1998/REC-CSS2-19980512/selector.html#q10) is too old.

Now the select on attribute values are defined in Selectors Level 3 spec.
http://www.w3.org/TR/css3-selectors/#attribute-selectors

Additionally in section 2.2:
>Most modern user agents support CSS attribute selectors ([CSS2])

Then in section F.2:
>[CSS2]
> Bert Bos; Tantek Çelik; Ian Hickson; Håkon Wium Lie et al. Cascading Style Sheets Level 2 Revision 1 (CSS 2.1) Specification. 7 June 2011. W3C Recommendation. URL: http://www.w3.org/TR/CSS2

Similarly should modify links to Selectors Level 3.

MichielBijl:
Work on this in: https://github.com/w3c/aria/tree/feature/issue-117

PR #120, opened 2015-12-12
Update attribute selector reference

MichielBijl:
Fixes issues in https://github.com/w3c/aria/issues/117


Joanmarie:
I didn't see any reason (or benefit) to have three commits when one is called for. Therefore I manually merged your changes after squashing them. (commit 431fe708a3)

MichielBijl:
No reason, just the way it worked out. I switch a lot between issues/things while I work.

Joanmarie:
Understood. You familiar with the `--amend` option of `git commit`?

MichielBijl:
I thought that was just to edit commit text?

Joanmarie:
Nope. If you make a change and then commit it, go do something else, then come back to the original task and make another change, you can do `git commit -a --amend`. That will add the new changes (assuming they were made to existing files) to the previous commit. It will also give you the opportunity to change the commit message, should you be so moved.

If the duration of the "something else" is significant, you can throw in the `--date` option, e.g. ``git commit -a --amend --date="`date`"`` does what I said above and also updates the timestamp of the commit to the present.

MichielBijl:
Cool! Thank you for the explanation :+1: Will do that next time.

Joanmarie:
Give it a try with issue 121 and let me know if you have any questions/problems. Thanks!

Issue #118, opened 2015-12-12
WAI-ARIA 1.1: Link to DOM2 Core


momdo:
In <a href="http://www.w3.org/TR/2015/WD-wai-aria-1.1-20151119/#intro_ria_accessibility">section 1.1</a>,
> Document Object Model [DOM-Level-2-Core],

Why does the draft refer DOM2? We know the <a href="http://www.w3.org/TR/dom/">W3C DOM4</a> spec is the latest W3C Recommendation.

MichielBijl:
Fixed in https://github.com/w3c/aria/pull/262

PR #121, opened 2015-12-12
Update DOM reference

MichielBijl:
Fixes issue in https://github.com/w3c/aria/issues/118


Joanmarie:
Why did you update the reference, but leave the associated links pointing to DOM2?

MichielBijl:
My bad. Will fix.

Issue #216, opened 2016-01-26
Remove or edit “HTML 5 - A look ahead”


MichielBijl:
> HTML 5 and the "serialized" XHTML version of HTML 5 are years away from standardization.

Source: [A2.5 HTML 5 - A look ahead](https://www.w3.org/TR/wai-aria-practices-1.1/#alookahead)

That XHTML version might be, but I'm pretty sure [HTML 5](https://www.w3.org/TR/html5/) is a recommendation ;) Should this be removed, or are there any bits that we need to save?

PR #250, opened 2016-02-09
Remove section A2.5 HTML 5 A look ahead

MichielBijl:
Fixes https://github.com/w3c/aria/issues/216


MichielBijl:
Two +1's and no further comments on the list: https://lists.w3.org/Archives/Public/public-aria/2016Feb/0183.html

PR #256, opened 2016-02-17
ReSpec event handler to fix scrolling when loading.

klown:
@Joanmarie @mcking65 @richschwer @michael-n-cooper 
This is an attempt at fixing the failure to position an ARIA spec document when loading from an URL with a fragment ID when ReSpec is in play.  We run into this problem when using rawgit URLs.

I've spot checked a few of the documents with a couple of browsers, but my tests are not exhaustive.  Could you check with the documents you edit to see if this works for all of them?  Here are some sample URLs.

- APG:  http://rawgit.com/klown/aria/scrollToFragId/practices/aria-practices.html#relations_labeling
- ARIA: http://rawgit.com/klown/aria/scrollToFragId/aria/aria.html#aria-multiline
- DPUB: http://rawgit.com/klown/aria/scrollToFragId/aria/dpub.html#references
- SVG-AAM: http://rawgit.com/klown/aria/scrollToFragId/svg-aam/svg-aam.html#glossary

@halindrome I put the code in "resolveReferences.js" simply because it is a file that is loaded by all of our ARIA documents.  So while it has nothing to do with resolving references, it was convenient to add it there.

### DIFF
--- a/common/script/resolveReferences.js
+++ b/common/script/resolveReferences.js
@@ -177,12 +177,10 @@ function fixIncludes(utils, content) {
 }
 
 // Fix the scroll-to-fragID problem:
-(function() {
-    respecEvents.sub("start", function (details) {
-        if (details === "core/location-hash") {
-            if(window.location.hash) {
-                window.location = window.location.hash;
-            }
+(function () {
+    respecEvents.sub("end-all", function () {
+        if(window.location.hash) {
+            window.location = window.location.hash;
         }
     });
 })();
--- a/common/script/resolveReferences.js
+++ b/common/script/resolveReferences.js
@@ -176,3 +176,13 @@ function fixIncludes(utils, content) {
     return (base.innerHTML);
 }
 
+// Fix the scroll-to-fragID problem:
+(function() {
+    respecEvents.sub("start", function (details) {
+        if (details === "core/location-hash") {
+            if(window.location.hash) {
+                window.location = window.location.hash;
+            }
+        }
+    });
+})();


halindrome:
FWIW respec already does this (in js/core/location-hash.js).  The message you are waiting for is the one that module emits.  However, that module will NOT change the location if you have already scrolled.  So if, for example, you were scrolled into the middle of a document and then reloaded, it will not move you.  Is this the problem you are trying to solve?  I ask because, for example, clicking on the link below:

http://rawgit.com/klown/aria/master/svg-aam/svg-aam.html#glossary

works perfectly the first time in right now - without your change

klown:
@halindrome I think svg-aam works because we (you and I) fixed this in the core-aam a while ago, and likely the svg-aam inherited the fix.  I'll take a closer look later.

However, compare these two.  The first should work, and the second shouldn't.
http://rawgit.com/klown/aria/scrollToFragId/aria/aria.html#aria-multiline
http://rawgit.com/klown/aria/master/aria/aria.html#aria-multiline
For that matter, compare the last one to (also doesn't work): http://rawgit.com/w3c/aria/master/aria/aria.html#aria-multiline

MichielBijl:
This seems to work, can we just merge it and see what happens?

klown:
@halindrome 

You wrote:
> if, for example, you were scrolled into the middle of a document and then reloaded, it will not move you. Is this the problem you are trying to solve?

No.  The problem is when first loading from rawgit, the contents do not scroll to the location given by the fragment ID, and you have to manually move around a bit to find the desired location.  This patch solves that.

I wrote:
> I think svg-aam works because we (you and I) fixed this in the core-aam a while ago, and likely the svg-aam inherited the fix. I'll take a closer look later.

I've confirmed that the svg-aam gets the fix we introduced to mapping-tables.js in 2014:
https://github.com/w3c/aria/commit/9f609e9907b170d04436e6c6ce61599e86f3a537

That's why the svg-aam works without my changes to resolveReferences.js.

halindrome:
Sorry that I don't have time to dig into this further - are you saying your
change is a generic fix for the specialized version we/you put into
mapping-tables.js?

On Fri, Feb 19, 2016 at 8:44 AM, Joseph Scheuhammer <
notifications@github.com> wrote:

> @halindrome <https://github.com/halindrome>
>
> You wrote:
>
> if, for example, you were scrolled into the middle of a document and then
> reloaded, it will not move you. Is this the problem you are trying to solve?
>
> No. The problem is when first loading from rawgit, the contents do not
> scroll to the location given by the fragment ID, and you have to manually
> move around a bit to find the desired location. This patch solves that.
>
> I wrote:
>
> I think svg-aam works because we (you and I) fixed this in the core-aam a
> while ago, and likely the svg-aam inherited the fix. I'll take a closer
> look later.
>
> I've confirmed that the svg-aam gets the fix we introduced to
> mapping-tables.js in 2014:
> 9f609e9
> <https://github.com/w3c/aria/commit/9f609e9907b170d04436e6c6ce61599e86f3a537>
>
> That's why the svg-aam works without my changes to resolveReferences.js.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/w3c/aria/pull/256#issuecomment-186241472>.
>



-- 
Shane McCarron
halindrome@gmail.com

klown:
@halindrome 

> are you saying your change is a generic fix for the specialized version we/you put into mapping-tables.js?

Not really.  It is a simpler fix since there is a lot more going on in the mapping-tables script where the fragID is handled.  With this pull request, the -aam documents run both fixes, which is, I suppose, mildly inefficient.  I could encapsulate the common code between the two into one function and make it "generic".

halindrome:
I dont think the efficiency is really an issue - once a spec is done it is
made static.  This is a developer / development problem.  Okay, I
understand what you are doing.  FWIW I would subscribe to the end-all
message instead of the one you are using - just in case someone after you
does something stupid and adjusts the DOM / Flows again.

On Fri, Feb 19, 2016 at 9:18 AM, Joseph Scheuhammer <
notifications@github.com> wrote:

> @halindrome <https://github.com/halindrome>
>
> are you saying your change is a generic fix for the specialized version
> we/you put into mapping-tables.js?
>
> Not really. It is a simpler fix since there is a lot more going on in the
> mapping-tables script where the fragID is handled. With this pull request,
> the -aam documents run both fixes, which is, I suppose, mildly inefficient.
> I could encapsulate the common code between the two into one function and
> make it "generic".
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/w3c/aria/pull/256#issuecomment-186256549>.
>



-- 
Shane McCarron
halindrome@gmail.com

MichielBijl:
This is an issue we run into frequently during calls; if you link to a branch version on Rawgit, you never end up where you’re supposed to be. Leads to unhappy and frustrated spec writers ;)

—Michiel

klown:
@MichielBijl 

> ... Leads to unhappy and frustrated spec writers ;)

Except *-aam writers, since 2014.  :-D

klown:
@halindrome Okay, I'll try end-all and see if there are any issues.  Thanks.

PR #297, opened 2016-03-14
fix link to menu button example

a11ydoer:
please add a link to new menu button example


MichielBijl:
Can you add the space back in line 192 (`omment archive</a>) or<a href="https:/`)?

And can you remove the title attribute from the link (`title="Menubutton" target="_blank">Menu button</a>`)?

a11ydoer:
I think I just copied other example link, which has title and target attribute(ie. link). Is not the way Spec done?

MichielBijl:
Yeah those need to go too, but we'll fix that after the heartbeat publication; I'll make an issue for it.

jnurthen:
Seeing as there is still a change to make here can you please close it and resubmit against Master with just the change to the menu link.
Thanks.

PR #300, opened 2016-03-15
Fixed some formatting errors

halindrome:
The new stylesheets have caused some changes to base styling, so
the stylesheet needed a tweak.

Also, the permalink styles in the local sheet were colliding with the
styles provided by respec.


halindrome:
Apologies.  This PR includes non-important changes to the aria-practices.html file.  I have no idea how those even happened.  I suspect my editor did something.  They are whitespace changes.  No need to merge that if you don't want to.  It is the stylesheet that is of interest.

MichielBijl:
Trailing spaces being removed is due to the .editorconfig file :)

So, looks good to me!

Issue #328, opened 2016-04-14
WAI-ARIA 1.1: Update Authoring Practices links to 1.1 ?


carmacleod:
March 17 WAI-ARIA 1.1 Working Draft
(https://www.w3.org/TR/wai-aria-1.1/ aka http://www.w3.org/TR/2016/WD-wai-aria-1.1-20160317/)

Should all links to "WAI-ARIA Authoring Practices" be updated now to point to 1.1 ?(https://www.w3.org/TR/wai-aria-practices-1.1/)
(Including links to sections in the Authoring Practices, such as "Using Tabindex to Manage Focus Among Widgets" and "Providing Keyboard Focus in WAI-ARIA Authoring Practices" and "Drag-and-Drop Support in the WAI-ARIA Authoring Practices").

Also note that some (but not all) references to "WAI-ARIA Authoring Practices" in WAI-ARIA 1.1 have the word "Guide" in them (i.e. "WAI-ARIA Authoring Practices Guide"). Should the word "Guide" be deleted, for consistency?

mcking65:
Looks like all links are good. There are several where fragments could direct the reader to specific sections, including links related to:
* combobox
* feed
* grid
* tabs
* tri-state checkbox

PR #332, opened 2016-04-15
Changed references to authoring practices to 1.1 version

jnurthen:
#328


MichielBijl:
Could you point them all to the https url? So: https://www.w3.org/TR/wai-aria-practices-1.1/ instead of http://www.w3.org/TR/wai-aria-practices-1.1/

jnurthen:
@MichielBijl I could but I think that would be a strange thing to do. The w3c server is already set up to serve requests through SSL if available. I fear that if I set the request to https and it was not available then it would fail.
The other linked documents are linking to non-ssl versions so this is consistent.

MichielBijl:
Fair enough.

PR #342, opened 2016-04-23
minor typos and missing words. feed was missing its [ARIA 1.1] marker.

carmacleod:
Just a few small things, including adding "[ARIA 1.1] " to the role-description for the "feed" role, which also adds "[ARIA 1.1] " to "feed" in the generated section "5.4 Definition of Roles".


Joanmarie:
Thanks!!

PR #348, opened 2016-04-25
Move scrollbar role section to correct position in alphabetical order…

carmacleod:
…. (This also fixes the order in the auto-generated section 5.4 Definition of Roles)

This is just 1 small change: cut scrollbar section out from between separator and slider, and paste it verbatim between rowheader and search.  :)


Joanmarie:
Thanks!

PR #354, opened 2016-04-28
Removing "golden hammer" language

bpmcneilly:
The 'golden hammer', removing elements with role="none" or "presentation" entirely from the accessibility tree, was rejected by the task force so I removed this text.

For stylistic reasons I removed "in addition" and added the sentence beginning "Certain roles" to the previous paragraph.

### DIFF
--- a/aria/graphics.html
+++ b/aria/graphics.html
@@ -229,19 +229,7 @@
 	 <p>
      Authors are given the ability to influence what is presented 
      to assistive technologies and to influence navigation through
-     the use of roles and properties. With graphics, there are many 
-     cases where presenting and navigating every element will make the 
-     graphic harder to understand and use. 
-     Authors may mark elements for exclusion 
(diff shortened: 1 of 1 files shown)


AmeliaBR:
Thanks for noticing this section, Brian.  It is problematic given the current interpretation.

That said, I'm not sure whether we really need to remove all that text, or whether we should keep it with new qualifications.  Describing how role=presentation normally works might still be useful, although perhaps it is less relevant since this is also the default behavior.  I'll take another look at it later, maybe integrate the text elsewhere.

bpmcneilly:
Good point. That sentence about presenting every element is still true (regardless of the golden hammer), but I couldn't gracefully find a way to integrate it in this section.

Perhaps this should find its way into the authoring practices document? Just a thought.

AmeliaBR:
Closing in favour of more comprehensive changes proposed in #365.

PR #361, opened 2016-05-04
Updated handling of rdef, pdef, and sdef elements

halindrome:
These elements are contained in divs in the source.
During load, these elements are converted into h4 elements and their
surrounding div element is changed into a section with the same
attributes + a class value of notoc so that they do not get section
numbers nor do they get included in the table of contents.

Note that this change has not yet been pushed into aria-child.js.  Presumably that needs similar changes, but I wanted someone to look at this first.

@mcking65 this is the change you requested - I would appreciate it if you could check it.


mcking65:
Looks perfect!! Thank you!!

PR #366, opened 2016-05-09
Remove redundant sentence from paragraph

carmacleod:
Authors probably do need to be reminded twice to use native host language semantics for form controls whenever possible... but not in the same paragraph.  :)


joanmarie:
Thanks for the fix!

Issue #137, opened 2016-01-12
SVG-AAM: Issues with the use of title attribute in name and description calculation


Joanmarie:
Section [10.1 Name and Description](https://rawgit.com/w3c/aria/master/svg-aam/svg-aam.html#mapping_additional_nd) has a couple of references to the "`xml:title` attribute". From some quick searching, I have yet to find any reference to that attribute; I have, however, found references and documentation for `xlink:title`. Typo?

Joanmarie:
(I just found another potential issue with the same section and same(?) attribute, so I'll mention it here.)

Item 2 states:

> Replace [step 2D](http://w3c.github.io/aria/accname-aam/accname-aam.html#step2D) with the following:

> [...]
> If performing a text alternative computation for an accessible description:
> * If the current node has at least one direct child `desc` element, select the appropriate description based on the language rules for the SVG specification, and return the concatenated text content of the description.
> * If the current node is a `use` element, and there was no child `desc` element, but it has a valid `href` or `xlink:href` attribute pointing to another element in the same document, set the current node to the referenced element and compute the text alternative from step 2 as if that element had been referenced with `aria-describedby`.

With this in mind, consider the following example (taken from the in-progress testable statements):

      // aria-label, xlink title
      if given `<a id="test" aria-label="booth" href="../index.html" xlink:title="table" ></a>` then accessible description = "table"

Given the replaced step 2D text above:

* The current node (the `a` element) does not have at least one direct child `desc` element.
* The current node (the `a` element)  is not a `use` element.

So as I understand it, we can't get the description for the current node using step 2D.

One possibility (I _believe_) might be to keep/apply/whatever the original 2D text as a fallback. That text states:

> Otherwise, if the current node's native markup provides an attribute or element that defines a text alternative, return that alternative as a flat string, unless the element is marked as presentational (role="presentation" or role="none").

So if the `xlink:title` attribute "defines a text alternative" (arguably it does), it could be used.

If that is not an appropriate solution, then perhaps 2H applies -- though it would need to be tweaked slightly. The current text is:

> Otherwise, if performing a text alternative computation for an accessible description, and the current node has at least one direct child `title` element, but the title elements were not used when generating the accessible name for the node, then select the appropriate title based on the language rules for the SVG specification, and return the title text content as the description.

In the example above, there is not the required "one direct child `title` element," but there is an attribute on the current node providing the title. And the title text from the attribute isn't being used for the name....

fredesch:
[...]

Issue #136, opened 2016-01-08
SVG-AAM / Core AAM: Consider explicit language regarding objects with presentational role and ARIA attributes


joanmarie:
In the SVG-AAM's [5.1.1 Excluding Elements from the Accessibility Tree](https://rawgit.com/w3c/aria/master/svg-aam/svg-aam.html#exclude_elements) there is the following note text:

> The other factors for excluding elements, as described in the Core Accessibility API Mappings, are as follows:
> - If the first mappable role provided by the author is none or presentation, the element must not be exposed.
> 
> [...]
> 
> Consult the original document [CORE-AAM] for the normative text.

One should, of course, consult the CORE-AAM as the good note states. But a reader of the above text couldn't be blamed all that severely if they read the above note text and concluded that given

```
<rect role="none" aria-label="Am I Here?" x="1" y="1" width="10"' height="10"/>
```

there should be no accessible object.

If one consults the CORE-AAM for the normative text, [5.1.1 Excluding Elements from the Accessibility Tree](https://rawgit.com/w3c/aria/master/core-aam/core-aam.html#exclude_elements2), they would find something quite similar:

> The following elements are not exposed via the accessibility API and user agents MUST NOT include them in the accessibility tree:
> - Elements with none or presentation as the first mappable role in the role attribute string, according to the rules for the none and the presentation role defined in Accessible Rich Internet Applications (WAI-ARIA) 1.0 [WAI-ARIA].

So unless the ARIA spec defines a rule for `none`/`presentation` that says otherwise, it would seem that the `rect` element in the example above does indeed need to be excluded.

There's a lot of text in the ARIA spec regarding the [presentation role](https://rawgit.com/w3c/aria/master/aria/aria.html#presentation), but for the question at hand this statement seems to be the most relevant:

> For any element with an explicit or inherited role of presentation, user agents MUST ignore any non-global, role-specific WAI-ARIA states and properties. **However, the user agent MUST always expose global WAI-ARIA states and properties to accessibility APIs, even if an element has an explicit or inherited role of presentation.** 

(emphasis added)

The `aria-label` attribute is global. So then the `rect` element in the example above would need to be exposed after all (right?).

I mention all of this for a couple of reasons:
[...]

Issue #134, opened 2016-01-06
SVG-AAM: Consider explicit language regarding inclusion of elements with child title or desc


joanmarie:
According to section 5.1.2:

> SVG user agents MUST provide an accessible object in the accessibility tree for rendered SVG elements that meet any of the following criteria:
> - It has at least one direct child ‘title’ element or ‘desc' element.
>   ...

Subsequent items explicitly state "non-empty". As a result, one might assume the above text means user agents MUST expose elements with a direct child title/desc -- even if the contents of that title/desc element is empty. Then again, if the provided title/desc is empty and there are no other criteria met justifying inclusion, why should there be an accessible object for the parent? (Not entirely rhetorical.)

Regardless of whether the statement quoted above applies always or only in the case of non-empty title/desc element children, I think it would be helpful if there were explicit language clarifying what implementors should do.

richschwer:
good point. We will discuss it at meetings. I see you assigned it to Fred for now.

Joanmarie:
@richschwer Actually, I didn't assign it. I had assigned the other issue (#133) to him. Then he told me that was fine, but he wouldn't be the one to fix it as he wasn't working on that particular area. As a result, I've stopped assigning for now. ;) (The other one I assigned to you as a correction.)

richschwer:
no problem. As long as we have an issue logged in tracker. This is a good one.

Joanmarie:
A somewhat-related question to this issue (and/or issue 136): Given

    <svg>
      <circle aria-labelledby="" cx="50" cy="100" r="15"/>
    </svg>

Were there not the `aria-labelledby`, the `circle` element would not be exposed according to the last non-note section in [5.1.1 Excluding Elements from the Accessibility Tree](https://rawgit.com/w3c/aria/master/svg-aam/svg-aam.html#exclude_elements).

Were there a _non-empty_ `aria-labelledby`, then that would of course be cause for exposure of the `circle` element.

But what should happen in the case of an _empty_ `aria-labelledby` attribute and no other semantic information justifying exposure as shown in the example above?

Regardless of what the answer is, perhaps some explicit language to address this would be of use.

Lastly, if the group consensus is to not include the `circle` element, the work-in-progress test assertions should probably be updated to remove the following assertion:

> // empty aria-labelledby only
> if given `<circle id="test" aria-labelledby="" cx='50' cy='100' r='15'/>` then accessible name = BLANK

because one cannot test the name of a non-exposed element.

[...]

PR #373, opened 2016-05-12
SVG-AAM: Use graphics roles, fix unclear requirements

AmeliaBR:
[Compiled version in RawGit](https://rawgit.com/w3c/aria/svg-aam/svg-aam/svg-aam.html)

- Lots of clean-up of unclear text, with a few normative changes based on discussions over the past few months.

- Changes the role mapping table to use the new ARIA Graphics roles as defaults.

Addresses issues #134, #136, and #137, along with various issues and actions on Tracker.

Comments welcome on these changes, or on any of the outstanding Editor's Notes, preferably before the next SVG Accessibility Task Force teleconference on 25 May 2016.


AmeliaBR:
Comments from Fred (@fesch):

- [Part 1](https://lists.w3.org/Archives/Public/public-svg-a11y/2016May/0021.html), with [Amelia's reply](https://lists.w3.org/Archives/Public/public-svg-a11y/2016May/0024.html)
- [Part 2](https://lists.w3.org/Archives/Public/public-svg-a11y/2016May/0022.html), with [Amelia's reply](https://lists.w3.org/Archives/Public/public-svg-a11y/2016May/0025.html)
- [Part 3](https://lists.w3.org/Archives/Public/public-svg-a11y/2016May/0023.html), & [Amelia's reply](https://lists.w3.org/Archives/Public/public-svg-a11y/2016May/0026.html)

Typos & wording fixes have been integrated in this branch.

The main outstanding debate is when should elements that are visually hidden with styles count as "aria-hidden" and excluded from the accessibility tree.  The current draft excludes all elements:

- that are not rendered due to SVG semantics. This won't change.

- that are not rendered due to `display: none`. This won't change by default; we can discuss whether it should be possible for an authors to over-ride this with `aria-hidden="false"`; however, we'd need commitments from user agents to go that path, as this is currently poorly supported in HTML.

- that are not visible (via `visibility: hidden`) and/or not painted (via `fill: none` and `stroke: none` on elements that are normally filled & stroked) **and** are not sensitive to pointer events based on the `pointer-events` property.  This could be simplified to only use `visibility` and `pointer-events`, although `pointer-events` itself relies on `fill` and `stroke`.

Other styles and attributes that can currently cause an element to be invisible (e.g., zero opacity, transparent fill, zero height/width, clipping, positioning off screen) are not assessed when determining if an element is "hidden" based on SVG/CSS semantics.  An author would need to explicitly add `aria-hidden="true"` to ensure that elements made invisible with these styles are ignored by assistive tech.  Conversely, if these styles were used to express a valid zero data value (like a zero-height bar on a bar chart), the element would still be included even though it doesn't have any corresponding painted pixels on screen.

AmeliaBR:
Resolutions from 25th May telcon of the SVG Accessibility Task Force:

- [Change SVG-AAM so that fill:none and stroke:none are not considered native semantics for hiding an element](https://www.w3.org/2016/05/25-svg-a11y-minutes.html#resolution01)

- [Accept Amelia's edits to SVG-AAM pending completion of [those changes]](https://www.w3.org/2016/05/25-svg-a11y-minutes.html#resolution02)

PR #378, opened 2016-05-20
Change to support heading level suppression

halindrome:
aria.js was updated to ensure that headings are used correctly within
the spec and so that the associated landmarks are appropriate for ATs.
Child specs probably need this change too.

@mcking65 can you look at this in conjunction with aria/dpub.html and ensure the heading levels make sense?


mcking65:
Shane, I looked at http://rawgit.com/w3c/aria/shane-childHeadingLevel/aria/dpub.html
and all the heading levels looked good to me.
Thank you for this fix!

PR #384, opened 2016-05-25
Added introduction to Graphics AAM

fredesch:
Copied most of the SVG AAM introduction to the Graphics AAM


AmeliaBR:
Maybe don't need quite as much repeat, but not a big deal.  However, at least two essential changes are required:

- The paragraph starting "This specification begins" needs to be removed and replaced with a description of the purpose of this specs.
- The list to other relevant specs needs to be updated to replace the reference to Graphics-AAM with a corresponding reference to SVG-AAM.

fredesch:
@AmeliaBR  wow, that is fast comments. I just noticed the reference to the Graphics AAM.

PR #385, opened 2016-05-25
Gm intro 

fredesch:
Added intro updated based on @AmeliaBR 's comments.


fredesch:
Adding Intro to Graphics AAM

PR #390, opened 2016-05-27
AXAPI aready in SpecRef

marcoscaceres:
- http://www.specref.org/?q=AXAPI


joanmarie:
Committed via manual merge. Thanks for the fix!

PR #392, opened 2016-05-27
Remove redudant jQuery

marcoscaceres:
ReSpec imports jquery for you.

### DIFF
--- a/accname-aam/accname-aam.html
+++ b/accname-aam/accname-aam.html
@@ -3,7 +3,6 @@
 <head>
 <title>Accessible Name and Description: Computation and API Mappings 1.1</title>
 <meta content="text/html; charset=utf-8" http-equiv="Content-Type"/>
-<script src="../common/script/jquery-1.9.0.min.js"></script>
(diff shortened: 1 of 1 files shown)


joanmarie:
Thanks for the pull request! (And for pull request #393). It looks like all of the -AAM specs have this; not just the two you cited. Before we remove those references -- along with [common/script/jquery-1.9.0.min.js](https://github.com/w3c/aria/blob/master/common/script/jquery-1.9.0.min.js):

@halindrome or @michael-n-cooper: Do you happen to know why we are maintaining [common/script/jquery-1.9.0.min.js](https://github.com/w3c/aria/blob/master/common/script/jquery-1.9.0.min.js)? It appears that this file arrived (in a different location) as part of the initial check-in of Core-AAM, and then got moved to the common tools during a repo restructuring.

marcoscaceres:
Actually, you might want to keep jQuery here. I plan on removing jQuery from ReSpec itself.

joanmarie:
So this PR and PR #393 are now obsolete? I am assuming "yes" and closing them. If my assumption is bogus, please re-open. Thanks again!

PR #395, opened 2016-05-31
Added detection of required states or properties.

### LINKED ISSUE #386: [SCRIPTS] Add (or restore/fix) programmatic way to mark required inherited properties
([Reported via Twitter](https://twitter.com/cookiecrook/status/735745101926780928))

In ARIA 1.0, required states and properties which were inherited were marked as "required" in the characteristics table. This is no longer happening. See, for example, [menuitemcheckbox for 1.0](https://www.w3.org/TR/2014/REC-wai-aria-20140320/complete#menuitemcheckbox) compared to [menuitemcheckbox for 1.1](https://www.w3.org/TR/2016/WD-wai-aria-1.1-20160317/#menuitemcheckbox). I've not yet researched how it happened before, and while I could go in and manually add "(required)" everywhere it belongs, it would be cool if it magically JustWorked(tm).

@halindrome: Do you happen to know off the top of your head if this used to be done by the scripts?

halindrome:
Now indicates if a state / property is required when inherited.
Fixes #386


Joanmarie:
Overall, LGTM.

Looking at the `switch` role, I see that `aria-checked` is both required and inherited. Not sure if that should be fixed by your script or by me not explicitly requiring states that will be required via inheritance. Probably the latter. If you agree, then ship it and I'll remove the explicit requirement of the checked state on `switch` Otherwise, I'll wait.

halindrome:
hmm.  well - what would happen if something were inherited NOT required, and a subclass added a requirement for it?

Joanmarie:
True.... In that case, the state would still appear twice, but the fix from your patch would not make the duplication jump out at me like it does now. :)

Aesthetically speaking, I think that states and properties that are explicitly supported on a particular role should not show up as inherited. I'm less convinced of that from a technical point of view.

On a related note, while looking to see if any other roles are impacted by this issue, I noticed that the `slider` role has three required properties (`aria-valuemax`, `aria-valuemin`, and `aria-valuenow`) which are also inherited, but they lack the "(required)" in the inherited row. If we decide we love duplication (or even if we don't), seems you may have an unhandled case. (?)

halindrome:
The core issue seems to be this:  Sometimes states / properties are inherited.  When that inheritance is from a non-abstract role, it may be the case that non-abstract role REQUIRED the state or property.  Then the subclass should clearly also indicate it is required.  It can do this by having the item in its own table row about required things, or it can be labelled at (required) in the inherited row.  It MUST NOT appear in two places.

halindrome:
This should be good now.  @Joanmarie r?

Joanmarie:
@halindrome: Looking at menuitemcheckbox, I'm not seeing aria-checked as required, either in its own row or via "(required)" in the inherited row.

halindrome:
I think this is fixed now - can you verify @Joanmarie ?

Joanmarie:
* "(required)" appears on:
  * `aria-checked` for `menuitemcheckbox` and `menuitemradio`
  * `aria-selected` for `treeitem`
* duplication handled:
  * `aria-valuemax`, `aria-valuemin`, `aria-valuenow` are only in the required row for `slider` (and not in inherited row)
  * `aria-checked` is only in the required row for `switch` (and not in the inherited row)
  * no other duplication found

r+

Thanks!

PR #396, opened 2016-06-02
Add warning about password role

MichielBijl:
Add warning about password role usage as [accepted in June 2nd's ARIA call](http://www.w3.org/2016/06/02-aria-minutes.html#resolution01).

**This is awaiting CfC**

### DIFF
--- a/aria/aria.html
+++ b/aria/aria.html
@@ -4561,8 +4561,8 @@
 			<rdef>password</rdef>
 			<div class="role-description">
 				<p>[ARIA 1.1] An <rref>input</rref> for entering a password.</p>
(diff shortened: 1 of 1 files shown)


Joanmarie:
Thanks for doing this. After any other language changes are made, could you please make the whitespace and element placement of the new paragraph consistent with all the other paragraphs in the spec? While I don't like the entire-paragraph-as-a-line thing we currently do, and plan to reformat the whole spec after we branch and separate repos, I hate inconsistency even more. ;)

MichielBijl:
I feel you, it's done.

PR #397, opened 2016-06-03
action-2065

cyns:
added text from action-2065


cyns:
@klown  merged manually

Issue #382, opened 2016-05-25
[SCRIPTS] Subclass roles missing from ARIA properties tables after 0460074


joanmarie:
Compare:
- https://cdn.rawgit.com/w3c/aria/0460074/aria/aria.html#input
- https://cdn.rawgit.com/w3c/aria/0460074^/aria/aria.html#input

In the properties table of the latter, the subclass roles are listed; in the former, that row is missing.

@halindrome, could you please take a look? Thanks!

halindrome:
Sure - might not be until late tomorrow though...  Is it urgent?

Joanmarie:
Not urgent (if it were, I'd dig into it and propose a patch for you to review). If you have time over the next few days, that would be great. Thanks!

halindrome:
Could this have been overcome by events?  Related to the general problem with ReSpec from last week?

Joanmarie:
I'm still seeing the problem. If I look at https://rawgit.com/w3c/aria/master/aria/aria.html#input, I see the following rows in the characteristics table:

- Is Abstract
- Superclass Role
- Related Concepts
- Inherited States and Properties
- Name From

If I look at https://cdn.rawgit.com/w3c/aria/0460074^/aria/aria.html#input, I see the following rows in the characteristics table:

- Is Abstract
- Superclass Role
- **Subclass Roles**
- Related Concepts
- Inherited States and Properties
- Name From

In the "Subclass Roles" row, I see:
- checkbox
- option
- password
- radio
- slider
- spinbutton
- textbox

PR #400, opened 2016-06-14
Addresses issue with subclass roles being dropped

halindrome:
As mentioned in issue #382, the subrole row of the role tables stopped
being populated when I wrapped the roles in section elements.

@Joanmarie if you agree please LMNK so I can merge this.


Joanmarie:
lgtm. Thanks!

PR #401, opened 2016-06-14
Added code to accumulate lists of where names are from

halindrome:
Also added placeholders in the aria spec for them in
sections 5.2.7.4 and 5.2.7.5 as per @Joanmarie 

Note that this addresses my action in Tracker 1723

### DIFF
--- a/aria/aria.html
+++ b/aria/aria.html
@@ -455,16 +455,18 @@
 				<h4>Text Alternative Computation</h4>
 				<p><a href="http://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_te">Text Alternative Computation</a> is defined in the Accessible Name and Description specification [[ACCNAME-AAM]].</p>
 			</section>
+            <!--
             <section id="namefromauthor">
-                <h4>Name From Author List</h4>
+                <h4>Name from Author List</h4>
                 <div id="index_fromauthor">
                 </div>
             </section>
             <section id="namefromcontent">
-                <h4>Name From Content List</h4>
+                <h4>Name from Content List</h4>
                 <div id="index_fromcontent">
                 </div>
             </section>
+            -->
 		</section>
 		<section id="childrenArePresentational">
 			<h3>Presentational Children</h3>
--- a/aria/aria.html
+++ b/aria/aria.html
@@ -455,6 +455,16 @@
 				<h4>Text Alternative Computation</h4>
 				<p><a href="http://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_te">Text Alternative Computation</a> is defined in the Accessible Name and Description specification [[ACCNAME-AAM]].</p>
 			</section>
+            <section id="namefromauthor">
+                <h4>Name From Author List</h4>
+                <div id="index_fromauthor">
+                </div>
+            </section>
+            <section id="namefromcontent">
+                <h4>Name From Content List</h4>
+                <div id="index_fromcontent">
+                </div>
+            </section>
 		</section>
 		<section id="childrenArePresentational">
 			<h3>Presentational Children</h3>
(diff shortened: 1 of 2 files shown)


Joanmarie:
Thanks for doing this! Observations:

1.  Nit: lowercase "f" in "from" (assuming we keep that section, which I'll come back to momentarily). This is according to the _Chicago Manual of Style_, which some doc somewhere indicates we follow.
2. I think we want to list the roles, but not the definitions. Compare with the [role categorization lists](https://rawgit.com/w3c/aria/master/aria/aria.html#roles_categorization).
3. I know I asked for the new sections. Now that I'm seeing it, I wonder if we also want to do something similar to the role categorization lists. Namely, within [5.2.7.1 (Name Computation)](https://rawgit.com/w3c/aria/master/aria/aria.html#namecalculation), add statements like:
    * "The following roles derive their accessible name from `aria-label` or `aria-labelledby`:"
    * "The following roles derive their accessible name from the element content when there is no author-provided name:"
4. I wonder if we also want to indicate when name is required (like we did for states and properties).

In terms of what is your problem versus what is mine: If you could do item 2, that would be great. The solution for 1 and 3 is probably to leave what you have, but comment it out. Then I can move it (if I choose), and put some text around it. Not sure about 4.

Thanks again!

halindrome:
1. I changed that.
2. I have made that change.
3. I have commented it out.  Note that means that my changes don't actually RUN so you can't see what I have done.  You will just have to trust me.
4. Required or not.... I don't think I know HOW to indicate that.  It is not indicated in the table content.  


P.S. The role "presentation" has a weird namefrom.  Can you suggest how to handle that?

P.P.S.  There are a few namefrom cells that say "n/a".  Is this intentional.  Normally a table row that doesn't apply is left blank and magically removed.

Joanmarie:
> I have commented it out. Note that means that my changes don't actually RUN so you can't see what I have done. You will just have to trust me.

Or uncomment it locally. Though I do trust you. :)

> Required or not.... I don't think I know HOW to indicate that. It is not indicated in the table content.

Could you use "Accessible Name Required"?

> P.S. The role "presentation" has a weird namefrom. Can you suggest how to handle that?

Hmmmm. Could you see if the value starts with "author"?

> P.P.S. There are a few namefrom cells that say "n/a". Is this intentional. Normally a table row that doesn't apply is left blank and magically removed.

Dunno. It's been that way since I've been an editor. We can ask some of the "old timers" if we're trying to make an explicit point.

halindrome:
>>Required or not.... I don't think I know HOW to indicate that. It is not indicated in the table content.

>Could you use "Accessible Name Required"?

LOL.  Clearly.  Never noticed that.  Yikes.  Now using that.  I have changed it to read (required) after the name it is required.  I am NOT looking at inheritance of name required.  There is such a row in the table and it is ALWAYS empty as far as I can tell.  

>>P.S. The role "presentation" has a weird namefrom. Can you suggest how to handle that?

>Hmmmm. Could you see if the value starts with "author"?

That's what I do now.  I was mostly wondering if there was any reason to annotate it specially.

>>P.P.S. There are a few namefrom cells that say "n/a". Is this intentional. Normally a table row that doesn't apply is left blank and magically removed.

>Dunno. It's been that way since I've been an editor. We can ask some of the "old timers" if we're trying to make an explicit point.

Thanks

Joanmarie:
LGTM. One minor further request (pending your thoughts) before you push: Instead of "(required)", what do you think about "(name required)"?

In the case of the states and properties, I think it's obvious that the state/property is the thing being required. In the case of names, what is required is the name, BUT the source can be the author or, in some cases, the content. And in the case where it's permissible to fall back on the content, "required" under the "Name from Author List" might be interpreted as "`aria-label` or `aria-labelledby` is required."

(And, yes, I should have thought of that before. Many apologies if it feels like this is becoming a game of "fetch me a rock.")

Regardless, go with your judgment regarding the above,  and I'll then add the lists Joseph wants. Thanks again!!

halindrome:
Happy to change to name required.  I note that (required) is inside a strong element in those other places. Do you want that here too?  &lt;strong&gt;(name required)&lt;/strong&gt;

Joanmarie:
No strong opinions regarding using the `strong` element. On the one hand, consistency; on the other hand, I think in the table we are emphasizing the requirement. In this list, I personally see it as a "btw." So whichever you think is fine with me.

PR #446, opened 2016-09-30
Issue444

richschwer:
Updated based on IA2 feedback and MacOSX feedback for updated dpub spec.


cookiecrook:
I'll raise the change requests as separate issues.

PR #467, opened 2016-11-04
ACTION-2120: UIA: clean up mapping for aria-haspopup

boggydigital:
Clean up the wording with the current UIA implementer expecation.
Currently UIA doesn't provide a way to expose richer semantic of
aria-haspopup values, so in all non-trivial cases we're using same
mapping as in "true" value case. For "false" the mapping would require
to not implement ExpandCollapse pattern which is reflected in new "Not
mapped" value

### DIFF
--- a/core-aam/core-aam.html
+++ b/core-aam/core-aam.html
@@ -1928,7 +1928,7 @@ var mappingTableLabels = {
                         <li>Expose as object attribute <code>haspopup:true</code></li>
                       </ul>
                   </td>
-                  <td>Expose state of the pop-up activities as <code>"expanded"</code> in the <code>ExpandCollapseState</code> property of the <code>ExpandCollapse</code> Control Pattern</td>
+                  <td>Implement <code>ExpandCollapse</code> Control Pattern, set the <code>ExpandCollapseState</code> property to match <a class="state-reference" href="#aria-expanded"><code>aria-expanded</code></a> (state)</td>
                   <td>
                     <ul>
                       <li>Expose <code>STATE_HAS_POPUP</code>
@@ -1947,9 +1947,7 @@ var mappingTableLabels = {
                     </ul>
                   </td>
                   <td>
-                    <ul>
-                      <li>If explicitly set to <code>"false"</code>, expose state of the pop-up activities as <code>"collapsed"</code> in the <code>ExpandCollapseState</code> property of the <code>ExpandCollapse</code> Control Pattern</li>
-                      <li>Otherwise, <a href="#not_mapped">Not mapped*</a></li>
+                    <a href="#not_mapped">Not mapped*</a></li>
                     </ul>
                   </td>
                   <td><a href="#not_mapped">Not mapped*</a></td>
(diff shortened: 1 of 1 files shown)


klown:
Thanks @boggydigital.

I have merged the changes into [master](https://github.com/w3c/aria/commit/d8f5e6d5245a473fe440b615dd764cacc80e18f0), and marked [ACTION-2120](https://www.w3.org/WAI/ARIA/track/actions/2120) as "pending review".  I will bring it up at next Tue's AAPI teleconference (Nov 29) for other members' comments.

richschwer:
This change will not work because aria-expanded is not required on widgets. For example a menu button can have aria-haspopup="true" and when you hit the space bar a menu popups up and focus typically moves to the menu and it does not matter whether the button is expanded. 

This won't work and we are not going go back to CR to require aria-expanded on menu buttons as it has never been needed.

boggydigital:
@richschwer It shouldn't be needed, perhaps you have a suggestion how to word this better?
The idea is the following:
aria-haspopup -> ExpandCollapsedPattern

additionally, to make this pattern actually mean something I've reused the earlier text that was there that maps ExpandCollapsedPattern.ExpandCollapseState to aria-expanded. By itself aria-haspopup definitely shouldn't require aria-expanded and would produce just ExpandCollapsePattern

richschwer:
So, if you have aria-haspopup, and you have an ExpandCollapsedPattern its existence means that something can be popped up but no expanded or collapsed will be set. Is that correct?

boggydigital:
@richschwer That is correct. I've thought about wording a bit more, how does the following proposal sound?

"Implement <code>ExpandCollapse</code> Control Pattern. If the element has <a class="state-reference" href="#aria-expanded"><code>aria-expanded</code></a> (state), set the <code>ExpandCollapseState</code> property to match it"

richschwer:
I am fine with that wording Bogdan.

klown:
@boggydigital Since I have already [merged](https://github.com/w3c/aria/pull/467#issuecomment-262844573) your changes into the Core-AAM master branch, I am willing to edit that to use [your new wording](https://github.com/w3c/aria/pull/467#issuecomment-262844573).  That is, no need for you to fiddle with this pull request, nor fix any conflicts.  The down side is you won't get credit in the git log, although I will note that you provided the text in the commit message.

If you want to make the change here in the pull request and fix the conflicts, you can -- let me know which way you want to go.

boggydigital:
@klown Thank you James, I wouldn't worry about git log credit if your manual edit is easier. Thank you!

klown:
@boggydigital 

> @klown Thank you James, I wouldn't worry about git log credit if your manual edit is easier. Thank you!

You're welcome.  BTW, I'm Joseph, not James.

boggydigital:
@klown Joseph, I'm sorry this was just a typo, I definitely known the correct name. It's been a long week, my apologies again!

klown:
@boggydigital  No worries!

boggydigital:
I believe this should be closed with https://github.com/boggydigital/aria/commit/3cc20bc582a31001c72526af3b285f96609a1071

PR #468, opened 2016-11-04
ACTION-2070: Add UIA Range Value Pattern to the range related roles in the role mapping table (e.g., slider).

boggydigital:
Ported original Cynthia's change, while cleaning it up for consistency.
Additional small tweak (which I hope to drive for UIA mapping
consistently in the future) is searchbox which I converted from a
sentence to a list.

### DIFF
--- a/core-aam/core-aam.html
+++ b/core-aam/core-aam.html
@@ -1208,7 +1208,12 @@ var mappingTableLabels = {
 				<tr id="role-map-scrollbar">
 					<th><a class="role-reference" href="#scrollbar"><code>scrollbar</code></a></th>
           <td><code>ROLE_SYSTEM_SCROLLBAR</code></td>
-          <td><code>scrollbar</code></td>
+          <td>
+            <ul>
+              <li>Control Type = <code>ScrollBar</code></li>
+              <li>Implement <code>Range</code> Control Pattern</li>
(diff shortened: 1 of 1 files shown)


core-aam/core-aam.html
```diff
@@ -1208,7 +1208,12 @@
 				<tr id="role-map-scrollbar">
 					<th><a class="role-reference" href="#scrollbar"><code>scrollbar</code></a></th>
           <td><code>ROLE_SYSTEM_SCROLLBAR</code></td>
-          <td><code>scrollbar</code></td>
+          <td>
+            <ul>
+              <li>Control Type = <code>ScrollBar</code></li>
```
klown: Nit:  The core-aam currently uses the style:
"Control type is `ScrollBar`"

That is, no "=" sign.  I prefer consistency with the current style.  Any objections?

boggydigital: Pardon my delay. No objections on the style I would gladly follow the established guidelines. I did the change as I've noticed many UIA items had inconsistent style already and wanted to pick some consistent style. Should have asked!

I'll resolve conflict soon, hopefully tomorrow...

boggydigital:
Should be addressed with https://github.com/w3c/aria/pull/495 as well

PR #469, opened 2016-11-04
ACTION-2103: Provide uia mappings for roles feed, figure, and term; and aria-colcount, aria-colspan, aria-colindex, aria-rowcount, aria-rowindex, aria-rowspan, and aria-details.

boggydigital:
Map outstanding ARIA 1.1 roles and properties to relevant UIA values.
Fortunately for the defined roles I beleive the mapping is really clean
and straightforward. I would expect the only question might be raised
for aria-details. We map this to UIA DescribedBy property that achieves
the desired effect - the aria-details element would be referenced, not
it's value won't be used in Name or Description computation.

### DIFF
--- a/core-aam/core-aam.html
+++ b/core-aam/core-aam.html
@@ -734,10 +734,10 @@ var mappingTableLabels = {
                         <td>
 				<code>ROLE_SYSTEM_GROUPING</code> + object attribute <code>xml-roles:feed</code></td>
                         <td>
-                            Need to verify. 
-			    <ul><li>Control Type: <code>Group</code></li>
-                                <li>Localized Control Type: <code>"feed"</code></li>
-                            </ul>
+			                    <ul>
+                            <li>Control Type = <code>Group</code></li>
+                            <li>Localized Control Type = "<code>feed</code>"</li>
+                          </ul>
                         </td>
                         <td>
                                 <code>ROLE_PANEL</code> + object attribute <code>xml-roles:feed</code>
@@ -752,10 +752,12 @@ var mappingTableLabels = {
         	<tr id="role-map-figure">
                 	<th><a class="role-reference" href="#figure"><code>figure</code></a> [ARIA 1.1]</th>                
 			<td><code>ROLE_SYSTEM_GROUPING</code> + object attribute <code>xml-roles:figure</code></td>
-          		<td><ul><li>Control Type: <code>Group</code></li> 
-				<li>Localized Control Type: <code>"figure"</code></li>
-			    </ul>
-			</td>
+          		<td>
+                <ul>
+                  <li>Control Type = <code>Group</code></li> 
+				          <li>Localized Control Type = "<code>figure</code>"</li>
+              </ul>
+          </td>
           		<td>
               			<code>ROLE_PANEL</code> + object attribute <code>xml-roles:figure</code>
           		</td>                         
@@ -1246,7 +1248,7 @@ var mappingTableLabels = {
           <td><code>ROLE_SYSTEM_TEXT</code> + object attribute <code>text-input-type:search</code></td>
           <td>
             <ul>
-              <li>Control Type is <code>Edit</code></li>
+              <li>Control Type = <code>Edit</code></li>
               <li>Localized Control Type is "<code>search box</code>"</li>
             </ul>
           </td>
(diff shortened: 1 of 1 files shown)


richschwer:
That is fine but you can't stringify the description. Often the description will involve alternate content. For example , it could be SVG or even MathML
 
 Sent from my iPad using IBM Verse
 
 On Nov 4, 2016, 6:30:06 PM, notifications@github.com wrote:

klown:
@boggydigital Thanks for this!

There is a chance that an element has both `aria-describedby` and `aria-details`.  Some accessbility APIs cannot handle the mulitple references.  The [spec](https://rawgit.com/w3c/aria/master/aria/aria.html#aria-details) states that in such cases, `aria-details` is used:

> In some user agents, multiple reference relationships for descriptive information are not supported by the accessibility API. In such cases, if both aria-describedby and aria-details are provided on an element, aria-details takes precedence.

Is UIA such an API?  If so, then I think the mappings should point this out.  If not,  how are they distinguished in UIA?

boggydigital:
@richschwer that is exactly he point and intent here. UIA DescribedBy is a semantic pointer, not to be confused destination for aria-describedby, that is used in description computation landing in UIA FullDescription string property.

@klown agree and it should work here for two reasons. DescribedBy is an array of pointers (see UIA_DescribedByPropertyId here https://msdn.microsoft.com/en-us/library/windows/desktop/ee684017(v=vs.85).aspx) and this array by itself doesn't have direct implications on FullDesciption property same way aria-describedby has. While currently aria-describedby is primarily used in FullDescription computation, it's also mapped as a pointer to element through DescribedBy, however this is absolutely unnecessary and we'll explore changing this in the future. With that said aria-details would add or create DescribedBy reference and there wouldn't be a hard conflict with aria-describedby

core-aam/core-aam.html
```diff
@@ -734,10 +734,10 @@
                         <td>
 				<code>ROLE_SYSTEM_GROUPING</code> + object attribute <code>xml-roles:feed</code></td>
                         <td>
-                            Need to verify. 
-			    <ul><li>Control Type: <code>Group</code></li>
-                                <li>Localized Control Type: <code>"feed"</code></li>
-                            </ul>
+			                    <ul>
+                            <li>Control Type = <code>Group</code></li>
+                            <li>Localized Control Type = "<code>feed</code>"</li>
```
klown: I noted in [PR 468](https://github.com/w3c/aria/pull/468#discussion_r89543308) that the current sytle was to use "is" not "=".  It appears the document inconsistently uses "is" in some cases and ":" (colons) in others.  Personally, I prefer the colons, but you are the UIA expert.  What would be best here, @boggydigital?  I will make a global change based on your advice.

core-aam/core-aam.html
```diff
@@ -1816,7 +1843,7 @@
 		   <p>IAccessible2: </p>
 		   <p>If the referenced object is in the accessibility tree, expose a pointer to the accessible object using <code>IA2_RELATION_DETAILS</code>, and expose reverse relations as described in <a href="#mapping_additional_relations">Relations</a>.</p>
 		  </td>
-                  <td>TBD.</td>
+                  <td><code>DescribedBy</code> property points to the referenced element</td>
```
klown: In light of the [previous discussion](https://github.com/w3c/aria/pull/469#issuecomment-260018111), how about changing this to:
"`DescribedBy` array property contains an entry pointing to the referenced accessible object".

richschwer:
The problem with not uniquely separating detail references and description references in the array is that details quite often is an alternative representation of the content (digital publishing industry) such as a ChemML rendering that cannot be processed by normal screen reading and the user may need to know that. So, this is much more than a description.

core-aam/core-aam.html
boggydigital: There is no reason not to match existing style guidelines, my fault - didn't know they exist and current UIA mappings were mixed. I'll definitely use "is" from now on

boggydigital:
@richschwer While I understand your concern there is no better option today - so we have basically to choose from adding it to DescribedBy array or no mapping. I would suggest to still go with the current proposed mapping and take the feedback to provide more semantic information on detail to UIA.

boggydigital:
I've opened a new revision that should be much cleaner and address the feedback - https://github.com/w3c/aria/pull/495

PR #487, opened 2016-11-23
Clarify keyshortcuts usage

MichielBijl:
Was: <q>Indicates keyboard shortcuts that an author has implemented to activate or give focus to an element.</q>

Changed into: <q>Indicates keyboard shortcuts that an author has implemented to activate or give focus to the currently focused element.</q>

@mcking65 @accdc


MichielBijl:
Maybe “current element” instead of “currently focused element”?

mcking65:
The element does not have to be focused for the keyshortcut property to be present or useful. And, current element is not a defined term.

I think the definition is very good the way it is. The only thing I would consider tweaking is changing "an element" to "the element" so it is consistent with similar definitions and makes it clear that the element that has the property is the element that will be activated or focused when the shortcut is executed.

MichielBijl:
Entire line has been rewritten by someone, so closing this.

PR #496, opened 2016-12-06
Provide UIA mappings for DPUB-AAM

boggydigital:
UIA mapping is to expose digital publication text entities in the UIA tree. This is achieved by mapping them to UIA Control Type "Text" and using role as Localized Control Type (without "doc-" prefix). 

Using "doc-toc" as example - it's mapped to the following:
- Control Type is Text
- Localized Control Type is "toc"

This is true for all text based roles, the only exception is "doc-cover" that is mapped to UIA Image Control Type. 

Last doc-pullquote is not mapped as agreed upon in W3C.

### DIFF
--- a/dpub-aam/dpub-aam.html
+++ b/dpub-aam/dpub-aam.html
@@ -236,8 +236,8 @@ var mappingTableLabels = {
 								</td>
                                                                 <td>
                                                                   <ul>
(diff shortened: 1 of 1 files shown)


richschwer:
Bogdan please review my comments.

boggydigital:
Thank you Richard, that makes sense and I believe there was W3C discussion that arrived to that result. I've updated the PR to reflect that. Updated change log to note the changes based on your feedback as well.

PR #499, opened 2016-12-14
one mapping-tables script that works with or without respec

jasonkiss:
A single mapping-tables.js that checks for respec context, and executes
accordingly. This is instead of having two versions, one for the respec
context in the master branch, and a different version for the snapshot
gh-pages branch. See related w3c/html-aam issue 74:
https://github.com/w3c/html-aam/issues/74

I'm sure my implementation can be improved, but is currently working.

We needed to do this to support auto-publishing of WD for the HTML-AAM. It also reduces the number of exceptions to publishing from a 'master' branch, since mapping-tables.js has only version, regardless of branch/context.

### DIFF
--- a/common/script/mapping-tables.js
+++ b/common/script/mapping-tables.js
@@ -1,7 +1,18 @@
-/* globals $, require */
+//check for require() and respec context
+if (typeof require !== "undefined") {
+	/* globals $, require */
+	require(["core/pubsubhub"], function(respecEvents) {
+		mapTables(respecEvents);
+	});
+} else {
+	$(document).ready(function() {
+		mapTables(false);
+	});
+}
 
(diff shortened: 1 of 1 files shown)


joanmarie:
Thanks @jasonkiss. I've added it to the core-aam repo (https://github.com/w3c/core-aam/commit/35d0d6257d455560db30d08231c50087d442c241z) and also done a pull request to aria-common which includes your patch (https://github.com/w3c/aria-common/pull/1).

I'm going to close this pull request because it doesn't apply to the ARIA spec itself, which is the only spec left in this repo.

michael-n-cooper:
The content from this pull request is in https://github.com/w3c/aria-common/blob/master/script/mapping-tables.js

PR #511, opened 2017-01-07
Clarify that owned-by author rules need explicit role attributes

sideshowbarker:
There are sentences in the ARIA spec which state document-conformance requirements for role values that have owned-by relationships with other role values. This change clarifies those sentences to state unambiguously that elements must have **explicit role attributes** containing those values in order for the requirements to be relevant.

For example, this change takes the following sentence:

> Authors MUST ensure elements with role `cell` are contained in, or owned by, an element with the role `row`. 

…and replaces that with this sentence:

> Authors MUST ensure elements with an explicit `role` attribute whose value contains `cell` are owned by an element with an explicit `role` attribute whose value contains `row`.

Also, the same sentences used the phrase “contained in, or owned by” but since the definition of “owned by” states that “An 'owned element' is any DOM *descendant* of the element, any element specified as a child via aria-owns, or any DOM *descendant* of the owned child” it’s not necessary to state “contained in, or owned by”; instead that phrase can be replaced by just “owned by”—since the definition of “owned by” already incorporates a “contained in” _descendant_-ancestor relationship as among what it means (so, “contained in, or owned by” is unnecessarily redundant).

Note that this change is relevant to the behavior for the HTML checker. For the context, see https://github.com/validator/validator/issues/425#issuecomment-269841829 and the discussion that follows that comment.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/511.html" title="Last updated on Jan 11, 2018, 1:13 PM GMT (2855dc6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/511/e81aa27...2855dc6.html" title="Last updated on Jan 11, 2018, 1:13 PM GMT (2855dc6)">Diff</a>

### DIFF
--- a/common/terms.html
+++ b/common/terms.html
@@ -134,7 +134,7 @@
     <dd>
       <p>Usable by users in ways they can control. References in this document relate to <cite><a href="http://www.w3.org/TR/WCAG20/#operable"><abbr title="Web Content Accessibility Guidelines">WCAG</abbr> 2.0 Principle 2: Content must be operable</a></cite> [[WCAG20]]. See <a>Keyboard Accessible</a>.</p>
     </dd>
-    <dt><dfn data-lt="owned element|owned|owned element's|owned elements">Owned Element</dfn></dt>
+    <dt><dfn data-lt="owned element|owns|owned|owned element's|owned elements">Owned Element</dfn></dt>
     <dd>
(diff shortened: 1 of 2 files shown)


sideshowbarker:
Rebased.

PR #532, opened 2017-03-07
Updated UIA mappings for role=heading

boggydigital:
Control Type=Text is the best current UIA target (no change), additionally use Localized Control Type=heading to capture "heading" semantic


klown:
Thanks @boggydigital.

Link to tracker ISSUE-738:
https://www.w3.org/WAI/ARIA/track/issues/738

Link to email about this mapping:
https://lists.w3.org/Archives/Public/public-aria/2017Mar/0010.html

klown:
@boggydigital  I've merged in your changes.  Let me know if you see any problems with the merge:
https://rawgit.com/w3c/aria/master/core-aam/core-aam.html#role-map-heading

klown:
@boggydigital @richschwer @Joanmarie Making a note here that, as I understand it, UIA does not implement, at this time, "Localized Control Type=heading".  If I'm wrong about that, let me know.  We are hoping that it will be implemented by the time testing of the mapping is done.

boggydigital:
Unfortunately we don't have ship vehicle to implement changes soon, however I've filed public bug https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11553099/ that would allow us to fix this in the next release we're currently working on.

PR #545, opened 2017-03-31
Core AAM: Fix ATK/AT-SPI2 mapping for ARIA listitem role

Joanmarie:
* Change ROLE_LISTITEM to ROLE_LIST_ITEM to reflect what is in both ATK and AT-SPI2.

* Remove "do not expose STATE_EDITABLE" because that is not necessarily accurate, nor is it specific to ARIA's listitem role: STATE_EDITABLE should be exposed for any element, regardless of role, if and only if the text of that element can be changed by the user.


Joanmarie:
Thanks for the review!

1. Would you like me to update this issue and proposed patch to *just* fix the spelling of the role?
1. Because this was a quick fix, I'd not noticed that "do not expose STATE_EDITABLE" appears in many places. Do you think we should (in a separate pull request) remove that statement?

Joanmarie:
Having found another spelling error, I'm going to do a new pull request to fix both this and the new one. Let's make the editable stuff a separate issue based on what you pointed out.

PR #549, opened 2017-04-10
DPUB-AAM: Map doc-pullquote for ATK, IA2, and UIA.

Joanmarie:
The "not mapped" mapping in this spec contradicts what is stated in the
DPub ARIA spec, namely that doc-pullquote will be exposed unless authors
explicitly mark it as presentational or hidden. See mailing list thread:
https://lists.w3.org/Archives/Public/public-dpub-aria/2017Apr/thread.html#msg0


Joanmarie:
Link to the rawgit/spec view: https://rawgit.com/w3c/aria/pullquote-mapping/dpub-aam/dpub-aam.html#role-map-pullquote

mattgarrish:
LGTM

Joanmarie:
@mattgarrish: Thanks! Since there have been no additional comments, will merge it in.

Issue #459, opened 2016-11-02
CORE-AAM 1.1: Create macOS Mappings for aria-colindex, -rowindex, -colcount, -rowcount


richschwer:
Moved from this action in tracker: https://www.w3.org/WAI/ARIA/track/actions/2119

richschwer:
Note: one issue we are discussing with NVDA and Freedom Scientific is the issue of dealing with aria-colindex and aria-rowindex, aria-rowspan, aria-colspan in the mappings when rows or columns are discontinuous.

cookiecrook:
What does that mean? rowspan and colspan by definition make the spanned columns discontiguous don't they?

cookiecrook:
These are mapped to: AXARIAColumnIndex, AXARIARowIndex, AXARIAColumnCount, AXARIARowCount. The *span attrs change the row/col associations.

klown:
Thanks @cookiecrook 

> These are mapped to: AXARIAColumnIndex, AXARIARowIndex, AXARIAColumnCount, AXARIARowCount. 

There are macOS mappiings of [aria-rowindex](https://rawgit.com/w3c/aria/master/core-aam/core-aam.html#ariaRowIndex) and [aria-colindex](https://rawgit.com/w3c/aria/master/core-aam/core-aam.html#ariaColIndex). I suspect these are wrong, but I wanted to confirm it with you:
`AXRowIndexRange: loc=<rowindex value> len=<rowspan value>`
`AXColumnIndexRange: loc=<column index value> len=<colspan value>`


>The *span attrs change the row/col associations.

I don't know how to word that in the mapping table.  What should go in the AX API cell for [aria-rowspan](https://rawgit.com/w3c/aria/master/core-aam/core-aam.html#ariaRowSpan) and [aria-colspan](https://rawgit.com/w3c/aria/master/core-aam/core-aam.html#ariaColSpan)?  Note, again, there already are entries:
`AXRowIndexRange: loc=<rowindex value> len=<rowspan value> `
`AXColumnIndexRange: loc=<column index> len=<colspan value> `

Joanmarie:
For the sake of documentation, etc.:
* [Nan did the implementation in October 2015](https://trac.webkit.org/changeset/190833).
[...]

PR #557, opened 2017-04-13
Core-AAM: Fix CFRange property names for AXAPI

Joanmarie:
AXAPI exposes aria-colspan and aria-rowspan through AXColumnIndexRange
and AXRowIndexRange respectively. These range values are of type
kAXValueCFRangeType which is a wrapper for CFRange [1]. Looking at the
documentation for CFRange [2], the properties are location and length;
not loc and len.

[1] https://developer.apple.com/reference/applicationservices/axvaluetype/kaxvaluecfrangetype
[2] https://developer.apple.com/reference/corefoundation/cfrange

### DIFF
--- a/core-aam/core-aam.html
+++ b/core-aam/core-aam.html
@@ -1908,9 +1908,9 @@ var mappingTableLabels = {
 			</ul>
                   </td>
                   <td><code>AXColumnIndexRange:<br />
-				loc=&lt;column index&gt;<br />
-				len=&lt;colspan value&gt;</code>,<br />
-				where <code>loc</code> is 0-based.
+				location=&lt;column index&gt;<br />
+				length=&lt;colspan value&gt;</code>,<br />
+				where <code>location</code> is 0-based.
 		  </td>
                 </tr>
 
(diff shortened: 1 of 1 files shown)


klown:
@Joanmarie I've merged in your changes:
https://github.com/w3c/aria/commit/6ef5fba702115141cf48ff1e0ec71922f41155e6

rawgit:

- https://rawgit.com/w3c/aria/b68faca5c2e5a69ab0bf9715d3327008e9bab265/core-aam/core-aam.html#ariaRowSpan
- https://rawgit.com/w3c/aria/b68faca5c2e5a69ab0bf9715d3327008e9bab265/core-aam/core-aam.html#ariaColSpan

Feel free to close, or let me know if there are issues.

Joanmarie:
@klown regarding this:

> The "len" and "loc" forms were confirmed in [gh-issue-459](https://github.com/w3c/aria/issues/459#issuecomment-272947803). Did the mappings change? Or was the documentation wrong back then?

TL;DR: It was more a case of not finding the documentation and believing tools like the Accessibility Inspector and other specs like HTML-AAM over the code. (In retrospect, that was extremely silly of me.)

Longer answer:  If you look at the comment you cite, you'll note I stated the following:

    And if you look at the associated code, you'll see the values come
    from AXRowIndexRange and AXColumIndexRange respectively. And for
    each of those, the value is a (range.location, range.length) pair.

So the code I reference shows that the mapping has the fully-written-out names. But then I said:

    ... Mind you, I was a tad surprised by the abbreviations "loc" and
    "len" because that seems really unWebKitish. But I just built Safari,
    loaded the test case, and examined it using Accessibility Inspector.
    It is indeed "loc" and "len". See attached screenshot... Lastly, all
    this jives with what's in the HTML-AAM.

Apologies for not having poked around more for the API docs which document what I found in the code.

PR #564, opened 2017-04-20
DPUB-AAM: Update mappings

Joanmarie:
A couple of changes were made in the WebKit (Safari and WebKitGtk)
implementation for DPub ARIA:

1. A new AXSubrole was added for ARIA groups. As a consequence, quite
   a few DPub ARIA roles which had no AXSubrole now have one. See:
   https://trac.webkit.org/changeset/214623

2. Several mappings were modified based on the Core ARIA role subclassed
   by DPub ARIA. See: https://trac.webkit.org/changeset/215554

As a result of these combined changes:

* For AXAPI, the AXSubrole is now AXApplicationGroup for the following
  roles: doc-abstract, doc-colophon, doc-dedication, doc-epigraph,
  doc-example, doc-footnote, doc-pullquote, and doc-qna

* For AXAPI, the AXRoleDescription of doc-abstract is now 'group'.

* For AXAPI, doc-pagebreak is now mapped to AXRole: AXSplitter,
  AXRoleDescription: 'splitter'

* For AXAPI, doc-notice and doc-tip are now mapped to AXSubrole:
  AXDocumentNote, AXRoleDescription: 'note'

* For ATK/AT-SPI2, doc-tip is now mapped to ROLE_COMMENT

### DIFF
--- a/dpub-aam/dpub-aam.html
+++ b/dpub-aam/dpub-aam.html
@@ -242,8 +242,8 @@ var mappingTableLabels = {
                                                                 </td>
                                                                 <td><p>Expose <code>ROLE_SECTION</code> and object attribute <code>xml-roles:doc-abstract</code>. </p></td>
                                                                 <td>AXRole: <code>AXGroup</code><br />
-                                                                        AXSubrole: <code>AXLandmarkRegion</code><br/>
-                                                                    AXRoleDescription: <code>'region'</code>
+                                                                        AXSubrole: <code>AXApplicationGroup</code><br/>
+                                                                    AXRoleDescription: <code>'group'</code>
                                                                 </td>
                                                         </tr>
                                                        <tr id="role-map-acknowledgments">
@@ -437,7 +437,7 @@ var mappingTableLabels = {
                                                                 </td>
                                                                 <td><p>Expose <code>ROLE_SECTION</code> and object attribute <code>xml-roles:doc-colophon</code>. </p></td>
                                                                 <td>AXRole: <code>AXGroup</code><br />
-                                                                        AXSubrole: <code>&lt;nil&gt;</code><br/>
+                                                                        AXSubrole: AXApplicationGroup</code><br/>
                                                                     AXRoleDescription: <code>'group'</code>
                                                                 </td>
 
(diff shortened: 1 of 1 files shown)


dpub-aam/dpub-aam.html
```diff
@@ -242,8 +242,8 @@
                                                                 </td>
                                                                 <td><p>Expose <code>ROLE_SECTION</code> and object attribute <code>xml-roles:doc-abstract</code>. </p></td>
                                                                 <td>AXRole: <code>AXGroup</code><br />
-                                                                        AXSubrole: <code>AXLandmarkRegion</code><br/>
-                                                                    AXRoleDescription: <code>'region'</code>
+                                                                        AXSubrole: <code>AXApplicationGroup</code><br/>
+                                                                    AXRoleDescription: <code>'group'</code>
```
cookiecrook: This diff looks wrong. Region should not have changed.

dpub-aam/dpub-aam.html
```diff
@@ -437,7 +437,7 @@
                                                                 </td>
                                                                 <td><p>Expose <code>ROLE_SECTION</code> and object attribute <code>xml-roles:doc-colophon</code>. </p></td>
                                                                 <td>AXRole: <code>AXGroup</code><br />
-                                                                        AXSubrole: <code>&lt;nil&gt;</code><br/>
+                                                                        AXSubrole: AXApplicationGroup</code><br/>
```
cookiecrook: These are probably right, but the specific role being changed is outside the diff lines, so it's difficult to be certain from this view.

dpub-aam/dpub-aam.html
Joanmarie: But I thought we agreed it wasn't a landmark.

dpub-aam/dpub-aam.html
Joanmarie: That's doc-colophon. I agree about the role and diff thing, but you can tell by the object attribute a couple of lines up. (IA2 and ATK require the actual/original role exposed in that fashion)

dpub-aam/dpub-aam.html
cookiecrook: Oh okay. I guess that's the same diff problem I mentioned below.

PR #575, opened 2017-04-27
Axapi popup button

Joanmarie:
As stated by Apple in this [comment](https://bugs.webkit.org/show_bug.cgi?id=171204#c3), "Presumably something still pops up  when the button is pressed, so we call it a popup button." Thus I believe my proposed change is correct.

Please review. Thanks!


Joanmarie:
Closing as a dup of #577.

PR #577, opened 2017-04-28
Add AXAPI mappings for button when aria-haspopup is not false

Joanmarie:
In both Safari and Chrome, a button where aria-pressed is undefined and
aria-haspopup is true has the following mapping:

    AXRole: AXPopUpButton
    AXSubrole: <nil>
    AXRoleDescription: 'pop up button'

This is also implemented in Safari for the new role-based values for
aria-haspopup.

This mapping is not reflected in the Core AAM.


klown:
@Joanmarie Squashed and merged:
https://github.com/w3c/aria/commit/280a1a721c26370bfbeda577e755ebd8373f1efd

Feel free to close this PR.

Joanmarie:
Thanks! Closing.

Issue #587, opened 2017-06-14
ARIA 1.1: Explicit guidance needed for non-pertinent errormessage elements


joanmarie:
I began implementing support for `aria-errormessage` in WebKitGtk, but was unsure how user agents should handle `aria-errormessage` referencing a valid, non-hidden element when that error message has not explicitly been marked as pertinent by the author. In particular:

1. The ARIA spec states, "Authors MUST use `aria-invalid` in conjunction with `aria-errormessage`. Initially, the object is in a valid state and either has `aria-invalid` set to `false` or no `aria-invalid` attribute, and the element referenced by `aria-errormessage` is not applicable."

    **Question:** What are user agents expected to do when `aria-invalid` is missing or `false`, but `aria-errormessage` references a valid, non-hidden element? Not expose the referenced element at all? Something else?

    **Proposal:** We should add a normative statement to the ARIA spec explaining the expectations of user agents. I would suggest something like, "User agents MUST NOT expose an element as a pertinent error message when the referencing element lacks `aria-invalid="true"`.

2. The ARIA spec states, "When `aria-errormessage` is pertinent, authors MUST ensure the content is not hidden and is included in a container that exposes the content to the user as it is expected that the assistive technology user will navigate to the content in order to access it."

    **Question:** When `aria-errormessage` is NOT pertinent, there is no statement regarding authors taking steps to cause the error message to be pruned from the accessibility tree. Is there any possibility the author will "hide" it visually in a way that it is still in the accessibility tree? If so, is it up to user agents to figure that condition out from aria-invalid's value? Hopefully not, but if that is the case, see item 1 above.

    **Proposal:** We should add a normative statement to the ARIA spec instructing authors to take steps to cause the error message to be pruned from the accessibility tree. Something like, "Authors MUST remove the error message when it is not pertinent, either by removing the element's id from `aria-errormessage` or by ensuring the referenced element is not rendered or has an `aria-hidden` value of `true`.

joanmarie:
Matt during meeting said he didn't like the idea of aria-hidden being set to true because that is not the purpose of aria-hidden="true".

joanmarie:
"Authors MUST remove the error message when it is not pertinent, either by removing the aria-errormessage attribute or its value, or by ensuring the referenced element is not rendered."

joanmarie:
[...]

PR #589, opened 2017-06-15
ARIA 1.2: Add explicit guidance for non-pertinent errormessage elements (#587)

joanmarie:
* When aria-errormessage is not pertinent, authors MUST either ensure
  the content is not rendered or remove the aria-errormessage attribute
  or its value.

* User agents MUST NOT expose an element as a pertinent error message
  when the referencing element lacks aria-invalid set to true.


joanmarie:
Spec-view: https://rawgit.com/w3c/aria/issue-587/aria/aria.html#aria-errormessage

joanmarie:
@richschwer and @mcking65, please review. Once we have consensus on the wording, I'll ping Mozilla and Google (since they've already implemented support) to do the same. Thanks!

joanmarie:
Thanks Matt.  I've merged your changes into this branch. Anything else needed?

mcking65:
@joanmarie, I gave it another read through; With the pull #590 merge, I think it is good to go.

joanmarie:
@asurkov and @minorninth: thoughts on this change? As far as user agents implementing support goes, it should boil down to the addition of a single sanity check:

> User agents MUST NOT expose aria-errormessage for an object when the object lacks aria-invalid set to true.

The benefit is that every single assistive technology doesn't have to do that same sanity check: If an error message is exposed by the user agent, by definition it is relevant/pertinent to the user.

minorninth:
Sounds good to me except that I think any non-false value for aria-invalid should be counted as "true", and  as currently worded it's a bit unclear what to do with aria-invalid=spelling or grammar or any future allowed enum value.

joanmarie:
> Sounds good to me except that I think any non-false value for aria-invalid should be counted as "true", and as currently worded it's a bit unclear what to do with aria-invalid=spelling or grammar or any future allowed enum value.

Argh. Good point. Thanks!!

joanmarie:
New language is:

> User agents MUST NOT expose `aria-errormessage` for an object with an `aria-invalid` value of `false`.

asurkov:
The behavior may be obscure for the authors or testing, when they have aria-invalid=false on an element for any reason, otherwise the change looks reasonable, especially if it makes AT life easier.

mcking65:
@joanmarie, I agree with the recent change to ensure that all ways of creating  a false value for aria-invalid are covered.

joanmarie:
The working group concluded that this could wait until 1.2. I'll merge after we branch.

PR #607, opened 2017-07-12
Modify 2D for clarity/alignment with HTML AAM.

richschwer:
step 2A we addressed earlier for clarity. You will note that I added text (as defined by the host language) in addressing this as elements like label have caviats in HTML AAM that say how to string them together. We really want the host language to define the details of how to handle these native host language attributes and elements as they will constantly be a moving target.


joanmarie:
Fine with me.

PR #610, opened 2017-07-21
Add aria-current mapping for values not in allowed list

richschwer:
addresses gap in core-aam for values that are not in the allowed list.


richschwer:
I made the change.

PR #611, opened 2017-07-21
modify 2F to process no role elements with content

richschwer:
Waiting for review by Steve Faulkner, Joseph, and Joanie.
https://www.w3.org/WAI/ARIA/track/issues/748

### DIFF
--- a/accname-aam/accname-aam.html
+++ b/accname-aam/accname-aam.html
@@ -349,7 +349,7 @@
               </details></div>
             </li>
             <li id="step2G">Otherwise, if the <code>current node</code> is a <a class="termref">Text node</a>, return its textual contents.</li>
-            <li id="step2H">Otherwise, if the <code>current node</code> is a descendant of an element whose <a class="termref">Accessible Name</a> or <a class="termref">Accessible Description</a> is being computed proceed to 2F.i.
+            <li id="step2H">Otherwise, if the <code>current node</code> is a descendant of an element whose <a class="termref">Accessible Name</a> or <a class="termref">Accessible Description</a> is being computed, and contains desendants, proceed to 2F.i.
 	    </li>
             <li id="step2I">Otherwise, if the <code>current node</code> has a <a class="termref">Tooltip attribute</a>, return its value. 
               <div><details>
--- a/accname-aam/accname-aam.html
+++ b/accname-aam/accname-aam.html
@@ -323,7 +323,7 @@
                 <pre class="example highlight"><code>&lt;div role="checkbox" aria-checked="false"&gt;Flash the screen &lt;span role="textbox" aria-multiline="false"&gt; 5 &lt;/span&gt; times&lt;/div&gt;</code></pre>
               </details></div>
               </li>
-            <li id="step2F">Otherwise, if the <code>current node's</code> <a class="termref">role</a> allows <a class="specref" href="#namefromcontent">name from content</a>, or if the current node has no defined role and contains content, or if the <code>current node</code> is referenced by <code>aria-labelledby</code>, <code>aria-describedby</code>, or is a native host language text alternative <a class="termref">element</a> (e.g. <code>label</code> in HTML):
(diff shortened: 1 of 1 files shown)


accname-aam/accname-aam.html
```diff
@@ -323,7 +323,7 @@
                 <pre class="example highlight"><code>&lt;div role="checkbox" aria-checked="false"&gt;Flash the screen &lt;span role="textbox" aria-multiline="false"&gt; 5 &lt;/span&gt; times&lt;/div&gt;</code></pre>
               </details></div>
               </li>
-            <li id="step2F">Otherwise, if the <code>current node's</code> <a class="termref">role</a> allows <a class="specref" href="#namefromcontent">name from content</a>, or if the <code>current node</code> is referenced by <code>aria-labelledby</code>, <code>aria-describedby</code>, or is a native host language text alternative <a class="termref">element</a> (e.g. <code>label</code> in HTML):
+            <li id="step2F">Otherwise, if the <code>current node's</code> <a class="termref">role</a> allows <a class="specref" href="#namefromcontent">name from content</a>, or if the current node has no defined role and contains content, or if the <code>current node</code> is referenced by <code>aria-labelledby</code>, <code>aria-describedby</code>, or is a native host language text alternative <a class="termref">element</a> (e.g. <code>label</code> in HTML):
```
klown: @richschwer 
The original text has three conditions, such that if any one condition holds, then the content of the node is used.  The conditions are:
- role allows name from content,
- node is referenced by `aria-labelledby` or `aria-describedby`,
- node is a native labelling element (e.g. html `<label>`)

The second condition almost does the job.  If the `<em>` node were referenced via `aria-labelledby`, then its contents would be used.  The difference is that in Cynthia's example the `<em>` is a child of a `<label>` -- it's implicitly "referenced" via a `<label>` element.  I suggest extending the second condition to include native host language labelling elements:

- node is referenced by `aria-labelledby` or `aria-describedby`, or is a descendant node of a native host language text alternative element (e.g., `<label>` in HTML).

Technically, how various descendant html elements contribute within a `<label>` element should be specified in the html-aam or the html spec itself, if it currently isn't.

richschwer: @joanie I don't think it does have the same problem as the HTML spec. does not call out <em>'s for name computation. Also <em> does not have an implicit role so it simply addresses the skipping of <em> in a recursive name computation. That said I think Joseph's wording is much tighter and explicit but I modified to it to include the text alternative as well as it was before. 

It now reads: Otherwise, if the current node's role allows name from content, or if the current node is referenced by aria-labelledby, aria-describedby, or is a native host language text alternative element (e.g. label in HTML), or is a descendant of a native host language text alternative element:

richschwer:
I modified a version of Joseph's last post to also include the text alternative element. I think we have this addressed now. please let me know if you agree. If so, pull it.

joanmarie:
At this moment, I don't have the time to do a thorough walk through using various test cases. However, from an admittedly extremely cursory glance, will this definitely solve the original problem? In particular, when the current node is the button, you have a role that supports name. But when you then build up that name from the content of child nodes, you have an `<em>`. That node does not support name from content, nor is it a text node or a tooltip. Doesn't the former condition prevent us from going back to the start of step 2 and descending the em? Doesn't the latter condition prevent us from getting the text under the em element?

I see now that you added text to 2F to prevent things like divs getting names, which is great. But as I _believe_ I stated somewhere (email? irc?) I think that prevention may need to be at the top of the algorithm rather than at 2F.

If I'm wrong, then please give me more time to do a proper review. I just didn't want to wait too long before responding.

richschwer:
That is why I had the previous change which you did not like. Let me did further to fill the other hole.

richschwer:
I am looking at the text and I there are far too many special cases that need to be addressed earlier in the name computation to move to an earlier step. For example, you have to address the role "none" and "presentational" and hidden text first. You also need to give precedence to aria-labelledby and aria-description as these must be processed as part of name and description first. What needs to happen is to continue with the name computation if you are a descendant of an element whose name is being computed and you will need to continue through its descendant children. This points to step 2F which appears to be the catch-all for embedded content.

joanmarie:
I've manually gone through the scenarios we've been discussing and the changes seem to work without introducing unintended side effects. So I'm fine with you merging it.

richschwer:
ok. merging

Issue #457, opened 2016-11-02
CORE-AAM 1.1: Add mappings for aria-errormessage on macOS


richschwer:
Moved from this action in tracker: https://www.w3.org/WAI/ARIA/track/actions/2119

cookiecrook:
https://bugs.webkit.org/show_bug.cgi?id=165190

joanmarie:
Whether or not it's implemented in WebKit/Safari, we need mappings for this ARIA feature stated in the Core AAM. Hence the summary/title change.

joanmarie:
@cookiecrook: As I mentioned in pull request #620, we really need to wrap up these remaining issues. If you could provide the language we should add to the Core AAM for macOS, that would be great. Thanks!

cookiecrook:
I don't have the answers at the moment. I'd like to cast my vote (again) that the AAM documents be Living standards rather than Rec track.

joanmarie:
> I'd like to cast my vote (again) that the AAM documents be Living standards rather than Rec track.

That -- or at least something closer to that -- has my vote too: Platform accessibility APIs change at a different rate (often a much more timely rate) than W3C standards. Thus the "correct" mappings (as far as the W3C is concerned) are in danger of becoming incorrect as far as platform accessibility APIs and ATs are concerned.

But **_with respect to the 1.1 cycle_**, I'm afraid going the living standard route is not possible. Not only is Core AAM 1.1 behind with respect to the milestones established for it, but ARIA 1.1 currently depends on Core AAM 1.1. And because of that, the ARIA spec's progress along the REC track is blocked.

Therefore, anything you/Apple can do to help us get Core AAM 1.1 moving on its current REC track would be greatly appreciated. During the 1.2 cycle let's see what we can do to get the mapping specs closer to being living standards -- along with trying to decouple ARIA from the Core AAM insofar as the REC track is concerned so that we don't find ourselves in this sort of bind again.

cookiecrook:
We are discussing a new property and notification name internally, but I don't have an answer for you yet. I'll confirm in the next few days if there are no objections.

cookiecrook:
Resolved by latest commit to PR #620

PR #620, opened 2017-08-02
Core-AAM: Add macOS mapping for aria-errormessage

joanmarie:
Expose the textual content of the referenced element as the value of
AXHelp.

Fixes github issue #457.

### DIFF
--- a/core-aam/core-aam.html
+++ b/core-aam/core-aam.html
@@ -2047,7 +2047,7 @@ var mappingTableLabels = {
                   <td>
 			<p>If the referenced object is in the accessibility tree, expose a pointer to the accessible object using <code>RELATION_ERROR_MESSAGE</code>, and expose reverse relations as described in <a href="#mapping_additional_relations">Relations</a>.</p>
 		  </td>
-                  <td>Expose the textual content of the referenced element as the value of <code>AXHelp</code>.</td>
+                  <td>Expose the textual content of the referenced element as the text alternative value of <code>AXValidationError</code>.</td>
 </td>
                 </tr>
                 <tr id="ariaExpandedTrue">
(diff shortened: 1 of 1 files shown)


joanmarie:
@cookiecrook: Please review. Thanks!

cookiecrook:
We usually don't recommend putting any critical information in AXHelp, since a lot of users will disable VoiceOver speaking help text until a command is issued. I think our options are:

- Append it to the label,
- Trigger an announcement at the time it is shown, or
- Expose it as a new TBD API.

stes-acc:
>>>>  … since a lot of users will disable VoiceOver speaking help text until a command is issued.

Is there a specific key or/and gesture to speak AXHelp on request for a given control?


-          Stefan

joanmarie:
> We usually don't recommend putting any critical information in AXHelp, since a lot of users will disable VoiceOver speaking help text until a command is issued. I think our options are:
> * Append it to the label,
> * Trigger an announcement at the time it is shown, or
> * Expose it as a new TBD API.

Fair enough. Thanks for the review!

Unfortunately, the lack of mapping is one of the things blocking the Core AAM from being able to enter CR. We **_really, really, really_** need to know what your platform's mapping is. Patches not only "welcome", but strongly encouraged. :smile:

cookiecrook:
@stes-acc wrote:

> Is there a specific key or/and gesture to speak AXHelp on request for a given control?

VO+Cmd+H will speak it on demand. Optionally, but by default, it will be spoken after a user-adjustable-length pause.

joanmarie:
> Works for me, but I added the last commits, so maybe you should review.

Thanks! What you did works for me. But I couldn't be bothered to figure out how to use github's UI to change authorship as part of a squash and merge, so I just did it the old-fashioned way. :smile:. a314c09.

Thanks again!

Issue #303, opened 2016-03-16
CORE-AAM: cell mapping specifies TableItem as UIA Control Type, but is a Control Pattern


jasonkiss:
UIA Control Type for role=cell is TableItem, but that is a Control Pattern, not a Control Type. See http://rawgit.com/w3c/aria/master/core-aam/core-aam.html#role-map-cell

Should it be DataItem ControlType?

joanmarie:
@boggydigital: Your platform. :)

If you could do a pull request providing any needed changes (or comment if no changes are needed), that would be great. Thanks in advance!

boggydigital:
https://github.com/w3c/aria/pull/623

joanmarie:
This issue was fixed through #623 and #638. As a result, https://rawgit.com/w3c/aria/master/core-aam/core-aam.html#role-map-cell lists the UIA mapping to the `DataItem` Control Type and `TableItem` Control Pattern.

PR #623, opened 2017-08-08
Update UIA mapping for role=cell

boggydigital:
As pointed out in https://github.com/w3c/aria/issues/303 TableItem is ControlPattern, not Control Type.

Additionally:
- Removed SelectionPattern requirement - it should not be required for this role.
- Removed MUST NOT implement InvokePattern. InvokePattern in UIA can be implemented for many reasons, including ancestry tree element that has click handlers. Not only it would be hard to guarantee lack of InvokePattern - this would be counterintuitive and not helpful for the users

### DIFF
--- a/core-aam/core-aam.html
+++ b/core-aam/core-aam.html
@@ -661,9 +661,7 @@ var mappingTableLabels = {
 					</td>
                     <td>
                       <ul>
-                        <li>Control Type is <code>TableItem</code></li>
-                        <li>Supports the <code>selection</code> pattern</li>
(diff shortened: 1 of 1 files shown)


joanmarie:
> As pointed out in #303 TableItem is ControlPattern, not Control Type.

Sanity check: The `cell` role should NOT support the TableItem pattern?

boggydigital:
Appreciate the check - and I believe you're right. Created PR to address: https://github.com/w3c/aria/pull/638 and added why this would be useful in the description.

PR #624, opened 2017-08-08
Update UIA mapping for role=alert

boggydigital:
CT=Group and LCT=alert to be consistent with other live regions (log, status)


boggydigital:
Forgot to add LiveSettingProperty value (Assertive) for consistency as well

PR #632, opened 2017-08-17
Add LCT for cell, gridcell, row

boggydigital:
Issue raised in the mail. While it's possible to distinguish between three based on additional Patterns and behaviors - LCT can greatly simplify this.


joanmarie:
Thanks! Merged: 9b454ff.

PR #105, opened 2015-11-09
Add alert example

MichielBijl:
Add an alert example to the practices.

PR #633, opened 2017-08-17
Remove ValuePattern from link role

boggydigital:
To be moved to HTML-AAM. As rightfully pointed out offline - ARIA doesn't have href attribute, so the existing mapping is not really correct...

### DIFF
--- a/core-aam/core-aam.html
+++ b/core-aam/core-aam.html
@@ -1118,9 +1118,7 @@ var mappingTableLabels = {
 						Interface: <code>IAccessibleHypertext</code>
 					</td>
 					<td>
-						Control Type: <code>HyperLink</code><br />
-						Control Pattern: <code>Value</code><br />
-						Property: Set <code>Value.Value</code> to the <code>href</code> attribute
+						Control Type: <code>HyperLink</code>
 					</td>
 					<td>
 						Role: <code>ROLE_LINK</code><br />


joanmarie:
I commented on the https://github.com/w3c/html-aam/pull/103. Here's what I stated for the sake of convenience:

> FWIW: The issue being addressed in this PR isn't limited to UIA. ATK and IA2 also expose the URI for "real" links (e.g. the `a` element with an `href` attribute), and cannot do so for ARIA links until we have role and property parity. :slightly_frowning_face: Thus if the conclusion of the Editors is that the HTML AAM shouldn't just point to the ARIA mappings for the `a` element, I think it makes sense to make that change across the board.
>
> That said, the HTML AAM seems to already have [mappings for exposure of the `href` attribute.](https://w3c.github.io/html-aam/#att-href). Those mappings just seem to have gotten a tad on the broken side. But once that's fixed (see [#105](https://github.com/w3c/html-aam/pull/105)), you'll have the mapping you need for the href attribute (Currently verifiable here: https://joanmarie.github.io/html-aam/#att-href).
>
> Long way of saying the following: One approach is this pull request. Another approach would be to leave the HTML AAM as-is (i.e. pointing to the ARIA mappings) and just remove the mention of the `href` attribute exposure from the Core AAM.

boggydigital:
@joanmarie , yes you're right, I completely missed the attribute mapping - just removing property line in Core-AAM should be sufficient and the right thing to do.

joanmarie:
> just removing property line in Core-AAM should be sufficient and the right thing to do.

Okie dokie. Thanks! Done in 95dea19.

PR #634, opened 2017-08-17
Change Range CT to RangeValue CT

boggydigital:
Pointed out offline - there is no Range Control Pattern, only RangeValue: https://docs.microsoft.com/en-us/dotnet/api/system.windows.automation.rangevaluepattern?view=netframework-4.7


joanmarie:
Thanks! Did another manual merge + combining of related pull requests: fa97996.

PR #635, opened 2017-08-17
Add RangeValue CP to separator (focusable)

boggydigital:
Pointed out offline:

The focusable separator, which in ARIA supports the aria-value* attributes and works like a window or pane splitter

It makes sense to add RangeValue Control Pattern.


joanmarie:
Thanks! Did another manual merge + combining of related pull requests: fa97996.

PR #636, opened 2017-08-17
Remove SelectionPattern for role=list

boggydigital:
Per offline discussion:

The list role mentions that the Selection pattern should be
   supported "When the list contains listitems". But the definition
   of the ARIA list role is "A section containing listitem elements."
   Those listitem elements can be immediate descendants or owned
   elements of the list, or they can be inside a group which is an
   immediate descendant or owned element of the list. But it is
   expected that the list will contain listitems.


joanmarie:
Thanks! I wound up merging this manually along with #637 because the conclusion from the combined pull requests seems to be the SelectionPattern was just on the wrong role. (ea3828a)

PR #637, opened 2017-08-17
Add Selection CP to listbox widget role

boggydigital:
Per offline discussion;

unlike the ARIA list role (which is not a widget), the listbox is a widget which supports selection


joanmarie:
Thanks! I wound up merging this manually along with #636 because the conclusion from the combined pull requests seems to be the SelectionPattern was just on the wrong role. (ea3828a)

PR #638, opened 2017-08-17
Add TableItem to role=cell mapping

boggydigital:
Please see https://github.com/w3c/aria/pull/623 - there is a valid point that it would be useful. Reading https://w3c.github.io/aria/aria/aria.html#cell in order to support states - we'll need TableItem for UIA


joanmarie:
Thanks. Merged in 944d104.

PR #639, opened 2017-08-18
Core-AAM: Remove LocalizedLandmarkType where there is semantic Landma…

boggydigital:
…rkType

This is redundant and not required for implementers Localized {Landmark, Control} Types are supposed to provide additional information, not duplicate the same available in {Landmark, Control} Type


joanmarie:
Thanks! Merged in c9732ef.

PR #640, opened 2017-08-20
removed editorial note relating to issue522

richschwer:


richschwer:
this is just cleanup. so I am squashing and merging.

Issue #492, opened 2016-12-02
[core-aam] Rules for excluding elements from the accessibility tree confuse definition of presentational elements


mcking65:
Section [5.1.1 Excluding Elements from the Accessibility Tree](http://w3c.github.io/aria/core-aam/core-aam.html#exclude_elements2) 
does not distinguish between hiding an element and hiding an element and its content.

While the first rule references the definition of the presentation role, it still states that the element is not exposed using the same language as the second rule for hidden content. That is, the rules seem to equate the treatment of presentational elements with the treatment of hidden elements.

Similarly, the description of user agent treatment for aria-hidden does not have any clear differences from the description of user agent treatment of elements that have children presentational true.

joanmarie:
Text designed to clarify issues related to presentational elements has been added to the ARIA spec and referenced in Section 5.1.1 of the Core AAM. In particular, the first rule/item now has the following content related to presentation-role elements:

> However, their exclusion is conditional and depends on other factors. In addition, the element's descendants and text content are generally included. These exceptions and conditions are documented in the section "[Presentational Roles Conflict Resolution](https://w3c.github.io/aria/aria/aria.html#conflict_resolution_presentation_none)" in Accessible Rich Internet Applications (WAI-ARIA) 1.1.

@mcking65: Does the above content from the Core AAM, combined with the new "Presentation Roles Conflict Resolution" content which was added to the ARIA spec, solve the confusion you reported?

mcking65:
It is closer for role presentation, but not for children presentational.

For role presentation, the added text, with some editorial improvement, could resolve this issue.

In particular, the "Presentational Roles Conflict Resolution" subsection in ARIA needs some editorial work to correct grammar errors and make it understandable. 

For instance, missing " of " in:

> There are a number ways presentational role conflicts are resolved.

And this paragraph needs a rewrite to make it intelligible:

> Host languages elements, having implicit presentational roles for which no roles, may be applied, must never be exposed to in the accessibility tree. With this exception, user agents must always expose global WAI-ARIA states and properties to accessibility APIs. In this case, the user agent ignores the presentation role and exposes the element according to its implicit native semantics. However, user agents must ignore any non-global, role-specific WAI-ARIA states and properties, unless it is on an inherited presentational role where an explicit role is applied.

The only way I came to understand it was because of the subsequent example.

WRT children presentational, it is still written as if children presentational true is the same as aria-hidden true. It does not say that the text nodes of elements with children presentational true are included in the accessibility tree.

[...]

PR #647, opened 2017-09-15
Core-AAM Issue 492: Clarify AX Tree Exclusion for Children Presentational True

mcking65:
For issue #492, modified section "5.1.1 Excluding Elements from the Accessibility Tree" in core-aam.html.

Clarified the criteria for elements with children presentational true to specify that:
1. Text content of excluded elements is included and;
2. Exceptions, such as focusability, also apply.


mcking65:
Not liking the details after the list, but breaking into multiple paragraphs seems like it might help. Does commit 7612372 
 work for you?

joanmarie:
I dig it. Merged. Thanks much!!

PR #661, opened 2017-11-10
Graphics AAM: Fix ATK mappings

joanmarie:
* Whether or not ROLE_DOCUMENT_FRAME exposes STATE_EDITABLE has nothing to
  to with the fact that it is a graphics-document. Removing this statement
  is also consistent with what is in the Core AAM for the document role:
  https://rawgit.com/w3c/aria/master/core-aam/core-aam.html#role-map-document

* The mapping of graphics-object to ROLE_GROUP is wrong because there is no
  such role in ATK. There is ROLE_GROUPING. However, for these sorts of
  containers, the preferred mapping is to ROLE_PANEL. ROLE_PANEL is the
  verified mapping of the ARIA group role (which is also how the SVG g
  element is exposed when inclusion in the accessibility tree is needed:
  https://rawgit.com/w3c/aria/master/core-aam/core-aam.html#role-map-group
  https://rawgit.com/w3c/aria/master/svg-aam/svg-aam.html#role-map-g

### DIFF
--- a/graphics-aam/graphics-aam.html
+++ b/graphics-aam/graphics-aam.html
@@ -358,7 +358,7 @@ var mappingTableLabels = {
                                                                     <li>Control type/role is <code>'Document'</code>.</li>
                                                                   </ul>
                                                                 </td>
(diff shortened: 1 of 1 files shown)


joanmarie:
@boggydigital: FYI.

joanmarie:
Thanks @AmeliaBR! I'll give some thought to the Graphics Document ATK role (I develop the platform's screen reader which will be consuming the mapping and role) and talk to my colleague (the ATK maintainer).

PR #662, opened 2017-11-10
Graphics AAM: Fix IAccessible2 mapping typo

joanmarie:
xml-roles:graphics-object; not xml-roles:graphics-obj


joanmarie:
@boggydigital: FYI

PR #664, opened 2017-11-10
Update UIA mappings for Graphics AAM

boggydigital:
- Removed AriaRole reference
- Mapped to sane Control Types: 
-- graphics-document -> Document
-- graphics-object -> Group
-- graphics-symbol -> Image

Localized Control Types are not mapped though those roles and I anticipate would come from SVG AAM, not here to preserve fidelity.


boggydigital:
@AmeliaBR Yes, AriaRole gets mapped automatically, including arbitrary strings. For example <div role="imprettysurethisroledoesntexist"> would produce AriaRole="imprettysurethisroledoesntexist"

Re LCT: 

There is no clear goal here and this shouldn't be blocking for this portion of the mapping - there would be no impact to graphics-aam either way. Pragmatically I would prefer to have a way to distinguish between different source SVG elements that all would be mapped to the same role, so there might be some fidelity loss - but it's currently not blocking any clients and there might be other solutions to this theoretical problem.

During TPAC we chatted very briefly with @joanmarie and there is a lot of value in the current clean mapping through the roles without additional information. We'll keep both options under consideration for UIA and I'll follow up with internal teams at Microsoft that might have opinions. 

As mentioned this shouldn't be blocking as this PR doesn't introduce any SVG-AAM changes, only implies we might need them in the future.

joanmarie:
I manually merged in the changes (commit 75e3e9f).

PR #666, opened 2017-11-10
Graphics AAM: Fix macOS mapping for graphics-symbol

joanmarie:
The graphics-symbol role functions as an image. Therefore it should be
exposed as an AXImage; not an AXGroup.

Also remove "?" appearing before graphics-document subrole.


joanmarie:
@boggydigital: fyi

cookiecrook:
> More generally (and specifically to James): is there any way to pass the "graphics" semantics to the API, while still allowing the element to have accessible children? 

It's possible to do that already in the API (AXImage can have children), but IIRC, it would require a VO update to have VO acknowledge those children.

PR #674, opened 2017-12-05
update SVG2 to CR

innovimax:


joanmarie:
Thanks for the fix. However, in this particular case I see no reason (editorial or technical) to not use specref. So I removed it. I believe that achieves the desired effect. If not, however, please let me know.

Thanks again!

innovimax:
Even better !

PR #675, opened 2017-12-05
update to xmlschema 1.1 instead of 1.0

innovimax:


joanmarie:
Thanks for this pull request. However from a quick search, it looks like there is another reference instance which needs to be updated in index.html as well. Plus it looks like there are even more (in the form of non-ref URLs) in there pointing to the 1.0 version. Bah.

If you have the spare cycles to sort out which references (or if it's all of them) need to be updated and fix the ones in need of fixing, that would be super. Otherwise, I'd prefer to keep this open (and unmerged) as a note to myself that it needs doing.

Thanks again either way!

innovimax:
I think I updated all of them.

joanmarie:
Works for me. Thanks!

PR #677, opened 2017-12-06
use specref version of EPUB-SSV

innovimax:


innovimax:
http://www.specref.org/?q=epub-ssv

joanmarie:
LGTM. Thanks!

PR #680, opened 2017-12-06
ARIA 1.2: Add explicit guidance for non-pertinent errormessage elemen…

joanmarie:
…ts (#587) (#589)

* ARIA: Add explicit guidance for non-pertinent errormessage elements (#587)

* When aria-errormessage is not pertinent, authors MUST either ensure
  the content is not rendered or remove the aria-errormessage attribute
  or its value.

* User agents MUST NOT expose an element as a pertinent error message
  when the referencing element lacks aria-invalid set to true.

* Proposed editorial revisions to proposal in pull #589 for issue #587. (#590)

* Treat any non-false value for aria-invalid as "true" for aria-errormessage

Otherwise, it is not clear what user agents should do with aria-invalid
set to spelling, grammar, or any future allowed enum value.


joanmarie:
This had to be manually merged due to repo split and other changes made to the document. 9979f86c

PR #761, opened 2018-05-17
Add blockquote, caption, and paragraph roles

joanmarie:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/761.html" title="Last updated on May 31, 2018, 5:50 PM GMT (1a78056)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/761/d037626...1a78056.html" title="Last updated on May 31, 2018, 5:50 PM GMT (1a78056)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -612,6 +612,8 @@
 			<ul>
 				<li><rref>application</rref></li>
 				<li><rref>article</rref></li>
+				<li><rref>blockquote</rref></li>
+				<li><rref>caption</rref></li>
 				<li><rref>cell</rref></li>
 				<li><rref>columnheader</rref></li>
 				<li><rref>definition</rref></li>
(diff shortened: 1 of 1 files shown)


prlbr:
I wonder whether it would be better to introduce a role `quotation` instead of `blockquote`, because one can also quote from another resource inline.

joanmarie:
> I wonder whether it would be better to introduce a role quotation instead of blockquote, because one can also quote from another resource inline.

We didn't think so. The `q` element is not something screen readers would likely want to provide navigation to using the same functionality as a blockquote. Ditto for providing context. Thus the `q` element is slated for the generic inline role with a TBD attribute. https://github.com/w3c/aria/wiki/Plans-regarding-role-parity

PR #774, opened 2018-05-26
Tristate value type: the default value should be "undefined"

### LINKED ISSUE #773: Tristate value type: the default value should be "undefined"
The definition of tristate attribute value type states:
> The default value for this value type is _false_ unless otherwise specified.

There are only two states of this type: the "aria-checked" state and the "aria-pressed" state. Both of them have "undefined" as a default value.
The tristate type is like applicable (true/false/undefined) type, but with additional "mixed" value. Applicable type's default is "undefined".
So I suggest to change that confusing definition, which must be:

"The default value for this value type is undefined unless otherwise specified.

aristov:
Closes #773.

***
<a href="https://pr-preview.s3.amazonaws.com/aristov/aria/pull/774.html" title="Last updated on May 26, 2018, 3:33 PM GMT (1f44051)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/774/c040ddc...aristov:1f44051.html" title="Last updated on May 26, 2018, 3:33 PM GMT (1f44051)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -8111,7 +8111,7 @@
 				<dt id="valuetype_true-false">true/false</dt>
 				<dd>Value representing either true or false. The default value for this value type is <code>false</code> unless otherwise specified.</dd>
 				<dt id="valuetype_tristate">tristate</dt>
-				<dd>Value representing true or false, with an intermediate "mixed" value. The default value for this value type is <code>false</code> unless otherwise specified.</dd>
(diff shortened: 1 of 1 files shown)


jnurthen:
Marked as non substantive for IPR from ash-nazg.

Issue #700, opened 2018-02-05
Should aria-selected be listed as required for option role? 


mcking65:
Originally posted by @kbae00 in w3c/aria-practices#594.

Currently aria-selected is listed in the Characteristics table as a required attribute for the option role: https://www.w3.org/TR/wai-aria-1.1/#option. But the description says "Elements with the role option have an implicit aria-selected value of false." Should aria-selected be listed as a supported attribute instead of required?

mcking65:
This is a good question. Section [5.2.2 Required States and Properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) says:

> Content authors MUST provide a non-empty value for required states and properties. Content authors MUST NOT use the value undefined for required states and properties, unless undefined is an explicitly-supported value of that state or property.

And, it also says:

> NOTE: A host language attribute with the appropriate implicit WAI-ARIA semantic fulfills this requirement.

Interesting that native attributes meet the requirement but the implied semantics associated with an ARIA role do not.
If the ARIA role has a default implied value, why would authors have to duplicate that implied value explicitly if the implied value is the correct value for the element?

We have this same type of requirement on other roles, aria-checked for example.

I think we should leave the role definitions the same but change the authoring requirement in section 5.2.2. Instead it should say:

> Content authors MUST provide a non-empty value for required states and properties except in conditions where the role that requires the state or property provides the appropriate value by default. Content authors MUST NOT use the value undefined for required states and properties, unless undefined is an explicitly-supported value of that state or property.

mcking65:
I have changed my view on how to fix this. I now think that we need to change aria-selected to a supported property for option.

There are two reasons. First, aria-selected includes language that saves authors from having to declare aria-selected=false on selectable elements in single-select widgets. The following from the aria-selected spec is baked into user agents:

>This attribute is used with single-selection and multiple-selection widgets:
>1. Single-selection containers where the currently focused item is not selected. The selection normally follows the focus, and is managed by the user agent.
>2. Multiple-selection containers. Authors SHOULD ensure that any selectable descendant of a container in which the aria-multiselectable attribute is true specifies a value of either true or false for the aria-selected attribute.
>
[...]

PR #799, opened 2018-08-15
Option role: Change aria-selected from required to supported and remove default value

mcking65:
For issue #700, made the following changes to the option role section in index.html:

1. Remove sentence from description that states there is an default value of false fore aria-selected on option elements.
2. Remove aria-selected from required states and properties row of attributes table.
3. Add aria-selected to supported row of attributes table.
4. Remove default value for aria-selected from implicit values row of attributes table.

Note: This cascades to treeitem so that aria-selected is not required on treeitem.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/799.html" title="Last updated on Oct 11, 2020, 2:19 AM UTC (de84a1f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/799/ff8cd29...de84a1f.html" title="Last updated on Oct 11, 2020, 2:19 AM UTC (de84a1f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4679,7 +4679,6 @@
 			<div class="role-description">
 				<p>A selectable item in a <rref>select</rref> list.</p>
 				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or owned by, an element with the <a>role</a>  <rref>listbox</rref> or <rref>group</rref>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
-				<p>Elements with the role <code>option</code> have an implicit <sref>aria-selected</sref> value of <code>false</code>.</p>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
@@ -4730,11 +4729,7 @@
 					</tr>
 					<tr>
 						<th class="role-required-properties-head">Required States and Properties:</th>
-						<td class="role-required-properties">
-							<ul>
-								<li><sref>aria-selected</sref></li>
-							</ul>
-						</td>
+						<td class="role-required-properties"></td>
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
@@ -4743,6 +4738,7 @@
 								<li><sref>aria-checked</sref></li>
 								<li><pref>aria-posinset</pref></li>
 								<li><pref>aria-setsize</pref></li>
+                <li><sref>aria-selected</sref></li>
 							</ul>
 						</td>
 					</tr>
@@ -4777,7 +4773,7 @@
 					</tr>
 					<tr>
 						<th class="implicit-values-head" scope="row">Implicit Value for Role:</th>
-						<td class="implicit-values">Default for <sref>aria-selected</sref> is <code class="default">false</code>.</td>
+						<td class="implicit-values"></td>
 					</tr>
 				</tbody>
 			</table>
--- a/index.html
+++ b/index.html
@@ -1,33 +1,24 @@
-<!--?xml version='1.0' encoding='UTF-8'?-->
 <!DOCTYPE html>
 <html xmlns="http://www.w3.org/1999/xhtml" lang="en-US" xml:lang="en-US">
 <head>
 <title>Accessible Rich Internet Applications (WAI-ARIA) 1.2</title>
 <meta content="text/html; charset=utf-8" http-equiv="Content-Type"/>
 <script src="https://www.w3.org/Tools/respec/respec-w3c-common" class="remove"></script>
-	<script src="common/biblio.js" class="remove"></script>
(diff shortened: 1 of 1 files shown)


aleventhal:
The way I always understood things were that containers such as listboxes had modes, single selection and multiple selection.
From Core AAM: "In the single selection case, selection follows focus (see the section "Focus States and Events Table" for information about focus events)..."

What this means is that if the author doesn't explicitly put aria-selected="false" on the currently focused item in a single selection container, it will be treated as selected by default. This mimics what you see in native widgets.

asurkov:
I think implicit value of aria-selected for role="option" makes sense, since it indicates a presence of selectable state on the element; in contrast to to other non-selectable-by-default roles like  role="columnheader".

Agreed, implicit aria-selected="false" sounds confusing in conjunction with "selection follows the focus" case, but probably the sentence should be reworded, something like "role option has an implict aria-selected value of 'false', until it is implicitly selected based on focus (link), in which case it defaults to 'true'.

melanierichards:
+1 to @asurkov's comment:

> I think implicit value of aria-selected for role="option" makes sense, since it indicates a presence of selectable state on the element; in contrast to to other non-selectable-by-default roles like role="columnheader".
> Agreed, implicit aria-selected="false" sounds confusing in conjunction with "selection follows the focus" case, but probably the sentence should be reworded, something like "role option has an implict aria-selected value of 'false', until it is implicitly selected based on focus (link), in which case it defaults to 'true'.

jnurthen:
@mcking65 This PR has been around for a while. Are there any of these comments that need actioning?

JAWS-test:
I argue that `aria-selected` is mandatory, at least for ARIA roles, which by default are intended to be selected like `option` and `tab` (see https://github.com/w3c/aria/issues/1094)

mcking65:
@melanierichards, @asurkov, @aleventhal, we need to account for the situations where authors are using aria-checked instead of selected. I'm not sure what the best way to do this is. I think we should discuss. There is not time to do so for ARIA 1.2.

We also need to discuss whether ARIA needs to support patterns that allow authors to use both checked and selected.

melanierichards:
cc @boggydigital

joanmarie:
@jcsteh: We'd like your opinion on this when you have a moment. Thanks!

jcsteh:
Oo. Quite a conundrum.

Talking only about selection, I agree with @asurkov's https://github.com/w3c/aria/pull/799#issuecomment-577814755. Saying there's no default value is misleading because in a single select widget, options without aria-selected get exposed as selected. In a multi select widget, options without aria-selected get exposed as not selected. ("Not selected" and "doesn't support selection" are two different things.) Doing otherwise is a backwards compat break. So, we should be clear that the default is different depending on single/multiple selection.

On the other hand, the fact that option/treeitem support aria-checked complicates things. It (probably?) doesn't make sense for something to be both selectable and checkable, since they both presumably convey the same meaning, just with slightly different presentation.

My suggestion:

1. Unless aria-checked is specified, aria-selected should default to true on a single select widget, false on a multi select widget.
2. If aria-checked is specified, aria-selected has no "default". Open mapping question: should platforms which have the selectable state (or equivalent) stop exposing it in this case? I think this is appropriate.

sinabahram:
Regarding selection and checked not being different. I want to make sure this accounts for the use case where noncontiguous multiple selection is available inside a tree view whose components can also be checked?

 

Imagine making blocks-based programming accessible via a tree view, and noncontiguous selection is available for group deletion, moving, etc., but items can be checked as well, for example, as part of controlling membership in a different set subject to different actions. I could also imagine this use case in certain digital audio workstation type scenarios, or other advanced applications in which more than one parallel track of actions is being controlled.

MarcoZehe:
> On the other hand, the fact that option/treeitem support aria-checked complicates things. It (probably?) doesn't make sense for something to be both selectable and checkable, since they both presumably convey the same meaning, just with slightly different presentation.

I am actually not sure about that. In a single-select tree view, items can be checkable, see certain types of Windows Explorer options or such that have checkable tree items. The Selected state moves along with the focused state, but is definitely there. That item can then be checked or not. Heck, some are even acting like radio buttons, where one of several can be checked at one time.

Where I agree is if a tree has multiple selectable items. Here, I feel like selectable and checkable are going hand in hand, although I currently don't remember ever seeing such a tree view with multiple selection myself. List boxes with option items, yes, and for that, this statement is true I guess as well, e. g. selection and checkmark go together in multi-select scenarios.

aleventhal:
@jcsteh do you mean the focused option in the single selection case defaults to selected? All of the unfocused options would default to not-selected.

I'm not sure why that rule would need to change due to the presence of checked or not. Honestly, we could probably even get rid of the selection follows focus rule without consequence. I'd be nervous to do so, but the reason it's in there is really historical. We were trying to make listbox/combobox widgets look like they did in MSAA.

sinabahram:
I think that selection should follow focus by default. To not do so violates a ton of assumptions/expectations.

mcking65:
Closing; replaced with #1340

PR #803, opened 2018-08-16
Add code fences to README

nschonni:


nschonni:
Marked as non substantive for IPR from ash-nazg.

Issue #812, opened 2018-08-31
Allowed structures for spinbutton widgets is ambiguous


mcking65:
The [ARIA spinbutton specification](https://www.w3.org/TR/wai-aria-1.1/#spinbutton) says: 

>Authors MAY create a spinbutton with children or owned elements, but MUST limit those elements to a textbox and/or two buttons.

It also says:

>When a spinbutton receives focus, authors SHOULD ensure focus is placed on the textbox element if one is present, and on the spinbutton itself otherwise. Authors SHOULD NOT include contained button elements in the primary navigation ring, e.g., the Tab ring in HTML, because they are superfluous for people using keyboard devices.

The above text is describing a composite structure like:

~~~
<div role="spinbutton">
  <button tabindex="-1">decrease</button>
  <input type="text">
  <button tabindex="-1">Increase</button>
</div>
~~~

Another structure supported by this spec language is:

~~~
<div role="spinbutton" tabindex="0">
  <span>Decrease</span>
  <span>text of value</span>
  <span>Increase</span>
</div>
~~~

However, the most common way of implementing a spinbutton widget that has a textinput has the increment and decrement buttons as siblings like:

~~~
<button tabindex="-1">decrease</button>
<input type="text" role="spinbutton">
<button tabindex="-1">Increase</button>
~~~

The description should more clearly permit building a spinbutton in this way.

This is related to APG issue w3c/aria-practices#826.

pkra:
Fixed by https://github.com/w3c/aria/pull/813

Issue #797, opened 2018-08-06
Make aria-valuenow supported instead of required for spinbutton role


mcking65:
The [spinbutton role](https://www.w3.org/TR/wai-aria-1.1/#spinbutton) requires aria-valuenow to be set. Not setting aria-valuenow is an author error, and  if not set, the implicit value of aria-valuenow is 0.

Section [5.2.2 Required States and Properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) says:

>Content authors MUST provide a non-empty value for required states and properties.

This means that a spinbutton with a textbox cannot be empty. Is this a reasonable limitation on spinbuttons? Does this mean that it is a requirement to disable delete and backspace on spinbuttons with text fields when there is only one character left in the field? Or, does it mean that pressing delete or backspace has to show a default value, e.g., 0.

Does this mean that it is not possible to make a spinbutton that is initially empty and has aria-required set to true in order to force the user to choose a value?

If it is reasonable to enable authors to make a spinbutton that can be empty, then it seems we need to remove the requirement to set aria-valuenow and instead change aria-valuenow to a supported property for the spinbutton role.

This approach would be supported by the [aria-valuenow description](https://www.w3.org/TR/wai-aria-1.1/#aria-valuenow), which says:

>If the current value is not known (for example, an indeterminate progress bar), the author SHOULD NOT set the aria-valuenow attribute. If the aria-valuenow attribute is absent, no information is implied about the current value.

jnurthen:
Agree that initial state is certainly a problem in the current spec. I'm in favour of making aria-valuenow not a required attribute. Indeed if someone specifies aria-valuetext this would be used instead of aria-valuenow and should certainly be allowed. requiring valuenow in this instance doesn't seem to add any value.

Issue #642, opened 2017-08-28
Inconsistent text for role=spinbutton


ZoeBijl:
The description for the [spinbutton role](https://www.w3.org/TR/wai-aria-1.1/#spinbutton) is a bit inconsistent. This piece of text says that the `aria-valuemin` and `aria-valuemax` attributes are not required.

> Authors must set the aria-valuenow attribute. Authors should set the aria-valuemin attribute when there is a minimum value, and the aria-valuemax attribute when there is a maximum value. If missing or not a number, the implicit values of these attributes are as follows:
>
> * The implicit value of aria-valuemin is that there is no minimum value.
> * The implicit value of aria-valuemax is that there is no maximum value.
> * The implicit value of aria-valuenow is 0.

The table that follows this text says:

> Characteristic | Value
> -- | --
> Required States and Properties: | aria-valuemax, aria-valuemin, aria-valuenow

My guess is that the table is geared towards implementers and not authors. Authors that read the table might get confused as to whether the listed attributes are required or not.

mcking65:
Required  attributes in the table are authoring requirements. The table and text need to be consistent.

joanmarie:
I'm open to suggestions. Some things to consider:

1. There are cases where there is (functionally) no minimum value and/or maximum value. If we change the authors SHOULD to an authors MUST, won't validators spit up when the author does not provide a value because the range is (functionally) boundless?
1. In cases where there is a known, bound minimum value, authors really need to provide it. Ditto for a known maximum value. To me "really need" means "required."
1. Some platform accessibility APIs expect (count on) a minimum and maximum value exposed by the user agent. So, again, `aria-valuemax` and `aria-valuemin` are really required -- whether the author knows those values or not.

With this in mind: What changes would you make to improve this situation?

mcking65:
on Aug 31, 2017

I'm open to suggestions. Some things to consider:

joanmarie commented:

>1. There are cases where there is (functionally) no minimum value and/or maximum value. If we change the authors SHOULD to an authors MUST, won't validators spit up when the author does not provide a value because the range is (functionally) boundless?

We need to accommodate the real world, which means that authors should not have to specify the min and max unless there is a min and max. That means, min and max cannot be required properties.

>2. In cases where there is a known, bound minimum value, authors really need to provide it. Ditto for a known maximum value. To me "really need" means "required."

This is typical territory for "authors SHOULD" throughout the spec ... it is as it should be.

[...]

PR #813, opened 2018-08-31
Role spinbutton: allow empty values, no min, no max, and structure with sibling steppers

mcking65:
For issue #797, allow empty spin buttons by:
1. Changing aria-valuenow to supported from required property.
2. Changing prose to author SHOULD from author MUST specify value for aria-valuenow.
3. Adding aria-valuetext as supported property.

For issue #642, allow min and max to be undefined:
1. Change aria-valuemin and aria-valuemax from required to supported properties.
2. Revise prose of paragraph with normative statements.

For issue #812:
1. Change description to clearly state that a text field with sibling buttons outside the spinbutton is permitted.
2. Related editorial change to consolidate keyboard requirements into a single paragraph.
3. Other related editorial revisions for clarity.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/813.html" title="Last updated on Aug 31, 2018, 7:48 PM GMT (9c07713)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/813/a5f7436...9c07713.html" title="Last updated on Aug 31, 2018, 7:48 PM GMT (9c07713)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6627,16 +6627,11 @@
 			<rdef>spinbutton</rdef>
 			<div class="role-description">
 				<p>A form of <rref>range</rref> that expects the user to select from among discrete choices.</p>
-				<p>A <code>spinbutton</code> typically allows the user to select from the given range through the use of an up and down button on the keyboard. Visibly, the current value is incremented or decremented until a maximum or minimum value is reached. Authors SHOULD ensure this functionality is accomplished programmatically through the use of <kbd>up</kbd> and <kbd>down</kbd> arrows on the keyboard.</p>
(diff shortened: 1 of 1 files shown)


joanmarie:
For future reference, let's aim for one pull request per issue.

From my initial triage of what we can already commit to the stable (Working Draft) branch, the issue #812 stuff doesn't strike me as something for which we need Core-AAM changes or tests; it's purely authoring.

On the other hand, this same pull request changed the default value of `aria-valuenow` from `0` to "that there is no current value" which should be tested. There may also be the need for a corresponding change in the Core AAM. 

Do we block merging this entire set of changes into stable because of the single line (but significant) change related to `aria-valuenow`? Do we manually break apart the changes and commit parts, but leave parts out? Neither is good imho. Solution: Please don't do this in the future. :) :)

Issue #811, opened 2018-08-28
New Accessibility Tree section needs intro


carmacleod:
The new [Section 7 Accessibility Tree](http://w3c.github.io/aria/#accessibility_tree) in the ARIA spec jumps into what must and must not be in the Accessibility Tree without introducing the tree first. Perhaps the paragraph in [Section 1.2 The Accessibility Tree and the DOM Tree](https://w3c.github.io/core-aam/#intro_treetypes) from the Core AAM would make a good intro.

PR #814, opened 2018-09-06
New Accessibility Tree section needs intro #811

carmacleod:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/814.html" title="Last updated on Sep 21, 2018, 11:28 PM GMT (54658c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/814/a5f7436...54658c7.html" title="Last updated on Sep 21, 2018, 11:28 PM GMT (54658c7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -11158,7 +11158,7 @@
 </section>
 <section class="normative" id="accessibility_tree">
 	<h1>Accessibility Tree</h1>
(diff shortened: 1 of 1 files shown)


joanmarie:
FWIW, "Generally if something can be trimmed out it will be, for reasons of performance and simplicity." is not necessarily the case. Gecko includes EVERY last div. I wish they didn't.

jnurthen:
@carmacleod I removed the last sentence. If we want to add something back in we can do so - but I think crafting something which describes the differing behaviours will be hard - probably better to remain silent in what is an introductory paragraph.

PR #820, opened 2018-10-05
Additional considerations regarding mapping implementation

plehegar:
This adds to the charter:
[[
No platform will be dropped out of the specification without prior consultation with that platform's owners. Every effort will be made to take member organizations' schedules into account and to ensure all platforms are included. 
]]


plehegar:
This doesn't go all the way, but I can't see how to go further without giving veto powers. @joanmarie, the wording comes from your latest proposal as far as I know.

PR #828, opened 2018-10-24
[WIP] Merge remaining changes from #813 into stable

joanmarie:
Most of the changes made for pull request #813 were editorial or
were specific to authoring and thus did not require any changes
to Core AAM or testing related to exit criteria. Two of the changes,
however did not fall into this category namely:

* Make aria-valuetext a supported property of spinbutton
* Change the default value of aria-valuenow from 0 to "there is no
  current value"

Once we have ensured mapping changes have been made, tests are in
place, and implementations present or planned, we can merge the above
into the stable branch.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/828.html" title="Last updated on Oct 24, 2018, 11:38 AM GMT (55d13dd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/828/ebae0d9...55d13dd.html" title="Last updated on Oct 24, 2018, 11:38 AM GMT (55d13dd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6668,6 +6668,7 @@
 								<li><pref>aria-valuemax</pref></li>
 								<li><pref>aria-valuemin</pref></li>
 								<li><pref>aria-valuenow</pref></li>
+								<li><pref>aria-valuetext</pref></li>
 							</ul>
 						</td>
 					</tr>
@@ -6700,7 +6701,7 @@
 						<td class="implicit-values">
 							Default for <pref>aria-valuemin</pref> is that there is no minimum value.<br/>
 							Default for <pref>aria-valuemax</pref> is that there is no maximum value.<br/>
-							Default for <pref>aria-valuenow</pref> is <code>0</code>.<br/>
+							Default for <pref>aria-valuenow</pref> is that there is no current value.<br/>
 						</td>
 					</tr>
 				</tbody>


jnurthen:
Once merged closes #812

jnurthen:
Looks like this is merged into stable with the force-push.... @joanmarie can you confirm we can close this?

joanmarie:
It wasn't merged into stable; if it were, we would see `aria-valuetext` listed as a supported property of `spinbutton`.  Ditto for the default value of `aria-valuenow` which should be "there is no current value"; not 0.

If you look in https://www.w3.org/TR/wai-aria-1.2/#spinbutton, you'll also note that the above changes were not made.

I'm not positive, but I believe what the force push cited above refers to is me fixing a silly error, a merge conflict, or something like that -- in my branch. Both master and stable should be protected from force merges.

That said, looking at both https://www.w3.org/TR/wai-aria-1.2/#spinbutton and https://www.w3.org/TR/wai-aria-1.1/#spinbutton, it turns out that `aria-valuetext` is already a supported property of `spinbutton` courtesy of inheritance. So I manually landed that change in stable: d36e35cf120627591abdd2819dd057a139129333.

That leaves the default value which will indeed need testing.

kbunge63:
Just want to make sure I am reading this correctly. You are stating that aria-valuenow should be set to an alphanumeric value?

joanmarie:
Here is the information on aria-valuenow: https://w3c.github.io/aria/#aria-valuenow.

This issue upon which you commented is a publication-specific issue.

carmacleod:
@joanmarie @jnurthen 

This PR still has a 1.2 milestone. Should it be moved to 1.3? Or maybe moved to Core AAM?

The remaining change that was not merged to stable (but is in [master](https://w3c.github.io/aria/#spinbutton)) is to change the default value of aria-valuenow for spinbutton from 0 to "there is no current value".

I believe it would still need a corresponding Core AAM change and testing related to exit criteria?

jnurthen:
Closing - pushed the one remaining change manually to stable.

Issue #843, opened 2018-11-15
Editorial: Change "Default value(s)" in "Handling Author Errors" into more corrective phrase


joanmarie:
Right now we have the phrase "Default value" and "Default values" in/around the table for what user agents need to expose for missing required attributes. We're also trying to disambiguate "required attributes" from "supported attributes with default values." With this in mind, perhaps we should change "Default value" to something more corrective-sounding. Perhaps:

* Fallback value
* Error correction
* (_your suggestion here_)

carmacleod:
+1 for "Fallback value"

PR #844, opened 2018-11-16
Editorial: Change Default Value to Fallback Value

jnurthen:
#843

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/844.html" title="Last updated on Nov 16, 2018, 6:30 PM GMT (10d6c65)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/844/0ce46ea...10d6c65.html" title="Last updated on Nov 16, 2018, 6:30 PM GMT (10d6c65)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -42,7 +42,7 @@
 		//previousDiffURI: "https://www.w3.org/TR/2014/REC-wai-aria-20140320/",
 
 		// if there a publicly available Editors Draft, this is the link
-		edDraftURI: "https://w3c.github.io/aria/", 
+		edDraftURI: "https://w3c.github.io/aria/",
 
 		// if this is a LCWD, uncomment and set the end of its review period
 		// lcEnd: "2012-02-21",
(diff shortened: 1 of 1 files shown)


joanmarie:
Works for me. Though whitespace fixes which are unrelated to what issue is being addressed are something I like to avoid.

jnurthen:
> Works for me. Though whitespace fixes which are unrelated to what issue is being addressed are something I like to avoid.

I was trying a new editor and didn't notice those changes. Like you I prefer to not make whitespace changes when not necessary. As there are only 4 other changes and they are really things that should be changed I proposed not rolling them back. Is that ok with you?

joanmarie:
> As there are only 4 other changes and they are really things that should be changed I proposed not rolling them back. Is that ok with you?

Sure. Please also cherry-pick the change into the "stable" branch.

Something I do before committing any change locally is a `git diff`. That helps me catch things like unrelated changes. You can, of course, also use `git show` after commit but before pushing the changes.

PR #846, opened 2018-11-28
Fixes #837 - Remove redundant paragraph from Presentational Roles Con…

### LINKED ISSUE #837: Remove redundant paragraph from Presentational Roles Conflict Resolution section
Spinning off from https://github.com/w3c/aria/issues/833#issuecomment-436669311

The [Presentational Roles Conflict Resolution](http://w3c.github.io/aria/#conflict_resolution_presentation_none) section says:
> ... With this exception, user agents MUST always expose global WAI-ARIA states and properties to accessibility APIs. In this case, the user agent ignores the presentation role and exposes the element according to its implicit native semantics. However, user agents MUST ignore any non-global, role-specific WAI-ARIA states and properties, unless it is on an inherited presentational role where an explicit role is applied.

and the identical wording appears again only 4 paragraphs later:
> User agents MUST always expose global WAI-ARIA states and properties to accessibility APIs, even if an element has an explicit or inherited role of presentation. In this case, the user agent ignores the presentation role and exposes the element according to its implicit native semantics. However, user agents MUST ignore any non-global, role-specific WAI-ARIA states and properties, unless it is on an inherited presentational role where an explicit role is applied.

I think the 2nd instance of the paragraph should be deleted.

carmacleod:
…flict Resolution section

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/846.html" title="Last updated on Nov 28, 2018, 5:11 AM GMT (9df569c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/846/0ce46ea...9df569c.html" title="Last updated on Nov 28, 2018, 5:11 AM GMT (9df569c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5051,7 +5051,6 @@
 					</pre>
 					<p>Explicit roles on a descendant or <a>owned</a> element override the inherited role of <code>presentation</code>, and cause the owned element to behave as any other element with an explicit role. If the action of exposing the implicit role causes the accessibility tree to be malformed, the expected results are undefined and the user agent MAY resort to an internal recovery mechanism to repair the accessibility tree.</p>
 					<p>If an element with a role of presentation is focusable, or otherwise interactive, user agents MUST ignore the normal effect of the role and expose the element with implicit native semantics, in order to ensure that the element is both <a>understandable</a> and <a>operable</a>.</p>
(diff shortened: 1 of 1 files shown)


carmacleod:
Argh, I forgot that I already fixed this in PR https://github.com/w3c/aria/pull/838, so I'm closing this.
@jnurthen we need to merge some PRs...  :)

PR #852, opened 2018-11-30
ARIA: Editorial cleanup needed to "Presentational Roles Conflict Reso…

### LINKED ISSUE #646: ARIA: Editorial cleanup needed to "Presentational Roles Conflict Resolution" section
Spinning this issue off from https://github.com/w3c/aria/issues/492#issuecomment-329386912, in which @mcking65 states:

> And this paragraph needs a rewrite to make it intelligible:
>
>> Host languages elements, having implicit presentational roles for which no roles, may be applied, must never be exposed to in the accessibility tree. With this exception, user agents must always expose global WAI-ARIA states and properties to accessibility APIs. In this case, the user agent ignores the presentation role and exposes the element according to its implicit native semantics. However, user agents must ignore any non-global, role-specific WAI-ARIA states and properties, unless it is on an inherited presentational role where an explicit role is applied.
>
> The only way I came to understand it was because of the subsequent example.

carmacleod:
…lution" section #646

Please see https://github.com/w3c/aria/issues/646#issuecomment-443045420 for reviewing suggestion.

Resolves #646.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/852.html#conflict_resolution_presentation_none" title="Last updated on Feb 28, 2020, 7:17 PM UTC (855f4df)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/852/2255df0...855f4df.html" title="Last updated on Feb 28, 2020, 7:17 PM UTC (855f4df)">Diff</a>


index.html
```diff
@@ -5041,16 +5041,19 @@ <h5>Note regarding the ARIA 1.1 <rref>none</rref> role.</h5>
...
+					<p>Host language elements having explicit or inherited presentational role MUST NOT be exposed in the accessibility tree, with these exceptions:</p>
+					<ul>
+						<li>If an element is focusable, or otherwise interactive, user agents MUST ignore the <code>presentation</code> role and expose the element with its implicit native semantics, in order to ensure that the element is both <a>understandable</a> and <a>operable</a>.</li>
+						<li>If an element has global WAI-ARIA states or properties, user agents MUST ignore the <code>presentation</code> role and expose the element with its implicit native semantics. However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</li>
```
jnurthen: Why do we need "unless the presentational role is inherited and an explicit non-presentational role is applied." 
If there is an explicit non-presentational role then this whole section would not apply as the section is about presentational roles conflict resolution.

carmacleod: I think it was intended to clarify that an explicit role wins over an inherited presentational role.
So if an author wanted to use `<table role="presentation">` for layout, with `<td role="checkbox">` descendants, then they would know that the td's inherited role of presentation is ignored and "checkbox" applies.

However, you are correct that it has nothing to do with any non-global attributes (i.e. `aria-checked` for example), so the "clarifying" phrase is in the wrong place.

Perhaps it merits a separate bullet? i.e. remove the phrase from where it currently is, and instead add:
- If a required owned element has an explicit non-presentational role, user agents MUST ignore an inherited presentational role

If you agree, then I would make it the 2nd of the 3 points (so that the point about global/non-global attributes comes last, right before Example 12).

carmacleod: Got it. I think it would actually be clearer to completely delete that paragraph, remove the phrase, and insert the following bullet in the "MUST NOT be exposed in the accessibility tree, with these exceptions" list (because I think it is another exception, and should be in the list):

- If a required owned element has an explicit non-presentational role, user agents MUST ignore an inherited presentational role and expose the element with its explicit role. If the action of exposing the explicit role causes the accessibility tree to be malformed, the expected results are undefined and the user agent MAY resort to an internal recovery mechanism to repair the accessibility tree.

Here's a preview of the [Presentational Roles Conflict Resolution section](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/852.html#conflict_resolution_presentation_none) with those changes.

index.html
```diff
@@ -5041,16 +5041,19 @@ <h5>Note regarding the ARIA 1.1 <rref>none</rref> role.</h5>
...
+					</ul>
 					<p>For example, <pref>aria-haspopup</pref> is a global attribute and would always be applied; <pref>aria-level</pref> is not a global attribute and would therefore only apply if the element was not in a presentational state.</p>
 					<pre class="example highlight">
-						<span class="comment">&lt;!-- 1. [role="presentation"] is ignored due to the global aria-haspopup property. --&gt;</span>
+<span class="comment">&lt;!-- 1. [role="presentation"] is ignored due to the global aria-haspopup property. --&gt;</span>
```
scottaohara: this example will need to be changed if #1023 lands since `aria-haspopup` will no longer be global.

maybe instead we have an example of an element that allows for naming, or an element with a tabindex for programmatic focusing, and how `role=presentation` should not be respected in those instances.

carmacleod: Ok, I've changed the example to use aria-label.
```html
<h1 role="presentation" aria-label="Sample Heading Content">Sample Content</h1>
```

carmacleod:
@joanmarie 
I really need your review on this, because I had to resolve conflicts with PR https://github.com/w3c/aria/pull/885 and I'm not 100% sure that I addressed all of your concerns. Particularly with that middle point there, about required owned elements.

Here's a preview link to the [Presentational Roles Conflict Resolution](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/852.html#conflict_resolution_presentation_none) section.

index.html
```diff
@@ -5041,16 +5041,19 @@ <h5>Note regarding the ARIA 1.1 <rref>none</rref> role.</h5>
...
-					<p>For example, <pref>aria-haspopup</pref> is a global attribute and would always be applied; <pref>aria-level</pref> is not a global attribute and would therefore only apply if the element was not in a presentational state.</p>
+					<p>Host language elements having explicit or inherited presentational role MUST NOT be exposed in the accessibility tree, with these exceptions:</p>
+					<ul>
+						<li>If an element is focusable, or otherwise interactive, user agents MUST ignore the <code>presentation</code> role and expose the element as described in the accessibility API mappings for presentational objects that remain in the accessibility tree, in order to ensure that the element is both <a>understandable</a> and <a>operable</a>.</li>
+						<li>If a <a href="#mustContain">required owned element</a> has an explicit non-presentational role, user agents MUST ignore an inherited presentational role and expose the element with its explicit role. If the action of exposing the explicit role causes the accessibility tree to be malformed, the expected results are undefined and the user agent MAY resort to an internal recovery mechanism to repair the accessibility tree.</li>
```
joanmarie: I'm still not in favor of mentioning heuristics that we don't define. Granted whether we mention them or not, user agents will do as they see fit. But why not just stop the sentence at "undefined"? Scare authors. :)

index.html
```diff
@@ -5041,16 +5041,19 @@ <h5>Note regarding the ARIA 1.1 <rref>none</rref> role.</h5>
...
+						<li>If an element has global WAI-ARIA states or properties, user agents MUST ignore the <code>presentation</code> role and expose the element as described in the accessibility API mappings for presentational objects that remain in the accessibility tree. However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</li>
```
joanmarie: Remove mention of API mappings.

scottaohara:
approved the changes per my comments, but +1 to @joanmarie's comments

index.html
```diff
@@ -5041,16 +5041,19 @@ <h5>Note regarding the ARIA 1.1 <rref>none</rref> role.</h5>
...
-					<p>For example, <pref>aria-haspopup</pref> is a global attribute and would always be applied; <pref>aria-level</pref> is not a global attribute and would therefore only apply if the element was not in a presentational state.</p>
+					<p>Host language elements having explicit or inherited presentational role MUST NOT be exposed in the accessibility tree, with these exceptions:</p>
+					<ul>
+						<li>If an element is focusable, or otherwise interactive, user agents MUST ignore the <code>presentation</code> role and expose the element as described in the accessibility API mappings for presentational objects that remain in the accessibility tree, in order to ensure that the element is both <a>understandable</a> and <a>operable</a>.</li>
```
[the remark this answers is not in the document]
carmacleod: So, remove " as described in the accessibility API mappings for presentational objects that remain in the accessibility tree" so that the sentence reads:
> If an element is focusable, or otherwise interactive, user agents MUST ignore the <code>presentation</code> role and expose the element.
?

index.html
carmacleod: So, remove " as described in the accessibility API mappings for presentational objects that remain in the accessibility tree" so that the sentence reads:
> If an element has global WAI-ARIA states or properties, user agents MUST ignore the <code>presentation</code> role and expose the element.

joanmarie: Yes.

index.html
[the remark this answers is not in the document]
joanmarie: That would work. However, now that you mention it, does ignoring the `presentation` role ensure the element is "understandable"? :) Really what it accomplishes is that ATs can see the object and will -- assuming no other implementation failures -- receive accessibility events. But that goes beyond the scope of our spec. So maybe "... in order to ensure that the element is operable."?

carmacleod:
Hi, @mcking65! This PR is ready for review.

Note that the Diff is messy because I reorganized words and moved things into a bulleted list.
The [Preview](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/852.html#conflict_resolution_presentation_none) is definitely easier to follow.

I only changed the prose in the Presentational Roles Conflict Resolution section (nothing else), so there's only 3 paragraphs plus the example to review (the bulleted list is in paragraph 2).

Note that the phrase "or otherwise interactive" is still being discussed in https://github.com/w3c/aria/issues/1192, so for now, I think we should just keep that phrase as-is in order to get this PR merged sooner.

mcking65:
The example helps a lot, but I am concerned about an example that goes against APG guidance. We caution against using aria-label on a heading.

carmacleod:
@mcking65 Notes for the latest commit:
- changed example to use aria-describedby on the heading instead of aria-label
- added "with its implicit role" in 2 places (for clarity on _how_ UA should expose element)

carmacleod:
@jnurthen Example is fixed up. Here's the ping you requested. :)

PR #880, opened 2019-01-10
Move aria-level on heading role from required section to supported section

### LINKED ISSUE #854: Heading has aria-level required, but there is a default value for aria-level
The default value is currently 2. The proposal is to remove the default value from the role definition and rely on the fallback value in https://w3c.github.io/aria/#document-handling_author-errors

melanierichards:
Fixes #854

A couple areas where I'd like to get feedback, please! :

* We're thinking of these "default" values as a repair technique rather than a default, so under the implicit values section I changed the language from "default" to "fallback". This introduces an inconsistency, do we want to keep the old language or update other implicit value sections?
* Another point there is do we just want to point people over to the fallback table instead of keeping the value inline (inline seems handy for the reader)
* The fallback table is prefaced with "Fallback values for missing required attributes". This is out of scope for this PR, but maybe we should modify the language here as we're using these fallback values for more than just missing required attributes. I don't think we're tracking this (though it could be referenced in another issue and I just missed it), but if people agree, I can open an issue for that change.

***
<a href="https://pr-preview.s3.amazonaws.com/melanierichards/aria/pull/880.html" title="Last updated on Jan 10, 2019, 4:56 PM UTC (52b576f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/880/2e2ccef...melanierichards:52b576f.html" title="Last updated on Jan 10, 2019, 4:56 PM UTC (52b576f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3028,11 +3028,11 @@
 					</tr>
 					<tr>
 						<th class="role-required-properties-head">Required States and Properties:</th>
-						<td class="role-required-properties"><pref>aria-level</pref></td>
+						<td class="role-required-properties"> </td>
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
-						<td class="role-properties"> </td>
+						<td class="role-properties"><pref>aria-level</pref></td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
@@ -3065,7 +3065,7 @@
 					</tr>
 					<tr>
 						<th class="implicit-values-head" scope="row">Implicit Value for Role:</th>
-						<td class="implicit-values">Default for <pref>aria-level</pref> is <code class="default">2</code>.</td>
+						<td class="implicit-values">Fallback for <pref>aria-level</pref> is <code class="default">2</code>.</td>
 					</tr>
 				</tbody>
 			</table>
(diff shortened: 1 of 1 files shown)


jnurthen:
> * We're thinking of these "default" values as a repair technique rather than a default, so under the implicit values section I changed the language from "default" to "fallback". This introduces an inconsistency, do we want to keep the old language or update other implicit value sections?

I think we have decided to drop the value from the table completely. The fallback is a repair technique so an author really should have no reason to ever rely on it. The fallback table is really for backwards compatibility rather than anything we want to expose to authors in the future.

> * Another point there is do we just want to point people over to the fallback table instead of keeping the value inline (inline seems handy for the reader)

Again - I don't want to encourage use of the fallbacks. It should be an author error if they are used.

> * The fallback table is prefaced with "Fallback values for missing required attributes". This is out of scope for this PR, but maybe we should modify the language here as we're using these fallback values for more than just missing required attributes. I don't think we're tracking this (though it could be referenced in another issue and I just missed it), but if people agree, I can open an issue for that change.

+1

melanierichards:
@jnurthen thanks for the feedback!

I just realized that I thought this was one of the ones where we wanted to move the prop from required to supported, but nope, that is not the case. Lesson learned: do not craft PRs before coffee :) Will fix and submit a new PR.

Issue #892, opened 2019-01-24
Range should subclass structure rather than widget


joanmarie:
In working on the new `meter` role, I had it subclass `range`. The problem with that is `range` subclasses `widget` and a `widget` is interactive.

The general consensus from the Working Group's meeting today is that `range` should subclass `structure`. Then the interactive roles which subclass `range` should also subclass `widget`.

joanmarie:
Regarding the interactive roles, it turns out that `slider` and `spinbutton` already subclass both `range` and `input`. And, of course, `input` subclasses `widget`.  So it's only `progressbar` and `scrollbar` whose parent class need to be changed.

PR #894, opened 2019-01-30
Change the superclass of range from widget to structure

joanmarie:
This is needed to support the aria-value* attributes on non-interative,
and potentially static objects such as a meter.

Because range now subclasses structure, we need to use multiple inheritance
with the widget roles which subclass range. Those roles are: progressbar,
scrollbar, slider, and spinbutton. The latter two already subclass both range
and input thus no further change is needed. Progressbar and scrollbar now
subclass both range and widget.

Addresses github issue #892.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/894.html" title="Last updated on Jan 31, 2019, 7:07 PM UTC (0d6f4d5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/894/60a9063...0d6f4d5.html" title="Last updated on Jan 31, 2019, 7:07 PM UTC (0d6f4d5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5161,7 +5161,12 @@
 					</tr>
 					<tr>
 						<th class="role-parent-head" scope="row">Superclass Role:</th>
-						<td class="role-parent"><rref>range</rref></td>
+						<td class="role-parent">
+							<ul>
+								<li><rref>range</rref></li>
+								<li><rref>widget</rref></li>
+							</ul>
+						</td>
 					</tr>
 					<tr>
 						<th class="role-children-head" scope="row">Subclass Roles:</th>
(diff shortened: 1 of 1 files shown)


joanmarie:
I just used github's "diff" tool to see what all changed in the actual spec content. (This is helpful because inherited states and properties get generated by scripts, rather than being hard-coded in the spec text.)

I don't see any unexpected changes related to inheritance. Or for that matter related to anything else. :smile: Thus I *believe* this change is safe regarding taxonomy fallout.

joanmarie:
Merged into master and stable after making the editorial change requested at today's meeting ("a structure"->"an element") and adding the ChangeLog entry.

index.html
```diff
@@ -5401,7 +5406,7 @@ <h5>Presentational Roles Conflict Resolution</h5>
 		<div class="role" id="range">
 			<rdef>range</rdef>
 			<div class="role-description">
-				<p>An input representing a range of values that can be set by the user.</p>
+				<p>An element representing a range of values that can be set by the user.</p>
```
mcking65: @joanmarie, thought of one other aspect of this definition that will not be accurate after meter is included ... this might also be true for progressbar: the user can not set the value.

joanmarie: Oops. Only just saw this today. I just removed "that can be set by the user" from the definition of range in both master and stable as it's an editorial change. Thanks for catching that!

PR #895, opened 2019-01-31
Add new time role

### LINKED ISSUE #878: need role for time element
per [plans for role parity](https://github.com/w3c/aria/wiki/Plans-regarding-role-parity#specific-role) need a role for equivalent to [`<time>`](https://www.w3.org/TR/html52/textlevel-semantics.html#the-time-element) in HTML

melanierichards:
Fixes #878 

(Moved the questions I had in this PR to the issue itself, for a single-point-of-truth for historic discussion)

***
<a href="https://pr-preview.s3.amazonaws.com/melanierichards/aria/pull/895.html" title="Last updated on Apr 2, 2019, 8:58 PM UTC (9ab3522)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/895/1546f84...melanierichards:9ab3522.html" title="Last updated on Apr 2, 2019, 8:58 PM UTC (9ab3522)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -7773,9 +7773,7 @@
 					</tr>
 					<tr>
 						<th class="role-namefrom-head" scope="row">Name From:</th>
-						<ul>
-							<li>author</li>
-						</ul>
+						<td class="role-namefrom">author</td>
 					</tr>
 					<tr>
 						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
--- a/index.html
+++ b/index.html
@@ -7686,7 +7686,41 @@
 			<rdef>time</rdef>
 			<div class="role-description">
 				<p>Represents a specific point in time.</p>
-				<p>Authors SHOULD limit text contents to a valid date- or time-related string as specified for the <a href="https://www.w3.org/TR/html52/textlevel-semantics.html#the-time-element"><abbr title="Hypertext Markup Language">HTML</abbr> <code>time</code></a> element.</p>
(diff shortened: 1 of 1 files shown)


melanierichards:
Just made a minor stylistic adjustment to the definition based on @mcking65's feedback a couple calls ago: "Represents…" becomes "An element that represents…" If a more specific term than "element" is desired, maybe "phrase"? Perhaps pending creation of a new phrase abstract role.

index.html
```diff
@@ -7680,6 +7681,120 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
+						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
+						<td class="role-properties"> </td>
+					</tr>
+					<tr>
+						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
+						<td class="role-inherited"> </td>
+					</tr>
+					<tr>
+						<th class="role-namefrom-head" scope="row">Name From:</th>
+						<ul>
```
jnurthen: When It is a single value we don't normally use a list in Name From. I wonder if this will fix it.

```suggestion
	<td class="role-namefrom">author</td>
```

jnurthen: (couldn't work out how to suggest a change for multiple lines - sorry but this should replace the next 2 too!)

melanierichards: @jnurthen weirdness is fixed! https://pr-preview.s3.amazonaws.com/melanierichards/aria/pull/895.html#time

I originally had more than one item in the list and just didn't simplify the markup when I made the change. But I think the missing `<td>` was more of the problem. Not sure how I missed that and was on the point of just running this through a validator, but in any case, fixed now. Thank you!

Issue #553, opened 2017-04-12
Inconsistency in rowgroup description


aleventhal:
Under rowgroup, the ARIA 1.1 spec says "Authors must ensure elements with role rowgroup are contained in, or owned by, an element with the role table or grid."

This would seem to indicate that a rowgroup cannot be in a treegrid.

However, in the table directly below it says "Required context role: table, grid, treegrid".

joanmarie:
Change landed. Closing. Thanks for the report!

PR #903, opened 2019-02-19
Fix issue #553: rowgroup role-description of required context roles mismatches characteristics table

dbjorge:
Per #553: The `rowgroup` description states that it *MUST* be owned by a `grid` or `table`, but the characteristics table for the `rowgroup` says that in addition to `grid` and `table`, a `treegrid` is an acceptable context role. Since `treegrid`'s definition is consistent in saying that a `treegrid` is a `grid` and `rowgroup`s are valid within it, this seems like it's just an omission in the role-description text.


joanmarie:
Merged into Editor's Draft and cherry-picked into Working Draft. Thanks much!!

PR #906, opened 2019-02-20
Add Translatable Attributes section

### LINKED ISSUE #755: define aria attributes that include a human readable text string are translatable attributes
from https://www.w3.org/Bugs/Public/show_bug.cgi?id=25729
> HTML defines a list of translatable attributes 
> http://www.w3.org/html/wg/drafts/html/master/dom.html#translatable-attributes
> 
>  ARIA should do the same
> 
> aria-label
> aria-valuetext
> aria-roledescription
> more?

jnurthen:
Resolves #755

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/906.html" title="Last updated on Mar 21, 2019, 11:04 PM UTC (b8e0790)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/906/49c8ef9...b8e0790.html" title="Last updated on Mar 21, 2019, 11:04 PM UTC (b8e0790)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -8501,7 +8501,7 @@
 	</section>
 	<section>
 		<h2>Translatable States and Properties</h2>
-		<p>The values of the following <a>states</a> and <a>properties</a> should be translated when a page is localized:</p>
+		<p>The HTML specification states that other specifications can define <a href="https://www.w3.org/TR/html52/dom.html#translatable-attributes">translatable attributes</a>. The values of the following <a>states</a> and <a>properties</a> are <a href="https://www.w3.org/TR/html52/dom.html#translatable-attributes">translatable attributes</a> and should be translated when a page is localized:</p>
 		<ul>
 			<li><pref>aria-label</pref></li>
 			<li><pref>aria-placeholder</pref></li>
--- a/index.html
+++ b/index.html
@@ -8494,20 +8494,20 @@
 			</dl>
 			<p>These are generic types for states and properties, but do not define specific representation. See <a href="#state_property_processing">State and Property Attribute Processing</a> for details on how these values are expressed and handled in host languages.</p>
 		</section>
-		<section>
-				<h3>Translatable Atttributes</h3>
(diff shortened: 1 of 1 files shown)


joanmarie:
The section looks fine.

Nit 1: I think it would make more sense to locate the section outside of 6.2. Functionally, it is an aspect of 6.3 (Values for States and Properties) because for those four properties, the value should be translated. So I'd put it under 6.3 (either as 6.3.1 or making it 6.4).

Nit 2: We're using "States and Properties" rather than "Attributes" for all headings in Section 6 (modulo 6.5.x for some reason). Therefore would it make sense to rename the new section "Translatable States and Properties" or "Translatable State and Property Values"?

jnurthen:
Update to reference HTML spec

index.html
```diff
@@ -8499,6 +8499,16 @@ <h3>Value</h3>
 		<h2>Values for States and Properties</h2>
 		<p>Many <a>states</a> and <a>properties</a> accept a specific set of tokens as <span>values</span>. The allowed values and explanation of their meaning is shown after the table of characteristics. The default value, if defined, is shown in <strong>strong</strong> type, followed by the parenthetical term 'default'. When a value is indicated as the default, the user agent MUST follow the behavior prescribed by this value when the state or property is empty or unspecified. Some <a>roles</a> also define what behavior to use when certain states or properties, that do not have default values, are not provided.</p>
 	</section>
+	<section>
+		<h2>Translatable States and Properties</h2>
+		<p>The values of the following <a>states</a> and <a>properties</a> should be translated when a page is localized:</p>
```
mcking65: Perhaps:

>In order to be understandable by assistive technology users, the values of the following states and properties need to be translated when a page is localized:

This revision:
1. clarifies the reason for translation.
2. Avoids normative sounding language when there is no clear subject for a normative requirement.

Issue #875, opened 2019-01-09
[Role Parity] need role for fieldset


jnurthen:
per [plans for role parity](https://github.com/w3c/aria/wiki/Plans-regarding-role-parity#specific-role) need a role for equivalent to [`<fieldset>`](https://www.w3.org/TR/html52/sec-forms.html#the-fieldset-element) in HTML

PR #911, opened 2019-02-25
Issue 875 need role for fieldeset/legend

jongund:
This issues is about fieldset parity, but the actual html5 fieldset element is handled easily be the group role. The main issue is the corresponding legend element.  This pull request defines a legend role that can be used to provide an accessible name to a group role.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/911.html" title="Last updated on Apr 5, 2019, 1:07 AM UTC (4dcc790)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/911/af105b5...4dcc790.html" title="Last updated on Apr 5, 2019, 1:07 AM UTC (4dcc790)">Diff</a>


jnurthen:
This issue was discussed in [a meeting](https://www.w3.org/2019/02/28-aria-minutes.html).

- No actions or resolutions
<details><summary><i class="differentiate">View the transcript</i></summary>
Issue 875 need role for fieldeset/legend<br/>
&lt;jamesn> <a href="https://github.com/w3c/aria/pull/911">https://github.com/w3c/aria/pull/911</a><br/>
<b>jamesn:</b> so yes, we need to add something<br/>
<b>mck:</b> this is a lot like label role in that it removes the need for aria-labelledby and would affect name calc for groups if it's present, right?<br/>
<b>jongund:</b> yes<br/>
<b>jamesn:</b> when we were thinking about this before, we said we would expand the caption role...<br/>
<b>mck:</b> did we add a caption role?<br/>
<b>joanie:</b> it covers HTML caption and figcaption<br/>
<b>&lt;jamesn&gt;</b> <a href="https://github.com/w3c/aria/wiki/Plans-regarding-role-parity#specific-role">https://github.com/w3c/aria/wiki/Plans-regarding-role-parity#specific-role</a><br/>
<b>Bryan:</b> from a grouping perspective, it'd implicitly reference the caption<br/>
<b>jamesn:</b> is there any reason we can't use caption for legend as well?<br/>
<b>Bryan:</b> can't think of one<br/>
<b>mck:</b> looking at current editors draft...<br/>
<b>&lt;jamesn&gt;</b> caption<br/>
<b>&lt;jamesn&gt;</b> On-screen descriptive text for a figure or table in the page.<br/>
<b>&lt;jamesn&gt;</b> Authors SHOULD reference the element with role caption by setting aria-describedby on the figure or table.<br/>
<b>joanie:</b> we need a nice way to connect the caption to the element it's describing, this will cause an accessible relationship to be formed<br/>
<b>mck:</b> i thought that was implicit?<br/>
<b>joanie:</b> i've seen captions that are longer than what you want for an accname<br/>
<b>mck:</b> if the caption is sometimes providing the name (or description)... well this is a little off topic<br/>
... real issue is can we do this for group? if purpose is to provide a name, we have to revisit -describedby<br/>
<b>Stefan:</b> currently, readers like JAWS seeing a form, these things are treated like regions. if we attempt something like this (groups), we may lose that ability to navigate by regions.<br/>
<b>joanie:</b> Jon is working on a label role. while a caption is a longer thing, the text in a figcaption is shorter, and this is closer to what Stefan was saying.<br/>
... i'm wondering if legend could add to label<br/>
<b>jamesn:</b> that makes sense, when you look at legend mapping, it maps to label in APIs. so having the legend relate to the fieldset makes a lot of sense<br/>
<b>jongund:</b> accname calc would be different<br/>
<b>mck:</b> captions go inside and labels go outside what they label<br/>
<b>jongund:</b> right<br/>
... there's also certain behaviors expected<br/>
<b>Bryan:</b> behaviorally, if legend is the description ... that information is tacked on to something. it's not the way you typically think of a name<br/>
<b>&lt;Devarshi&gt;</b> Isn't Legend the 'group' label for the fieldset?<br/>
<b>jamesn:</b> bringing this back, Jon currently has a legend role being defined. we have 3 potential options<br/>
... a legend role, a label, or a caption<br/>
... is that a fair summary?<br/>
<b>mck:</b> if we reuse caption, do we need the relationship?<br/>
<b>jamesn:</b> yes<br/>
<b>mck:</b> I thought the point was that we didn't need to explicitly define?<br/>
<b>jamesn:</b> so if you had caption... you need to have a relationship.<br/>
<b>mck:</b> a caption labels its closest ancestor... right?<br/>
<b>jamesn:</b> if we're doing it with aria, does it have to be a descendent relationship?<br/>
<b>mck:</b> well if we're talking about role parity... this might be confusing for authors<br/>
<b>Bryan:</b> aria-labelledby already works like that, Jon is saying the legend would be recognized by the group it's in<br/>
... but what if they put a group with a mixture of elements like radios etc? what gets the label?<br/>
<b>mck:</b> so we'd have to make it clear to use the same approach as in HTML<br/>
... it's possible to create conflicts.<br/>
... or at least might be redundant<br/>
<b>jamesn:</b> why are we trying to mimic the HTML structure? isn't the point to allow role parity w/o enforcing the same structure?<br/>
<b>jongund:</b> a lot of times, people are fixing stuff. most stuff out there is made out of divs and spans<br/>
<b>mck:</b> the driver of HTML parity has been web components and virtual dom nodes etc.<br/>
<b>jamesn:</b> so... why would we create something where a legend has to be a child of a group? rather than specifying what the relationship is?<br/>
<b>mck:</b> the legend role isn't revealed from an a11y standpoint - it's a label. if you're pointing at it with -labelledby, then why make it a legend?<br/>
<b>jamesn:</b> right? So why not make it a label<br/>
<b>mck:</b> or any role?<br/>
<b>jongund:</b> the main advantage of my proposal is an auth technique that doesn't require labels<br/>
<b>&lt;jamesn&gt;</b> <a href="https://www.w3.org/TR/html-aam-1.0/#details-id-84">https://www.w3.org/TR/html-aam-1.0/#details-id-84</a><br/>
<b>jongund:</b> if aria is bigger than just html, this provides a group label with visible text in an easy way<br/>
... it simulates the features of an HTML5 legend element<br/>
... so if role=legend is on something and it's not descendent of a group, and not ref'd, it does nothing<br/>
<b>Bryan:</b> i understand the value from parity perspective, but it's the same from an accname perspective<br/>
<b>mck:</b> conceptually, the algorithm is similar. so if you have a fieldset, its browsers are already calculating it's accname from the legend?<br/>
... what jamesn was proposing was spec'ing out more flavors<br/>
<b>Bryan:</b> but we don't have a legend role<br/>
<b>jamesn:</b> which is what the PR is attempting to do<br/>
... straw poll: 3 options<br/>
... legend role, expand caption role, or use label role<br/>
<b>&lt;mck&gt;</b> +1 for legend role<br/>
<b>jamesn:</b> +1 for legend role?<br/>
<b>&lt;jongund&gt;</b> +1<br/>
<b>&lt;jamesn&gt;</b> +1<br/>
<b>&lt;HarrisSchneiderman&gt;</b> +1<br/>
<b>&lt;melanierichards&gt;</b> +1<br/>
<b>&lt;jemma&gt;</b> +1<br/>
<b>&lt;mck&gt;</b> +1 for bryan on legend role too<br/>
<b>&lt;Stefan&gt;</b> +1<br/>
<b>&lt;pkra&gt;</b> +1<br/>
<b>&lt;Devarshi&gt;</b> +1<br/>
<b>jamesn:</b> do i need to do the others? no i don't think so<br/>
... any other +1's?<br/>
<b>joanie:</b> 0 for all<br/>
... i need to do some more thinking<br/>
<b>&lt;Stefan&gt;</b> +1 for label as alternative<br/>
<b>jamesn:</b> joanie, can you look at the PR and add some comments then? all else, put comments online on PR<br/>
</details>

joanmarie:
While I still don't understand why we cannot use the `label` role, I'm not opposed to a `legend` role if that's what everyone else thinks is best from an authoring standpoint.

From an implementation-in-browsers standpoint, I don't think `legend` versus `label` makes a difference. 

From an AT point of view, I also think `legend` versus `label` makes no difference. If you look at the mappings in the HTML-AAM, [`legend`](https://w3c.github.io/html-aam/#el-legend) has essentially the same mappings as [`label`](https://w3c.github.io/html-aam/#el-label).

jongund:
@joanmarie Where it will make a difference is the in the accessible name calculation.  The LEGEND role will be used to label groups as a descendant element, where LABEL will be used to label controls and widgets through encapsulation.   If we used LABEL for both it could lead to unexpected accessibility problems when there were authoring errors and complicate accessible name calculation for browser developers.

jongund:
@jnurthen 
I updated the pull request to harmonize the style of the definition with the changes to the label definition.

joanmarie:
Regarding:
> legend: name comes from the text value of the first descendant element node with the role of legend. Although this may be allowed in addition to "author" in some roles, this is used in content only if higher priority "author" features are not provided.

I find "Although this may... this is used in content" a bit vague.

I assume the first "this" is "legend". I assume the second "this" is the "the legend". So what about:

> Although "legend" may be allowed in addition to "author" in some roles", it will only be used for the name if higher priority "author" features are not provided."

Or something like that.

jongund:
@joanmarie I made the suggested edit on the definition of encapsulation

jnurthen:
note - I think we might want to expand legends applicability in future (not as part of this PR). I see no reason in the future we shouldn't be able to use legend to label any of the landmark or window roles. Again - this is not a comment for this PR. I'll open a new issue for this.

jnurthen:
resolves #875

jongund:
@jnurthen I have cleaned up the conflicts with the label role, so this pull request should be ok for merging

PR #916, opened 2019-03-19
Add the deletion and insertion roles (w3c#526)

aleventhal:
[Preview of insertion](https://raw.githack.com/w3c/aria/24b545eb68ee08246b52b214af6604835d5fb019/index.html#insertion)
[Preview of deletion](https://raw.githack.com/w3c/aria/24b545eb68ee08246b52b214af6604835d5fb019/index.html#deletion)

### DIFF
--- a/index.html
+++ b/index.html
@@ -662,6 +662,7 @@
 				<li><rref>cell</rref></li>
 				<li><rref>columnheader</rref></li>
 				<li><rref>definition</rref></li>
+        <li><rref>deletion</rref></li>
 				<li><rref>directory</rref></li>
 				<li><rref>document</rref></li>
 				<li><rref>feed</rref></li>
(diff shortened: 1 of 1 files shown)


aleventhal:
Matt, I basically used your text except:
- removed the word "recently", since the recency is not important
- removed the word "element", since although a role is used on an element we aren't defining an element type here

index.html
```diff
@@ -2209,6 +2211,80 @@ <h2>Definition of Roles</h2>
...
+            <th class="role-required-properties-head">Required States and Properties:</th>
+            <td class="role-required-properties"> </td>
+          </tr>
+          <tr>
+            <th class="role-properties-head" scope="row">Supported States and Properties:</th>
+            <td class="role-properties"> </td>
+          </tr>
+          <tr>
+            <th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
+            <td class="role-inherited">Placeholder</td>
```
melanierichards: I was confused by the "Placeholder" value here for a second (given that there are HTML/ARIA concepts of a placeholder), maybe we should note "TBD, pending final Superclass Role" to make it clearer to the casual reader and our future selves? Not something I'd block merge on, though.

index.html
```diff
@@ -3224,6 +3300,80 @@ <h2>Definition of Roles</h2>
...
```
melanierichards: Same comment as `deletion` Inherited States + Props

Issue #834, opened 2018-11-01
Remove the string reflection properties that will be replaced by incompatible element reflection properties


cookiecrook:
Some attribute string reflection properties will likely be replaced by TBD and potentially incompatible object reference properties, so in ARIA 1.2, we should remove the string reflection properties for any attributes referencing IDREF or IDREFS based values.

cookiecrook:
Example: String Element.ariaActiveDescendant may be replaced by Element Element.ariaActiveDescendantElement, which could make it difficult to keep string reflection supported and working as authors expect. 

Note: there would be no change to reflection of attributes that take number, token, or string values.

cookiecrook:
Yes, I'm suggesting we remove these _before_ publishing 1.2.

```
	attribute DOMString? ariaActiveDescendant;
	attribute DOMString? ariaControls;
	attribute DOMString? ariaDescribedBy;
	attribute DOMString? ariaDetails;
	attribute DOMString? ariaErrorMessage;
	attribute DOMString? ariaFlowTo;
	attribute DOMString? ariaLabelledBy;
	attribute DOMString? ariaOwns;
```

jnurthen:
1.2 is still quite a way off. But if these will be removed and not consistently supported then removing them seems like a good idea.

cookiecrook:
Note: The prose of the "disambiguation pattern" section references `ariaDescribedBy` so that could be replaced with another came-cased example such as `ariaHasPopup`.

cookiecrook:
No need to wait. We plan to [unship these from WebKit](https://webkit.org/b/197764). I do not believe any other implementor will ship these, because element reflection is the likely path forward.

jnurthen:
@joanmarie Looks like this hasn't been merged into master. I don't believe there is anything left to do here... Should we merge?

joanmarie:
@jnurthen which PR specifically hasn't been merged? Or is another PR needed?

jnurthen:
@joanmarie I was confused.... There is no PR... forget I said anything....

cookiecrook:
Looks like this has been superseded with later edits by Alice and Aaron. I think this issue can be closed. Do you concur @jnurthen?

cookiecrook:
NM. Ignore that. We'll remove them now and add them back in 1.3.

PR #920, opened 2019-03-22
Use element reflection for IDREF/IDREF list

alice:
This depends on a [pull request on WHATWG](https://github.com/whatwg/html/pull/3917), which depends on implementations, which depends on something to implement, so we do have a bit of a dependency cycle here.

In any case, we are planning to send out an intent to implement this on Chrome; it should be fine to do that based on a PR, but if you have any feedback it would be much appreciated!

### DIFF
--- a/index.html
+++ b/index.html
@@ -11747,7 +11747,7 @@
 		<h2>Interface Mixin <dfn>AriaAttributes</dfn></h2>
 		<pre class="idl">
 			interface mixin AriaAttributes {
-				attribute DOMString? ariaActiveDescendant;
+				attribute Element? ariaActiveDescendantElement;
 				attribute DOMString? ariaAtomic;
 				attribute DOMString? ariaAutoComplete;
 				attribute DOMString? ariaBusy;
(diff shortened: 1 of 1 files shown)


BenevolentHighPriestess:
Thank you

jnurthen:
Thanks @alice 
We will discuss in our next chair meeting how to proceed.

alice:
@cookiecrook FYI

foolip:
The changes here have reached web-platform-tests in https://github.com/web-platform-tests/wpt/pull/16463, and I noticed in https://wpt.fyi/results/wai-aria/idlharness.window.html?diff&filter=ADC&run_id=6543921261838336&run_id=4861101535657984 (via [check](https://github.com/web-platform-tests/wpt/pull/16463/checks?check_run_id=109462258)) that Safari was previously passing all the tests. This is the case for Safari stable (12.1) as well:
https://wpt.fyi/results/wai-aria/idlharness.window.html?run_id=6277854882955264&run_id=5140432048619520

Was this considered when renaming? Is there a WebKit bug to follow this rename? The risk otherwise is of course that both names will rename and that web developers will have to deal with both for a long time.

alice:
> Was this considered when renaming? 

The fact that Safari had shipped the ARIA properties?

Note that it's less of a rename and more of a removal/add, since the type has also changed.

> Is there a WebKit bug to follow this rename?

@cookiecrook to comment.

> The risk otherwise is of course that both names will rename and that web developers will have to deal with both for a long time.

I doubt the old names saw a lot of use in production; is there a way to check?

carmacleod:
Is Safari the only one implementing so far?
@LJWatson was asking in https://github.com/w3c/aria/issues/957

cookiecrook:
As I mentioned in #834, the string reflection variants should be removed from ARIA 1.2. We'll likely unship these from the WebKit implementation before the element reflection bits go in.

foolip:
> I doubt the old names saw a lot of use in production; is there a way to check?

Searching HTTP Archive using BigQuery is the method I'd use, [HTTP Archive for web compat decision making](https://docs.google.com/document/d/1cpjWFoXBiuFYI4zb9I7wHs7uYZ0ntbOgLwH-mgqXdEM/edit#heading=h.3701s9nnsb27) has some instructions. However, depending on the expected usage this might be overkill, I'll leave it to the WebKit folks trying to remove this :) (If someone does want to run a query I can help though, just let me know.)

cookiecrook:
Moving this back from "Done" to "In progress" because we commented these lines for the 1.2 release.

cookiecrook:
I think this is all superseded now. Should be okay to close.

Issue #765, opened 2018-05-22
Consider addition of property(ies) to expose braille string


joanmarie:
A couple of use cases have come up in which it would be desirable to give assistive technologies braille strings which they could display as-is:

* Abbreviated/localized role descriptions
* Nemeth rendering for math content

With respect to the former, we created `aria-roledescription` so that authors could customize what is presented to the user by assistive technologies. A side effect/consequence of this feature is that it stomps on the screen reader's ability to provide abbreviated role names. For instance, if the ARIA role is `button`, a screen reader could abbreviate that as "btn" in order to be able to show more information at once on the refreshable braille display. But if an element has an `aria-roledescription` value of "Pizza slice" (to borrow an ETS example), that value overrides what the screen reader should display -- and occupies potentially more braille cells than may be desired. It would be handy if an author could specify "ps" or "slice" or whatever as the braille role name.

With respect to the latter, there are tools which can produce a spoken representation of math expressions. Assistive technologies which can take advantage of such a tool can pass the string along to be spoken. But that spoken representation is likely not what should be displayed in braille. Hypothetically, such a tool could also provide a Nemeth string which should be similarly passed along unmodified. But there's no standard means to expose this information to assistive technologies.

jnurthen:
This feels like a very similar problem to ruby. Maybe we could use whatever solution we come up with for that too.

arnog:
For a given expression, authoring tools might want to produce multiple representations of that expression, for example a spoken representation, a Nemeth Braille representation, a UEB Braille representation and a (plain) braille representation (particularly useful for users that know Braille but may still be learning Nemeth Braille). Having a mechanism to author this would be very useful.

pkra:
@joanmarie @jnurthen could we put this on the WG agenda some time?

pkra:
Some initial ideas: https://github.com/w3c/aria/wiki/*-braille-properties-for-exposing-custom-Braille-descriptions

garconvacher:
I dream of an `aria-label-braille`!
Example with International Phonetic Alphabet:
```
<p>[<span aria-label-braille="⠃⠢⠸⠝⠡⠒⠝⠢">bəˈnɑːnə</span>] est la prononciation du mot "banane" en anglais.</p>
```
Currently, we use:
```
<span class="sr-only">⠃⠢⠸⠝⠡⠒⠝⠢</span><span aria-hidden="true">bəˈnɑːnə</span> 
```
[...]

PR #923, opened 2019-04-03
feat: aria-braillelabel

pkra:
***
<a href="https://pr-preview.s3.amazonaws.com/pkra/aria/pull/923.html#aria-braillelabel" title="Last updated on Mar 4, 2020, 8:09 AM UTC (0605572)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/923/76474e5...pkra:0605572.html" title="Last updated on Mar 4, 2020, 8:09 AM UTC (0605572)">Diff</a>


joanmarie:
Githack link for the current (only) commit: https://raw.githack.com/w3c/aria/5b3de91/index.html#aria-label-braille

jasonjgw:
I support the proposal in general. It isn't clear what "uncontracted braille ASCII" means (and I'm very familiar with braille and braille-related technology). I would suggest either eliminating (my preference) or defining that possibility so that assistive technology developers know precisely what to expect in the string.

johnfoliot:
I have a small concern about introducing a new "pattern" into the aria lexicon - specifically "double-hyphenated" attributes (i.e. aria-label-braille and aria-roledescription-braille). A quick check of the existing ARIA attributes, and this will be the first time that this WG would be introducing that pattern, and so more than anything else - why? 

Previously, even "multi-name" attributes have all been concatenated into one string (ref: Position In Set = aria-posinset), and I would strongly encourage the ARIA WG to maintain that pattern (aria-labelbraille, aria-roledescriptionbraille) if these new attributes are to move forward.

Additionally, if the only reason for the double-hyphen is "readability", the values can always be presented (instructionally) in CamelCase, as elements and attributes are case insensitive: (aria-LabelBraille, aria-RoleDescriptionBraille - although that goes counter to existing attributes as well)

pkra:
Thanks for the comments so far!

@jasonjgw  wrote 

>  It isn't clear what "uncontracted braille ASCII" means (and I'm very familiar with braille and braille-related technology). 
> I would suggest either eliminating (my preference) or defining that possibility so that assistive technology developers know precisely what to expect in the string.

Thanks for calling this out - it is something I wanted to bring up in the meeting. 

I've been trying to capture the previous discussions on the mailing list (in particular with AT developers) and the call in February around this. But I'm not as well versed at Braille Ascii encodings.I got "uncontracted" [from RNIB](https://www.rnib.org.uk/braille-and-other-tactile-codes-portal-braille-codes/uncontracted-grade-1-braille-explained). Would just "Grade 1" make more sense? 

The basic idea from the discussions was that AT should be able to push the content of the Braille label to a Braille display as directly as possible.  A key goal for this SHOULD statement is to avoid additional markup to indicate encoding (in particular to avoid authoring errors).


@johnfoliot wrote

> I have a small concern about introducing a new "pattern" into the aria lexicon
> - specifically "double-hyphenated" attributes (i.e. aria-label-braille and aria-roledescription-braille).
> ... 
>  I would strongly encourage the ARIA WG to maintain that pattern (aria-labelbraille, 
> aria-roledescriptionbraille) if these new attributes are to move forward.

I generally agree. However, I left it as before to connect to previous discussions and because the question had come up whether a more general Braille (or tactile) namespace might be preferential. If I remember correctly, this came up in the context of potential outcomes from the APA pronounciation task force which might face similar questions.

johnfoliot:
@pkra I again caution you on attempting a micro-syntax in ARIA. I'll +1 Jon Gunderson's suggestions (i.e. aria-braillelabel, aria-brailleroledescription) for addressing the need of signifying braille-centric attributes.

jasonjgw:
The problem is that there is no single standard mapping even from the printable ASCII subset of Unicode to a braille representation. There are different ASCII->braille tables used in different countries (e.g., the North american ASCII->braille table is different from the German table, and similarly for other locales - and I'm not even sure that either is officially standardized anywhere).

"Grade 1" doesn't help in this context, as it wouldn't specify any mapping of ASCII characters to braille cells. The only absolutely standard and unambiguous way to do it is with the Unicode block set aside for this purpose, which is why I would prefer to eliminate the ASCII-based option. The only alternative would be to find a suitable normative reference that specifies the mapping of printable ASCII to six-dot braille. In effect, you would be choosing one of the mappings that have been implemented.

pkra:
@johnfoliot  

> @pkra I again caution you on attempting a micro-syntax (i.e. namespacing) in ARIA.

FWIW I agree. I'm documenting the discussion in the WG here.

@jasonjgw wrote 

>  I would prefer to eliminate the ASCII-based option. 

FWIW I would prefer that as well. I'm documenting the discussion in the WG here.

carmacleod:
+1 to @johnfoliot's +1 to @jongund's attribute naming suggestions. :)

Just looking a bit further down the road, and applying the "remove hyphen; camelcase the rest" pattern, they would be reflected as follows:

- aria-braillelabel = ariaBrailleLabel
- aria-brailleroledescription = ariaBrailleRoleDescription

which would sort them together in developer tooling (more useful than having "-braille" at the end).

cc @cookiecrook for confirmation

charmarkk:
+1 to @johnfoliot's and @jongund's suggestions. +1 to @carmacleod's adjustments/formatting.

johnfoliot:
To be clear...

CamelCasing is useful for "instructional" reasons, however attributes **SHOULD** be case-insensitive, and earlier compounded attributes in the spec have not traditionally been rendered in CamelCase  (eg. aria-posinset vs. aria-PosInSet, aria-labeledby vs. aria-LabeledBy, etc.). Consistency in naming conventions should also be considered.

joanmarie:
@johnfoliot and @charmarkk: I believe @carmacleod was specifically talking about the "reflected" properties which is for AOM. Thus the official properties added to ARIA would NOT be camel-cased.

cookiecrook:
I'm having trouble understanding the use case here. Screen reader braille implementations are usually complex, allowing users to customize braille verbosity, use different translation tables in different contexts, etc. 

Perhaps someone could explain the "proposal" since no description was included above. Whatever the case, this edit should not be included without clear example illustrating the utility of the feature, and a clear description as to when/why/where this should be used instead of letting the screen reader do its thing.  

PS. This caveat should definitely be removed as it's not internationalizable.
> or of a string of uncontracted (Grade 1) Braille ASCII.

pkra:
@cookiecrook  thanks for your comment.

Some references to previous discussions:

* https://github.com/w3c/aria/issues/765
* meeting minutes from meetings where this topic was discussed
  * https://lists.w3.org/Archives/Public/public-aria/2019Jan/0028.html
  * https://lists.w3.org/Archives/Public/public-aria/2019Apr/0015.html 
* other email discussions
  * https://lists.w3.org/Archives/Public/public-aria/2019Feb/0003.html
  * https://lists.w3.org/Archives/Public/public-aria/2019Apr/0005.html

> PS. This caveat should definitely be removed as it's not internationalizable.

Yes; this was the agreement during last week's meeting.

cookiecrook:
Commented in detail on #765

mraccess77:
while in general I am not in favor of separate labels I believe this could be useful for situations like a math where we need to show unified English braille or Nemeth code but have something different announced in speech.

pkra:
I've updated the branch with a re-write that stays close to the language in #1097.

index.html
```diff
@@ -10300,6 +10300,56 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
...dition, assistive technologies will be able to customize such braille output according to user preferences. Using only the accessible name, e.g., from content or via <code>aria-label</code> is <strong>almost always</strong> the better user experience and authors are <strong>strongly discouraged</strong> from using <code>aria-braillelabel</code> to replicate <code>aria-label</code>. Instead, <code>aria-braillelabel</code> is meant be used only the accessible name cannot provide an adequate braille representation, i.e., when a specialized braille description is very different from a text description converted to Braille. It is very important to note that when using <code>aria-braillelabel</code> authors are solely responsible to align the attribute value with the document language and clearly communicate the use of this attribute to the user. This is even more important when the value consists of Unicode Braille Patterns because <a>Assistive Technologies</a> will pass such content directly to the user without applying user specific braille translations; in general, authors are <strong>strongly discouraged</strong> from using Unicode Braille Patterns in <code>aria-braillelabel</code>.
```
carmacleod: Change "only the" in this phrase:
> is meant be used only the accessible name cannot provide an adequate braille representation

to use "only **if** the":
> is meant be used only if the accessible name cannot provide an adequate braille representation

carmacleod:
Regarding:
> As required by the accessible name and description computation, user agents give precedence to aria-labelledby over aria-braillelabel when computing the accessible Braille name property.

I don't understand why we wouldn't change the accessible name and description computation to allow `aria-braillelabel` to override `aria-labelledby`?

We have said that:
> The aria-braillelabel property gives authors the ability to override how assistive technologies localize and express the accessible name of an element in Braille.

So it seems kind of strange to me that `aria-braillelabel` can override the accessible name if it was provided by content, or by an `aria-label`, or presumably by host language labelling techniques such as `<label>`, but we wouldn't allow it to override the one case where the accessible name happens to come from `aria-labelledby`? Why is that?

(Of course, I know very little about Braille, and it is very likely that I am missing something here).

sinabahram:
I wish to state an enthusiastic agreement about confusion around why aria-labelledby is overriding aria-braillelabel. Aria-braillelabelledby should absolutely override aria-braillelabel, yes, but cross-modality priority for aria-labelledby seems wrong in my opinion. If we have an element with aria-braillelabel and aria-labelledby, the author has made a clear choice to override the label in braille and to provide an accessible label for non-braille users.

pkra:
Thanks, @carmacleod , @sinabahram! I agree that this is a bit confusing (and the first draft had marked this part TBD because I felt the same way).

When we discussed it on a WG call, the initial agreement was that we wouldn't need "braille labelled by" (or described by) since a braille label can/should be on the target of the labelledBy. This also makes sense since that is where the element gets its name from. Otherwise, we end up "splitting up" names to come from different elements which might be problematic for implementors and authors alike.

That's just to recap -- I'm happy to reconsider.

index.html
pkra: Thanks!

pkra:
I also noticed another piece I had intended to remove that might cause confusion -- there's no such thing as "accessible Braille name property" -- but I'll wait until after today's meeting.

carmacleod:
I'm not asking for "braille labelled by".
I'm just thinking that "braille label" should probably override "labelled by".

pkra:
@carmacleod thanks for clarifying! (And sorry for not responding to each comment individually!)

sinabahram:
Same over here, I also don’t mean to imply the existence/immediate call for an aria-braillelabelledby, though again, for the record, I don’t agree that a need for this does not exist, simply that it’s far less urgent or necessary than aria-braillelabel

jongund:
Isn't `aria-braillelabel` suppose to be a parallel labeling system?   It should be used in addition to the standard label for use by refreshable Braille devices.   So it should not be part of the standard labeling algorithm, and it seems that should not be used to satisfy a labeling requirement for a control or widget.

mraccess77:
braillelabel is an alternative representation that is meant to be display in braille only when the name used for speech is not appropriate and thus we absolutely do not want it to be used in the accessible name calculation unless as an last resort fallback where no other name can be calculated using the algorithm.

carmacleod:
Ok. So I think we're saying that the following sentence should just be completely deleted:
> As required by the accessible name and description computation, user agents give precedence to aria-labelledby over aria-braillelabel when computing the accessible Braille name property.

index.html
```diff
@@ -11053,6 +12346,64 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
...ill be able to customize such Braille output according to user preferences. Using only <code>aria-roledescription</code> is <strong>almost always</strong> the better user experience and authors are <strong>strongly discouraged</strong> from using <code>aria-brailleroledescription</code> to replicate <code>aria-roledescription</code>. Instead, <code>aria-brailleroledescription</code> is meant be used only when <code>aria-roledescription</code> cannot provide an adequate Braille representation, i.e., when a specialized Braille description is very different from a text description converted to Braille. It is very important to note that when using <code>aria-brailleroledescription</code> authors are solely responsible to align the attribute value with the document language and clearly communicate the use of this attribute to the user. This is even more important when the value consists of Unicode Braille Patterns because <a>Assistive technologies</a> will pass such content directly to the user without applying user specific Braille translations; in general, authors are <strong>strongly discouraged</strong> from using Unicode Braille Patterns in <code>aria-brailleroledescription</code>.
```
carmacleod: Change:
> authors are solely responsible to align the attribute value 

to:
> authors are solely responsible for aligning the attribute value 

Also, I don't understand this phrase - how can authors do this? Can this phrase just be deleted?
> and clearly communicate the use of this attribute to the user

mraccess77:
My initial thought is that it's more than just preference over aria-labelledby but over any other accessible name e.g. title attribute, text from content, etc.  That is braillelabel could only be an accessible name when no other name can be calculated using native semantics as well.

jongund:
I am concerned about authors being confused about the use of `aria-braillelabel`, we should be very careful in not making it as any part of the accessible name computation.  There maybe a note about when no accessible name exists it could be one source for identifying an accessible name.    My understanding it is an abbreviated version of the accessible name suitable for refreshable Braille devices.

sinabahram:
Just summarizing my thoughts on aria-braillelabel and accessible name.

 

*	It should not be part of accessible name calculation, as that information is often spoken aurally, and the information that would be most appropriate for Braille is orthogonal in almost all cases to what is appropriate for speech
*	Aria-labelledby should not; however, override aria-braillelabel by exactly the same logic.

 

Therefore, if a node has an aria-braillelabel and an accessible name, independent of how that is calculated, then the accessible name should be spoken and the aria-braillelabel should be brailled.

 

Do we all agree on the above? It sounds like we mostly do, but there may possibly be some objections to my second assertion?

joanmarie:
My observations:

1. I think we should remove any mention of the accessible name calculation. The `aria-braillelabel` value will be exposed to ATs as an attribute on the accessible object to which it applies.

    The accessible name will still be calculated like it always is. The screen readers would speak the accessible name (like they already do). But when displaying that very same object in braille, they will use platform-specific accessibility API to ask if there is a braille label. If there is, they will display that braille label instead of the accessible name.
2. Given that `aria-braillelabel` won't participate in the accessible name calculation, and given that ATs will expect to find the braille label as an attribute on the object being presented, I think the simplest thing to do is have authors place the braille label on the element it applies to and NOT place it on the `aria-labelledby` element which might also apply to the element.

sinabahram:
Joanie, I agree with both of those.

pkra:
Thanks everyone for the many and thoughtful comments.

I'll make the changes suggested by @joanmarie in https://github.com/w3c/aria/pull/923#issuecomment-575312963 ahead of the next meeting.

I have two follow-up questions: 

* What should happen when an element uses aria-labelledby and the target uses aria-braillelable? 
* What should happen when an element uses aria-braillelabel and also aria-labelledby and the target also uses aria-braillelabel?

joanmarie:
> I have two follow-up questions:
> 
> * What should happen when an element uses aria-labelledby and the target uses aria-braillelable?

This one I'm not sure about. But once we have consensus, I think it would be good to add language to the spec about it.

> * What should happen when an element uses aria-braillelabel and also aria-labelledby and the target also uses aria-braillelabel?

My thoughts: `aria-braillelabel` on the element wins for the element presentation. But if the user later navigates into the content/target referenced by `aria-labelledby`, presentation of that content in braille would respect its `aria-braillelabel`.

jongund:
Since only the author knows what the format Braille (e.g. Grade 2, Nemeth Math,...) the `aria-braillelabel` should NOT be included in any part of the accessible name or description calculation.   It provides authors a technique to provide an alternative label (e.g. abbreviated) for a refresh-able Braille devices.

joanmarie:
> Since only the author knows what the format Braille (e.g. Grade 2, Nemeth Math,...) the `aria-braillelabel` should NOT be included in any part of the accessible name or description calculation. It provides authors a technique to provide an alternative label (e.g. abbreviated) for refresh-able Braille devices.

To whom and what specifically is this in response? (I think we all agree that it would not participate in the accessible name calculation.)

sinabahram:
Here’s how I think of aria-braillelabel, aria-label, aria-labelledby, etc.

 

At the end of the day, the accessible name is being set. We must introduce one more concept (just for discussion, not anywhere technical), and that is the Braille name.

 

So, at the AT’s discretion, the accessible name and braille name are presented to the user via various modalities that the user and AT have agreed to and that we do not need to care about.

 

What we care about is when something should modify these things. For accessible name, I won’t even try to summarize that because we have an entire accessible name calculation algorithm, so that’s what can modify accessible name. For our concept of braille name, all we are saying is that if a braillelabel exists, then it defines braille name’s value, so that for example it is not presumably the same as accessible name or an abbreviation therein, but again, not our concern on the output side.

 

So, if we simply state a single rule, which is that label, labelledby, etc. in no way effect braille name, then all other questions can be answered from first principles.

 

If we want a braille label to be referenced, then we need to do the work and define aria-braillelabelledby. Without it, we absolutely do not need to be having braillelabel being brought into the picture for an element because it uses aria-label, etc., otherwise, we have an absolute mess on our hands, or so I humbly feel.

jongund:
@sinabahram 
Rather than try to go the route of another complex labelledby exercise, which is still be refined for `aria-labelledby`, maybe the next step should be the AOM route, with having a DOM property to set/read an `ariaBrailleLabel` property through javascript.

pkra:
The latest commit addresses @joanmarie's comment at https://github.com/w3c/aria/pull/923#issuecomment-575312963 and includes some minor fixes.

Let's discuss labelledby on the next call (and maybe merge #1097 as well).

sinabahram:
I just wish to reiterate something that has now been repeatedly said. I don’t believe anyone is advocating for braillelabelledby at this time, so I don’t think we need to do anything about that at the moment. I just want to make sure we have total consensus that aria-labelledby has nothing what-so-ever to do with braillelabel. That is all.

pkra:
As per today's discussion, we'll think about examples around labelledby situations to have more practical discussions.

pkra:
@sinabahram thank you for the review! I'll work those in for next week.

Regarding the complicated sentence on braille patterns, I've been wondering if we should add a term to https://w3c.github.io/aria/#terms. At least we could have Unicode Braille Pattern there (so we don't have to repeat the range). Maybe we can even add something that describes "non-trivial" Braille patterns (i.e., not just Dot 0).

sinabahram:
That sure sounds wonderful to me, Peter, especially if we can have a definition for something like non-empty Braille string or what not, so we can massively improve the readability here. I 100% support that idea, and feel free to loop me into any way of helping that e.g. assigning ticket or collaborating on definition or whatever.

pkra:
@sinabahram I've just pushed an update hopefully addressing your review.

The one exception was the point about using h5 elements since I don't know what regions you were referring to. Also, the entire spec has only three h5's so  I'm not sure what the broader editorial policy is.

Regarding moving the complicated sentence on braille patterns to the terms section, I'd like to first merge the two PRs.

index.html
pkra: I missed this earlier -- sorry @carmacleod. I'll be happy to make some changes but since this particular note is identical to the one for brailleroledescription, would you be ok with revisiting it after both PRs are merged? That way changes are more easily done for both.

As for the question, this came out of the various discussion for the note. The idea is to try to warn authors that when they use these attributes, they risk causing confusion and they need to mitigate that confusion. I'd compare this to role=application where authors are told to provide instructions without a lot of specifics.

pkra:
@sinabahram I've updated the code example here as well so this is hopefully ready.

Unfortunately, I'm not sure what happened to this PR -- a lot of commits from other people are now showing up. The link to the diff preview is hopelessly noisy but I hope the preview is ok but if it causes an issue with review, let me know.

sinabahram:
Ok, a few things on this one:

 

1, you have:

Note that  <https://pr-preview.s3.amazonaws.com/pkra/aria/pull/923.html#dfn-assistive-technology> Assistive Technologies with braille support can convert the acessible name to Braille

 

‘accessible’ is what you mean

 

2, you say:

“The following example shows the use of aria-braillelabel to indicate that a button's description has a particular braille contraction.”

 

This needs to be reworded, and I’m sorry but I do feel pretty strongly about this editorial comment. A “Braille contraction” is a term of art. It means contractions used in higher grades of Braille e.g. ‘p’ by itself in a sentence means ‘people’ and dots-2-3-4-6 means ‘the’. We must not use the word contraction here.

 

3, I feel that wording and example are confusing in general. The real power comes from being able to put Nemeth in there, but we obviously don’t wish to complicate the spec with that. I get it. However, a plain language example would be something like “4 stars” for an aria-label and “****” for a braillelabel. I can also come up with some other examples too, but that’s a really simple case where the Braille and verbal messages really differ from one another, and it comes from the real world.

 

That’s all I got on this.

carmacleod:
> “The following example shows the use of aria-braillelabel to indicate that a button's description has a particular braille contraction.”

@sinabahram What word would you use instead of "contraction"? Maybe "short form", "shorthand", "abbreviation"?

@pkra In the same sentence, I think the word "description" should be changed to either "name" or "label" so that there's no possible confusion with the button's accessible description.

sinabahram:
Oh, good catch on label VS description. That’s excellent!

 

As for something other than contraction, I do like short form or shorter form or abbreviated, though that gets into contraction, and short form is far more plain language conformant than abbreviated is.

pkra:
Thanks, @sinabahram and @carmacleod for the additional comments. Before I get started to address them, a quick question.

@sinabahram could you confirm that I've addressed the issues raised in your review (cf. https://github.com/w3c/aria/pull/923#pullrequestreview-347731101)?

@carmacleod did you see my (very late) response to your inline review comment? Again, I'm sorry for missing it earlier.

mraccess77:
We could say braille code.

sinabahram:
Braille code, I think is also confusing. The point being conveyed is that it’s shorter. That’s all, not that it’s in Braille, which is a secondary point being conveyed.

mraccess77:
Hi @sinabahram I'm not sure why it has to be smaller -- the purpose of this property is to provide something that can be displayed in braille and not sent to text to speech.  It's possible that uncontracted braille could be used and the code in fact could be longer than what is meant for text to speech when you apply code changers, etc.

sinabahram:
I can’t agree more that this will be used for different, not just shorter, Braille labels, but I was trying to satisfy the tone of the original text.

pkra:
How about: 

"The following example shows the use of aria-braillelabel to customize a button's name in braille output."

Followed by

```
<button aria-label="4 stars" aria-braillelabel="****">
  <img alt="" src="images/stars.jpg">
</button>
```

index.html
carmacleod: I have opened https://github.com/w3c/aria/issues/1205 to address this after aria-braillelabel is merged.

pkra: Thanks!

carmacleod:
@pkra 
> carmacleod did you see my (very late) response to your inline review comment? Again, I'm sorry for missing it earlier.

I've opened https://github.com/w3c/aria/issues/1205 to address it after this PR is merged. Sorry I didn't answer earlier!

pkra:
@sinabahram @carmacleod @mraccess77 did you have any thoughts on my proposed wording and example in my comment at https://github.com/w3c/aria/pull/923#issuecomment-588148647 ?

mraccess77:
The word output may imply that it has already been translated - which is likely not the case.   Code is more clear to me.

carmacleod:
Maybe change the example code to use the img alt for the accessible name?
```html
<button aria-braillelabel="****">
  <img alt="4 stars" src="images/stars.jpg">
</button>
```

sinabahram:
I agree with the suggestion of using alt on the img instead of aria-label.

pkra:
@carmacleod thanks for the suggestion, I'm happy with that, too.

If anyone has a better suggestion than "braille output" then I'm all ears.

sinabahram:
I really think we should keep it as “Braille output”. I’m not a fan of code because that means specific things in Braille.

pkra:
I've just pushed an update that modifies the code example as per the discussion; I also adjusted the sentence after the pre element to match the changes.

I think this addresses all comments so far - with the exception of #1205. Please let me know if I missed a comment and if there are any additional comments we should address before merging. (And #1205 will ensure an opportunity for further commens later on, too.)

pkra:
@jnurthen I requested a review from you as per yesterday's telco. Should I clean up the branch before or after your review? (I must have rebased badly at some point, introducing more than 100 unrelated commits).

jnurthen:
> @jnurthen I requested a review from you as per yesterday's telco. Should I clean up the branch before or after your review? (I must have rebased badly at some point, introducing more than 100 unrelated commits).
Can you please fix... I can't really review in this state.

pkra:
@jnurthen this was a bit weird. I tried a new PR from the the same branch and github's preview shows none of those commits. 

 Anyway, I'll force push and see if that improves things.

pkra:
@jnurthen the git history should be reasonable now.

pkra:
@jnurthen is it reviewable now or do you need me to try a completely squashed version?

jnurthen:
> @jnurthen is it reviewable now or do you need me to try a completely squashed version?

Yes - only 1 file changed now so all good.

PR #924, opened 2019-04-03
feat: aria-brailleroledescription

pkra:
***
<a href="https://pr-preview.s3.amazonaws.com/pkra/aria/pull/924.html" title="Last updated on Sep 27, 2019, 12:52 PM UTC (2579a05)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/924/f2caeec...pkra:2579a05.html" title="Last updated on Sep 27, 2019, 12:52 PM UTC (2579a05)">Diff</a>


joanmarie:
Githack link for the current (only) commit: https://raw.githack.com/w3c/aria/47de3e9/index.html#aria-roledescription-braille

jasonjgw:
I have the same comment as in #923.

johnfoliot:
(Duplicate from #923)

I have a small concern about introducing a new "pattern" into the aria lexicon - specifically "double-hyphenated" attributes (i.e. aria-label-braille and aria-roledescription-braille). A quick check of the existing ARIA attributes, and this will be the first time that this WG would be introducing that pattern, and so more than anything else - why?

Previously, even "multi-name" attributes have all been concatenated into one string (ref: Position In Set = aria-posinset), and I would strongly encourage the ARIA WG to maintain that pattern (aria-labelbraille, aria-roledescriptionbraille) if these new attributes are to move forward.

Additionally, if the only reason for the double-hyphen is "readability", the values can always be presented (instructionally) in CamelCase, as elements and attributes are case insensitive: (aria-LabelBraille, aria-RoleDescriptionBraille - although that goes counter to existing attributes as well)

pkra:
I've responded at https://github.com/w3c/aria/pull/923#issuecomment-479789757

charmarkk:
Responsed on #923.

NSoiffer:
Apologies if this has already been discussed...

The spec currently has these two clauses about the value of `aria-brailleroledescription`:

> 2. The value of aria-brailleroledescription is not empty or does not contain only whitespace characters.
> 3. The value of aria-brailleroledescription SHOULD consist of a string of Unicode Braille Patterns (U+2800..U+28FF).

These seem partially redundant as '3' says whitespace characters should not be part of the value.
These two clauses can be consolidated into just:

- The value of `aria-brailleroledescription` SHOULD consist of a non-empty string of Unicode Braille Patterns (U+2800..U+28FF).

Also, I'm a little dubious that the phrase "Unicode Braille Patterns" is proper. The block is called "Braille Patterns" and the characters have names that include that phrase, but technically, a string consists of _characters_, so "Unicode Braille Pattern characters" would be more correct. Also correct would be "characters from the Unicode Braille Patterns block". I suggest the current two clauses should be replaced with:
- The value of `aria-brailleroledescription` SHOULD consist of a non-empty string of Unicode Braille Pattern characters (U+2800..U+28FF).

pkra:
@NSoiffer:

> These seem partially redundant as '3' says whitespace characters should not be part of the value.

Happy to change if the WG agrees but they don't seem redundant to me. Point 2 is (essentially) a MUST (cf. also AT guidance to ignore it otherwise), point 3 is a SHOULD. 

> "characters from the Unicode Braille Patterns block"

Sounds good to me but happy to get more feedback.

pkra:
@joanmarie I've incorporated the changes from TPAC (as per https://github.com/w3c/aria/issues/765#issuecomment-531626355). Please take a look.

cookiecrook:
At TPAC we agreed that this should not require (and should actively discourage) using "characters from the Unicode Braille Patterns block" because, in almost all instances, the web author will not know what these should be, even if they are an expert in braille.

As the simplest example, the web author has no way of knowing whether the user reads uncontracted (Grade 1) or contracted (Grade 2) braille, so this attribute would be formatted as a non-braille unicode string. For example, the screen reader could translate `aria-roledescription= "denomination" aria-brailleroledescription="den"` into the user's preferred Grade 1 (⠙⠊⠝) or Grade 2 (⠙⠔) braille rendering. 

Other examples author-encoded braille strings include when a user is fluent in multiple languages, but only fluent in one variant of braille. For example, a member of my team is fluent in reading, typing, and speaking English, but only reads braille using his native German braille table. VoiceOver can provide that to him, but a web author cannot.

cookiecrook:
We also agreed that it may be okay to allow unicode braille with sufficient warnings, and no examples of this usage. The only context I can think of where this is necessary is testing, where a user may be learning tactile braille through remote instruction. E.g. a student learning the differences between US English versus UEB, in Grade 1 or 2.

cookiecrook:
I am skeptical of suggesting the unicode braille block is necessary even in the case of Math polyfills like MathJax. For the same reasons listed above, the polyfill can't know whether the user prefers to read Math in Nemeth, UEB Math, or Computer Braille.

pkra:
@cookiecrook thanks for re-iterating those points from TPAC. Is there a particular change you have in mind for the latest changes in this PR?

zorkow:
@cookiecrook It does make a difference for polyfills like MathJax.
As of v3, MathJax exposes both speech and Braille via Unicode in its accessibility extension. As aria-label however we can currently only expose speech. Its language depends on the localisation the user (or page author) chooses, as we also don't know if the user wants to read their maths in English, French, or Spanish. So selecting a Braille dialect is just another l10n option choice. 
The aria-label depends on that choice, so will the aria-braille.
I think the last thing readers will want is the maths speech translated into Braille, whatever grade.

index.html
```diff
@@ -11053,6 +11053,64 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="property" id="aria-brailleroledescription">
+			<pdef>aria-brailleroledescription</pdef>
+			<div class="property-description">
+				<p>Defines a human-readable, author-localized Braille description for the <a>role</a> of an <a>element</a>.</p>
```
cookiecrook: I think this could be better described as: "Defines a human-readable, author-localized short description for the role of an element, which is intended to be converted into Braille."

index.html
```diff
@@ -11053,6 +11053,64 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Some <a>assistive technologies</a>, such as screen readers, present the role of an element as part of the user experience. Such assistive technologies typically localize the name of the role, and they may customize it as well. Users of these assistive technologies depend on the presentation of the role name, such as "region," "button," or "slider," for an understanding of the purpose of the element and, if it is a widget, how to interact with it.</p>
+				<p>The <code>aria-brailleroledescription</code> property gives authors the ability to override how <a>assistive technologies</a> localize and express the name of a role in Braille. Thus inappropriately using <code>aria-brailleroledescription</code> may inhibit users' ability to understand or interact with an element on Braille interfaces. Authors SHOULD limit use of <code>aria-brailleroledescription</code> to clarifying the purpose of non-interactive container roles like <rref>group</rref> or <rref>region</rref>, or to providing a <em>more specific</em> description of a <rref>widget</rref> in a Braille context.</p>
```
cookiecrook: Editorial nit: The proper nouns: Braille, Web, Internet, should be capitalized, however, their adjective counterparts should not. E.g. "A web page on the Web." "She read Braille on a braille display." IOW, the first use of the word on this line ("…the role in Braille.") is capitalized correctly. The second instance ("…in a Braille context.") should be lowercased.

index.html
```diff
@@ -11053,6 +11053,64 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Authors MUST NOT use <code>aria-brailleroledescription</code> without providing a <em>more specific</em> <code>aria-roledescription</code>. In general, <code>aria-brailleroledescription</code> should only be used in rare cases when a <code>aria-roledescription</code> does not suffice.</p>
```
cookiecrook: s/does not suffice/is excessively verbose when rendered in Braille./

index.html
```diff
@@ -11053,6 +11053,64 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
...ill be able to customize such Braille output according to user preferences. Using only <code>aria-roledescription</code> is <strong>almost always</strong> the better user experience and authors are <strong>strongly discouraged</strong> from using <code>aria-brailleroledescription</code> to replicate <code>aria-roledescription</code>. Instead, <code>aria-brailleroledescription</code> is meant be used only when <code>aria-roledescription</code> cannot provide an adequate Braille representation, i.e., when a specialized Braille description is very different from a text description converted to Braille. It is very important to note that when using <code>aria-brailleroledescription</code> authors are solely responsible to align the attribute value with the document language and clearly communicate the use of this attribute to the user. This is even more important when the value consists of Unicode Braille Patterns because <a>Assistive technologies</a> will pass such content directly to the user without applying user specific Braille translations; in general, authors are <strong>strongly discouraged</strong> from using Unicode Braille Patterns in <code>aria-brailleroledescription</code>.
```
cookiecrook: I don't really understand this note. It seems to be burying the fact that most authors should never use EITHER aria-roledescription OR aria-brailleroledescription.

index.html
```diff
@@ -11053,6 +11053,64 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				</p>
+				<p>User agents MUST NOT expose the <code>aria-brailleroledescription</code> property if any of the following conditions exist:</p>
+				<ol>
+					<li>The element to which <code>aria-brailleroledescription</code> is applied does not have a valid WAI-ARIA <code>role</code> or does not have an implicit WAI-ARIA role semantic.</li>
+					<li>The value of <code>aria-brailleroledescription</code> is empty or contains only whitespace characters or contains only Braille Pattern dots-0 (U+2800).</li>
+					<li>The element to which <code>aria-brailleroledescription</code> is applied does not have a valid WAI-ARIA <code>aria-roledescription</code>.</li>
+					<li>The element to which <code>aria-brailleroledescription</code> is applied has a valid WAI-ARIA <code>aria-roledescription</code> that is identical to a WAI-ARIA <code>role</code> or an implicit WAI-ARIA role semantic.</li>
```
cookiecrook: This line invalidates using `img` as a short braille description. Was that intentional? Same thing with "toc" if that one makes it from the DPUB module into the core spec.

index.html
```diff
@@ -11053,6 +11053,64 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				</ol>
+				<p><a>Assistive technologies</a> SHOULD use the value of <code>aria-brailleroledescription</code> when presenting the role of an element in Braille, but SHOULD NOT change other functionality based on the role of an element that has a value for <code>aria-brailleroledescription</code>. For example, an assistive technology that provides functions for navigating to the next <rref>region</rref> or <rref>button</rref> SHOULD allow those functions to navigate to regions and buttons that have an <code>aria-brailleroledescription</code>.</p>
```
cookiecrook: I don't understand the relevance of the example.

index.html
```diff
@@ -11053,6 +11053,64 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p><a>Assistive technologies</a> SHOULD expose the <code>aria-brailleroledescription</code> property as follows:</p>
+				<ol>
+					<li>If the value of <code>aria-brailleroledescription</code> consists only of Unicode characters that are not Unicode Braille Patterns, translate the value according to the user's preferred translation table.</li>
+					<li>If the value of <code>aria-brailleroledescription</code> consists only of Unicode Braille Patterns characters, pass the value to the user without translation.</li>
+				</ol>
+				<p>The following two examples show the use of <code>aria-brailleroledescription</code> to indicate that a button's description has a particular Braille contraction.</p>
```
cookiecrook: > "the following two examples"

"the following example"

index.html
cookiecrook: Perhaps "abbreviated role description"

index.html
cookiecrook: There are a bunch of other instances of this throughout the PR.

index.html
pkra: Thanks, @cookiecrook, I'll make a new PR to address this.

index.html
pkra: Thanks.

index.html
pkra: Thanks.

index.html
pkra: This note has been expanded several times over the lifetime of this PR. I agree it's a bit long-winded. I'll think about possible improvements (first guess: moving the second part to the front), but I'd be happy for any suggestions.

index.html
pkra: I admit I struggle to answer that. The relevant commit is from June and the commit message explicitly mentions this change. I would have said I added it after feedback from the group, but I can't find any minutes around the time that match this idea.

index.html
pkra: This was copied from aria-roledescription. Does it make sense there but not here?

index.html
pkra: Thanks.

pkra:
@cookiecrook I've created https://github.com/w3c/aria/pull/1097 with the first couple of changes.

index.html
joanmarie: What about:

> As a general rule, representing an object's role is a job best left to the assistive technologies which have been configured by the user based on their specific needs and preferences. Because `aria-roledescription` overrides this representation, it should only be used when absolutely necessary.
>
> `aria-brailleroledescription` makes it possible for authors who are knowledgeable about Braille in general, and about their users' needs in particular, to provide a better experience than would be had from a braille translation of the `aria-roledescription` value. Authors who lack this knowledge should not use `aria-brailleroledescription` because a long, but correctly-translated, braille string is preferable to a shorter, but incorrect one.
>
> Authors are also strongly discouraged from using Unicode Braille Patterns. Providing the value of `aria-brailleroledescription` as text makes it possible for assistive technologies to apply the user's preferred translation tables thus increasing the likelihood that the result will be closer to what the user expects.

pkra: Sounds good to me. @cookiecrook what do you think?

index.html
cookiecrook: Okay, I think I understand now. I might suggest something that references the comment in the context of the non-braille roledescription. ~"As with `aria-roledescription`, [change the role presentation, but don't change the functionality…]"

pkra: Thanks, @cookiecrook. I'll add that.

NSoiffer:
I commented on the use of the phrase "Unicode Braille Patterns" earlier and the distinction between the block name and an individual _character_ -- strings consist of _characters_ not patterns, symbols, or glyphs. I'm happy to see that in one instance that distinction is properly used. However, since the time of my comment, there have been several changes to text and I think improper usage has creeped in again.

Examples where I think changes are warranted so that the sentence is technically correct:
1. The value of aria-brailleroledescription consists of a string of either Unicode with no Unicode Braille Patterns (U+2800..U+28FF) or consists of a string of only Unicode Braille Patterns (U+2800..U+28FF) while not only containing Braille Pattern dots-0 (U+2800).
2. If the value of aria-brailleroledescription consists only of Unicode characters that are not Unicode Braille Patterns, translate the value according to the user's preferred translation table

I suggest these become:
1. The string value of aria-brailleroledescription does not contain any characters in Unicode Braille Patterns (U+2800..U+28FF) or consists of only characters in the range U+2801..U+28FF (Unicode Braille Pattern block excluding dots-0).
  - Note: the previous statement mentioned it must contain "either Unicode or...". In HTML, all attributes value strings much contain Unicode values, so that statement is not a restriction. In fact, [not all Unicode values are legal,](https://www.w3.org/TR/2012/WD-html-markup-20120329/syntax.html#syntax-text) so it isn't even correct.
2.  If the value of aria-brailleroledescription does not contain Unicode Braille Pattern characters, translate the value according to the user's preferred translation table.
  - Note: The above is the first clause. I think the second clause would be better written as "Otherwise, pass the value to the user without translation." I don't think it is helpful to write a condition on the second clause as there is no alternative third clause (i.e., AT should do one of those two things).

pkra:
@NSoiffer copied to https://github.com/w3c/aria/pull/1097#issuecomment-562260784

PR #942, opened 2019-04-11
Remove implicit value of aria-level from heading role

melanierichards:
Addresses issue #854

***
<a href="https://pr-preview.s3.amazonaws.com/melanierichards/aria/pull/942.html" title="Last updated on Apr 25, 2019, 3:57 AM UTC (490eb68)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/942/78b9178...melanierichards:490eb68.html" title="Last updated on Apr 25, 2019, 3:57 AM UTC (490eb68)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3146,10 +3146,6 @@
 						<th class="role-presentational-inherited-head" scope="row">Inherits Presentational:</th>
 						<td class="role-presentational-inherited"> </td>
 					</tr>
-					<tr>
-						<th class="implicit-values-head" scope="row">Implicit Value for Role:</th>
-						<td class="implicit-values">Default for <pref>aria-level</pref> is <code class="default">2</code>.</td>
-					</tr>
 				</tbody>
 			</table>
 		</div>
--- a/index.html
+++ b/index.html
@@ -3070,7 +3070,7 @@
 			<rdef>heading</rdef>
 			<div class="role-description">
 				<p>A heading for a section of the page.</p>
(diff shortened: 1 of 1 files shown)


jnurthen:
[link for review](https://raw.githack.com/melanierichards/aria/heading-default/index.html#heading)

jnurthen:
I think 

> If headings are organized into a logical outline, the aria-level attribute is used to indicate the nesting level.
needs rewording now that aria-level is required.

carmacleod:
Something like:
> Headings SHOULD be organized into a logical outline, and the aria-level attribute MUST be used to indicate the nesting level.

?

melanierichards:
@jnurthen @carmacleod 

In the spirit of simplifying shoulds/musts, how about:

> To ensure headings are organized into a logical outline, authors MUST use the `aria-level` attribute to indicate the proper nesting level.

melanierichards:
Made the suggested change!

accdc:
As a quick question about the accessibility tree mappings, what heading level will a heading have if there is no implicit level?
Thanks, Bryan

jnurthen:
> As a quick question about the accessibility tree mappings, what heading level will a heading have if there is no implicit level?
> Thanks, Bryan

http://w3c.github.io/aria/#document-handling_author-errors_states-properties states

> If a required WAI-ARIA attribute for a given role is missing, user agents SHOULD process the attribute as if the values given in the following table were provided.
 In the table the fallback for aria-level is 2

So not supplying a heading level is now an author error, but if a browser is following the fallback guidance it should still expose it with a level of 2.

accdc:
Excellent, thanks, that was my only concern with this.

Issue #853, opened 2018-12-03
Combobox has aria-expanded required, but there is a default value for aria-expanded


jnurthen:
The default value is currently false. The proposal is to remove the default value from the role definition and rely on the fallback value in https://w3c.github.io/aria/#document-handling_author-errors

PR #943, opened 2019-04-11
Remove default value for aria-expanded from combobox role

melanierichards:
Addresses #853


jnurthen:
[link for review](https://raw.githack.com/melanierichards/aria/combobox-default/index.html#combobox)

jnurthen:
Need to remove 

> Elements with the role combobox have an implicit aria-expanded value of false.

carmacleod:
Also, consider changing:
> Typically, the default state of a combobox is collapsed.

to:
> Typically, the initial state of a combobox is collapsed.

melanierichards:
Updated per @jnurthen and @carmacleod's feedback!

PR #944, opened 2019-04-11
Move some properties from "required" to "supported" on scrollbar role

melanierichards:
Addresses #858

***
<a href="https://pr-preview.s3.amazonaws.com/melanierichards/aria/pull/944.html" title="Last updated on May 30, 2019, 3:07 PM UTC (5a09286)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/944/78b9178...melanierichards:5a09286.html" title="Last updated on May 30, 2019, 3:07 PM UTC (5a09286)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6376,16 +6376,19 @@
 						<td class="role-required-properties">
 							<ul>
 								<li><pref>aria-controls</pref></li>
-								<li><pref>aria-orientation</pref></li>
-								<li><pref>aria-valuemax</pref></li>
-								<li><pref>aria-valuemin</pref></li>
 								<li><pref>aria-valuenow</pref></li>
 							</ul>
 						</td>
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
-						<td class="role-properties"> </td>
+						<td class="role-properties">
+							<ul>
+								<li><pref>aria-orientation</pref></li>
+								<li><pref>aria-valuemax</pref></li>
+								<li><pref>aria-valuemin</pref></li>
+							</ul>
+						</td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
@@ -6417,7 +6420,6 @@
 							Default for <pref>aria-orientation</pref> is <code class="default">vertical</code>. <br/>
 							Default for <pref>aria-valuemin</pref> is <code class="default">0</code>. <br/>
 							Default for <pref>aria-valuemax</pref> is <code class="default">100</code>. <br/>
-							Default for <pref>aria-valuenow</pref> is half way between <code class="default">aria-valuemax</code> and <code class="default">aria-valuemin</code>.
 						</td>
 					</tr>
 				</tbody>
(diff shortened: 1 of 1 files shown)


jnurthen:
[link for review](https://raw.githack.com/melanierichards/aria/scrollbar-defaults/index.html#scrollbar)

carmacleod:
This sentence:
> Authors MUST set the aria-valuemin, aria-valuemax, and aria-valuenow attributes.

should probably just be:
> Authors MUST set the value of the aria-valuenow attribute.

The list that follows the above sentence may need rewording as well:
> If missing, their implicit values follow the same rules as the HTML range input type:
> - If aria-valuemin is missing or not a number, it defaults to 0 (zero).
> - If aria-valuemax is missing or not a number, it defaults to 100.
> - If aria-valuenow is missing or not a number, it defaults to the value half way between aria-valuemin and aria-valuemax.
> - If aria-valuenow is present but less than aria-valuemin, it defaults to the value of aria-valuemin.
> - If aria-valuenow is present but greater than aria-valuemax, it defaults to the value of aria-valuemax.

melanierichards:
Methinks I was a little too literal in addressing the issues for these and missed some prose update opportunities, thanks @carmacleod for all your feedback! :)

Made some changes to the prose, left repair techniques in there for now because Core-AAM doesn't seem to cover the overflow repair techniques (where valuenow is less than valuemin, for example). I imagine this may warrant further discussion, especially given that, from a quick test, the Windows browsers aren't following these repair techniques: https://jsfiddle.net/r39sqne2/2/

index.html
```diff
@@ -6318,10 +6318,13 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
+				<p>Authors MAY set <pref>aria-valuemin</pref> and <pref>aria-valuemax</pref> to indicate the minimum and maximum thumb position. Otherwise, their implicit values follow the same rules as the <a href="https://www.w3.org/TR/html5/forms.html#range-state-%28type=range%29" title="Range state">HTML range input type</a>:</p>
+				<ul>
+				    <li>If <code>aria-valuemin</code> is missing or not a <a href="#valuetype_number">number</a>, it defaults to 0 (zero).</li>
+				    <li>If <code>aria-valuemax</code> is missing or not a <a href="#valuetype_number">number</a>, it defaults to 100.</li>
+				</ul>
+				<p>Authors MUST set the <pref>aria-valuenow</pref> attribute to indicate the current thumb position. If this attribute is missing or an unexpected value, the repair techniques follow the same rules as the <a href="https://www.w3.org/TR/html5/forms.html#range-state-%28type=range%29" title="Range state">HTML range input type</a>:</p>
```
mcking65: As specified in the issue, we are removing the default value and resorting to optional browser fallbacks specified in section 9.2. This paragraph and the following list are worded as there is a default. Suggest removing the list and rewording the paragraph to capture both the author and browser requirements as follows:

>Authors MUST set the aria-valuenow attribute to a value that represents the current thumb position. If aria-valuenow is missing or has an unexpected value, browsers MAY implement the repair techniques specified in the section describing handling author errors in states and properties, which are equivalent to the repair techniques for the HTML range input type.

Finally, aria-valuemin and aria-valuemax need to be removed from the table in section 9.2.

melanierichards: Thanks for the suggestion, I like this messaging suggestion. Made the update and moved the reqs to the fallback table (was light on details before, like `slider`).

Similarly to `slider`, #949 asks what to do with fallback techniques for supported but not required attrs in the table.

melanierichards: Removed supported but not required attrs from the fallback table as discussed on the call

melanierichards:
Should be up to date now based on 2019-05-23 call.

Issue #845, opened 2018-11-19
Editorial: Merge redundant points in Handling Author Errors section


carmacleod:
The first part of the [Handling Author Errors](https://w3c.github.io/aria/#document-handling_author-errors) section contains a sample list of logical validation items that UA are not responsible for.

Point 2. "Correct usage with regard to DOM tree structure, such as an aria-activedescendant being a DOM-descendant of the element with the relation."

and point 5. "Determining whether aria-activedescendant actually points to a descendant or another owned element."

are very similar, and should probably just be merged into a single point.

PR #947, opened 2019-04-11
Editorial: Merge redundant points in Handling Author Errors section #845

carmacleod:
Reworded Point 2 using a different example that does not intersect with Point 5, as follows:

Point 2. Correct usage with regard to DOM tree structure, such as an element being owned by more than one other element.

Point 5. Determining whether aria-activedescendant actually points to a DOM descendant or owned element.


jnurthen:
fixes #845

jnurthen:
@joanmarie I think this is editorial but want a 2nd opinion please.

joanmarie:
Yeah, I think it's editorial.

Issue #841, opened 2018-11-08
Tree Inclusion focus/activedescendant requirements need clarification


cookiecrook:
From: 
https://w3c.github.io/aria/#tree_inclusion

This bullet covers two disparate requirements and should therefore be split into separate bullets with clarifications: one regarding focus, and the other regarding activedescendant.  

> Elements that are focusable, or have an ID attribute and an ancestor with the aria-activedescendant attribute that matches the implicit or explicit semantics of the required context role. In either case, the element may receive focus and may fire an accessibility API focus event.

Comments individually:

> Elements that are focusable, […] may receive focus and may fire an accessibility API focus event.

Hidden elements that are focusable generally are not exposed unless they become focused. I'd suggest that this be change to include something like both "visible elements that are focusable" and ~"hidden elements if/when they become focused."  

> Elements that have an ID attribute and an ancestor with the aria-activedescendant attribute that matches the implicit or explicit semantics of the required context role. 

This is confusing. I can't parse "aria-activedescendant attribute that matches the implicit or explicit semantics of the required context role." The aria-activedescendant attribute value is an IDREF, so I don't know how an IDREF would match "the implicit or explicit semantics of the required context role."

carmacleod:
@cookiecrook 

This similar paragraph in [Core-AAM General rules for exposing WAI-ARIA semantics](https://w3c.github.io/core-aam/#mapping_general) seems easier to parse:

> A DOM element is included if it has semantic significance. As a first approximation, it is significant if it has an explicit or implicit WAI-ARIA role, state, or property, is focusable or interactive, or bears a relationship with some other element (e.g., the element referenced by aria-controls).

We are planning to move some of the Core-AAM content into the ARIA spec, i.e. replacing or augmenting what is currently there. (See https://github.com/w3c/core-aam/issues/28)

carmacleod:
@cookiecrook 
> I can't parse "aria-activedescendant attribute that matches the implicit or explicit semantics of the required context role."

I can't parse that either. :smile: Must be a cut/paste error. I changed it to:
> "aria-activedescendant attribute that matches the ID"

So the single bullet split into 2 separate bullets now reads:

- Elements that are focusable may receive focus and may fire an accessibility API focus event.
- Elements that have an ID attribute, and are owned by or have an ancestor with the aria-activedescendant attribute that matches the ID, may receive focus and may fire an accessibility API focus event.

I added the bit about "owned by" because I think it needs to be there - correct me if I am wrong.

[...]

PR #950, opened 2019-04-11
Tree Inclusion focus/activedescendant requirements need clarification…

carmacleod:
… #841

Link for review: https://raw.githack.com/w3c/aria/car/issue841/index.html#tree_inclusion

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/950.html" title="Last updated on Oct 17, 2019, 5:30 PM UTC (cf13666)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/950/78b9178...cf13666.html" title="Last updated on Oct 17, 2019, 5:30 PM UTC (cf13666)">Diff</a>


carmacleod:
Ping @cookiecrook - are you able to give this a quick review?

The only change is the 2nd bullet in [7.2 Including Elements in the Accessibility Tree](https://raw.githack.com/w3c/aria/car/issue841/index.html#tree_inclusion) - the one with 3 sub-bullets. I think it's clearer now (I grouped the "may fire an event" points together).

Here's the [old content](http://w3c.github.io/aria/#tree_inclusion), for reference.

index.html
```diff
@@ -11872,9 +11872,14 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
...
-            <li>Elements that may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>.</li>
-            <li>Elements that are focusable, or have an ID <a>attribute</a> and an ancestor with the <pref>aria-activedescendant</pref> attribute that matches the implicit or explicit <a>semantics</a> of the required context role. In either case, the element may receive focus and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event.</li>
+            <li>Text nodes.</li>
+            <li>Elements that may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>.
+              <ul>
+                <li>Elements that are focusable may receive focus and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event.</li>
+                <li>Elements that have an ID <a>attribute</a>, and are owned by or have an ancestor with the <pref>aria-activedescendant</pref> attribute that matches the ID, may receive focus and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event.</li>
```
mcking65: In ARIA 1.1, we allow aria-activedescendant on a textbox to point to a descendant of an element controlled by the textbox. This was a critical change for the combobox pattern. So, this text should probably change to reflect that as well. Generally, those elements would probably in the tree for another reason, but we probably can't assume that is always the case.

index.html
```diff
@@ -11872,9 +11872,14 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
 	  <h2>Including Elements in the Accessibility Tree</h2>
           <p>If not already excluded from the accessibility tree per the rules above in <a href="#tree_exclusion">Excluding Elements in the Accessibility Tree</a>, user agents MUST provide an <a>accessible object</a> in the <a>accessibility tree</a> for <abbr title="Document Object Model">DOM</abbr> <a>elements</a> that meet any of the following criteria:</p>
           <ul>
-            <li>Text nodes</li>
-            <li>Elements that may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>.</li>
-            <li>Elements that are focusable, or have an ID <a>attribute</a> and an ancestor with the <pref>aria-activedescendant</pref> attribute that matches the implicit or explicit <a>semantics</a> of the required context role. In either case, the element may receive focus and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event.</li>
+            <li>Text nodes.</li>
```
joanmarie: I realize that the above is ancient text which was merely moved from Core-AAM as-is, but....

As long as we're clarifying things, what exactly is a text node? Given `<p>foo</p>` is the "text node" the `p` element itself, or the static text leaf node with "foo"? On my platform (ATK), we want the `p` accessible but NOT the static text leaf node. Mind you, some platforms (AXAPI) *do* expose the static text leaf node.

Long way of saying that if we're going to clean this text up, we might as well either clarify or remove "text node" from the user agents MUST provide an accessible object for....

joanmarie: One possible resolution, if we don't want to explicitly define "text nodes" is to change the MUST to a MAY for text nodes and then add a MUST for the other items.

index.html
```diff
@@ -11872,9 +11872,14 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
...
+            <li>Elements that may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>.
+              <ul>
+                <li>Elements that are focusable may receive focus and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event.</li>
+                <li>Elements that have an ID <a>attribute</a>, and are owned by or have an ancestor with the <pref>aria-activedescendant</pref> attribute that matches the ID, may receive focus and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event.</li>
+                <li>Elements, including their descendants, that receive focus and fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event even though they have <sref>aria-hidden</sref> set to <code>true</code>.</li>
```
jnurthen: This doesn't seem right to me. In my reading this states that if there is an element which can receive focus and fire an accessibility API focus event then all of its descendents MUST be exposed.

mcking65: I think that is a correct reading. It is probably necessary to ALWAYS include descendants of focusable elements. I imagine a lot of marginally accessible things would break if that were not the case.

index.html
carmacleod: Can use some of the words in aria-activedescendant.

cookiecrook: Concern: I'm not certain these changes are readily implementable in a performant manner.

For example, it's achievable to update the tree on events, such as focus... so if an otherwise inaccessible element were focused, it could become accessible at that time. Several implementations do this today. However, it's not as straightforward to determine what may or may not trigger some relevant event at some point in the future. Likewise, it's not clear how to expose some elements referenced by ARIA relationship attributes. It's reasonable to calculate a label from `aria-labelledby` referencing a hidden element, but exposing a hidden element referenced by `aria-controls` or `aria-owns` would be unexpected. I'm not convinced it's what authors or users would want.

One example: most combobox implementations use both `aria-owns` for the menu relationship and `aria-activedescendant` for the selected menu item. The visibility of this "popup" is usually controlled by setting `display:none` on the menu when it is irrelevant or not displayed. I've never seen an implementation that removes the `aria-owns` relationship each time the menu is shown or hidden. Authors usually leave these attributes in place.

This spec language seems to indicate that most combobox implementations should always expose the sub menu, even if it is not displayed. I don't believe anyone actually wants that.

Minor nit: Every use of the term "owned by" should link to the definition.

index.html
cookiecrook: > Elements, including their descendants, that receive focus and fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event even though they have <sref>aria-hidden</sref> set to <code>true</code>.

Editorial: I might phrase this:

Any _rendered_ element that becomes focused, even if the element or one of its ancestor elements has its <sref>aria-hidden</sref> attribute set to <code>true</code>.

cookiecrook: The "rendered" bit may be important, because it's ambiguous what it means to focus an element that is not in the render tree. Normally this clarification would not be necessary, but juxtaposed with this reference to `aria-hidden` the spec should be clear.

index.html
carmacleod: The ARIA spec has this for [text node](http://w3c.github.io/aria/#dfn-text-node):
> Type of DOM node that represents the textual content of an attribute or an element. A Text node has no child nodes.

So that's the "leaf". So I moved "Text nodes" out of the list, and added a MAY paragraph afterwards.

index.html
carmacleod: Changed to use your words, with the "fire an accessibility API focus event" bit back in.

carmacleod:
@cookiecrook @jnurthen @joanmarie @mcking65 

I believe I have addressed all of your concerns.
If you have a moment to re-review this section, the direct link is:
[Including Elements in the Accessibility Tree](https://raw.githack.com/w3c/aria/car/issue841/index.html#tree_inclusion)

carmacleod:
Apparently deleting a branch closes the related pull request... (of course it does)... so I created a new pull request for the new branch that I just created. Please review the cleaned-up branch in PR https://github.com/w3c/aria/pull/1100 instead.

PR #953, opened 2019-04-18
Add strong and emphasis roles

joanmarie:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/953.html" title="Last updated on Apr 25, 2019, 4:54 PM UTC (c2c6045)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/953/4c9acba...c2c6045.html" title="Last updated on Apr 25, 2019, 4:54 PM UTC (c2c6045)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -667,6 +667,7 @@
         <li><rref>deletion</rref></li>
 				<li><rref>directory</rref></li>
 				<li><rref>document</rref></li>
+				<li><rref>emphasis</rref></li>
 				<li><rref>feed</rref></li>
 				<li><rref>figure</rref></li>
 				<li><rref>group</rref></li>
@@ -686,6 +687,7 @@
 				<li><rref>rowgroup</rref></li>
 				<li><rref>rowheader</rref></li>
 				<li><rref>separator</rref> (when not focusable)</li>
+				<li><rref>strong</rref></li>
 				<li><rref>subscript</rref></li>
 				<li><rref>superscript</rref></li>
 				<li><rref>table</rref></li>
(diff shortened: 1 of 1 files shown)


joanmarie:
Links for review:
* https://raw.githack.com/w3c/aria/issue-905/index.html#emphasis
* https://raw.githack.com/w3c/aria/issue-905/index.html#strong

index.html
```diff
@@ -2529,6 +2531,85 @@ <h2>Definition of Roles</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="emphasis">
+			<rdef>emphasis</rdef>
+			<div class="role-description">
+				<p>One or more emphasized characters. See related <rref>strong</rref>.</p>
+				<p>The <code>emphasis</code> role is intended to be used to indicate stress or emphasis of its contents; not for typographical presentation for presentation's sake. In general, authors SHOULD use this role only if the absence of the emphasis would change the meaning of the content.</p>
```
mcking65: A nit, but a role does not have content. An element that has a role applied to it has content. Since a role does recast, we could say an "emphasis element". But, I think we can simplify the wording without that.

Also, do we want to say "in general" at the start of a normative statement? If there are valid reasons for using emphasis to indicate purely presentational typography, then we should remove the normative statement.

Suggested wording:

>The purpose of the emphasis role is to stress or emphasize content. It is not for communicating changes in typographical presentation that do not impact the meaning of the content. Authors SHOULD use the emphasis role only if its absence would change the meaning of the content.

index.html
```diff
@@ -7345,6 +7426,85 @@ <h5>Presentational Roles Conflict Resolution</h5>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="strong">
+			<rdef>strong</rdef>
+			<div class="role-description">
+				<p>Content which is important, serious, or urgent. See related <rref>emphasis</rref>.</p>
+				<p>The <code>strong</code> role is intended to be used to indicate strong importance, seriousness, or urgency for its contents; not for typographical presentation for presentation's sake. Authors SHOULD use this role only if the absence of the importance would change the meaning of the content.</p>
```
mcking65: For the same reasons I gave for changing the language used to describe the emphasis role, I suggest:

>The purpose of the strong role is to communicate strong importance, seriousness, or urgency. It is not for communicating changes in typographical presentation that are not important to the meaning of the content. Authors SHOULD use the strong role only if its absence would change the meaning of the content.

mcking65:
github is messing with me ... I can't seem to find my last review comment so adding it again. appologize if it is a dupe:

Like the emphasis role, the strong role also needs information describing whether there are any constraints on descendant content.

index.html
joanmarie: Sounds good to me. Thanks again! Changes made.

joanmarie:
@mcking65: Regarding this:

> Like the emphasis role, the strong role also needs information describing whether there are any constraints on descendant content.

I think that makes good sense. But I am not an author, so I'll leave it to the group to decide and then update this PR accordingly.

carmacleod:
HTML strong and em are allowed to contain [phrasing content](http://w3c.github.io/html/dom.html#phrasing-content), which includes anchor, button, input, select, etc.
So they shouldn't have children presentational true.

carmacleod:
@mcking65 
> Like the emphasis role, the strong role also needs information describing whether there are any constraints on descendant content.

They are not supposed to contain [block-level content](https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements), which includes sectioning content, headings, div, paragraph, etc. (so for example, an entirely emphasized paragraph would need to have the em as a child of the p and not the other way around).

However, other than "Required Context Role" or "Required Owned Elements" (which I don't think apply here?) and related prose, are there other places in the ARIA spec where we discuss constraints on descendant content? i.e. Maybe this type of information belongs in the APG?

scottaohara:
@carmacleod beat me to pushing this comment, i had to leave it alone for awhile to attend to some client work.  So apologies for repeating / expanding on some same ideas Carolyn.


`strong` and `em` each allow for [phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2) as descendants. So there are restrictions in HTML to the allowed contents of these elements, or what they can be descendants of, but not to the point to indicate descendants of these roles should be presentational.

Regarding appropriate use and constraints, using Matt's example of `role=emphasis` wrapping a `main`, in HTML an `<em>` cannot wrap a `<main>`. Therefore the `emphasis` role should also not be allowed to wrap a `<main>` or `role=main`. It should only be allowed to be a parent to other roles which would map to HTML Phrasing Content elements / text.

With that said, how should this be outlined in ARIA?  Appropriate use and constraints seems to be  mentioned in regards to roles that have required owned elements, or a required context role.
But take `role=button`, for example.  HTML `<button>` allows phrasing content as child elements to buttons, and also notes that other interactive elements are not allowed.  The ARIA specification does not presently indicate such expectations or restrictions.  

To outline these considerations for `strong` and `emphasis` in ARIA, I'd imagine we'd need to do the same for other roles lacking this sort of information as well?

This could get quite cumbersome if individuals roles were outlined whether in the prose or as part of the characteristics table.  To mitigate some of this, roles could be categorized as they are in HTML, and then indicate the allowed roles via their categorization (e.g. phrasing content).

OR, a simpler solution to this could be to mention the role should match the HTML element's existing definitions / allowances, and to refer to the related concepts (which already link to the HTML definition)?

mcking65:
@carmacleod commented:

>However, other than "Required Context Role" or "Required Owned Elements" (which I don't think apply here?)

Correct, they do not apply because, as defined today, they are not restrictive. Required owned elements definition does not say that other types of elements are not allowed. This, btw, has been a point of author confusion.

@carmacleod continues:

>and related prose, are there other places in the ARIA spec where we discuss constraints on descendant content?

Just prose and then children presentational. While children presentational true doesn't say you cannot include child elements of any specific type, it wipes out all child semantics. When it comes to the accessibility tree then, children presentational true is equivalent to a restriction.

@carmacleod asks:

>i.e. Maybe this type of information belongs in the APG?

APG is not normative, and this is an important validation issue. I'm pretty certain ARIA should not allow ARIA declarations that conflict with HTML. I think role parity needs to also mean parity of allowed usage.

@scottaohara commented:

>With that said, how should this be outlined in ARIA? Appropriate use and constraints seems to be mentioned in regards to roles that have required owned elements, or a required context role.

The [definition of required owned elements](http://w3c.github.io/aria/#mustContain) does not prohibit other type of elements in a context. Perhaps it should though.

@scottaohara continues:

>But take role=button, for example. HTML <button> allows phrasing content as child elements, and also note that other interactive elements are not allowed.

While ARIA does not restrict what children a button can have, it does restrict what accessibility semantics will be exposed because children presentational is true. From an accessibility perspective, this is almost equivalent because the accessibility tree is built as if the button contains only text nodes, which is even more restrictive than HTML.

That said, if you create a button with a focusable child, focus will go inside the button. ARIA probably should explicitly prohibit interactive descendants of button. So, I agree that there is a gap here.

@scottaohara continues:

>Or, a simpler solution to this could be to mention the role should match the HTML element's existing definitions / allowances, and to refer to the related concepts (which already link to the HTML definition)?

Unfortunately, that would create a normative dependency on HTML, which we do not want.

We might need to:

1. Add language to the definition of required owned elements so it becomes a restrictive characteristic.
2. Add another role characteristic that is equivalent to the HTML content model concept.
3. Examine whether or current categorization of roles could be utilized for enumerating allowed content.

scottaohara:
Thanks for the detailed reply @mcking65

Largely agree with the points you raised, and per your closing summary:  


1. Per modifying the required owned elements definition, seems to me it might be more straight forward to add a new "restricted owned elements" definition. But I agree with you that something should happen here. It just needs to be easy for people to identify what roles are required vs not allowed.

2. Agree.

3. Re: current role categories, I've re-reviewed these and I don't think they'd be suitable.  As a quick example, `emphasis` and `strong` would allow for `button` or `img` as allowed child elements.  However, many of the other Widget and Document roles would not be allowed. 



Circling back to this PR, should these new ideas be blockers for merging the new roles as they are now?  I think with the updates you requested, and @joanmarie implemented, `emphasis` and `strong` are no more or less clear about appropriate use and constraints than many other roles in the spec.

I bring this up because if this information is a blocker here, then it likely stands to be a blocker for other new roles that are to be added as well.

Issue #833, opened 2018-10-26
Consider prohibiting author naming certain roles


joanmarie:
At the present time, [5.2.7.4 Roles Supporting Name from Author](https://w3c.github.io/aria/#namefromauthor) says:

> All roles support name from author with two exceptions. The roles that do not support name from author are `presentation` and `none`.

Prior to role parity, that might have been fine. But as we work on the "generic" (real name to be confirmed) role and similar roles, I question whether or not we want the author to be able to name everything (e.g. what about paragraph?).

Related: I personally don't think a `div` or `span` element which lacks an ARIA role should be able to be given a name via `aria-label` or `aria-labelledby`. But I'm pretty sure the language of the AccName spec and the HTML-AAM permit doing so. (To what avail, who knows? Screen readers might not see fit to present -- or even check for -- a name on a `div` or `span`. And if they do, I am not aware of any guidance regarding the expected presentation.)

jnurthen:
(comments welcome - this is only my personal opinion - no group consensus is implied)

Taking the current ARIA Roles I would propose we prohibit names on the following
- blockquote
- caption
- listitem
- paragraph
- presentation (already prohibited)
- none (already prohibited)

Potentially add the following after resolving other issues
- definition (filed issue #836 ) - potentially move to prohibited list based on outcome

carmacleod:
**Is** name currently prohibited on presentation/none?

The [Presentational Roles Conflict Resolution](http://w3c.github.io/aria/#conflict_resolution_presentation_none) section says (twice!):
> Host languages elements, having implicit presentational roles for which no roles, may be applied, MUST never be exposed to in the accessibility tree. With this exception, user agents MUST always expose global WAI-ARIA states and properties to accessibility APIs. In this case, the user agent ignores the presentation role and exposes the element according to its implicit native semantics. 

and again:
> User agents MUST always expose global WAI-ARIA states and properties to accessibility APIs, even if an element has an explicit or inherited role of presentation. In this case, the user agent ignores the presentation role and exposes the element according to its implicit native semantics.

(We should remove one of those paragraphs because it's redundant), however the point is that even if they are prohibited on presentation/none, aria-label, aria-labelledby and aria-describedby, being global WAI-ARIA properties, are supposed to be exposed, which puts them in some kind of grey area (not-allowed-but-used-anyhow).

jnurthen:
> Is name currently prohibited on presentation/none?

[...]

Issue #940, opened 2019-04-09
Does the math role, as written, make sense?


joanmarie:
The current definition of the `math` role states the following:

> Content with the role math is intended to be marked up in an accessible format such as MathML [MathML3], or with another type of textual representation such as TeX or LaTeX, which can be converted to an accessible format by native browser implementations or a polyfill library.

Why is an ARIA role required for polyfills to work? Unless I'm missing something, it's not. Nor is it, IMHO, our job to provide a means for others to hack around lack of MathML support in user agents.

The rest of our roles (with the exception of `presentation`/`none`) do a couple of things:
* Cause the element with that role to appear in the accessibility tree
* Provide a means whereby ATs can know what something is.

I think the `math` role should do the above two things and only the above two things.

Over time, we may want to consider adding other math related roles (e.g. to achieve parity with MathML or the "refresh" in progress of MathML). HOWEVER, that work, if agreed upon, would fall into the 1.4 cycle. Plus, APA is currently working on knowledge domain accessibility. They may come up with some solution which we'll want to consider.

In the meantime, here is my proposal for the 1.2 cycle:
1. We remove from the spec what is currently [Example 6](https://www.w3.org/TR/wai-aria-1.2/#ex-6-example-6-note-use-a-javascript-polyfill-library-to-ensure-this-renders-in-user-agents-that-do-not-support-mathml-the-math-element-has-an-implicit-role-math-math-xmlns-http-www-w3-org-1998-math-mathml-mrow-mi-x-mi-mo-mo-mfrac-mrow-mo-form-prefix-mo-mi-b-mi-mo-mo-msqrt-msup-mi-b-mi-mn-2-mn-msup-mo-mo-mn-4-mn-mo-x2062-invisibletimes-mo-mi-a-mi-mo-x2062-invisibletimes-mo-mi-c-mi-msqrt-mrow-mrow-mn-2-mn-mo-x2062-invisibletimes-mo-mi-a-mi-mrow-mfrac-mrow-annotation-encoding-tex-x-frac-b-pm-sqrt-b-2-4ac-2a-annotation-math). It's a large MathML example without an reference to ARIA other than the comment 'The math element has an implicit role="math".'
2. We replace it with some simpler example which actually uses ARIA and demonstrates how some other host language than MathML (it could be HTML, SVG, whatever) combined with the `math` role would make sense for authors to consider. This example should have child elements.
[...]

PR #967, opened 2019-04-29
Make aria-label and labelledby non global

jnurthen:
Draft PR to make aria-label and labelledby non global

For #833

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/967.html" title="Last updated on Jun 13, 2019, 10:14 PM UTC (fd7f617)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/967/a8e3167...fd7f617.html" title="Last updated on Jun 13, 2019, 10:14 PM UTC (fd7f617)">Diff</a>


carmacleod:
Wow. This PR removes `aria-label` and `aria-labelledby` from all of the following roles:
- alert
- blockquote
- caption
- cell
- definition
- deletion
- heading
- insertion
- label
- legend
- listitem
- log
- marquee
- math
- note
- paragraph
- presentation
- roletype
- rowgroup
- section
- sectionhead
- status
- structure
- subscript
- superscript
- term
- time
- timer
- tooltip

That's a lot of roles.  :)

At first glance, I'm wondering about `heading`, because @DavidMacDonald's [testing of aria-label, etc on static elements](http://davidmacd.com/blog/does-aria-label-override-static-text.html) says:
> If its a `heading` role , `aria-label` or `aria-labelledby` overrides it on NVDA, VoiceOver and Talkback. JAWS ignores the `aria-label` or `aria-labelledby`.

So we would break some expectations. Not sure of actual use in the wild.
Would sure be nice to have [that data](https://github.com/w3c/html-aam/issues/160) before we commit this PR.

jnurthen:
This is a Draft PR. Its purpose is to provoke discussion. It is not going to be merged as-is.

carmacleod:
No worries. My comment was supposed to be the beginning of the discussion.  :)
Edited to remove "before we commit this PR", so the final sentence now reads "Would sure be nice to have that data.", because that's all I really meant.

carmacleod:
Not sure if we should remove aria-label/aria-labelledby from math role?
Here's some [discussion on labelling math role](https://github.com/w3c/aria/issues/940#issuecomment-481940303) from @AmeliaBR and @joanmarie

accdc:
Hi,
Just for the record, I wanted to document my concerns that I brought up during the meeting, and to expound on some primary concerns about the proposal here.

First, here is a basic example of a standard visual technique that is used all across the web to represent pagination buttons, itemized shopping cart buttons, new message buttons, date picker next and previous month and year buttons, and many other types.

```
<button><span aria-label="Next " class="nextArrowIcon"></span>Email</button>
```

This, plus the prevention of labeling on generic roles, will break all such implementations everywhere.

Regarding the points I made about the AccName and ARIA specs, it will be a very bad idea to say that these attributes will remain valid regardless of role within the AccName spec, but that the same attributes are prohibited in the ARIA spec on specific roles.

As was mentioned, I could indeed say that certain uses are valid when traversing child nodes within the AccName spec, however these will still be documented as prohibited in the ARIA spec with no such exceptions, thus leading to the same contradictions between the two normative specs.

If the purpose that we are trying to solve here is to prevent authors from being stupid, there is nothing that will ever prevent people from doing stupid things. As John Cleese said once: "You can't fix stupid." 

However, if the intent is to get validators to flag improper uses of ARIA, this is best solved by adding these combinations to the HTML-ARIA spec because this is what the HTML validator actually maps to directly, and not the ARIA spec where such implementations might be the only way to fix specific accessibility issues when encountered. This is very important when remediating enterprise CMS platforms with legacy code bases.

There is another ramification when separating out a bunch of valid versus invalid roles when computing the accessible name, which will likely have an impact on the process speed and efficiency of the naming computation itself within browsers. If every iteration of the algorithm has to check every node against a comprehensive list of valid versus invalid role and element combinations, and this must be done on everything on the page to compute accessible names for everything, it will likely cause some pretty significant process lags and performance hits. This will already need to be done to a minor extent for roles that support name from content as opposed to those that don't when processing child nodes as we've previously discussed, but checking against a total index of all possible roles and implicit element types on every node during this traversal process will be expensive.

With all of these considerations in mind, I recommend leaving these attributes as global, and simply add recommended usage patterns to the HTML-ARIA spec instead to guide user authoring practices that way, rather than by messing with the actual accessibility of elements at their fundamental levels by breaking present day solutions that are already accessible.

All the best,
Bryan

OwenEdwards:
I'd like to echo @accdc's concern about any change to ARIA that would be a breaking change in the wild. I hope the ARIA Working Group will carefully consider the impact of any potential breaking changes to the ARIA spec; there seems to be considerable danger that web developers who are not well versed in the intricacies of ARIA, and for whom it took considerable effort to persuade them to adopt the additional markup in their pages, will throw their hands up and abandon ARIA if non-backward compatible changes are introduced into the markup specification.

In particular, developers often use frameworks such as Modernizr to allow them to support a wide range of browsers, versions, and platforms that are part of the business requirements; however, it is impossible for these frameworks to detect which version of assistive technology (especially screen reader) is being used, so it is not possible for web pages to modify their behavior to adapt to older and newer versions of AT. This would make any breaking changes in ARIA especially impactful.

jeffsmith:
Also echoing @accdc's concerns and wanted to highlight that I also agree that the following is the best way to begin steering devs and authors away from unintended usage:

> "...this is best solved by adding these combinations to the HTML-ARIA spec because this is what the HTML validator actually maps to directly, and not the ARIA spec where such implementations might be the only way to fix specific accessibility issues when encountered."

Validation should be the carrot that leads folks away from improper usage, rather than breaking changes being introduced to the spec.

jnurthen:
Decided yet again not to go this way. closing

PR #984, opened 2019-05-22
Generalize AccessibilityRole/AriaAttributes IDL

domenic:
This generalizes the AccessibilityRole/AriaAttributes IDL mixins to allow them to have different behaviors per host interface. For Element, they have the reflection behavior specified, but e.g. for ElementInternals, HTML can use this framework to define different behavior.

In the process, this consolidates the two mixins into one ("AccessibilityMixin"), since this makes it easier to define their behavior uniformly, and consumers should generally not mix in one without the other.

/cc @alice @tkent-google.

My next step is to work on a counterpart pull request on the HTML side, to prototype how we would use this more general framework for ElementInternals to allow custom elements to have native roles/states/properties. It's probably not worth reviewing this PR in too much depth before that is ready, in case I find out that something needs to be tweaked.

***
<a href="https://pr-preview.s3.amazonaws.com/domenic/aria/pull/984.html" title="Last updated on Sep 14, 2020, 7:04 PM UTC (70dcd60)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/984/19e5b99...domenic:70dcd60.html" title="Last updated on Sep 14, 2020, 7:04 PM UTC (70dcd60)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -13600,8 +13600,8 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
 		<p>Interfaces that include <code>ARIAMixin</code> must provide the following algorithms:</p>
 
 		<ul>
-			<li><dfn>get the accessibility IDL attribute</dfn>, which takes the host interface instance, IDL attribute name, and content attribute name, and must return a string value; and
-			<li><dfn>set the accessibility IDL attribute</dfn>, which takes the host interface instance, IDL attribute name, content attribute name, and string value, and must return nothing.</li>
+			<li><dfn><code>ARIAMixin</code> getter steps</dfn>, which takes the host interface instance, IDL attribute name, and content attribute name, and must return a string value; and
+			<li><dfn><code>ARIAMixin</code> setter steps</dfn>, which takes the host interface instance, IDL attribute name, content attribute name, and string value, and must return nothing.</li>
 		</ul>
 
 		<p>For every IDL attribute <var>idlAttribute</var> defined in <code>ARIAMixin</code>, on getting, it must perform the following steps:</p>
@@ -13609,7 +13609,7 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
 		<ol>
 			<li><p>Let <var>contentAttribute</var> be the ARIA content attribute determined by looking up <var>idlAttribute</var> in the ARIA Attribute Correspondence table.</p></li>
 
-			<li><p>Return the result of <a>get the accessibility IDL attribute</a>, given this, <var>idlAttribute</var>, and <var>contentAttribute</var>.</p></li>
+			<li><p>Return the result of running the <a><code>ARIAMixin</code> getter steps</a>, given this, <var>idlAttribute</var>, and <var>contentAttribute</var>.</p></li>
 		</ol>
 
 		<p>Similarly, on setting, it must perform the following steps:</p>
@@ -13617,7 +13617,7 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
 		<ol>
 			<li><p>Let <var>contentAttribute</var> be the ARIA content attribute determined by looking up <var>idlAttribute</var> in the ARIA Attribute Correspondence table.</p></li>
 
-			<li><p><a>Set the accessibility IDL attribute</a>, given this, <var>idlAttribute</var>, <var>contentAttribute</var>, and the given value.</p></li>
+			<li><p>Run the <a><code>ARIAMixin</code> setter steps</a>, given this, <var>idlAttribute</var>, <var>contentAttribute</var>, and the given value.</p></li>
 		</ol>
 
 		<p class="note">This very general framework is motivated by the desire for different host interfaces, such as <code>Element</code> and <code>ElementInternals</code>, to give these IDL attributes different behaviors. The alternative is requiring each host interface to duplicate the IDL attributes independently, so that they can specify independent behaviors, but that comes with a high risk of them getting out of sync.</p>
@@ -13713,8 +13713,8 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
 
 		<p>For <code>Element</code>:</p>
 		<ul>
-			<li><p>To <a>get the accessibility IDL attribute</a> given <var>element</var>, <var>idlAttribute</var>, and <var>contentAttribute</var>, return the result of the getter algorithm for <var>idlAttribute</var> <a href="https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflect">reflecting</a> <var>contentAttribute</var> on <var>element</var>.</p></li>
-			<li><p>To <a>set the accessibility IDL attribute</a> given <var>element</var>, <var>idlAttribute</var>, <var>contentAttribute</var>, and <var>value</var>, perform the setter algorithm for <var>idlAttribute</var> <a href="https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflect">reflecting</a> <var>contentAttribute</var> on <var>element</var>, given <var>value</var>.</p></li>
+			<li><p>The <a><code>ARIAMixin</code> getter steps</a> given <var>element</var>, <var>idlAttribute</var>, and <var>contentAttribute</var> are to return the result of the getter algorithm for <var>idlAttribute</var> <a href="https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflect">reflecting</a> <var>contentAttribute</var> on <var>element</var>.</p></li>
+			<li><p>The <a><code>ARIAMixin</code> setter steps</a> given <var>element</var>, <var>idlAttribute</var>, <var>contentAttribute</var>, and <var>value</var> are to perform the setter algorithm for <var>idlAttribute</var> <a href="https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflect">reflecting</a> <var>contentAttribute</var> on <var>element</var>, given <var>value</var>.</p></li>
 		</ul>
 
 		<p class="note">In practice, this means that, e.g., the <code>role</code> IDL on <code>Element</code> reflects the <code>role</code> content attribute; the <code>ariaValueMin</code> IDL attribute reflects the <code>aria-valuemin</code> content attribute; etc.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -12611,6 +12630,24 @@ <h3>IDL Attribute Name Notes or Exceptions</h3>
 		</section>
 	</section>
 
+	<section id="idl_element">
+		<h2><code>AccessibilityMixin</code> Mixed in to <code>Element</code></h2>
+
+		<p>User agents must include <code>AccessibilityMixin</code> on <code>Element</code>:</p>
```
carmacleod: Do you want lowercase must or uppercase MUST?

domenic: I'm not sure what this group's convention is; editorial help is welcome. For other web specs we prefer lowercase, per https://infra.spec.whatwg.org/#conformance, but perhaps things are different here.

jnurthen: For RFC2119 as stated at https://w3c.github.io/aria/#conformance we require uppercase MUST.

caridy:
@domenic thanks a lot for putting the time to get this ready! wonderful stuff that we are eager to try out.

jnurthen:
@domenic based on your comment 

> It's probably not worth reviewing this PR in too much depth before that is ready, in case I find out that something needs to be tweaked.

are we to assume this is not ready to consider merging? If so do you mind if I put a WIP label on it?

In order to consider merging in the future we'll need to the RFC2119 format changed to caps.

Once merged into our Editor's draft we'll need tests before we can add into a WD - the IDL section previously had tests supplied from outside the ARIA WG - would it be possible for this to happen with this change too?

domenic:
Sorry for the delay in response! Please do put a WIP label on it. I'll fix the caps RFC2119 keyword.

As a status update, I did the HTML pull request in https://github.com/whatwg/html/pull/4658, and this pull request seems pretty good. So at least technically I think it's ready to go, but it should still be marked as WIP until we have rough consensus on the HTML pull request.

For tests, I think this PR doesn't actually introduce any new normative requirements that would need testing. Instead, it restructures the existing normative requirements in a way that allows their underlying infrastructure to be reused by HTML, in https://github.com/whatwg/html/pull/4658. And https://github.com/whatwg/html/pull/4658 will definitely get tests before landing, as a required part of the WHATWG working mode; my team at Chrome is happy to commit to that.

cookiecrook:
I have concerns about combining `AccessibilityRole` and `AriaAttributes` into a single `AccessibilityMixin`

1. Role is not used solely in the context of ARIA. 
2. ARIA attributes to not represent the entirety of mixins that might be used to augment Accessibility, so this naming convention seems to limit the ability to mixin more Accessibility-specific extensions later. For example, some of the items in AOM would not necessarily need to be lock-step with an ARIA update.

My preference would be to change this back to two separate mixins: AccessibilityRole and AriaAttributes.

domenic:
@cookiecrook thanks for raising these concerns. I'd like to try to understand them better. Note that how these mixins are structured, and their name, is largely an editorial detail, and is not observable to web developers.

The reason I merged these is because the specification infrastructure operates in the exact same way on both role and on the attributes. In particular, the "get/set the accessibility IDL attribute" infrastructure, and the steps for the getters/setters for each IDL attribute. See the spec text below the IDL block [in this pull request](https://pr-preview.s3.amazonaws.com/domenic/aria/pull/984.html#AccessibilityMixin), and in [the corresponding HTML pull request](https://whatpr.org/html/4658/custom-elements.html#native-accessibility-semantics-map).

We could certainly re-separate them. However, this would just make all the specification text more awkward: we would need to have separate cases for `role`, which are largely copying-and-pasting the spec text for the other accessibility attributes. Given the fact that these mixins are unobservable, I thus thought it would be better to keep them together.

To your points,

> Role is not used solely in the context of ARIA.

This is definitely true, but I'm not sure how it relates to the PR.

Perhaps you are anticipating in the future mixing `AccessibiltyRole` into places that don't get `AriaAttributes` mixed in? Do you have any concrete examples in mind? Note that we can always re-separate them in the future, with no observable consequences and just a bit more work for spec writers.

> ARIA attributes to not represent the entirety of mixins that might be used to augment Accessibility, so this naming convention seems to limit the ability to mixin more Accessibility-specific extensions later. For example, some of the items in AOM would not necessarily need to be lock-step with an ARIA update.

This isn't the intention. As with any Web IDL mixin, the only actual constraints on augmenting the `AccessibilityMixin` are that doing so automatically augments its mixin targets. Currently those mixin targets are `ElementInternals` and `Element`.

AOM could add lots of functionality, which could target `ElementInternals`, `Element`, or both. The intention of this mixin is for things that end up targeting both. If AOM wanted to add functionality that targeted only one of them, there are definitely ways to do that.

If your concern is mostly about naming, I would be happy with any naming convention. Keeping in mind that the name is only observable to spec authors, we could call it `AccessibilityElementOrElementInternals` (similar to the `DocumentOrShadowRoot` or `WindowOrWorkerGlobalScope` mixins). Or maybe `RoleAndARIAAttributesElementOrElementInternals`, if you think in the future we might want to add other accessibility-related mixins that apply to both `Element` and `ElementInternals`. (Although in that case we probably should just expand this mixin?) Let me know!

cookiecrook:
Based on that explanation–thanks btw–I think it'd be okay to keep them together and just rename `AccessibilityMixins` to `AriaMixins`.

Does that work for you?

domenic:
For sure! I've renamed it and will make the corresponding HTML-side changes now. (I used capitalized ARIAMixin instead of Pascal-case AriaMixin per https://w3ctag.github.io/design-principles/#casing-rules.)

index.html
```diff
@@ -13020,64 +13011,90 @@ <h2>Interface Mixin <dfn>AriaAttributes</dfn></h2>
...
+
+			<li><p><a>Set the accessibility IDL attribute</a>, given this, <var>idlAttribute</var>, <var>contentAttribute</var>, and the given value.</p></li>
+		</ol>
+
+		<p class="note">This very general framework is motivated by the desire for different host interfaces, such as <code>Element</code> and <code>ElementInternals</code>, to give these IDL attributes different behaviors. The alternative is requiring each host interface to duplicate the IDL attributes independently, so that they can specify independent behaviors, but that comes with a high risk of them getting out of sync.</p>
+	</section>
+
+	<section id="accessibilityroleandproperties-correspondence" class="normative" data-dfn-for="ARIAMixin" data-link-for="ARIAMixin">
+		<h2>ARIA Attribute Correspondence</h2>
+		<p>The following table provides a correspondence between IDL attribute names and content attribute names, for use by <code>ARIAMixin</code>.</p>
```
annevk: Does it make sense to talk about content attributes here given that they're not content attributes for all consumers? From the wording in the HTML PR it seems these are really "state or properties" (that happen to match the casing conventions for content attributes and are used as such when this mixin is included by elements).

domenic: I am mostly just reflecting preexisting text (where the previous version of this table said "Reflected ARIA Content Attributes"), but I agree with your sentiment so I'll change it.

domenic: Hmm, nope. Upon going to make the edit, I realized that because "role" is included in this table, "content attribute" is now the best super-category. ARIA defines a number of content attributes, namely role, a bunch of states, and a bunch of properties. So I think this is reasonable as-is.

There is a bigger editorial meta-issue about how ARIA seems to be largely defined in terms of content attributes instead of having underlying models, but let's not get in to that here :).

domenic:
ARIA editors, I think this is ready to merge. Tests are available for the new stuff that this enables (which is specced in https://github.com/whatwg/html/pull/4658) at https://github.com/web-platform-tests/wpt/blob/master/custom-elements/form-associated/ElementInternals-accessibility.html.

Also, I realized I didn't make this super clear before: from the ARIA spec's perspective this is a refactoring without observable consequences. But it does support infrastructure like https://github.com/whatwg/html/pull/4658.

domenic:
Ping to ARIA editors. I've rebased this on master, and the corresponding HTML side pull request in https://github.com/whatwg/html/pull/4658 for adding accessibility semantics to custom elements is blocked on this.

carmacleod:
Thanks for the ping, @domenic - it's looking good!
A couple of little things:
1. The role link in the correspondence table is broken, because `role` is defined in a completely different document.
Probably the nicest fix is to just change `<pref>role</pref>` to: `<a href="#introroles">role</a>`.

2. Wondering if we need an "Editor's Note" about the missing element reflection attributes in the mixin and the correspondence table? Here's an [example Editor's Note in combobox](https://w3c.github.io/aria/#combobox) and the corresponding markup on [index.html#L2098](https://github.com/w3c/aria/blob/master/index.html#L2098):
   ```
   <p class="ednote" title="Major Changes to combobox role in ARIA 1.2">
   The Guidance for <rref>combobox</rref> has changed significantly ...
   </p>
   ```

Other than those, I think this is good to merge!

domenic:
> The role link in the correspondence table is broken, because role is defined in a completely different document.

Fixed!

> Wondering if we need an "Editor's Note" about the missing element reflection attributes in the mixin and the correspondence table?

Although that makes some sense to me, I'm not sure how it relates to this PR; this PR just consolidates what was previously two mixins (one for role and one for everything else) into one.

domenic:
Thanks for all your help! Please feel free to merge at your convenience :).

index.html
```diff
@@ -13604,65 +13595,93 @@ <h2>Interface Mixin <dfn>AriaAttributes</dfn></h2>
...
-				<tr><td><dfn>ariaValueText</dfn></td><td><pref>aria-valuetext</pref></td></tr>
-			</table>
-			<p class="note">Note: Attributes <pref>aria-dropeffect</pref> and <sref>aria-grabbed</sref> were deprecated in ARIA 1.1 and do not have corresponding IDL attributes.</p>
-		</section>
+
+		<p>Interfaces that include <code>ARIAMixin</code> must provide the following algorithms:</p>
+
+		<ul>
+			<li><dfn><code>ARIAMixin</code> getter steps</dfn>, which takes the host interface instance, IDL attribute name, and content attribute name, and must return a string value; and
+			<li><dfn><code>ARIAMixin</code> setter steps</dfn>, which takes the host interface instance, IDL attribute name, content attribute name, and string value, and must return nothing.</li>
```
carmacleod, line 13604: Not sure if "takes" needs to be "take"?
i.e.
...getter steps, which take ...
...setter steps, which take ...

domenic: Fixed, thanks!

domenic:
Gentle ping! I'd love to get this and the corresponding HTML spec pull request merged.

PR #985, opened 2019-05-23
Prohibit label on certain roles

jnurthen:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/985.html" title="Last updated on Jun 28, 2019, 11:28 PM UTC (7d26326)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/985/971b418...7d26326.html" title="Last updated on Jun 28, 2019, 11:28 PM UTC (7d26326)">Diff</a>

### DIFF
--- a/common/script/aria.js
+++ b/common/script/aria.js
@@ -319,14 +319,8 @@ require(["core/pubsubhub"], function( respecEvents ) {
                                     name = item.textContent || item.innerText;
                                 }
                                 var type = (item.localName === "pref" ? "property" : "state");
-                                var req = false;
-                                if ($(node).hasClass("role-required-properties") ) {
-                                    req = true;
-                                }
-                                var dis = false;
-                                if ($(node).hasClass("role-disallowed") ) {
-                                    dis = true;
-                                }                                
+                                var req = $(node).hasClass("role-required-properties");
+                                var dis = $(node).hasClass("role-disallowed");                              
                                 attrs.push( { is: type, name: name, required: req, disallowed: dis } );
                                 // remember that the state or property is
                                 // referenced by this role
--- a/common/script/aria.js
+++ b/common/script/aria.js
@@ -172,8 +172,11 @@ require(["core/pubsubhub"], function( respecEvents ) {
                     propList[title] = { is: type, title: title, name: content, desc: desc, roles: [] };
                     var abstract = container.querySelector("." + type + "-applicability");
                     if ((abstract.textContent || abstract.innerText) === "All elements of the base markup") {
-                        globalSP.push({ is: type, title: title, name: content, desc: desc });
-                    }
+                        globalSP.push({ is: type, title: title, name: content, desc: desc, prohibited: false });
+                    } 
+                    else if ((abstract.textContent || abstract.innerText) === "All elements of the base markup except for some roles or elements that prohibit its use") {
+                        globalSP.push({ is: type, title: title, name: content, desc: desc, prohibited: true });
+                    } 
                     
                     // the rdef is gone.  if we are in a div, convert that div to a section
 
@@ -222,6 +225,9 @@ require(["core/pubsubhub"], function( respecEvents ) {
                             globalSPIndex += "<sref title=\"" + lItem.name + "\">" + lItem.name + " (state)</sref>";
                         } else {
                             globalSPIndex += "<pref>" + lItem.name + "</pref>";
+                        }
+                        if (lItem.prohibited) {
+                            globalSPIndex += " (Except where prohibited)";
                         }
                         globalSPIndex += "</li>\n";
                     }
--- a/common/script/aria.js
+++ b/common/script/aria.js
@@ -172,11 +172,8 @@ require(["core/pubsubhub"], function( respecEvents ) {
                     propList[title] = { is: type, title: title, name: content, desc: desc, roles: [] };
                     var abstract = container.querySelector("." + type + "-applicability");
                     if ((abstract.textContent || abstract.innerText) === "All elements of the base markup") {
-                        globalSP.push({ is: type, title: title, name: content, desc: desc, prohibited: false });
-                    } 
-                    else if ((abstract.textContent || abstract.innerText) === "All elements of the base markup except for some roles or elements that prohibit its use") {
-                        globalSP.push({ is: type, title: title, name: content, desc: desc, prohibited: true });
-                    } 
+                        globalSP.push({ is: type, title: title, name: content, desc: desc });
+                    }
                     
                     // the rdef is gone.  if we are in a div, convert that div to a section
 
@@ -226,9 +223,6 @@ require(["core/pubsubhub"], function( respecEvents ) {
                         } else {
                             globalSPIndex += "<pref>" + lItem.name + "</pref>";
                         }
-                        if (lItem.prohibited) {
-                            globalSPIndex += " (Except where prohibited)";
-                        }
                         globalSPIndex += "</li>\n";
                     }
                     parentNode = document.querySelector("#global_states");
@@ -265,9 +259,6 @@ require(["core/pubsubhub"], function( respecEvents ) {
                 var roleIndex = "";
                 var fromAuthor = "";
                 var fromContent = "";
-                var fromEncapsulation = "";
-                var fromLegend = "";
-                var fromProhibited = "";
 
                 $.each(document.querySelectorAll("rdef"), function(i,item) {
                     var container = item.parentNode;
@@ -319,7 +310,7 @@ require(["core/pubsubhub"], function( respecEvents ) {
                     }
                     // are there supported states / properties in this role?  
                     var attrs = [];
-                    $.each(container.querySelectorAll(".role-properties, .role-required-properties, .role-disallowed"), function(i, node) {
+                    $.each(container.querySelectorAll(".role-properties, .role-required-properties"), function(i, node) {
                         if (node && ((node.textContent && node.textContent.length !== 1) || (node.innerText && node.innerText.length !== 1))) {
                             // looks like we do
                             $.each(node.querySelectorAll("pref,sref"), function(i, item) {
@@ -328,9 +319,11 @@ require(["core/pubsubhub"], function( respecEvents ) {
                                     name = item.textContent || item.innerText;
                                 }
                                 var type = (item.localName === "pref" ? "property" : "state");
-                                var req = $(node).hasClass("role-required-properties");
-                                var dis = $(node).hasClass("role-disallowed");                              
-                                attrs.push( { is: type, name: name, required: req, disallowed: dis } );
+                                var req = false;
+                                if ($(node).hasClass("role-required-properties") ) {
+                                    req = true;
+                                }
+                                attrs.push( { is: type, name: name, required: req } );
                                 // remember that the state or property is
                                 // referenced by this role
                                 propList[name].roles.push(title);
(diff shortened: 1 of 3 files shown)


carmacleod:
- The Prohibited section for blockquote is before the Inherited section, but for all of the others (caption, code, deletion, emphasis, insertion, note, paragraph, presentation, strong, subscript, superscript), the Prohibited section is after the Inherited section.

- Consider removing "Name from: N/A" rows from the characteristics table for consistency with other N/A characteristics rows.

jnurthen:
>     * Consider removing "Name from: N/A" rows from the characteristics table for consistency with other N/A characteristics rows.

I considered this but thought it might be worth calling it out explicitly. I'm certainly willing to go either way on it, but I was also thinking of adding a corresponding type in the section https://pr-preview.s3.amazonaws.com/w3c/aria/pull/985.html#namecalculation corresponding with whatever we decide we would call the N/A value.

carmacleod:
> I considered this but thought it might be worth calling it out explicitly.

Right. Hard to say. On the one hand, "Prohibited States and Properties: aria-label, aria-labelledby" is pretty explicit and right above where "Name from:" would be, but on the other hand I know what you mean - it is probably worth doing as much as possible to point out this particular change.

> I was also thinking of adding a corresponding type in the section https://pr-preview.s3.amazonaws.com/w3c/aria/pull/985.html#namecalculation corresponding with whatever we decide we would call the N/A value.

Right. If we do that, then we need to keep Name From.
So, we would have author, contents, encapsulation, legend, and ... none? prohibited?

jnurthen:
@stevefaulkner very interested in your feedback on this

stevefaulkner:
@jnurthen from looking at the changes it is unclear to me which roles are effected, is there a list of roles or a rendered view of the branch I can look at?

carmacleod:
@stevefaulkner 
Here's the list of the roles that prohibit aria-label[ledby] in this PR:

- blockquote
- caption
- code
- deletion
- emphasis
- insertion
- note
- paragraph
- presentation
- strong
- subscript
- superscript

It's mostly new roles (except presentation) that we recently added for HTML role parity, so fairly minimal impact at the moment. This is paving the way for adding the "generic" role, which is intended to be the default role for div and span (and [some others](https://github.com/w3c/aria/wiki/Plans-regarding-role-parity#generic-role-but-may-need-an-attribute-for-full-parity)). Our current thinking is that generics should be nameless.

Here's the PR [Preview](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/985.html). If you look for the word "prohibited" in the preview, then you will hit all of the salient changes.

common/script/aria.js
```diff
@@ -323,7 +323,11 @@ require(["core/pubsubhub"], function( respecEvents ) {
                                 if ($(node).hasClass("role-required-properties") ) {
                                     req = true;
                                 }
-                                attrs.push( { is: type, name: name, required: req } );
+                                var dis = false;
```
schne324: coding nitpick:

```js
var dis = $(node).hasClass("role-disallowed");
```

index.html
```diff
@@ -12325,10 +12439,11 @@ <h3>States and Properties</h3>
 					</ul>
 				</li>
 				</ul>
+			<p>If the author specfies a WAI-ARIA property on a role where that property is prohibited, the user agent SHOULD treat the property as if it were allowed. </p>
```
joanmarie: Perhaps we could put an editor's note after this statement to further discourage authors from ignoring the prohibition? Candidate text:

> Note: At the present time, user agents have no obligation to ignore prohibited states and properties. However, this may change in a future version of ARIA in order to ensure interoperability.

jnurthen: We agreed at the meeting to remove the text from the author errors section I believe. Could this go after line 498 instead?

jnurthen:
I will need to split this into 2 (aria-common and aria) - leaving as one until ready for convenience

index.html
```diff
@@ -492,6 +492,11 @@ <h3>Supported States and Properties</h3>
 			<h3>Inherited States and Properties</h3>
 			<p>Informative list of properties that are inherited onto a <a>role</a> from superclass roles. <a>States</a> and <a>properties</a> are inherited from superclass roles in the role <a>taxonomy</a>, not from ancestor <a>elements</a> in the <abbr title="Document Object Model">DOM</abbr> tree. These properties are not explicitly defined on the role, as the inheritance of properties is automatic. This information is provided to facilitate reading of the specification. The set of supported states and properties combined with inherited states and properties forms the full set of states and properties supported by the role.</p>
 		</section>
+		<section id="prohibitedattributes">
+				<h3>Prohibited States and Properties</h3>
+				<p>List of properties that are prohibited on a <a>role</a>. Authors MUST NOT specify a prohibited state or property.</p>
```
carmacleod: Change "List of properties"
to "List of states and properties"
(for consistency).

index.html
```diff
@@ -12328,7 +12442,7 @@ <h3>States and Properties</h3>
 			<p>If a <abbr title="Accessible Rich Internet Application">WAI-ARIA</abbr> property contains an unknown or disallowed value, the user agent SHOULD expose to platform <a>accessibility APIs</a> as follows:</p>
```
carmacleod: Change "property contains"
to "state or property contains".

index.html
```diff
@@ -12340,7 +12454,7 @@ <h3>States and Properties</h3>
 				</ul>
 				</li>
 				<li>Otherwise, ignore the value and treat the property as not present.</li>
```
carmacleod: Change "property"
to "attribute".
(for consistency with other points in this list).

index.html
```diff
@@ -12340,7 +12454,7 @@ <h3>States and Properties</h3>
...
-			</ul>
+			</ul>	
 			<p class="note">In <abbr title="User Interface Automation">UIA</abbr>, the user agent might leave the corresponding property set to &quot;unsupported.&quot;</p>
```
carmacleod: Change "property"
to "attribute".
(for consistency with other points in this list).

index.html
jnurthen: This line has been removed

index.html
jnurthen: happy to accept a PR cleaning these up but I don't want to make this any larger than it has to be.

index.html
jnurthen: happy to accept a PR cleaning these up but I don't want to make this any larger than it has to be.

stevefaulkner:
@jnurthen wondering about the reasons for choosing paragraph/note/blockquote. is there related discussion?

jnurthen:
@stevefaulkner there is some discussion in https://github.com/w3c/aria/issues/833 - although note looks like a mistake - thanks for catching it. I am removing that - it was intended to be on none - but that doesn't follow the normal role structure.

scottaohara:
shouldn't `term` be changed to prohibited as well, especially since it's no longer related to `dt`?

`definition` could be given it's name from the `term`

```html
<span role="term" id="t">Word</span>
<span role="definition" aria-labelledby="t">
  words to define word.
</span>
```

joanmarie:
> `definition` could be given it's name from the `term`
> 
> ```
> <span role="term" id="t">Word</span>
> <span role="definition" aria-labelledby="t">
>   words to define word.
> </span>
> ```

When something has a valid name, screen readers might present that name and NOT present the associated text. So if there's no `aria-labelledby` the term "word" has a presented-by-the-screenreader definition of "words to define word". But with the `aria-labelledby` present, some screen readers might present the definition of "word" as "word". Unless they have smarts to say `definition`, like `paragraph` and friends really shouldn't have a name.

TL;DR: Should we prohibit naming of `definition`?

jnurthen:
> 
> TL;DR: Should we prohibit naming of `definition`?

We would have to change `definition`'s definition
> Authors SHOULD identify the element being defined by giving that element a role of term and referencing it with the aria-labelledby attribute or by making the element with role term a descendant of the element with role definition.

scottaohara:
@joanmarie I had originally had that thought about `definition` when doing the review for the revised `term` and `definition` work.  But I put that aside because I assumed `definition` would work like `group`, where it could have a name, and its contents still remain accessible.

For example, a definition that consists of two paragraphs:

```html
<span role="term" id="f">
  Funny
</span>
<div role="definition" aria-labelledby="f">
  <p>Causing laughter or amusement.</p>
  <p>Making or given to making amusing jokes or witticisms.</p>
</div>
```

joanmarie:
It will be interesting to see how screen readers present `definition`.

stevefaulkner:
Of the list provided the only definitely questionable one is `blockquote` I can envisage scenarios where it would be useful to associate a label/description for a `blockquote` pertaining to it source (for example). In a quick [test](https://thepaciellogroup.github.io/AT-browser-tests/test-files/blockquote-cite.html#blk-003) I found that both JAWS and NVDA announce the presence of a `blockquote` and NVDA also announces the  accessible name if it has one.

I am on the fence about `<p>` as i could envisage situation where annotating a paragraph could be useful, although I have not encountered AT support for announcing the presence of `<p>` elements.

mcking65:
@stevefaulkner wrote:

> Of the list provided the only definitely questionable one is `blockquote` I can envisage scenarios where it would be useful to associate a label/description for a `blockquote` pertaining to it source (for example). In a quick [test](https://thepaciellogroup.github.io/AT-browser-tests/test-files/blockquote-cite.html#blk-003) I found that both JAWS and NVDA announce the presence of a `blockquote` and NVDA also announces the accessible name if it has one.
> 
> I am on the fence about `<p>` as i could envisage situation where annotating a paragraph could be useful, although I have not encountered AT support for announcing the presence of `<p>` elements.

I agree that it is feasible for screen readers to support named blockquotes. Most screen readers present the boundaries of blockquotes in reading mode, which provides a reasonable way of supporting names without confusing users. While it doesn't seem like a practice to recommend, prohibiting it may be unnecessary.

However, I strongly believe prohibiting named paragraphs is necessary.

index.html
```diff
@@ -9399,7 +9504,7 @@ <h2>Translatable States and Properties</h2>
 	</section>
 	<section id="global_states">
 		<h2>Global States and Properties</h2>
-		<p>Some <a>states</a> and <a>properties</a> are applicable to all host language <a>elements</a> regardless of whether a <a>role</a> is applied. The following global states and properties are supported by all roles and by all base markup elements.</p>
+		<p>Some <a>states</a> and <a>properties</a> are applicable to all host language <a>elements</a> regardless of whether a <a>role</a> is applied. The following global states and properties are supported by all roles and by all base markup elements unless otherwise prohibited. Any global states and properties which are prohibited on any role will be listed in the Prohibited States and Properties section.</p>
```
mcking65: >Any global states and properties which are prohibited on any role will be listed in the Prohibited States and Properties section.

This sounds contradictory. The second "on any" could be interpreted to mean "from all," meaning that the state or property cannot be used at all on any element. In addition, the statement is future tense instead of present tense. And, the location of listing is not clear. Suggest:

>If a role prohibits use of any global states or properties, those states or properties are listed as prohibited in the characteristics table included in the section that defines the role.

index.html
```diff
@@ -9399,7 +9504,7 @@ <h2>Translatable States and Properties</h2>
...
 		<p class="placeholder">Placeholder for global states and properties</p>
```
mcking65: If a state or property in this list is prohibited from use on some roles, the script should add "(prohibited from use on some roles)"

index.html
```diff
@@ -9399,7 +9504,7 @@ <h2>Translatable States and Properties</h2>
...
 		<p>Global states and properties are applied to the role <rref>roletype</rref>, which is the base role, and therefore inherit into all roles. To facilitate reading, they are not explicitly identified as either supported or inherited states and properties in the specification. Instead, the inheritance is indicated by a link to this section.</p>
```
mcking65: I think this paragraph should be deleted. It is confusing. The reality is that global states and properties are included in the list of inherrited states and properties in every role definition.

In addition, line 10937, in the definition of aria-label had the following in the row for "used in roles":

><td class="property-applicability">All elements of the base markup</td>

Should be changed to:

<td class="property-applicability">All elements of the base markup except for some elements that prohibit its use</td>

The same needs to be done for aria-labelledby.

jnurthen:
https://github.com/w3c/accname/pull/53 shows how accname could change to enable this

index.html
jnurthen: Sounds editorial to me. I'd like to make this happen as agree this adds clarity. @mcking65 Would you be ok pushing this to a future issue if this were the only remaining comment?

index.html
```diff
@@ -531,6 +536,7 @@ <h3>Accessible Name Calculation</h3>
...
             <li>encapsulation: name comes from the text value of the <a>element</a> node with role <code>label</code> that is the closest ancestor. Although "encapsulation" may be allowed in addition to "author" and "contents" in some <a>roles</a>, "encapsulation" is used only if higher priority "author" features are not provided. Priority is defined by the <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_te">accessible name and description computation</a> algorithm.  [[ACCNAME-1.1]]</li>
             <li>legend: name comes from the text value of the first descendant <a>element</a> node with the role of <code>legend</code>.  Although "legend" may be allowed in addition to "author" in some <a>roles</a>, "legend" is used in content only if higher priority "author" features are not provided. Priority is defined by the <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_te">accessible name and description computation</a> algorithm.  [[ACCNAME-1.1]]</li>
+            <li>prohibited: the element has no name. Authors MUST NOT use the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attributes to name the element.</li>
```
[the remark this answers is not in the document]
jnurthen: Added scripting to build a list for each labelling technique

carmacleod:
I think this PR is good to go.
The [initial set of roles that will prohibit naming](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/985.html#namefromprohibited) makes sense, and more can be added as/if needed.
The wording also allows for other attributes to be prohibited on specific roles in future, as/if needed.

index.html
carmacleod: Agreed. I looked into writing a cleanup PR, but it's a bit of a pandora's box, so leaving it for now.
Resolving this conversation.

index.html
carmacleod: Resolving conversation as mentioned in [previous comment](https://github.com/w3c/aria/pull/985#discussion_r296823383).

index.html
jnurthen: (did this in the script in the end)

jnurthen:
I think this is ready to go. @joanmarie do you want to merge or should I?

joanmarie:
> I think this is ready to go. @joanmarie do you want to merge or should I?

You did all the work. :) Go for it! Do please update the wiki so we can add it to the list of things which will need tests.

PR #1020, opened 2019-07-22
Issue1019 initial draft for accessible name using caption role

jongund:
[Proposed definition of accessible name from caption](https://raw.githack.com/w3c/aria/issue1019-accname-from-caption/index.html#namecalculation)

[Proposed list of roles that support name from caption](https://raw.githack.com/w3c/aria/issue1019-accname-from-caption/index.html#namefromcaption)

[proposed changes to `figure` definition](https://raw.githack.com/w3c/aria/issue1019-accname-from-caption/index.html#figure)

[proposed changes to `grid` definition](https://raw.githack.com/w3c/aria/issue1019-accname-from-caption/index.html#grid)

[proposed changes to `table` definition](https://raw.githack.com/w3c/aria/issue1019-accname-from-caption/index.html#table)

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jan 20, 2021, 11:00 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Faria%2F692220c7656b3de75d3b156d18329a70609dd636%2Findex.html%3FisPreview%3Dtrue)

```

Specification: https://rawcdn.githack.com/w3c/aria/692220c7656b3de75d3b156d18329a70609dd636/index.html?isPreview=true&publishDate=2021-01-20
ReSpec version: 25.6.0
File a bug: https://github.com/w3c/respec/
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 0ms.
    at ExecutionContext._evaluateInternal (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:217:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:106:16)
    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:208:12)
    at async toHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:87:18)
    at async Object.generate [as respec] (/u/spec-generator/generators/respec.js:13:40)
    at async /u/spec-generator/server.js:89:44

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231020.)._
</details>


jongund:
Updated the pull request by restricting the use of the caption technique to first child element.

index.html
```diff
@@ -3109,7 +3115,8 @@ <h2>Definition of Roles</h2>
...
+				<p>Authors SHOULD provide a reference to the <code>figure</code> from the main text, but the <code>figure</code> need not be displayed at the same location as the referencing element. Authors MAY reference text serving as a caption using <pref>aria-describedby</pref>.</p>
+				<p>Authors MAY give each element with role table a brief label that describes the purpose of the content in the table.  Authors MAY use a visible caption as the first or last child element of the figure to provide a label.  The caption MAY be an instance of the standard host language <code>caption</code> element or an instance of an element with role <rref>caption</rref>.  If a visible label is present, but not contained in the first or last child element with a role of <code>caption</code>, authors SHOULD reference the visible label with <pref>aria-labelledby</pref>.</p>
```
scottaohara: swap out references to "table" with "figure"

scottaohara:
looking at this some more and the caption role definition in general, i'm curious as to why it states:
>Authors SHOULD reference the element with role caption by setting aria-describedby on the figure or table.

in HTML, caption and figcaption provide accessible names for these elements, not descriptions.

jongund:
Suggestion from ARIA call today to add treegrid to the list of roles and to merge legend and caption to create a single role

jongund:
@scottaohara @jnurthen @mcking65 

I have updated the pull request to make caption and legend synonyms, but there is one difference that may need to be addressed.  Elements with the role caption should be able to provide the accessible name for elements with the role figure, from either being the first or last child element with role of caption.  In the updated pull request I define caption as providing an accessible name from either being the first or last child element, and legend only from being the first child element.

carmacleod:
@jongund 
> In the updated pull request I define caption as providing an accessible name from either being the first or last child element, and legend only from being the first child element.

Actually, according to the HTML spec, [table caption](https://html.spec.whatwg.org/multipage/tables.html#the-caption-element) needs to be the first child also:

> Contexts in which this element can be used:
As the first element child of a table element.

It has the same requirement as [legend](https://html.spec.whatwg.org/multipage/form-elements.html#the-legend-element):
> Contexts in which this element can be used:
As the first child of a fieldset element.

It's only [figcaption](https://html.spec.whatwg.org/multipage/grouping-content.html#the-figcaption-element) that can be first or last:
> Contexts in which this element can be used:
As the first or last child of a figure element.

So:
> ...there is one difference that may need to be addressed...

Agreed. If we define caption and legend roles as synonyms, then HTML caption, figcaption, and legend elements would all default to a single "caption/legend role", and this would likely lead to some strange ambiguities regarding whether or not "caption/legends" are allowed at the end of their parent elements.

index.html
```diff
@@ -531,9 +531,9 @@ <h3>Accessible Name Calculation</h3>
 				<dd>One of the following values:
 					<ol>
 						<li>author: name comes from values provided by the author in explicit markup features such as the <pref>aria-label</pref> attribute, the <pref>aria-labelledby</pref> attribute, or the host language labeling mechanism, such as the <code>alt</code> or <code>title</code> attributes in <abbr title="Hypertext Markup Language">HTML</abbr>, with HTML <code>title</code> attribute having the lowest precedence for specifying a text alternative.</li>
+            <li>caption/legend: name comes from the text value of the first descendant <a>element</a> node with the role of <code>legend</code> or <code>caption</code>.  In the case of the role <code>caption</code> name is also allowed from the last child node of the <a>element</a>.  Although "caption" or "legend" may be allowed in addition to "author" in some <a>roles</a>, "caption" or "legend" is used in content only if higher priority "author" features are not provided. Priority is defined by the <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_te">accessible name and description computation</a> algorithm.  [[ACCNAME-1.1]]</li>
```
scottaohara: is it necessary to state here where the caption/legend is placed beyond it must be a direct child of the parent element?  I would think we can just indicate the appropriate placement for the elements specifically in their definition / their parent element's definition.

e.g. mention in `group`, `radiogroup`, `grid` and `table` that a `caption/legend` must be the first child element.  mention in `figure` that `caption/legend` must be the first or last child element.

index.html
```diff
@@ -1636,8 +1636,18 @@ <h2>Definition of Roles</h2>
 		<div class="role" id="caption">
 			<rdef>caption</rdef>
 			<div class="role-description">
-				<p>On-screen descriptive text for a figure or table in the page.</p>
-				<p>Authors SHOULD reference the element with role <code>caption</code> by setting <pref>aria-describedby</pref> on the figure or table.</p>
+				<p>A visible name or caption for an image, table or grid roles.  See synonym <rref>legend</rref>.</p>
```
scottaohara: `caption` *would not* provide a name to an image. This should say `figure`.

index.html
```diff
@@ -1636,8 +1636,18 @@ <h2>Definition of Roles</h2>
...
+				<p>An element with role <code>caption</code> MAY provide an accessible name for  <a href="#namefromcaptioncaption">elements that support the "caption/legend" naming technique</a> by using one of three methods:</p>
+				<ul>
+					<li>caption: The element with role <code>caption</code> is the first child of the element it is naming.</li>
```
scottaohara: per my comment from line 534, this would be where we designate that `figure` can have a caption as first or last child.  `table` would require it to be first child.

this is assuming the naming via caption/legend convention is going to be part of 1.2.  If not, then this is moot for now and authors would have to associate via `aria-labelledby`

index.html
```diff
@@ -1636,8 +1636,18 @@ <h2>Definition of Roles</h2>
...
+					<li>caption: The element with role <code>caption</code> is the last child of the element it is naming.</li>
+					<li>reference: The element with role <code>caption</code> is referenced by the element it names via the <pref>aria-labelledby</pref> attribute.</li>
+				</ul>
+				<p>Authors SHOULD ensure that:</p>
+				<ul>
+					<li>All elements with role <code>caption</code> are associated with one or more elements with roles that support the "caption/legend" naming technique.</li>
+					<li>When an element with role <code>caption</code> is activated by touch or a pointer, focus moves to the element that is the first focusable user interface component which is part of the sequential tab order of the page that is a descendant of the associated table, grid or treegrid.</li>
```
scottaohara: as mentioned before, this functionality of moving focus should be removed as it's not how `caption` or `legend` work in HTML, and is not something authors SHOULD do.

index.html
```diff
@@ -3109,7 +3124,8 @@ <h2>Definition of Roles</h2>
 			<rdef>figure</rdef>
 			<div class="role-description">
 				<p>A perceivable <rref>section</rref> of content that typically contains a <a>graphical document</a>, images, code snippets, or example text. The parts of a <code>figure</code> MAY be user-navigable.</p>
-				<p>Authors SHOULD provide a reference to the <code>figure</code> from the main text, but the <code>figure</code> need not be displayed at the same location as the referencing element. Authors MAY reference text serving as a caption using <pref>aria-describedby</pref>. Authors MAY provide a label using <pref>aria-label</pref> or MAY reference text serving as a label using <pref>aria-labelledby</pref>.</p>
+				<p>Authors SHOULD provide a reference to the <code>figure</code> from the main text, but the <code>figure</code> need not be displayed at the same location as the referencing element. Authors MAY reference text serving as a caption using <pref>aria-describedby</pref>.</p>
```
scottaohara: this is confusing now:
>Authors MAY reference text serving as a caption using `aria-describedby`

as the `caption` is meant to provide the accessible name of the `figure`, we shouldn't also say it should be referenced by `aria-describedby`.

index.html
```diff
@@ -3109,7 +3124,8 @@ <h2>Definition of Roles</h2>
...
+				<p>Authors MAY give each element with role figure a brief label that describes the purpose of the content in the figure.  Authors MAY use a visible caption as the first or last child element of the figure to provide a label.  The caption MAY be an instance of the standard host language <code>caption</code> element or an instance of an element with role <rref>caption</rref>.  If a visible label is present, but not contained in the first or last child element with a role of <code>caption</code>, authors SHOULD reference the visible label with <pref>aria-labelledby</pref>.</p>
```
scottaohara: >The caption MAY be an instance of the standard host language <code>caption</code> element 

needs to be changed to
>The caption MAY be an instance of the standard host language <code>figcaption</code> element 
a `<caption>` outside of the context of a `<table>` would be invalid HTML.

This paragraph also gives leniency to where a `caption` may live in the DOM, while we also state it should be a child of the element it is providing the accessible name to.   if a `caption` can be place outside of a parent `table`, `figure`, `group` then should it not be indicated it should be associated via `aria-owns` to facilitate the expected parent/child relationship?

index.html
```diff
@@ -3380,6 +3401,7 @@ <h2>Definition of Roles</h2>
...
+				<p>Authors MUST give each element with role grid a brief label that describes the purpose of the content in the gird.  Authors MAY use a visible caption as the first child element of the gird to provide a label.  The caption MAY be an instance of the standard host language <code>caption</code> element or an instance of an element with role <rref>caption</rref>.  If a visible label is present, but not contained in the first child element with a role of <code>caption</code>, authors SHOULD reference the visible label with <pref>aria-labelledby</pref>.</p>
```
scottaohara: again, should `aria-owns` be used in tandem with `aria-labelledby`?

index.html
```diff
@@ -4155,20 +4182,15 @@ <h2>Definition of Roles</h2>
...
-				<p>A element with role <code>legend</code> element can provide an accessible name for elements with grouping roles if it is progammatically associated with the element.  Authors MAY associate an element with role <code>legend</code> with an element by using one of two methods:</p>
+				<p>A visible name or caption for a group of related user interface components.  See synonym <rref>caption</rref>.</p>
+				<p>An element with role <code>legend</code> MAY provide an accessible name for <a href="#namefromcaptioncaption">elements that support the "caption/legend" naming technique</a> by using one of two methods:</p>
 				<ul>
-					<li>legend: The element contains an element with role <code>legend</code> element.  If the element contains more than one element with role <code>legend</code>, only the first descendant element with role <code>legend</code> is used for computing the accessible name.</li>
+					<li>legend: The element with role <code>legend</code> is the first child of the element it is naming.</li>
 					<li>reference: The element with role <code>legend</code> is referenced by the element it names via the <pref>aria-labelledby</pref> attribute.</li>
```
scottaohara: again should also include mention of `aria-owns`?

index.html
```diff
@@ -1636,8 +1636,18 @@ <h2>Definition of Roles</h2>
...
 			<rdef>caption</rdef>
 			<div class="role-description">
-				<p>On-screen descriptive text for a figure or table in the page.</p>
-				<p>Authors SHOULD reference the element with role <code>caption</code> by setting <pref>aria-describedby</pref> on the figure or table.</p>
+				<p>A visible name or caption for an image, table or grid roles.  See synonym <rref>legend</rref>.</p>
+				<p>An element with role <code>caption</code> MAY provide an accessible name for  <a href="#namefromcaptioncaption">elements that support the "caption/legend" naming technique</a> by using one of three methods:</p>
+				<ul>
+					<li>caption: The element with role <code>caption</code> is the first child of the element it is naming.</li>
+					<li>caption: The element with role <code>caption</code> is the last child of the element it is naming.</li>
+					<li>reference: The element with role <code>caption</code> is referenced by the element it names via the <pref>aria-labelledby</pref> attribute.</li>
```
scottaohara: mention `aria-owns` to keep the caption/legend associated with the element it names?  (more on this in comments further down the document)

index.html
```diff
@@ -4155,20 +4182,15 @@ <h2>Definition of Roles</h2>
...
-				<p>The legend method of naming is supported only if the element being named has one of the following grouping roles:</p>
-				<ul>
-					<li><pref>group</pref></li>
-					<li><pref>radiogroup</pref> </li>
-				</ul>
 				<p>Authors SHOULD ensure that:</p>
 				<ul>
-					<li>All elements with role <code>legend</code> are associated with one or more elements with grouping roles.</li>
+					<li>All elements with role <code>legend</code> are associated with one or more elements with roles that support the "caption/legend" naming technique.</li>
 					<li>When an element with role <code>legend</code> is activated by touch or a pointer, focus moves to the first focusable element in the associated group of user interface components.</li>
```
scottaohara: as mentioned before, this functionality of moving focus should be removed as it's not how caption or legend work in HTML, and is not something authors SHOULD do.

index.html
```diff
@@ -8438,6 +8460,7 @@ <h5>Presentational Roles Conflict Resolution</h5>
 			<div class="role-description">
 				<p>A <rref>section</rref> containing data arranged in rows and columns. See related <rref>grid</rref>.</p>
 				<p>The <code>table</code> role is intended for tabular containers which are not interactive. If the tabular container maintains a selection state, provides its own two-dimensional navigation, or allows the user to rearrange or otherwise manipulate its contents or the display thereof, authors SHOULD use <rref>grid</rref> or <rref>treegrid</rref> instead.</p>
+				<p>Authors MUST give each element with role table a brief label that describes the purpose of the content in the table.  Authors MAY use a visible caption as the first child element of the table to provide a label.  The caption MAY be an instance of the standard host language <code>caption</code> element or an instance of an element with role <rref>caption</rref>.  If a visible label is present, but not contained in the first child element with a role of <code>caption</code>, authors SHOULD reference the visible label with <pref>aria-labelledby</pref>.</p>
```
scottaohara: change:
>but not contained in the first child element with a role of <code>caption</code>

to:
>but not the first child element of an element with a role of <code>table</code>

and again, mention `aria-owns` as well as labelling?

index.html
scottaohara: resolving per previous comment about required context roles

index.html
scottaohara: resolving per previous comment about required context roles

index.html
jongund: @scottaohara  I removed the authoring should's for the touch and pointer behaviors

index.html
jongund: @scottaohara made the change to "but not the first child element of an element with a role of `table`"

index.html
jongund: @scottaohara This should be resolved now with my latest push

index.html
jongund: @scottaohara The reference to an image has been removed in the latest draft

index.html
jongund: @scottaohara I think it gives authors a better understanding of the usage of the caption/legend technique and is pretty succinct, so I would advocate leaving it.  It will help people understand how it is different form other labeling techniques that are in the list of naming techniques, without having to read the details for specific roles that can be labeling with caption/legend.

scottaohara: understand the desire to leave it in, but then there should be better clarity that this statement only applies to `figure`:
>In the case of the role `caption` name is also allowed from the last child node of the element

index.html
scottaohara: the incorrect html element is still referenced in the latest update to the text.

index.html
jongund: @scottaohara changed to say "in some cases" for the last child role

jnurthen:
Closing as this feature is no longer intended to ship

PR #1023, opened 2019-07-31
remove 4 Global States and properties (#999)

### LINKED ISSUE #999: Revising what ARIA attributes should be considered global
Per mention in the WG call today, and spurred by the discussions started with #990, I've created the following break down for review of the attributes I believe should remain global, and those that should become either prohibited on roles where they don't make sense, or should be removed from being global attributes all together.  

The following has been worded under the assumption the attributes would be removed from being global:

## Attributes to remain global
* `aria-atomic`
* `aria-busy` (state)
* `aria-current` (state)
* `aria-dropeffect` - "deprecated"
* `aria-flowto`
* `aria-grabbed` (state) - "deprecated"
* `aria-hidden` (state)
* `aria-live`
* `aria-owns`
* `aria-relevant`
* `aria-roledescription`

## Attributes to remain global, but prohibited
Per other work already happening for `role=generic` and accName, `aria-label` and `aria-labelledby` have consensus of staying global, but should be noted as prohibited due to wanting to advice against naming certain roles.
* `aria-label`
* `aria-labelledby`

I'd also submit that if a role shouldn't be named, it also likely shouldn't have a description associated with it...
* `aria-describedby`
* `aria-details`

## Undecided
`aria-keyshortcuts`  
I'm not exactly sure what to do with this one.  Is there actual value in allowing keyboard shortcuts for non-interactive elements?  I can't think of a reason, but maybe someone else?

## Attributes to reallocate to specific roles
Pending further discussion, especially regarding the roles that are listed with a "(?)", the following attributes should *not* be global, or should be noted as prohibited on all applicable roles where they would not be expected.

### `aria-disabled` should be allowed on:
* Widget Container Roles  
* Widget roles *excluding*:
  * link
  * meter
  * progressbar
  * <s>scrollbar</s>
  * <s>separator</s>
  * tabpanel

#### Discussion point
`aria-disabled` _could_ be allowed on the following Document Structure roles as well:
* application
* group
* toolbar

Doing so would allow for parity with HTML's allowance of `<fieldset disabled>`. Though allowing on these could also be more trouble than its worth.  Really interested to hear what others think.

### `aria-errormessage` and `aria-invalid` should be allowed on:
* `checkbox`
* `combobox`
* `listbox`
* `tree` 
* `gridcell`
* `radiogroup`
* `searchbox`
* `slider`
* `switch` 
* `textbox`

### `aria-controls` should be allowed on:
* `button`
* `checkbox`
* `link`
* `menuitem` (?)
* `menuitemcheckbox` (?)
* `menuitemradio` (?)
* `option` (?)
* `radio`
* `scrollbar`
* `searchbox`
* `slider`
* `spinbutton`
* `switch`
* `tab`
* `textbox` (?)
* `treeitem` (?)

Note, #995 may mean `aria-controls` should remain global until a larger discussion on if any redefinition of the attribute should take place so that AT will be more inclined implement it / not ignore it.

### `aria-haspopup` should be allowed on:
* `button`
* `combobox`
* `gridcell` (?)
* `menuitem`
* `searchbox`
* `slider` (?) - datalist is allowed on HTML's input but treated differently
* `textbox`
* `treeitem` (?)

## (edit) Final note:
Whatever is decided here, I'd imagine a good path forward would be to not necessarily act on this as a single issue, but to create sub-issues for any reallocation. Then make PRs on those sub-tasks individually, instead of tackling this all at once.

jnurthen:
closes #999

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1023.html" title="Last updated on Oct 22, 2019, 4:56 PM UTC (61782df)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1023/690285a...61782df.html" title="Last updated on Oct 22, 2019, 4:56 PM UTC (61782df)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -994,11 +994,15 @@
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
-            <ul>
-              <li><pref>aria-activedescendant</pref></li>
-              <li><sref>aria-expanded</sref></li>
-            </ul>
-            </td>
+							<ul>
+								<li><pref>aria-activedescendant</pref></li>
+								<li><sref>aria-disabled</sref></li>
+								<li><pref>aria-errormessage</pref></li>
+								<li><sref>aria-expanded</sref></li>
+								<li><pref>aria-haspopup</pref></li>
+								<li><sref>aria-invalid</sref></li>
+							</ul>
+						</td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
--- a/index.html
+++ b/index.html
@@ -3639,7 +3639,12 @@
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
-						<td class="role-properties"><pref>aria-activedescendant</pref></td>
+						<td class="role-properties">
+							<ul>
+								<li><pref>aria-activedescendant</pref></li>
+								<li><sref>aria-disabled</sref></li>
+							</ul>
+						</td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
--- a/index.html
+++ b/index.html
@@ -7801,6 +7801,8 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
+								<li><pref>aria-errormessage</pref></li>
+								<li><sref>aria-invalid</sref></li>
 								<li><pref>aria-readonly</pref></li>
 								<li><pref>aria-required</pref></li>
 								<li><pref>aria-valuemax</pref></li>
--- a/index.html
+++ b/index.html
@@ -10906,6 +10906,10 @@
 						<td class="property-applicability">Placeholder</td>
 					</tr>
 					<tr>
+							<th class="property-descendants-head" scope="row">Inherits into Roles:</th>
+							<td class="property-descendants">Placeholder</td>
+					</tr>					
+					<tr>
 						<th class="property-value-head" scope="row">Value:</th>
 						<td class="property-value"><a href="#valuetype_idref">ID reference</a></td>
 					</tr>
--- a/index.html
+++ b/index.html
@@ -10747,7 +10747,7 @@
 				<p>Indicates that the <a>element</a> is <a>perceivable</a> but disabled, so it is not editable or otherwise <a>operable</a>. See related <sref>aria-hidden</sref> and <pref>aria-readonly</pref>.</p>
 				<p>For example, irrelevant options in a radio group may be disabled. Disabled elements might not receive focus from the tab order. For some disabled elements, applications might choose not to support navigation to descendants. In addition to setting the <sref>aria-disabled</sref> attribute, authors SHOULD change the appearance (grayed out, etc.) to indicate that the item has been disabled.</p>
 				<p>The <a>state</a> of being disabled applies to the current element and all focusable descendant elements of the element on which the <sref>aria-disabled</sref> <a>attribute</a> is applied.</p>
-                <p class="note">While <sref>aria-disabled</sref> and proper scripting can successfully disable an element with role <rref>link</rref>, fully disabling a host language equivalent can be problematic. Authors are advised not to use <sref>aria-disabled</sref> on elements that cannot be fully disabled through author scripting alone.</p>
+                <p class="note">While <sref>aria-disabled</sref> and proper scripting can successfully disable an element with role <rref>link</rref>, fully disabling a host language equivalent can be problematic. Authors are advised not to use <sref>aria-disabled</sref> on elements that cannot be disabled through features of the host language alone.</p>
 			</div>
 			<table class="state-features">
 				<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -1596,6 +1596,8 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
+								<li><sref>aria-disabled</sref></li>
+								<li><pref>aria-haspopup</pref></li>
 								<li><sref>aria-expanded</sref></li>
 								<li><sref>aria-pressed</sref></li>
 							</ul>
@@ -1872,7 +1874,9 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
+								<li><pref>aria-errormessage</pref></li>
 								<li><sref>aria-expanded</sref></li>
+								<li><sref>aria-invalid</sref></li>
 								<li><pref>aria-readonly</pref></li>
 								<li><pref>aria-required</pref></li>
 							</ul>
@@ -2184,9 +2188,12 @@
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
-						<td class="role-properties">
+						<td class="role-properties">					
 							<ul>
 								<li><pref>aria-autocomplete</pref></li>
+								<li><pref>aria-errormessage</pref></li>
+								<li><pref>aria-haspopup</pref></li>
+								<li><sref>aria-invalid</sref></li>
 								<li><pref>aria-readonly</pref></li>
 								<li><pref>aria-required</pref></li>
 							</ul>
@@ -2424,7 +2431,12 @@
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
-						<td class="role-properties"><pref>aria-activedescendant</pref></td>
+						<td class="role-properties">
+							<ul>
+								<li><pref>aria-activedescendant</pref></li>
+								<li><sref>aria-disabled</sref></li>
+							</ul>
+						</td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
@@ -3529,7 +3541,11 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
-                <li><sref>aria-expanded</sref></li>
+								<li><sref>aria-disabled</sref></li>
+								<li><pref>aria-errormessage</pref></li>
+								<li><sref>aria-expanded</sref></li>
+								<li><pref>aria-haspopup</pref></li>
+								<li><sref>aria-invalid</sref></li>
 								<li><pref>aria-readonly</pref></li>
 								<li><pref>aria-required</pref></li>
 								<li><sref>aria-selected</sref></li>
@@ -3855,7 +3871,7 @@
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
-						<td class="role-properties"> </td>
+						<td class="role-properties"><sref>aria-disabled</sref></td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
@@ -4498,7 +4514,9 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
+								<li><pref>aria-errormessage</pref></li>
 								<li><sref>aria-expanded</sref></li>
+								<li><sref>aria-invalid</sref></li>
 								<li><pref>aria-multiselectable</pref></li>
 								<li><pref>aria-readonly</pref></li>
 								<li><pref>aria-required</pref></li>
@@ -5361,7 +5379,9 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
+								<li><sref>aria-disabled</sref></li>
 								<li><sref>aria-expanded</sref></li>
+								<li><pref>aria-haspopup</pref></li>
 								<li><pref>aria-posinset</pref></li>
 								<li><pref>aria-setsize</pref></li>
 							</ul>
@@ -6430,6 +6450,8 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
+								<li><pref>aria-errormessage</pref></li>
+								<li><sref>aria-invalid</sref></li>
 								<li><pref>aria-readonly</pref></li>
 								<li><pref>aria-required</pref></li>
 							</ul>
@@ -6780,6 +6802,7 @@
 						<td class="role-properties">
 							<ul>
 								<li><pref>aria-colindex</pref></li>
+								<li><sref>aria-disabled</sref></li>
 								<li><sref>aria-expanded</sref></li>
 								<li><pref>aria-level</pref></li>
 								<li><pref>aria-posinset</pref></li>
@@ -7076,6 +7099,7 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
+								<li><sref>aria-disabled</sref></li>
 								<li><pref>aria-orientation</pref></li>
 								<li><pref>aria-valuemax</pref></li>
 								<li><pref>aria-valuemin</pref></li>
@@ -7554,6 +7578,7 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
+								<li><sref>aria-disabled</sref> (if focusable)</li>
 								<li><pref>aria-orientation</pref></li>
 								<li><pref>aria-valuemax</pref> (if focusable)</li>
 								<li><pref>aria-valuemin</pref> (if focusable)</li>
@@ -7661,6 +7686,9 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
+								<li><pref>aria-errormessage</pref></li>
+								<li><pref>aria-haspopup</pref></li>
+								<li><sref>aria-invalid</sref></li>
 								<li><pref>aria-orientation</pref></li>
 								<li><pref>aria-readonly</pref></li>
 								<li><pref>aria-valuemax</pref></li>
@@ -8390,6 +8418,7 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
+								<li><sref>aria-disabled</sref></li>
 								<li><sref>aria-expanded</sref></li>
 								<li><pref>aria-posinset</pref></li>
 								<li><sref>aria-selected</sref></li>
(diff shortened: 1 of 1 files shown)


carmacleod:
Summary of changes: 

Removes the following attributes from global attributes:
- aria-disabled
- aria-errormessage
- aria-haspopup
- aria-invalid

Adds the above as supported attributes on:
- https://pr-preview.s3.amazonaws.com/w3c/aria/1023/690285a...758d784.html#aria-disabled
- https://pr-preview.s3.amazonaws.com/w3c/aria/1023/690285a...758d784.html#aria-errormessage
- https://pr-preview.s3.amazonaws.com/w3c/aria/1023/690285a...758d784.html#aria-haspopup
- https://pr-preview.s3.amazonaws.com/w3c/aria/1023/690285a...758d784.html#aria-invalid

jnurthen:
> Here's the only things that jumped out at me:
> `menubar` allows for `aria-disabled`. being a related concept to `menubar`, I think `toolbar` should for `aria-disabled` as well.

Happy to add that

> 
> I question if `menuitemcheckbox` and `menuitemradio` should allow for `aria-errormessage` or `aria-invalid`.

I think I agree but I can't change that in this PR due to the inheritance model. This PR does not make things worse as they already allow these attributes. Are you ok making this change and potentially tackling this in 1.3?

scottaohara:
>Are you ok making this change and potentially tackling this in 1.3?

yes if you remove the word "potentially" from that statement :D

carmacleod:
>> I question if menuitemcheckbox and menuitemradio should allow for aria-errormessage or aria-invalid.

> I think I agree but I can't change that in this PR due to the inheritance model. 

Does it make sense to list `aria-errormessage` and `aria-invalid` in "Prohibited States and Properties:" section for `menuitemcheckbox` and `menuitemradio`? That way, the inheritance model doesn't have to change.

The one that felt weird to me is `aria-haspopup` on a `slider`? So I googled it, and discovered that [whatsock recommends using `aria-haspopup` for a context menu](http://whatsock.com/training/matrices/). Not sure if this advice is old, or still valid? If valid, then aria-haspopup may need to stay global?

Also want to note that if a `group` is representing a [fieldset](https://html.spec.whatwg.org/multipage/form-elements.html#the-fieldset-element), then it _can_ be disabled. But in other contexts, it can't. Not quite sure how to handle that.

One final note... not sure why [Level Access tests `aria-haspopup` on links and radios](https://labs.levelaccess.com/index.php/ARIA_Haspopup_property)?
Hmmm... looking a bit further, it looks like we might have to allow `aria-haspopup` on links because people seem to be creating dropdown navs like this [a11y-101 nested nav example](https://a11y-101.com/development/nested-navigation) and this [ingresso-bootstrap dropdown example](https://ingresso.github.io/ingresso-bootstrap/components/dropdowns/).

scottaohara:
@carmacleod thanks for making me think some more :)

>Does it make sense to list aria-errormessage and aria-invalid in "Prohibited States and Properties:" section for menuitemcheckbox and menuitemradio? That way, the inheritance model doesn't have to change.

That might be a reasonable approach, yet at the same time I feel a bit of apprehension about doing it so as to avoid changing other things.  If others think this is an ok thing to do, then sure... but I also think it's just as reasonable to leave as is and then tackle in 1.3.

>The one that felt weird to me is aria-haspopup on a slider? 

I also went back and forth on whether a slider should have an `aria-haspopup`  but I also don't have a strong argument for or against, so imo better to stay with current allowances here. 

>So I googled it, and discovered that whatsock recommends using aria-haspopup for a context menu. Not sure if this advice is old, or still valid? If valid, then aria-haspopup may need to stay global?

[ARIA 1.0 defines `aria-haspopup`](https://www.w3.org/TR/wai-aria-1.0/complete#aria-haspopup) as:
>Indicates that the element has a popup context menu or sub-level menu...

"context menu" is absent from the definition as of ARIA 1.1. While that does not indicate it can't be one, the definition now includes:
>For the popup element to be keyboard accessible, authors SHOULD ensure that the element that can trigger the popup is focusable, that there is a keyboard mechanism for opening the popup

Per the updated definition it seems to me that the focusable expectation precludes this attribute from being global.

>One final note... not sure why Level Access tests aria-haspopup on links and radios?

Level Access has a bunch of tests for ARIA support.  Being a global attribute it's reasonable for there to be tests for it.  I'm sure I have some more buried somewhere as well :)

>Hmmm... looking a bit further, it looks like we might have to allow aria-haspopup on links because people seem to be creating dropdown navs like this a11y-101 nested nav example and this ingresso-bootstrap dropdown example.

Regarding those examples, I'd submit that improper use of the attribute does not indicate that it should be continued to be allowed.  If anything I'd point to those examples as evidence of author confusion about the attribute, and why it shouldn't be allowed on links to help mitigate misuse.


>Also want to note that if a group is representing a fieldset, then it can be disabled. But in other contexts, it can't. Not quite sure how to handle that.

Yeh, I was also thinking about that... the flexible nature of the `group` role does make that a bit weird, but if we kept `aria-disabled` as allowed on `group` it would continue to keep parity with a `group` being used as a stand-in for `fieldset`, and again would be no worse off than it presently is for other potentially dubious usage.  @jnurthen I think it does make sense to continue to allow `aria-disabled` on `group` if possible.

carmacleod:
@scottaohara Thanks for the thoughtful responses.

> If others think this is an ok thing to do, then sure... but I also think it's just as reasonable to leave as is and then tackle in 1.3.

👍 

> I also went back and forth on whether a slider should have an aria-haspopup but I also don't have a strong argument for or against, so imo better to stay with current allowances here.

👍 

> "context menu" is absent from the definition as of ARIA 1.1. 

I'd like to ask the team what the history of this is.

> Per the updated definition it seems to me that the focusable expectation precludes this attribute from being global.

`tabindex="0"` on any element? Anyhow, first need to determine whether aria-haspopup is needed for context menu, or not.

> If anything I'd point to those examples as evidence of author confusion about the attribute, and why it shouldn't be allowed on links to help mitigate misuse.

👍 

> if we kept aria-disabled as allowed on group it would continue to keep parity with a group being used as a stand-in for fieldset

👍

accdc:
Personally, I'm opposed to removing aria-disabled, specifically from role=link as we discussed today on the call.

We have current clients using similar implementations as outlined below, which is a valid use of aria-disabled.

There is a web portal that includes multiple components, each of which has specific actions that can be applied to each region. These include "Edit", "Remove", "Move", and so on.

Each of these action elements are not standard active elements, but clickable icons made of span tags plus relevant visual styling.

To ensure keyboard accessibility and screen reader accessibility, they also include tabindex="0" and role="link", plus relevant scripting so that keyboard behavior matches mouse click functionality.

These use role="link" instead of role="button", because there is a submit button at the end of the form that needs to be easily navigable by pressing "b" using relevant desktop screen readers, or in the swipe order for such controls on mobile devices, and the use of role="button" for all of these action elements would interfere with the user’s ability to do so without having to go through every such control prior to that button. Thus role="link" is preferable for this purpose instead.

Moreover, based on certain conditions, not all of these action elements are enabled at the same time, so they are sometimes disabled and sometimes not. They are visually styled to convey this disabled state to sighted users, but there also needs to be a way for screen reader users to receive the same information. So, aria-disabled="true" is dynamically added to the relevant role="link" elements that are thus effected for this purpose.

If support for aria-disabled is removed from role="link", then it will be impossible for screen reader users to receive the same level of information as sighted users, which defeats the intended purpose of ARIA.

scottaohara:
>If support for aria-disabled is removed from role="link", then it will be impossible for screen reader users to receive the same level of information as sighted users, which defeats the intended purpose of ARIA.

This is not an accurate statement.  Buttons, which can be disabled, could have been used here and would have allowed screen reader users to receive the same level of information as sighted users. However, as noted they were not used in an effort to simplify the UX for screen reader users navigating by button elements. The two issues should not be conflated.

That said, given the choice to continue to allow the attribute on `role=link` for 1.2, or to impede other progress this PR would make, I'll go with progress.  I suggest we have a grounded discussion about this again for attribute parity in 1.3, and for now note that while `aria-disabled` may be used on a `role=link`, it is not recommended to be used on `<a href>` for the reasons I stated on the call, and then some.

stes-acc:
As a matter of fact, the discussion for the host language is pretty old:
see e.g. `https://discourse.wicg.io/t/adding-a-disabled-attribute-to-a-s/1116/2`

accdc:
"That said, given the choice to continue to allow the attribute on role=link for 1.2, or to impede other progress this PR would make, I'll go with progress.
I suggest we have a grounded discussion about this again for attribute parity in 1.3, and for now note that while aria-disabled may be used on a role=link,
it is not recommended to be used on <a href> for the reasons I stated on the call, and then some."

What has been concerning me for a while now, is that there seems to be a strong push to make all ARIA usage synchronous with HTML, which is being justified by the role parity project. ARIA is not designed to be only used with HTML.

scottaohara:
I understand that ARIA is not for use with just HTML.  It is, however, designed to be in conformance with the allowances of the host language.  Hence my request stating that "disabling a host language equivalent may not be allowed"

accdc:
"I understand that ARIA is not for use with just HTML. It is, however, designed to be in conformance with the allowances of the host language. Hence my
request to note that "disabling a host language equivalent may not be allowed""

I understand that, and for the intended purpose of the APG, that is fine, because the APG is non-normative and is a series of recommended best practices that are not meant to restrict the accessibility API mappings in any way.

However, what is being proposed here is the removal of support from the ARIA spec itself, which does remove support from all other technologies, not just HTML.

It is this last part that is concerning me greatly, because this will impact the mappings everywhere, and this is essentially saying that the only thing that matters for ARIA is what happens in HTML.

scottaohara:
What else is being removed here that you take issue with?  

As already noted, `aria-disabled` isn't being removed from `role=link` for 1.2, and though I'm requesting we talk about it again in 1.3, clearly that doesn't mean it would even be removed then either.  

So, if there's something else, please indicate it specifically so we can work that out.

carmacleod:
+1 for [moving the discussion of disabled links to ARIA 1.3](#1099).

carmacleod:
Here is a new summary of the changes this PR would introduce as of October 11:

Removes the following 4 attributes from global attributes:
- aria-disabled
- aria-errormessage
- aria-haspopup
- aria-invalid

Adds the above attributes back as supported attributes on certain roles (and their subclasses), as follows:
- [aria-disabled](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1023.html#aria-disabled)
  - application
  - button
  - composite
  - gridcell
  - group
  - input
  - link
  - menuitem
  - row * see note, below
  - scrollbar
  - separator
  - tab
  - toolbar * see note, below

- [aria-errormessage](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1023.html#aria-errormessage)
  - application
  - checkbox 
  - combobox 
  - gridcell 
  - listbox 
  - radiogroup 
  - slider 
  - textbox 
  - tree 

- [aria-haspopup](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1023.html#aria-haspopup)
  - application
  - button 
  - combobox 
  - gridcell 
  - menuitem 
  - slider
  - tab
  - textbox 
  - treeitem 

- [aria-invalid](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1023.html#aria-invalid)
  - application
  - checkbox 
  - combobox 
  - gridcell 
  - listbox 
  - radiogroup 
  - slider 
  - textbox 
  - tree 

NOTE: Now that `group` supports `aria-disabled`, `row`, `select`, and `toolbar` don't need to explicitly support it because they are subclasses of group and will inherit `aria-disabled`.

index.html
```diff
@@ -10751,6 +10751,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				<p>Indicates that the <a>element</a> is <a>perceivable</a> but disabled, so it is not editable or otherwise <a>operable</a>. See related <sref>aria-hidden</sref> and <pref>aria-readonly</pref>.</p>
 				<p>For example, irrelevant options in a radio group may be disabled. Disabled elements might not receive focus from the tab order. For some disabled elements, applications might choose not to support navigation to descendants. In addition to setting the <sref>aria-disabled</sref> attribute, authors SHOULD change the appearance (grayed out, etc.) to indicate that the item has been disabled.</p>
 				<p>The <a>state</a> of being disabled applies to the current element and all focusable descendant elements of the element on which the <sref>aria-disabled</sref> <a>attribute</a> is applied.</p>
+                <p class="note">While <sref>aria-disabled</sref> and proper scripting can successfully disable an element with role <rref>link</rref>, fully disabling a host language equivalent can be problematic. Authors are advised not to use <sref>aria-disabled</sref> on elements that cannot be fully disabled through author scripting alone.</p>
```
smhigley: This bit is a little confusing to me:
>  elements that cannot be fully disabled through author scripting alone

since most things (including links) can be made impossible to interact with through author scripting, and my gut developer reaction to the wording is "challenge accepted" 😆 ,

Would something like this be worth considering?:
>  elements that cannot be fully disabled through HTML alone

with the intent that native HTML elements that do not support the `disabled` attribute also shouldn't support `aria-disabled` (though that's also my editorial opinion 😄)

smhigley: To clarify what I mean by links can be disabled through author scripting:

In addition to `pointer-events: none` and preventing click/keydown/etc, I could work around the context menu issue by cloning the `<a>`, removing the `href`, and programmatically calling a `contextmenu` event on the cloned node. Then you'd get a comparable context menu without the "open link" options.

It's a completely silly exercise of course, and no one would do it in an actual product, but it's why I'm leery of saying "this is impossible with scripting"

scottaohara: how about 
>elements that cannot be disabled through features of the host language alone.

that keeps this agnostic rather than singling out HTML links, and yes, removes the "challenge" aspect of the advisement

stes-acc:
We definitely should address things like

`<a href="[href]" aria-disabled="true">`

without explicit role="link" declaration with respect to what HTML browsers should map in these cases.

However, this is part of  the larger discussion "using native roles + aria properties without explicit aria role declaration".

In my opinion, for the aforementioned example browsers should just map disabled attribute to accessibility API as there were a role defined. Explicitly disabling will happen by author JS/CSS code or /and native element href manipulation.

index.html
carmacleod: Or possibly (fewer words `<g>`)
> elements that cannot be disabled using the host language.

carmacleod:
@stes-acc 
> We definitely should address things like `<a href="[href]" aria-disabled="true">`...

Please move this comment to #1099 where we will discuss "disabled links" (and their related mappings, if any) in the ARIA 1.3 time frame.

accdc:
"What else is being removed here that you take issue with?"

There isn't, the link role was what concerned me since we have clients using this right now based on prior support.

Also, one of our clients has a software-based web UI development platform that has some major restrictions in what their developers are allowed to do within that framework. One of these restrictions is that they cannot change the href attributes on any links, and the platform does support disabling links, which are styled differently and do nothing when clicked. This platform does allow for the addition of supporting attributes however, so aria-disabled was added to make these links equally accessible for non-sighted screen reader users. Moreover, these changes were made as part of a lawsuit, and if this suddenly stopped working it would put the client back in legal jeopardy.

If this attribute is not removed from link, then I support the pull.

stes-acc:
> "What else is being removed here that you take issue with?"
> 
> There isn't, the link role was what concerned me since we have clients using this right now based on prior support.
> 
> Also, one of our clients has a software-based web UI development platform that has some major restrictions in what their developers are allowed to do within that framework. One of these restrictions is that they cannot change the href attributes on any links, and the platform does support disabling links, which are styled differently and do nothing when clicked. This platform does allow for the addition of supporting attributes however, so aria-disabled was added to make these links equally accessible for non-sighted screen reader users. Moreover, these changes were made as part of a lawsuit, and if this suddenly stopped working it would put the client back in legal jeopardy.
> 
> If this attribute is not removed from link, then I support the pull.

+1

LJWatson:
I realise that support for -disabled on elements with role="link" isn't going to be dropped for the moment, but it worries me that the rationale for not dropping it is apparently based on a single use case - and a use case that is, I think, not a particularly strong one.

@accdc wrote:
>There is a web portal that includes multiple components, each of which has specific actions that can be applied to each region. These include "Edit", "Remove", "Move", and so on.>

>These use role="link" instead of role="button", because there is a submit button at the end of the form that needs to be easily navigable by pressing "b" using relevant desktop screen readers, or in the swipe order for such controls on mobile devices, and the use of role="button" for all of these action elements would interfere with the user’s ability to do so without having to go through every such control prior to that button. Thus role="link" is preferable for this purpose instead.>

Arguably these should be buttons, given that they appear to invoke an action on the current interface, rather than fetch a resource. Making them links might ease screen reader navigation to the submit button if that's the intended action, but it equally makes navigation to these links harder if one of them is the intended action.

We push developers to code semantically and to use the appropriate HTML (code) for the purpose, but this approach suggests it's ok not to do that when it suits accessibility, and that makes it hard to keep insisting that it's important at all.

I think we need strong use cases for making changes, or deciding not to make changes, to standards, and that those use cases need to be backed up by a reasonable amount of evidence to support their validity in the wild.

accdc:
"I realise that support for -disabled on elements with role="link" isn't going to be dropped for the moment, but it worries me that the rationale for not dropping it is apparently based on a single use case - and a use case that is, I think, not a particularly strong one."

I don't understand this, because people are already doing this.

E.G

Microsoft: Turn on or off links in email messages - Outlook
https://support.office.com/en-us/article/turn-on-or-off-links-in-email-messages-2d79b907-93b6-4774-82e6-1f0385cf20f8
"If the Junk Email Filter doesn’t consider a message to be spam but does consider it to be phishing, the message is left in the Inbox, but any links in the message are disabled and you can’t use the Reply and Reply All commands."

How to disable a link using only CSS? - GeeksforGeeks
https://www.geeksforgeeks.org/how-to-disable-a-link-using-only-css/

People have already been doing this in the wild for years.

I don't understand this reasoning that, because links can't be disabled in HTML, nobody is allowed to have disable links anywhere. The reason being, that people are already implementing disabled links, they are just not accessible because they do not convey their disabled state to non-sighted screen reader users.

So what I keep hearing here, is that blind people shouldn't have a need to know when a link is disabled, because it doesn't match HTML. We had the same discussion when it was proposed to remove support for aria-checked on role=option, because it doesn't match HTML.

I disagree with that. I have every right as a blind person to get the same meaning from a control as a sighted person does visually, and right now, for links, using aria-disabled is the only option available for doing this.

As I mentioned before, it is fine to add guidance to the APG to discourage people from doing this as a general practice. However, it is not okay to remove the ability to do this from all technologies everywhere just because some people don't agree with it.

jnurthen:
>     1. Make support for aria-disabled on `row` conditional on the `row` being in a `grid` or `treegrid`.

Logged https://github.com/w3c/aria/issues/1101 . As these were previously global this is not adding support here. This can be addressed in 1.3

jnurthen:
>     2\. Since `aria-errormessage` and `aria-invalid` are supported by slider, make them also supported on spinbutton.

resolved

jnurthen:
> In the preview, aria-errormessage is not showing that it is inherrited into any roles. Shouldn't it be inherrited by searchbox? Interestingly, aria-invalid is showing as inherrited into 7 roles. It should be possible to use aria-errormessage on any element that supports aria-invalid. Maybe this is a script bug?

Was missing the placeholder section in the property

jnurthen:
> Less important: the abstract role `select` is listed as inherrited for aria-disabled. Is that a bug in the script propagating inherited states and properties?

no - aria-disabled is on group so inherits to select and all of its child roles.

PR #1036, opened 2019-08-21
Remove should from tooltip role definition

### LINKED ISSUE #1035: Remove normative should from tooltip role
https://github.com/w3c/aria-practices/issues/1099 has an inconsistency between APG and spec. We think the best way to resolve is to remove the normative should from the aria spec about tooltip delay.

jnurthen:
resolves #1035


jnurthen:
ugh - started with unclean repo

Issue #1046, opened 2019-08-29
Does combobox require an accessible name?


WilcoFiers:
In the axe-core 3.3.0 we added a rule to check for the accessible name of non-native form fields, including combobox. Since than I've had a few people point me to this example in the ARIA practices:

https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

```html
<label id="ex2-label" class="combobox-label">
  Choice 2 Fruit or Vegetable
</label>
<div class="combobox-wrapper">
  <div role="combobox"
       aria-expanded="false"
       aria-owns="ex2-listbox"
       aria-haspopup="listbox"
       id="ex2-combobox">
    <input type="text"
           aria-autocomplete="list"
           aria-controls="ex2-listbox"
           aria-labelledby="ex2-label"
           id="ex2-input">
  </div>
  <ul aria-labelledby="ex2-label"
      role="listbox"
      id="ex2-listbox"
      class="listbox hidden">
  </ul>
</div>
```

This example has three elements with a role that, according to ARIA 1.1 require an accessible name: `combobox`, `textbox`, and `listbox`. In this example, the `textbox` and `listbox` have the same accessible name, and the `combobox` has no accessible name.

I'm guessing combobox probably doesn't need to have an accessible name, unless role=combobox is used on a native text box. Should this be changed?

accdc:
I'm not sure that's right though, having unlabeled comboboxes makes it impossible for screen readers to jump between combobox widgets and intuitively differentiate them for non-sighted users.

jnurthen:
I believe in the linked issue at the last F2F we decided that the combo and its child textinput should simply have the same label.

aleventhal:
@jnurthen or why not have the browser copy one to the other when one of them is nameless. Or, my favorite, I suggest the AT is responsible for doing this. NVDA will look at the parent of an nameless input and for a combobox name and use that.

accdc:
The issue with having browsers do it, is that I then need to add this to the AccName spec, which is annoying because it doesn't follow any form of paradigm existing elsewhere in the spec for any other widget type. I would have to make a step just for dealing with the combobox role.

mcking65:
[...]

Issue #998, opened 2019-06-19
The Combobox 1.0 design pattern references incorrectly include aria-owns instead of aria-controls.


accdc:
As we discussed in the APG meeting last week, the APG design pattern for the Combobox 1.0 markup does not take into account that aria-owns was changed to aria-controls to correct long standing accessibility API issues when applied to elements that do not support child elements, such as the HTML input element.

It was the filing of this issue years ago against the 1.0 spec that led to the creation of the 1.1 Combobox design pattern, yet none of the 1.0 design pattern references include this change within the 1.1 ARIA spec. As a result, this issue continues to negatively impact the accessibility of this widget globally, just because the spec was not changed to reflect this going forward.

This is primarily an editorial change, to simply replace the aria-owns references to aria-controls instead, and that is all that needs to be done. By changing this within real world implementations, the accessibility of these widgets suddenly become accessible across all mainstream platforms, whereas if this change is not done, serious accessibility issues will always continue to exist otherwise.

This change requires no action by the browser implementors, because doing this within the markup already results in the desired accessible results across all mainstream browsers already.

Example:
http://whatsock.com/tsg/Coding%20Arena/ARIA%20Comboboxes/ARIA%20Comboboxes%20(Simulated,%20Readonly)/demo.htm

accdc:
Another example that includes the use of standard HTML input elements can be tested here as well:
http://whatsock.com/tsg/Coding%20Arena/ARIA%20Comboboxes/ARIA%20Comboboxes%20(Native%20Inputs,%20Editable%20and%20Readonly)/demo.htm

jnurthen:
I'm still not clear on what needs to change. If it is as simple as you are proposing a PR would help clarify.

mcking65:
While I agree with @accdc on the fundamentals of this issue, I don't think this is editorial. We would not have the 1.1 pattern if this change would have been permitted in ARIA 1.1. Some members of the ARIA WG, during the 1.1 cycle, were concerned that the controls relationship was insufficient for establishing a relationship between the input of a combobox and the popup of a combobox. Because of those concerns, the 1.1 pattern was added.

The downside is that nearly two years after 1.1, there still is no support in screen readers for the 1.1 pattern. Ironically though, if you make the simple tweak to the 1.0 pattern that Bryan suggests, we get all the benefits of the 1.1 pattern, and we get them now because it already works with screen readers. I don't think the people who objected to just making the simple tweak foresaw this.

[...]

Issue #909, opened 2019-02-23
Clarify combobox label placement


carmacleod:
The [ARIA 1.1 spec for combobox](https://www.w3.org/TR/wai-aria-1.1/#combobox) needs to clearly specify whether to put the label on the parent (combobox) element, or the child (textbox) element.

Example 5 does show it on the combobox element and not on the textbox, but are examples normative?

An argument for having the label on the combobox element is that AT need to know the label of the combobox in order to list them by name in a form fields dialog, otherwise you have problems like [this "unlabelled comboboxes" JAWS issue](https://github.com/FreedomScientific/VFO-standards-support/issues/77).

An argument for having the label on the textbox element is that it takes focus, so AT will want to speak the name.

I don't think it makes sense to ask authors to put the label on both.

Unfortunately, neither JAWS, NVDA, nor VoiceOver seem to fully support ARIA 1.1 comboboxes yet.
Focus goes to the textbox, and the screen readers just say "edit, type in text". The user won't even know they are in a combobox. (If the textbox has a value, that is read as well, but still no indication of combobox).

carmacleod:
Related: https://github.com/w3c/aria/issues/893

carmacleod:
Closing. This discussion is covered in #893.

Issue #893, opened 2019-01-25
The ARIA 1.1 Combobox design pattern endorses unlabelled form fields as normative spec.


accdc:
Within the ARIA 1.1 Combobox design pattern, the following markup is specified as the proper syntax for the new 1.1 combobox structure.

```
<div aria-label="Tag" role="combobox" aria-expanded="true" aria-owns="owned_listbox" aria-haspopup="listbox"> 
    <input type="text" aria-autocomplete="list" aria-controls="owned_listbox" aria-activedescendant="selected_option">
</div>
<ul role="listbox" id="owned_listbox"> 
    <li role="option">Zebra</li> 
    <li role="option" id="selected_option">Zoom</li>
</ul>
```

Specifically, the following:

```
    <input type="text" aria-autocomplete="list" aria-controls="owned_listbox" aria-activedescendant="selected_option">
```

This is going to lead to major accessibility issues across devices since the combobox role is not a substitute for a labelling mechanism according to the AccName spec.

carmacleod:
> the combobox role is not a substitute for a labelling mechanism according to the AccName spec.

Would it work if we add an item to the beginning of the list in AccName step 2E to say something like:
- If the embedded control has role textbox, and it is a descendant of an element with role combobox, then compute the text alternative for the combobox beginning with step 2, and set the result to that text alternative. Append the value of the textbox control to the result, with a space. Return the result.

@joanmarie Is this close? You are so good with AccName words! :smile:

carmacleod:
Here are [3 more combobox test cases](https://carmacleod.github.io/playground/combo-test.html) that I would like to ensure would work with any accname changes that we make. Note that none of these have aria-autocomplete.
- Combo 1 is basically the same as @accdc's example, above, except with a visible label
- Combo 2 is the same as Combo 1 except read-only
- Combo 3 uses a span for the textbox, tabindex="0" on the combobox div, and it "pops up" a dialog instead of a listbox

One problem that is noticeable is that JAWS and NVDA don't even recognize Combo 1 and Combo 2 as a combobox, let alone labelling them correctly. The same problem exists with the APG ARIA 1.1 combobox examples, but it is not as noticeable because those examples all have aria-autocomplete and they are spoken as "Edit, has autocomplete", so the user might miss that there's no mention of "combobox".

stes-acc:
> One problem that is noticeable is that JAWS and NVDA don't even recognize Combo 1 and Combo 2 as comboboxes....

[...]

PR #1051, opened 2019-09-11
Fix combobox issues

mcking65:
The purpose of this pull request is to revise a collection of combobox issues by better aligning the specification with what has been long supported by both browsers and assistive technologies.

The rationale for these changes and an explanation of how they resolve the issues listed below are available on this [wiki page about resolving ARIA 1.1 combobox issues](https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues).

### Summary of changes in this pull request

1. Revise definition of combobox to encompass implementations that do not support text editing, such as some implementations of the HTML select element by browsers.
2. Revise description of the structure of combobox so the element with the combobox role is the focused element instead of a wrapper that is not focusable and the relationship with the popup is defined by aria-controls.
3. Change the super class to input from select so it describes the actual structure; select is a composite and the revised combobox structure does not include any focusable descendants. This also has the beneficial side effect of removing aria-orientation, which is not applicable.
4. Add aria-activedescendant as a supported property because it is no longer inherrited from its super class.
5. Remove all required owned elements. A combobox no longer has any descendants.
6. Change section 2.3 (id="managingfocus") to remove combobox from the list of composite widgets and for clarity and compatibility with revised combobox structure.
7. Editorial revisions to the definition and description of the aria-activedescendant property to reflect the changes to the combobox structure.

### Issues resolved by this pull request

* #998: The Combobox 1.0 design pattern references incorrectly include aria-owns instead of aria-controls.
* #776: aria-controls a required property for Combobox? 
* #716: Combobox: aria-controls or owns should be on input, not on combobox element, and only when combobox is expanded
* #909: Clarify combobox label placement
* #1046: Does combobox require an accessible name? 
* #893: The ARIA 1.1 Combobox design pattern endorses unlabelled form fields as normative spec.
* #742: role combobox either misses text or is incorrect regarding role searchbox
* #817: Update combobox spec as per html spec/html-aam mapping for `<select>`* [HTML-AAM: select element should map to listbox role regardless of presentation 

### Example Implementation

[View an test an example implementation of the new pattern in the Authoring Practices Guide](https://raw.githack.com/w3c/aria-practices/aria1.2-combobox-proposal/examples/combobox/combobox-autocomplete-list.html)

### Preview Changes

[View the rendered changes in the compare branch for this pull request](https://raw.githack.com/w3c/aria/fix-combobox/index.html#combobox)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1051.html" title="Last updated on Oct 31, 2019, 10:52 PM UTC (6e38169)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1051/c05acab...6e38169.html" title="Last updated on Oct 31, 2019, 10:52 PM UTC (6e38169)">Diff</a>


accdc:
Thanks, it's true that Level Access points to the APG for general guidance, but not for the combobox widget. All of the reasons why are discussed in the following article, which I've instituted as a company policy for Level Access.
https://www.levelaccess.com/differences-aria-1-0-1-1-changes-rolecombobox/

If your 1.0 comboboxes were constructed using this criteria, they would not be inaccessible. These have been tested across all mainstream AT and browser combinations on all supporting platforms to ensure this.

index.html
```diff
@@ -2107,23 +2115,50 @@ <h2>Definition of Roles</h2>
...
+          A <code>combobox</code> is said to be expanded when both the <code>combobox</code> element showing its current value and its associated popup element are visible.
+          Authors MUST set <sref>aria-expanded</sref> to <code>true</code> on an element with role <code>combobox</code> when it is expanded and <code>false</code> when it is collapsed.
+        </p>
+				<p>
+          Authors MUST ensure the popup element associated with a <code>combobox</code> has a role of <rref>listbox</rref>, <rref>tree</rref>, <rref>grid</rref>, or <rref>dialog</rref>.
+          Authors MUST set <pref>aria-controls</pref> on a <code>combobox</code> element to a value that refers to the <code>combobox</code> popup element.
+        </p>
+				<p>
+          Elements with the role <code>combobox</code> have an implicit <pref>aria-haspopup</pref> value of <code>listbox</code>.
+          If the <code>combobox</code> popup element has a role other than <rref>listbox</rref>, authors MUST specify a value for <pref>aria-haspopup</pref> that corresponds to the type of its popup.
```
carmacleod: Maybe change:
> corresponds to the type of its popup.

to:
> corresponds to the role of its popup.

(because it feels more precise)

index.html
```diff
@@ -2107,23 +2115,50 @@ <h2>Definition of Roles</h2>
...
+          When a descendant of the popup element is active, authors MAY set <pref>aria-activedescendant</pref> on the <code>combobox</code> to a value that refers to the active element within the popup while focus remains on the <code>combobox</code> element.
+        </p>
 				<pre class="example highlight">
-				  &lt;div aria-label="Tag" role="combobox" aria-expanded="true" aria-owns="owned_listbox" aria-haspopup="listbox"&gt;
-				      &lt;input type="text" aria-autocomplete="list" aria-controls="owned_listbox" aria-activedescendant="selected_option"&gt;
-				  &lt;/div&gt;
-				  &lt;ul role="listbox" id="owned_listbox"&gt;
-				      &lt;li role="option"&gt;Zebra&lt;/li&gt;
-				      &lt;li role="option" id="selected_option"&gt;Zoom&lt;/li&gt;
+		      &lt;input type="text" aria-label="Tag"
```
carmacleod: This example would be friendlier if it used a `<label for="the-text-input">Tag</label>` instead of an invisible `aria-label="Tag"`. :)

index.html
```diff
@@ -2107,23 +2115,50 @@ <h2>Definition of Roles</h2>
...
+				<p>
+          The ARIA 1.0 specification describes a <code>combobox</code> pattern where the input element with the <code>combobox</code> role references the popup element with <pref>aria-owns</pref> instead of <pref>aria-controls</pref>.
+          <a>User agents</a>, <a>assistive technologies</a>, and conformance checkers SHOULD continue to support the ARIA 1.0 pattern so that existing implementations of the ARIA 1.0 pattern remain functional.
+          The ARIA 1.1 specification describes a <code>combobox</code> pattern where the element with role <code>combobox</code> is a composite container instead of a focusable input.
+          <a>User agents</a> and <a>assistive technologies</a> did not adequately support the ARIA 1.1 pattern, and this pattern supersedes the ARIA 1.1 pattern, which authors SHOULD abandon.
```
carmacleod: Not sure if "this pattern" is specific enough in the following phrase:
> and this pattern supersedes the ARIA 1.1 pattern

Consider changing to:
> so the ARIA 1.2 pattern supersedes the ARIA 1.1 pattern

index.html
```diff
@@ -2163,15 +2198,7 @@ <h2>Definition of Roles</h2>
 					</tr>
```
carmacleod: Weird typo in "Related Concepts:" row (was there before, but trivial to fix):
- Delete `/a>` after `[[HTML]]`

index.html
```diff
@@ -313,14 +313,19 @@ <h2>Managing Focus and Supporting Keyboard Navigation</h2>
...
+            User agents are not expected to validate that the active descendant is a descendant of the container element.
+            It is the responsibility of the user agent to ensure that keyboard events are processed at the <a>element</a> that has <abbr title="Document Object Model">DOM</abbr> focus.
+            Any keyboard events directed at the active descendant bubble up to the <abbr title="Document Object Model">DOM</abbr> element with focus for processing.
+          </p>
 	  <section id="managingfocus_authors">
 	    <h3>Information for Authors</h3>
 	    <p>If the author removes the element with focus, the author SHOULD move focus to a logical element. Similarly, authors SHOULD not scroll the element with focus off screen unless the user performed a scrolling action.</p>
 	    <p>Authors SHOULD ensure that all interactive <a>elements</a> are focusable and that all parts of composite widgets are either focusable or have a documented alternative method to achieve their function.</p>
 	    <p>Authors MUST manage focus on the following container roles:</p>
 	    <ul>
```
carmacleod: > Authors MUST manage focus on the following container roles:

Doesn't combobox need to stay in this list? Authors do still need to manage focus on them.
Perhaps change the list intro sentence to read:

> Authors MUST manage focus on the following roles:

And add `<li><rref>combobox</rref></li>` back into the list?

carmacleod:
Hi @eps1lon 

Thanks for the good feedback on mouse behavior problems in the combobox example!

The current issue is about updating the ARIA combobox specification. The above Example Implementation was based off of an [APG combobox example](http://w3c.github.io/aria-practices/examples/combobox/aria1.0pattern/combobox-autocomplete-list.html). Issues can be opened in the [APG issue tracker](https://github.com/w3c/aria-practices/issues).

I suspect that the mouse problems you are experiencing may exist in some of the other APG combobox examples as well.

index.html
mcking65: No, this is one of the more important changes. It changes combobox from a composite, which manages focus, to an input. Inputs include button, radio, textbox, etc. These elements do not have focusable descendants.

authors manage focus in the popup that is controlled by the combobox, but there is nothing to manage in the combobox itself. In this way, combobox becomes much like a menubutton. It also makes it more like a native select.

mcking65:
@jhausler1266 commented:

> The 1.1 structure makes perfect sense as combobox is a _combination_ of an input and a list, tree, grid, etc.

I get this point totally ... there is a logiclal appeal to it. It is that logical appeal that convinced me to write the 1.1 structure into the spec even though I didn't really want to do so. 3 years later I am kicking myself for not pushing back harder, which would have probably meant we would have not made changes and stuck with 1.0 structure. maybe that would have been better than being in the position we are in now.

> Also, in the case where we are mimicking an html we place readonly on the so that users cannot type. In this case the input is readonly, but the combobox is not, and therefore should communicate that it can take input.

I hoped 1.1 would help with this, but in reality there is no way that it can. Since the mid-90's, screen readers have presented the textbox as the combobox, not as something separate or as something inside the combobox.

> Can we not move the AT vendors to move towards 1.1?

It is not impossible to get some of the bugs fixed. However, I don't think we will ever get screen readers to present the combobox as a composite widget -- never. I think the primary argument will always be that it will be really confusing for users because it has never been a true composite. It would also mean we have to also change how html:select@size=1 gets presented, which is also not proving straightforward.

On the other hand, changing a combobox from a composite to an input has pretty much 0 inertia on the screen reader side because it has always been presented that way.

index.html
carmacleod: Curious why not? As a sighted user, I prefer when a combobox has a visible label. I think we should model the most inclusive behavior?

index.html
carmacleod: > authors manage focus in the popup that is controlled by the combobox, but there is nothing to manage in the combobox itself.

Ok. I was thinking that "managing `aria-activedescendant` in the combobox" was the same thing as "managing focus in the combobox".

carmacleod:
In case it's interesting, NVDA merged a PR 3 days ago to [support ARIA 1.1 comboboxes](https://github.com/nvaccess/nvda/pull/10046) in Chrome and FF. Not saying this changes anything. Just for info.

index.html
mcking65: That was supposed to be a thumbs up ... guess I somehow chose wrong emoji. Making the change.

mcking65:
@jhausler1266,

With input from Microsoft and Freedom Scientific, I've been investigating and reviewing other ways of resolving the above listed issues. They get pretty complex. The changes in this pull request still seem to be the most straightforward.

Please read this wiki page:
https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues

index.html
```diff
@@ -10022,14 +10022,14 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p> If an element has <code>aria-autocomplete</code> set to <code>list</code> or <code>both</code>, authors MUST ensure both of the following conditions are met:</p>
 				<ol>
 					<li>The element has a value specified for <pref>aria-controls</pref> that refers to the element that contains the collection of suggested values.</li>
-					<li>Either the element or a containing element with role <rref>combobox</rref> has a value for <pref>aria-haspopup</pref> that matches the role of the element that contains the collection of suggested values.</li>
+					<li>The element has a value for <pref>aria-haspopup</pref> that matches the role of the element that contains the collection of suggested values.</li>
```
jnurthen: Can we change this? While I get that we don't want people using the 1.1 combobox pattern this would make them invalid and I'm concerned about that

JAWS-test:
Unfortunately I do not understand the following explanation in the [WIKI page](https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues)

> No select support: The input in the above code can be replaced with a div or span whose content displays the current value. This exactly replicates the functionality of combobox widgets that do not have a textbox.

If a `div` is used instead of `input`, no text can be entered. For AT users the difference is not noticeable, because in both cases the HTML element has the role `combobox`. So how should the difference between a combo box with text input and a combo box without text input be perceptible in the future?

mcking65:
> Unfortunately I do not understand the following explanation in the [WIKI page](https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues)
> 
> > No select support: The input in the above code can be replaced with a div or span whose content displays the current value. This exactly replicates the functionality of combobox widgets that do not have a textbox.
> 
> If a `div` is used instead of `input`, no text can be entered. For AT users the difference is not noticeable, because in both cases the HTML element has the role `combobox`. So how should the difference between a combo box with text input and a combo box without text input be perceptible in the future?

@JAWS-test wrote:

> Unfortunately I do not understand the following explanation in the [WIKI page](https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues)
> 
> > No select support: The input in the above code can be replaced with a div or span whose content displays the current value. This exactly replicates the functionality of combobox widgets that do not have a textbox.

This is a way to recreate the functional and accessibility equivalent of a `select` element with the size attribute set to 1. You cannot do this with the ARIA 1.1 pattern because there is no way for the AT to know which element contains the value.

> If a `div` is used instead of `input`, no text can be entered. For AT users the difference is not noticeable, because in both cases the HTML element has the role `combobox`. So how should the difference between a combo box with text input and a combo box without text input be perceptible in the future?

Exactly the same way it is done today . Compare the experience of [this combobox with a textbox implemented using the pattern in this PR](https://raw.githack.com/w3c/aria-practices/aria1.2-combobox-proposal/examples/combobox/combobox-autocomplete-list.html) with the experience of a `select` element with size set to 1 in either Chrome or Firefox on Windows. The combobox with a textbox has the editable text interface so a screen reader can announce it as an edit combobox. The select does not support editable text so it is announced as a combobox.

index.html
mcking65: Since this an authoring requirement, not a browser requirement, making the 1.1 implementations invalid is helpful. The few existing implementations of 1.1, which already do not work well, will become invalid. That will promote fixing them so they work better for more users in more browsers. And, at the same time, nothing here changes browser behavior in a way that would make the existing implementations of 1.1 worse than they already are.

JAWS-test:
@mcking65 Thank you very much for your explanation. I assumed that when using an ARIA role, the screenreader always outputs the same role regardless of the HTML element used (ARIA roles override HTML semantics). I have tested the [new example](https://raw.githack.com/w3c/aria-practices/aria1.2-combobox-proposal/examples/combobox/combobox-autocomplete-list.html) in Chrome with JAWS and it is indeed the case that at the `input` "edit combo" and at the `span` "combobox" is output, which surprises me very much. It's great that it works like this. But then you might have to update https://w3c.github.io/core-aam/ to display the different output depending on the HTML element 
However, it might be necessary to update [Core Accessibility API Mappings](https://w3c.github.io/core-aam/#details-id-14) to reflect the different output depending on the HTML element.

JAWS-test:
The new 1.2 combobox pattern is supported by JAWS and NVDA as follows:
* NVDA and Chrome, Firefox and IE: combobox with and without text input is not output with different roles, but with different status ("changeable" for combobox with text input)
* JAWS and Chrome: ok ("edit combo" or "combobox")
* JAWS and Firefox: combobox with and without text input is not output with different role
* JAWS and IE 11: wrong output because `combobox` and IE 11+JAWS never worked (none of the 1.0, 1.1 and 1.2 patterns)

JAWS-test:
Actually I like the ARIA 1.2 combobox design. However, I think there are 2 problems:
* `combobox` would be the only ARIA role that does not override the semantics of the host language, but enriches it. That would be a precedent that changes the previous role concept.
* For this reason, the `combobox` role can't fully work in markup languages that don't have a native element with the `textbox` role (such as SVG, I believe)  because the 1.2 design relies on the input field of the host language.

mcking65:
@JAWS-test wrote:

> Actually I like the ARIA 1.2 combobox design. However, I think there are 2 problems:
> 
> * `combobox` would be the only ARIA role that does not override the semantics of the host language, but enriches it. That would be a precedent that changes the previous role concept.

It does override the role of the element just like every other ARIA role. Putting role combobox on an input:type=text overrides the textbox role.

Lots of ARIA roles effectively extend or enrich native semantics. The grid role can be used to extend or enrich table. The searchbox role can be used to extend or enrich textbox. I'd say it is the case for the majority of widget roles.

> * For this reason, the `combobox` role can't fully work in markup languages that don't have a native element with the `textbox` role (such as SVG, I believe)  because the 1.2 design relies on the input field of the host language.

It would be a lot more work to make an editable combobox in any host language that does not have a text input that supports the accessible text interface. But, that was just as true for ARIA 1.1 and 1.0. And, it is just as true for creating an element with the searchbox role in one of those languages.

However, if you wanted to make a combobox that does not support editable text, it is now supported with ARIA 1.2. And, that is a big advantage if your host language does not have an element that supports the accessible text interface.

mcking65:
@JAWS-test wrote:

> @mcking65 Thank you very much for your explanation. I assumed that when using an ARIA role, the screenreader always outputs the same role regardless of the HTML element used (ARIA roles override HTML semantics). I have tested the [new example](https://raw.githack.com/w3c/aria-practices/aria1.2-combobox-proposal/examples/combobox/combobox-autocomplete-list.html) in Chrome with JAWS and it is indeed the case that at the `input` "edit combo" and at the `span` "combobox" is output, which surprises me very much. It's great that it works like this.
> 
> However, it might be necessary to update [Core Accessibility API Mappings](https://w3c.github.io/core-aam/#details-id-14) to reflect the different output depending on the HTML element.
> 
> The same applies to the following sentence in [Using ARIA](https://w3c.github.io/using-aria/#nativesemantics), which in my understanding would then also no longer be valid:
> 
> > Adding an ARIA role **overrides** the native role semantics in the accessibility tree which is reported via the accessibility API, ....
> 
> And maybe also the [ARIA specification](https://w3c.github.io/aria/#introroles):
> 
> > When WAI-ARIA roles **override** host language semantics, ...

I don't think we need those updates. It is up to the assistive technology to decide how they want to present the availability of editable text. It doesn't have to be part of the role announcement.

Of course, we always have the option of doing something in ARIA to specify what AT should do when the input is editable. I'd rather let AT attempt to solve this without specific direction from the ARIA specification. I think there is plenty of history that shows they will because it has been done in native applications. Apple's native approach is distinctly different from Microsoft's, and I don't see a reason to force them to be the same.

JAWS-test:
@mcking65 
> It does override the role of the element just like every other ARIA role. Putting role combobox on an input:type=text overrides the textbox role.
> Lots of ARIA roles effectively extend or enrich native semantics. The grid role can be used to extend or enrich table. The searchbox role can be used to extend or enrich textbox. I'd say it is the case for the majority of widget roles.

As far as I have understood, each ARIA role overrides the HTML element. I can use `role=grid` (with all necessary child roles like `row` and `gridcell`) on a `span` or `table` element and it makes no difference which HTML element I use.

With `role=combobox` this would be different according to the new specification:
* at an input field it would be an element like the HTML input with `datalist`
* at a `span` element it would be an element like the HTML `select`.

These are two different elements with different operation modes, but with the same ARIA role `combobox`.

mcking65:
@JAWS-test wrote:

> As far as I have understood, each ARIA role overrides the HTML element. I can use `role=grid` (with all necessary child roles like `row` and `gridcell`) on a `span` or `table` element and it makes no difference which HTML element I use.
> 
> With `role=combobox` this would be different according to the new specification:
> 
> * at an input field it would be an element like the HTML input with `datalist`

Not quite, but pretty close.

> * at a `span` element it would be an element like the HTML `select`.

Correct if size=1 and the operating system is Windows or Linux.
 
> These are two different elements with different operation modes, but with the same ARIA role `combobox` (see [w3c/html-aam#230](https://github.com/w3c/html-aam/issues/230)).

Whether the element with role combobox is an HTML input or a span, the ARIA role overrides the implicit semantics and makes the element into a combobox. This is identical to other ARIA roles.

The interactions conferred by the combobox role are identical regardless of whether the input is editable. The combobox role communicates that the input controls a popup and the value of the input can be set via the popup.

And, like other ARIA roles, ARIA properties are used to add additional information about the interactions supported for a particular combobox, e.g., aria-haspopup and aria-autocomplete. This is like the button role. If you add aria-haspopup or aria-pressed, it adds additional information about supported interactions for the widget. Similarly, aria-haspopup adds interaction information to the menuitem role and aria-multiselectable adds interaction information to widgets with a variety of roles.

JAWS-test:
> Whether the element with role combobox is an HTML input or a span, the ARIA role overrides the implicit semantics and makes the element into a combobox. This is identical to other ARIA roles.

I don't think I get it yet. Because role=combobox at the input element leads to the output of "edit combobox" in the screenreader and role=combobox at the span element leads to the output of "combobox". Those seem to be two different roles to me. In my opinion these are relatively different elements with different operation.

PR #1059, opened 2019-09-26
Correct text to require aria-expanded on the tab

### LINKED ISSUE #1040: discussion of aria-expanded on multi select tabpanels
There is a sentence in the [section on tablists](https://www.w3.org/TR/wai-aria-1.1/#tablist) that I found curious.

> For a multi-selectable tablist, authors SHOULD ensure each visible tabpanel has its aria-expanded attribute set to true, and that the remaining hidden tabpanel elements have their aria-expanded attributes set to false.

1) is it right to have aria-expanded on a tablist?
2) If so, wouldn't aria-expanded go on the trigger element not the destination element role="tab" rather than tabpanel?
3) If more than one tabpanel can be displayed at once, then why not just have aria-selected on all of them, instead of aria-expanded?

jnurthen:
closes #1040

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1059.html" title="Last updated on Sep 26, 2019, 10:21 PM UTC (c1b214b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1059/2b5a43e...c1b214b.html" title="Last updated on Sep 26, 2019, 10:21 PM UTC (c1b214b)">Diff</a>


carmacleod:
Looks good.

PR #1061, opened 2019-09-26
Remove roledescription examples

### LINKED ISSUE #83: WAI-ARIA 1.1: aria-roledescription
Filed at https://lists.w3.org/Archives/Public/public-pfwg-comments/2015JulSep/0007.html by Sailesh Panchang spanchang02@yahoo.com 

Refer to example #27/28 where this ARIA property has been set to "slide" on an element identified by an aria-labelledby.
Should the aria-labelledby text not identify this as a slide  like " Quarterly Report ...slide 2 of 12"?
Else, there may be some preceding content before the slides that informs the user that  a slide presentation follows and the slide identifier is "Quarterly Report ... 2 of 12".
And in example 29/30: <button aria-roledescription="attachment button">:
Generally one advises developers not to label a button element as "Register button" because screen readers will voice the text "button" twice: once as a role identifier and once as the element's name.
So in this example will button not be duplicated?
In this case, "attachment button" does not convey the button's purpose. Should it be "download attachment"?
In which case, this should be conveyed as part of the button's accessible name, no?
So really from the description of the property or examples, I am not clear about the usage of aria-roledescription.
Or, is the aria-roledescription  meant to override  the element's role that is exposed to AT? If yes, this is not clear from the description.
Is it meant to work like a "hint" as we see in iOS and OSX with VoiceOver? At present the aria-description is exposed as a hint. Will this new property cause confusion for makers of user agents / At?
It is quite likely that developers will be very confused. Many are overwhelmed and unsure of what text goes into the different attributes for naming / describing elements and what takes precedence.
Thanks,
Sailesh Panchang

jnurthen:
Closes #83

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1061.html" title="Last updated on Sep 26, 2019, 11:04 PM UTC (8b06c14)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1061/2b5a43e...8b06c14.html" title="Last updated on Sep 26, 2019, 11:04 PM UTC (8b06c14)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -12062,10 +12062,6 @@
 &lt;!-- remaining slide contents --&gt;
 &lt;/section&gt;</pre>
 				<p>In the previous examples, a screen reader user may hear "Quarterly Report, slide" rather than the more vague "Quarterly Report, region" or "Quarterly Report, group."</p>
-				<p>The following examples show the use of <code>aria-roledescription</code> to indicate that a <rref>button</rref> in a web-based email client is associated with an "attachment."</p>
-				<pre class="example highlight">&lt;div role="button" tabindex="0" aria-roledescription="attachment button"&gt;family_reunion.jpg&lt;/div&gt;</pre>
-				<pre class="example highlight">&lt;button aria-roledescription="attachment button"&gt;family_reunion.jpg&lt;/button&gt;</pre>
-				<p>In the previous two examples, because "button" is part of the localized description, a screen reader user should still understand how to interact with that control.</p>
 			</div>
 			<table class="property-features">
 				<caption>Characteristics:</caption>


carmacleod:
I agree that those 2 examples should be removed.
Looks good.

mcking65:
I disagree with removing the slide example. Why wasn't there discussion of this?

joanmarie:
@mcking65 The slide example was *not* removed. The attachment-button examples were removed.

joanmarie:
Here's the link to the two examples related to the slides which continue to be in the spec: https://w3c.github.io/aria/#example-24 https://w3c.github.io/aria/#example-25

PR #1062, opened 2019-09-26
Add directory Required Context role to listitem

### LINKED ISSUE #1030: `listitem` role seems to be missing `directory` as required context role
`listitem` (role) has this definition:

> `listitem` (role) 
> A single item in a list or directory.

However, it doesn't have `directory` listed as one of its required context roles. 

Is it a mistake that `directory` is not listed as a required context role?
Or is it a mistake that the definition says `listitem` can also be an item in a directory?

The discussion around this came up when trying to write an Accessibility Conformance Testing (ACT) Rule for required context roles: https://github.com/act-rules/act-rules.github.io/pull/255/files

jnurthen:
resolves #1030

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1062.html" title="Last updated on Sep 26, 2019, 11:22 PM UTC (98b20f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1062/2b5a43e...98b20f5.html" title="Last updated on Sep 26, 2019, 11:22 PM UTC (98b20f5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4579,6 +4579,7 @@
 						<th class="role-scope-head" scope="row">Required Context Role:</th>
 						<td class="role-scope">
 							<ul>
+								<li><rref>directory</rref></li>
 								<li><rref>group</rref></li>
 								<li><rref>list</rref></li>
 							</ul>


scottaohara:
Per [James's mention of low usage](https://github.com/w3c/aria/issues/1030#issuecomment-530493318), and the fact that `role=directory` is merely exposed as a list [with no tested pairing indicating it as a "directory"](https://scottaohara.github.io/tests/aria-directory/), it seems to me that we should seriously consider deprecating the role as it does nothing that couldn't be accomplished with a standard list.

joanmarie:
> Per [James's mention of low usage](https://github.com/w3c/aria/issues/1030#issuecomment-530493318), and the fact that `role=directory` is merely exposed as a list [with no tested pairing indicating it as a "directory"](https://scottaohara.github.io/tests/aria-directory/), it seems to me that we should seriously consider deprecating the role as it does nothing that couldn't be accomplished with a standard list.

Do we have an open issue for the deprecation?

scottaohara:
Nope @joanmarie.  just the mentioning of it in the linked issue.  I can/will create a new issue for this though.

PR #1069, opened 2019-09-28
removes the old default aria-checked value from properties table

### LINKED ISSUE #826: Checkbox has aria-checked required, but there is a default value for aria-checked
Proposal is to leave aria-checked required, remove the default value and only provide a fallback default value in the Core-AAM default values table.

scottaohara:
closes #826

removes the previous default state of undefined for the default state of “false” from the properties table, fully closing this issue.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1069.html" title="Last updated on Oct 16, 2019, 10:16 PM UTC (d415409)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1069/2b5a43e...d415409.html" title="Last updated on Oct 16, 2019, 10:16 PM UTC (d415409)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1908,10 +1908,6 @@
 						<th class="role-presentational-inherited-head" scope="row">Inherits Presentational:</th>
 						<td class="role-presentational-inherited"> </td>
 					</tr>
-						<tr>
-							<th class="implicit-values-head" scope="row">Implicit Value for Role:</th>
-							<td class="implicit-values">Default for <sref>aria-checked</sref> is <code class="default">false</code>.</td>
-						</tr>
 				</tbody>
 			</table>
 		</div>
--- a/index.html
+++ b/index.html
@@ -10195,7 +10195,7 @@
 				</thead>
 				<tbody>
 					<tr>
-						<th class="value-name" scope="row"><strong class="default">false (default)</strong></th>
+						<th class="value-name" scope="row">false</th>
 						<td class="value-description">The element supports being checked but is not currently checked.</td>
 					</tr>
 					<tr>
@@ -10208,7 +10208,7 @@
 						<td class="value-description">The element is checked.</td>
 					</tr>
 					<tr>
-						<th class="value-name" scope="row">undefined</th>
+						<th class="value-name" scope="row"><strong class="default">undefined</strong> (default)</th>
 						<td class="value-description">The element does not support being checked.</td>
 					</tr>
 				</tbody>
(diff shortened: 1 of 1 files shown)


jnurthen:
"Implicit Value for Role" needs removing on the Checkbox role too

scottaohara:
Ah will get on that. Thanks

jnurthen:
> Ah will get on that. Thanks

Not right now :)

index.html
```diff
@@ -1908,10 +1908,6 @@ <h2>Definition of Roles</h2>
 						<th class="role-presentational-inherited-head" scope="row">Inherits Presentational:</th>
 						<td class="role-presentational-inherited"> </td>
 					</tr>
-						<tr>
-							<th class="implicit-values-head" scope="row">Implicit Value for Role:</th>
-							<td class="implicit-values">Default for <sref>aria-checked</sref> is <code class="default">false</code>.</td>
-						</tr>
```
joanmarie: I don't think we want to delete these; we have scripts which do the removal when the spec is viewed. See, for example, https://w3c.github.io/aria/#code (shows no implicit values) and https://github.com/w3c/aria/blob/d5d2883/index.html#L2007 (has rows for implicit values).

scottaohara: thanks @joanmarie i'll update

scottaohara: OK.  should be good to go now.

index.html
```diff
@@ -10207,10 +10207,6 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 						<th class="value-name" scope="row">true</th>
 						<td class="value-description">The element is checked.</td>
 					</tr>
-					<tr>
-						<th class="value-name" scope="row"><strong class="default">undefined (default)</strong></th>
```
mcking65: I think we need to allow for undefined to represent that the element is not checkable. This is important for option and treeitem, which may or may not be checkable.

The default value should be defined for the specific roles, not in the property definition. If we want a default for checkbox and radio, it can be defined for those roles. We do not need to say the default value here.

jnurthen: aria-checked is required so it cannot have a default value as it is a validation error if it is not specified.

agenda+ for Tomorrow

scottaohara: if the element is not checkable, then should it just not have the `aria-checked` attribute?

treeitem and option do not require this attribute.  For treeitem aria-selected is required on the role, and i still am unconvinced that aria-checked should even be allowed on option.

index.html
```diff
@@ -10207,6 +10207,10 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 						<th class="value-name" scope="row">true</th>
 						<td class="value-description">The element is checked.</td>
 					</tr>
+					<tr>
+						<th class="value-name" scope="row">undefined</th>
```
jnurthen: undefined needs to be the default not false

Issue #723, opened 2018-05-03
Set up links to APG for each feature


michael-n-cooper:
Ideally using automation, and making sure links don't break across versions. Ported from [ACTION-2082](https://www.w3.org/WAI/ARIA/track/actions/2082).

joanmarie:
I think we want to prioritize this -- when feasible. See this thread https://twitter.com/jamiebuilds/status/1017114795189854209.

jnurthen:
Bulk closing stale issues - please reopen if necessary

PR #1071, opened 2019-09-28
Make Form Accessible Name Required True

### LINKED ISSUE #778: Consider treating nameless forms similar to nameless regions
In ARIA 1.1, we added the requirement that authors MUST name their regions. Then, in the Core AAM, we added conditional mappings: Expose as a landmark if it has an explicit name; don't expose as a landmark if it lacks a name.

It has been suggested that we do something similar for the form role. See #514). This issue was closed and moved to Core-AAM for the conditional mappings (https://github.com/w3c/core-aam/issues/11). However, we cannot take steps on that Core-AAM issue until we decide what to do in the ARIA spec itself.

jnurthen:
resolves #778

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1071.html" title="Last updated on Oct 10, 2019, 3:51 AM UTC (3475764)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1071/2b5a43e...3475764.html" title="Last updated on Oct 10, 2019, 3:51 AM UTC (3475764)">Diff</a>


index.html
```diff
@@ -3190,7 +3190,9 @@ <h2>Definition of Roles</h2>
...
+				<p>A form may be a mix of host language form controls, scripted controls, and hyperlinks. Authors are reminded to use native host language semantics to create form controls, whenever possible. For search facilities, authors SHOULD use the <rref>search</rref> role and not the generic <code>form</code> role.</p>
+				<p>Authors MUST give each element with role <code>form</code> a brief label that describes the purpose of the content in the form. Authors SHOULD reference a visible label with <pref>aria-labelledby</pref> if a visible label is present. Authors SHOULD include the label inside of a heading whenever possible. The heading MAY be an instance of the standard host language heading element or an instance of an element with role <rref>heading</rref>.</p>
```
joanmarie: Just out of curiosity, why is it desirable for authors to include the label inside of a heading? (I see that the same language is included for `region`; I'm questioning its applicability to `form`.)

scottaohara: if the WebAIM survey results are anything to go by, the reporting that navigating by landmarks has been declining year to year, while navigating by headings remains a consistently preferred method of navigation, would be one reason to call this out.

jnurthen: I had the same thoughts. Both of these seem more appropriate for APG rather than for the spec.
I think the benefit is in giving people using different navigation methods in screen readers the ability to get at the content. I don't think it is a bad pattern but I do question its relevance for the spec itself.

joanmarie: Ok, I can go either way. (i.e. make a decision, mark this convo as resolved, and land). Thanks for the thoughts!

jongund:
@jnurthen @joanmarie I don't think this is a good idea to make accessible name required for `role=form`, because it would then generate a landmark region which may be redundant with the main region on many pages (e.g. contribute to landmark pollution), and it would also require HTML form elements to also have an accessible names in validators.

I think it would be better to offer an authoring suggestion if there is more than one form on a page, then form region should have an accessible name.

jnurthen:
> @jnurthen @joanmarie I don't think this is a good idea to make accessible name required for `role=form`, because it would then generate a landmark region which may be redundant with the main region on many pages (e.g. contribute to landmark pollution), and it would also require HTML form elements to also have an accessible names in validators.

No. HTML `form` would not change. If a name is supplied it is a landmark if there is no name supplied it is not a landmark (just like `section` behaves today). Nothing would change in the HTML logic.
This change gives `role=form` parity with `role=region` where it is not allowed to use it without a name. Just like `role=region` if the author has not given the element a name then in order to resolve an ARIA validator then they should either add a name (if it should be a landmark) or remove the role (if it should not be a landmark)

> I think it would be better to offer an authoring suggestion if there is more than one form on a page, then form region should have an accessible name.

That is not what this change is intended to do. An un-named form region is not useful to anyone so should not be a landmark. This change clarifies this behaviour.

jongund:
@jnurthen OK thank you for the clarification, but I still think there should be some authoring guidance, like:
This role should only be used when there is more than one form on a page, or something like that.  

It seems `role=form` can be easily abused, since developers see the role name and think it should be used for any set of form controls on a page without understanding the user experience of using it.

jnurthen:
@jongund Isn't authoring guidance more appropriate for the APG - we have https://www.w3.org/TR/wai-aria-practices-1.1/#aria_lh_form

Rather than essentially replicating things which should be in APG in the spec I'd prefer us to resolve https://github.com/w3c/aria/issues/723 instead.

index.html
```suggestion
				<p>Authors MUST give each element with role <code>form</code> a brief label that describes the purpose of the form. Authors SHOULD reference a visible label with <pref>aria-labelledby</pref> if a visible label is present. Authors SHOULD include the label inside of a heading whenever possible. The heading MAY be an instance of the standard host language heading element or an instance of an element with role <rref>heading</rref>.</p> 
```

PR #1072, opened 2019-09-30
Aria-relevant clarifications for #1048

### LINKED ISSUE #1048: Description at aria-relevant incomprehensible and wrong
https://rawgit.com/w3c/aria/master/#aria-relevant:

> When text changes are denoted as relevant, user agents MUST monitor any descendant node change that affects the accessible name and description computation of the live region as if the accessible name were determined from contents (nameFrom: contents). 

* Does this mean that the elements in the live region should be treated as if their name is derived from the content? I.e. <code>button aria-label=button</code> should be output if its text content changes and not be output if its aria-label changes? This would be quite misleading, because then a change in the source code would be output, which would not be noticeable anywhere on the page when navigating with the screenreader.
* Or does this mean, only elements, which are labelled by content, should be output when changing? Then many updates would not be noticeable, especially since some elements cannot be labeled via content at all according to the specification, but only by the author.
* "When text changes are denoted as relevant": Does this mean that this only applies to aria-relevant=text and not to aria-relevant=additions? Should the Accessible name be output instead of the text content for aria-relevant=additions?  This would be very strange, because by default aria-relevant has the value "additions text", which means that the switch button above would be output twice when inserted into the DOM: once with the label in the aria label and once with the label in the text content.

> For example, a text change would be triggered if the HTML alt attribute of a contained image changed. However, no change would be triggered if there was a text change to a node outside the live region, even if that node was referenced (via aria-labelledby) by an element contained in the live region.

The example with the alt attribute seems to me to be wrong, because when I follow the link (https://rawgit.com/w3c/aria/master/#namecalculation), it explains that the alt attribute is name from author and not name from content

See also: https://github.com/FreedomScientific/VFO-standards-support/issues/274

jnurthen:
closes #1048

***

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Faria%2Fb3004ec92be7832e12546129eddc2890b905ea77%2Findex.html%3FisPreview%3Dtrue)

```
waiting for function failed: timeout 30000ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231072.)._
</details>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1072.html#aria-relevant" title="Last updated on Aug 25, 2022, 11:01 PM UTC (907cb8b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1072/c803925...907cb8b.html" title="Last updated on Aug 25, 2022, 11:01 PM UTC (907cb8b)">Diff</a>


index.html
```diff
@@ -11912,14 +11912,14 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>When text changes are denoted as relevant, user agents MUST monitor any descendant node change that affects the <a href="#textalternativecomputation">accessible name and description computation</a> of the live region as if the accessible name of the live region itself were determined from contents (<a href="#namecalculation">nameFrom: contents</a>). For example, a text change would be triggered if the HTML <code>alt</code> attribute of a contained image changed. However, no change would be triggered if there was a text change to a node outside the live region, even if that node was referenced (via <pref>aria-labelledby</pref>) by an element contained in the live region.</p>
```
joanmarie: So this clarifies things a bit, but the paragraph in question still strikes me as confusing.

Regarding the sentence you're changing, could we instead say:

    When text changes are denoted as relevant, user agents MUST monitor
    all descendant nodes for changes to their content and name.

?

If you agree with the above change, I would be tempted to remove the sentence beginning with "However". After all, my proposed change says to monitor all *descendant* nodes. Pointing out that nodes which are not descendant nodes are not monitored seems superfluous.

carmacleod:
Unfortunately, pr-preview is failing again.
Here's the [Diff](https://pr-preview.s3.amazonaws.com/w3c/aria/1072/01da023...b3004ec.html#aria-relevant) preview for this.

pkra:
@jnurthen ping.

jnurthen:
messed up the rebase. closing

PR #1086, opened 2019-10-07
update caption role for 1.2

### LINKED ISSUE #901: Add additional details regarding the new caption role
Now that I'm implementing some related support for our new `caption` role in a user agent, I noticed some things we (I) potentially neglected to address in the initial commit:

* There is currently not a "Required Context Role" for `caption`. I don't think we want `caption` instances floating around in unexpected places in the accessibility tree. I think we should add `table` and `figure` as the required context roles.
* HTML states that the `caption` element should appear as the first element. I don't know if we want to do that or not. But even if we address the first point, we don't know where the caption will appear in the accessibility tree (other than it will descend from a `table` or `figure`.

### LINKED ISSUE #1085: caption role should be associated with aria-labelledby not describedby
In lieu of future work to finalize the `caption` role, one thing I would like to get squared away for ARIA 1.2 is the current definition stating:
>Authors SHOULD reference the element with role caption by setting `aria-describedby` on the figure or table.

We should change this for the 1.2 release as `caption` should provide the `table` or `figure` its accessible name, not a description.

edit: ooops.... clearly was in the wrong repo when i made the issue...

scottaohara:
closes #1085

a `caption` and `figcaption` are meant to provide an accessible name to their parent `table` and `figure` respectively.

This PR changes the current direction to set `aria-describedby` on the parent element to `aria-labelledby` to better match how a `caption` should be used.

Additionally, this PR looks to close #901 as well.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1086.html" title="Last updated on Oct 22, 2019, 4:01 PM UTC (a0922ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1086/ff73742...a0922ac.html" title="Last updated on Oct 22, 2019, 4:01 PM UTC (a0922ac)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1636,10 +1636,14 @@
 		<div class="role" id="caption">
 			<rdef>caption</rdef>
 			<div class="role-description">
-				<p>A visible name or caption for a figure, table or grid in the page.</p>
-				<p>A <code>caption</code> SHOULD be a direct child of the figure, table or grid it is used to name. If a child of a table or grid the <code>caption</code> SHOULD be the first child. If a child of a figure the <code>caption</code> SHOULD be the first or last child.</p>
-				<p>Authors SHOULD reference the element with role <code>caption</code> by setting <pref>aria-labelledby</pref> on the parent element.</p>
-				<p>If a <code>caption</code> contains content that would serve as both a name and description to the parent element, authors MAY use <pref>aria-labelledby</pref> to reference an element within the <code>caption</code> to produce a concise accessible name, and <pref>aria-describedby</pref> to reference the descriptive content.</p>
+				<p>Visible content that names, and may also describe, a <rref>figure</rref>, <rref>table</rref>, or <rref>grid</rref>.</p>
+				<p>When using <code>caption</code> authors SHOULD ensure:</p>
+				<ul>
+					<li>The <code>caption</code> is a direct child of a <rref>figure</rref>, <rref>table</rref>, or <rref>grid</rref>.</li>
+					<li>The <code>caption</code> is the first child of a <rref>table</rref> or <rref>grid</rref>.</li>
+					<li>The <code>caption</code> is the first or last child of a <rref>figure</rref>.</li>
+				</ul>
+				<p>Authors SHOULD set <pref>aria-labelledby</pref> on the parent <code>figure</code>, <code>table</code>, or <code>grid</code> to reference the element with role <code>caption</code>. However, if a <code>caption</code> contains content that serves as both a name and description for its parent, authors MAY instead set <pref>aria-labelledby</pref> to reference an element within the <code>caption</code> that contains a concise name, and set <pref>aria-describedby</pref> to reference an element within the <code>caption</code> that contains the descriptive content.</p>
 
 				<pre class="example highlight">
 					&lt;div role="table" aria-labelledby="name" aria-describedby="desc"&gt;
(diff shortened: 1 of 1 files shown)


joanmarie:
Also does it close #901 ?

scottaohara:
I think a sentence to discourage long captions, particularly in tables/grids, is a good add.

I also see this change as correcting the misalignment with how HTML's `caption` and `figcaption` presently work, where they do provide the accessible names of their parents.  Since this misalignment wouldn't be rectified until 1.3 with Jon's larger updates, I do hope we can come to consensus here.

joanmarie:
> I think a sentence to discourage long captions, particularly in tables/grids, is a good add.

This would go a long way to make me feeling happy. :smile: 

> I also see this change as correcting the misalignment with how HTML's `caption` and `figcaption` presently work, where they do provide the accessible names of their parents.

Maybe we're correcting the misalignment in the wrong way? Consider the two examples found here: https://html.spec.whatwg.org/multipage/grouping-content.html#the-figcaption-element. The accessible names will be:

* A duck. Photograph courtesy of 🌟 News.
* Average rent for 3-room apartments, excluding non-profit apartments Zürich’s Statistics Office — 14 November 2017

With respect to the first one, it's not super huge, and "A duck" is a perfectly acceptable name. But "Photograph courtesy of  🌟 news" strikes me as descriptive info.

I think the second one is definitely descriptive info, and too long for a name.

But wait! There's more!! https://html.spec.whatwg.org/multipage/tables.html#the-caption-element

Here's the caption:

```
<caption>
<p>Table 1.
<p>This table shows the total score obtained from rolling two
six-sided dice. The first row represents the value of the first die,
the first column the value of the second die. The total is given in
the cell that corresponds to the values of the two dice.
</caption>
```

Again: "Table 1" perfectly reasonable name." But "This table shows the total score obtained from rolling two six-sided dice. The first row represents the value of the first die, the first column the value of the second die. The total is given in the cell that corresponds to the values of the two dice." has ZERO business being an accessible name IMHO.

When I did the `caption` role text, I didn't accidentally type `aria-describedby`. Instead, I deliberately refrained from typing `aria-labelledby`. And the reason why is stuff like the above.

Personally, I think the way to bring things into alignment is to change how HTML-AAM does stuff. For instance, in all of the examples, the caption/figcaption element is composed of two child elements. A nice short namey one followed by a much longer descriptiony one. Maybe HTML-AAM could somehow specify that when that's the case, use the first one as the name and the second one as the description?

joanmarie:
Thinking about this some more, a way which would address my concerns more fully is to include an actual example in the spec. What about this:

* Use the text of the `caption` element example in my previous comment (turning the elements into `div` instances)
* Put `aria-labelledby` for the first div child of the `role='caption'`
* Put `aria-describedby` for the second div child of the `role='caption'`
* Write some text reminding authors that names should be short.

Thoughts?

scottaohara:
I'd be happy to provide an example along the lines of what you're describing.  I think it's a good way to call out how this particular role could be used prior to the larger changes coming in 1.3 (or if these might alter course there as well).

Re: HTML AAM, I also think we have some thinking to do on that and we can move that part of the convo over there :)

scottaohara:
Thank you again for the review and feedback @mcking65.

I've largely incorporated your suggestions with some additional wording revisions to reduce the length of some of the paragraphs without losing their meaning.

jnurthen:
@mcking65 @joanmarie please re-review

PR #1090, opened 2019-10-10
clarify use of alertdialog over alert roles

scottaohara:
a notification message, which could be an alert, could very well be persistent until a user decides to dismiss it.

this text change is to help indicate that it’s OK for an alert to have a dismissal button to remove it from the DOM.  But to specifically indicate that if an alert obsures content / requires a user’s focus move to the alert to close it to unobscure the content, that an `alertdialog` should be considered instead.

additionally updates `alertdialog` definition to say:
>such as a form edit field or an OK button
to
>such as a form control or confirmation button

to indicate that any form control would be sufficient, and that confirming may extend beyond the bounds of “OK”.  :)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1090.html" title="Last updated on Oct 21, 2019, 12:28 AM UTC (c15d4fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1090/4228b98...c15d4fc.html" title="Last updated on Oct 21, 2019, 12:28 AM UTC (c15d4fc)">Diff</a>


index.html
```diff
@@ -761,15 +761,15 @@ <h3>Window Roles</h3>
...
+				<p>Alerts are assertive live regions and will be processed as such by assistive technologies. Neither authors nor user agents are required to set or manage focus to them in order for them to be processed. Since alerts are not required to receive focus, content authors SHOULD NOT require users to close an alert. If the operating system allows, the <a>user agent</a> SHOULD fire a system alert <a>event</a> through the accessibility API when the <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> alert is created. If an alert obscures content and requires a user's focus be moved to close the alert, then content authors SHOULD use <rref>alertdialog</rref> instead.</p>
```
jnurthen: Not sure I agree on this.... Can we say if the alert takes focus when it is created then .....
To me - that is when alertdialog should be used - it isn't about whether it obscures other content

scottaohara: yeah i think that makes sense, though I think the bit about closing it isn't necessary then.  Because if it's not obscuring content, then there's no reason to even mention that it needs to be closed.

scottaohara: @jnurthen updated

index.html
```diff
@@ -761,15 +761,15 @@ <h3>Window Roles</h3>
 	</section>
 	<section id="role_definitions">
 		<h2>Definition of Roles</h2>
-		<p>Below is an alphabetical list of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> <a>roles</a> to be used by rich internet application authors.</p>
+		<p>Below is an alphabetical list of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> <a>roles</a> to be used by rich Internet application authors.</p>
```
carmacleod: Probably don't need to capitalize internet here.

index.html
```diff
@@ -761,15 +761,15 @@ <h3>Window Roles</h3>
...
+				<p>Alerts are assertive live regions and will be processed as such by assistive technologies. Neither authors nor user agents are required to set or manage focus to them in order for them to be processed. Since alerts are not required to receive focus, content authors SHOULD NOT require users to close an alert. If the operating system allows, the <a>user agent</a> SHOULD fire a system alert <a>event</a> through the accessibility API when the <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> alert is created. If an alert requires a user's focus be moved to it once invoked, then content authors SHOULD use <rref>alertdialog</rref> instead.</p>
```
mcking65: The original paragraph was somewhat incoherent. It contained two different sets of ideas and topics. I think we should take advantage of this opportunity to improve the readability.

I also don't think the double speak in the original wording that said that if an alert requires focus then use alertdialog. If it requires focus, then in this context, it is not an alert.

```suggestion
				<p>Alerts are assertive live regions, which means they cause immediate notification for assistive technology users. If the operating system allows, the <a>user agent</a> SHOULD fire a system alert <a>event</a> through the accessibility API when the <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> alert is created.</p>
				<p>Neither authors nor user agents are required to set or manage focus to an alert in order for it to be processed. Since alerts are not required to receive focus, authors SHOULD NOT require users to close an alert. If an author desires focus to move to a message when it is conveyed, the author SHOULD use <rref>alertdialog</rref> instead of <code>alert</code>.</p>
```

index.html
```diff
@@ -761,15 +761,15 @@ <h3>Window Roles</h3>
...
 		<h2>Definition of Roles</h2>
-		<p>Below is an alphabetical list of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> <a>roles</a> to be used by rich internet application authors.</p>
+		<p>Below is an alphabetical list of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> <a>roles</a> to be used by authors.</p>
 		<p>Abstract roles are used for the ontology. Authors MUST NOT use abstract roles in content.</p>
 		<p id="index_role">Placeholder for compact list of roles</p>
 		<div class="role" id="alert">
 			<rdef>alert</rdef>
 			<div class="role-description">
 				<p>A type of <a>live region</a> with important, and usually time-sensitive, information. See related <rref>alertdialog</rref> and <rref>status</rref>.</p>
 				<p>Alerts are used to convey messages to alert the user. In the case of audio warnings this is an accessible alternative for a hearing-impaired user. The <code>alert</code> <a>role</a> goes on the node containing the alert message. Alerts are specialized forms of the <rref>status</rref> role, which will be processed as an atomic <a>live region</a>.</p>
```
mcking65: I've never liked the overuse of the word alert here; it is like using a word to define itself. I know this is not directly related to the purpose of this PR, but it is nonetheless editorial clarification, which is in the spirit of the PR.

In addition, there is a missing comma in the second sentence that I fixed in this suggestion.

```suggestion
				<p>Alerts are used to convey messages that may be immediately important to users. In the case of audio warnings, alerts provide an accessible alternative for hearing-impaired users. The <code>alert</code> <a>role</a> is applied to the element containing the alert message. An <code>alert</code> is a specialized form of the <rref>status</rref> role, which is processed as an atomic <a>live region</a>.</p>
```

scottaohara:
thanks @mcking65, pulled in your changes, I think those make this a better update.

PR #1097, opened 2019-10-15
brailleroledescription update

pkra:
Continues #924

Addressing review comments, part 1
* rephrase opening line - https://github.com/w3c/aria/pull/924#discussion_r334697274
* Braille/braille capitalization - https://github.com/w3c/aria/pull/924#discussion_r334698128
* suffice => overly verbose - https://github.com/w3c/aria/pull/924#discussion_r334698332
* two examples show => example shows - https://github.com/w3c/aria/pull/924#discussion_r334700411

***
<a href="https://pr-preview.s3.amazonaws.com/pkra/aria/pull/1097.html" title="Last updated on Feb 18, 2020, 4:00 PM UTC (c6db17f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1097/18522e3...pkra:c6db17f.html" title="Last updated on Feb 18, 2020, 4:00 PM UTC (c6db17f)">Diff</a>


index.html
```diff
@@ -12107,17 +12107,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p>Authors MUST NOT use <code>aria-brailleroledescription</code> without providing a <em>more specific</em> <code>aria-roledescription</code>. In general, <code>aria-brailleroledescription</code> should only be used in rare cases when a <code>aria-roledescription</code> does not suffice.</p>
+				<p>The <code>aria-brailleroledescription</code> property gives authors the ability to override how <a>assistive technologies</a> localize and express the name of a role in Braille. Thus inappropriately using <code>aria-brailleroledescription</code> may inhibit users' ability to understand or interact with an element on braille interfaces. Authors SHOULD limit use of <code>aria-brailleroledescription</code> to clarifying the purpose of non-interactive container roles like <rref>group</rref> or <rref>region</rref>, or to providing a <em>more specific</em> description of a <rref>widget</rref> in a braille context.</p>
```
mcking65: I think this may be easier to understand by cutting a few words.

```suggestion
				<p>The <code>aria-brailleroledescription</code> property gives authors the ability to override how <a>assistive technologies</a> localize and express the name of a role in Braille. Thus inappropriately using <code>aria-brailleroledescription</code> may inhibit understanding or interaction by people who rely on a braille interface. Authors SHOULD limit use of <code>aria-brailleroledescription</code> to clarifying the purpose of non-interactive container roles like <rref>group</rref> or <rref>region</rref>, or to providing a <em>more specific</em> description of a <rref>widget</rref>.</p>
```

index.html
```diff
@@ -12107,17 +12107,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Authors MUST NOT use <code>aria-brailleroledescription</code> without providing a <em>more specific</em> <code>aria-roledescription</code>. In general, <code>aria-brailleroledescription</code> should only be used in rare cases when a <code>aria-roledescription</code> is excessively verbose when rendered in Braille.</p>
```
mcking65: I don't think we know that the need is rare. In general, I would imagine that role descriptions may be longer than normal role names.

Also, "excessively verbose" is more appropriate for speech. In braille, it is not the number of sylables that matter, but the number of cells. The only way to know if a brailleroledescription is needed is to know if it would take more cells than the screen reader usually provides for displaying roles.

```suggestion
				<p>Authors MUST NOT use <code>aria-brailleroledescription</code> without providing a <em>more specific</em> <code>aria-roledescription</code>. In general, <code>aria-brailleroledescription</code> should only be used in cases where the braile rendering of a <code>aria-roledescription</code> would require significantly more cells than are usually allocated for displaying the role of an element.</p>
```

index.html
```diff
@@ -12107,17 +12107,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p> When using <code>aria-brailleroledescription</code>, authors SHOULD also ensure that:</p>
 				<ol>
 					<li>The element to which <code>aria-brailleroledescription</code> is applied has a valid <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> role or has an implicit WAI-ARIA role semantic.</li>
 					<li>The value of <code>aria-brailleroledescription</code> is not empty or does not contain only whitespace characters.</li>
 					<li>The value of <code>aria-brailleroledescription</code> consists of a string of either Unicode with no Unicode Braille Patterns (U+2800..U+28FF) or consists of a string of only Unicode Braille Patterns (U+2800..U+28FF) while not only containing Braille Pattern dots-0 (U+2800).</li>
```
mcking65: The word "either" is logically misplaced.

I don't understand the last phrase:
>while not only containing Braille Pattern dots-0 (U+2800)

I assume that is the code for an empty cell and this means that the string cannot be only empty cells.

If this is correct, I suggest the following:

```suggestion
					<li>The value of <code>aria-brailleroledescription</code> either consists of a string of Unicode that does not include any Unicode Braille Patterns (U+2800..U+28FF) or consists of a string including only Unicode Braille Patterns (U+2800..U+28FF) that would not be an empty string if all Braille Pattern dots-0 (U+2800) were removed.</li>
```

index.html
```diff
@@ -12107,17 +12107,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
...ill be able to customize such braille output according to user preferences. Using only <code>aria-roledescription</code> is <strong>almost always</strong> the better user experience and authors are <strong>strongly discouraged</strong> from using <code>aria-brailleroledescription</code> to replicate <code>aria-roledescription</code>. Instead, <code>aria-brailleroledescription</code> is meant be used only when <code>aria-roledescription</code> cannot provide an adequate braille representation, i.e., when a specialized braille description is very different from a text description converted to Braille. It is very important to note that when using <code>aria-brailleroledescription</code> authors are solely responsible to align the attribute value with the document language and clearly communicate the use of this attribute to the user. This is even more important when the value consists of Unicode Braille Patterns because <a>Assistive Technologies</a> will pass such content directly to the user without applying user specific braille translations; in general, authors are <strong>strongly discouraged</strong> from using Unicode Braille Patterns in <code>aria-brailleroledescription</code>.
```
mcking65: The following adds a missing comma, corrects some other grammar, and makes minor wording changes that hopefully improve clarity.

```suggestion
				<p class="note"><a>Assistive technologies</a> that support braille output can convert <code>aria-roledescription</code> values to Braille. In addition, assistive technologies may also apply user preferences, such as for a specific grade of braille,  to the conversion. Using only <code>aria-roledescription</code> will <strong>almost always</strong> provide the best user experience, and authors are <strong>strongly discouraged</strong> from using <code>aria-brailleroledescription</code> to replicate <code>aria-roledescription</code>. Instead, <code>aria-brailleroledescription</code> is meant to be used only when <code>aria-roledescription</code> cannot provide an adequate braille representation, i.e., when an ideal braille description would be very different from a text description converted to Braille. It is very important to note that when using <code>aria-brailleroledescription</code>, authors are solely responsible for aligning the attribute value with the document language and clearly communicating use of this attribute to the user. This is even more important when the value consists of Unicode Braille Patterns because <a>Assistive Technologies</a> will pass such content directly to the user without applying user specific braille translation preferences. In general, authors are <strong>strongly discouraged</strong> from using Unicode Braille Patterns in <code>aria-brailleroledescription</code>.
```

cookiecrook:
LGTM including Matt's editorial.

pkra:
Thanks so much @mcking65 and @cookiecrook! I'll merge those changes in before the next telco and also adopt @joanmarie's phrasing from https://github.com/w3c/aria/pull/924#discussion_r335572228.

index.html
pkra: I had missed this earlier. Since @mcking65 and @cookiecrook have indicated approval of this version, I will leave this as proposed here (i.e., I will **not** replace it with @joanmarie 's version as I wrote in an earlier comment of mine). I hope that's ok with @joanmarie as well.

index.html
```diff
@@ -12107,17 +12107,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 		<div class="property" id="aria-brailleroledescription">
```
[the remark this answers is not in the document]
pkra: Thanks! This should be fixed in the latest draft.

pkra:
I think this is ready.

As mentioned, I've left the author warning as edited by @mcking65. 

There's one more question leftover from https://github.com/w3c/aria/pull/924#discussion_r334699713 where @cookiecrook had asked:

> This line invalidates using img as a short braille description. Was that intentional? Same thing with "toc" if that one makes it from the DPUB module into the core spec.

and I had responded:

> I admit I struggle to answer that. The relevant commit is from June and the commit message explicitly mentions this change. I would have said I added it after feedback from the group, but I can't find any minutes around the time that match this idea.

Maybe we can decide that on the next call.

index.html
joanmarie: But I worked so hard on it. :stuck_out_tongue: 
If everyone else is happy with your revisions, then I'm happy too.

pkra:
@jnurthen @joanmarie can we add this to the next call's agenda? There's only one minor question left that I'd like to bring up with the group.

pkra:
@neil at https://github.com/w3c/aria/pull/924#issuecomment-561968850

> I commented on the use of the phrase "Unicode Braille Patterns" earlier and the distinction between the block name and an individual _character_ -- strings consist of _characters_ not patterns, symbols, or glyphs. I'm happy to see that in one instance that distinction is properly used. However, since the time of my comment, there have been several changes to text and I think improper usage has creeped in again.
> 
> Examples where I think changes are warranted so that the sentence is technically correct:
> 
>     1. The value of aria-brailleroledescription consists of a string of either Unicode with no Unicode Braille Patterns (U+2800..U+28FF) or consists of a string of only Unicode Braille Patterns (U+2800..U+28FF) while not only containing Braille Pattern dots-0 (U+2800).
> 
>     2. If the value of aria-brailleroledescription consists only of Unicode characters that are not Unicode Braille Patterns, translate the value according to the user's preferred translation table
> 
> 
> I suggest these become:
> 
>     1. The string value of aria-brailleroledescription does not contain any characters in Unicode Braille Patterns (U+2800..U+28FF) or consists of only characters in the range U+2801..U+28FF (Unicode Braille Pattern block excluding dots-0).
> 
> 
>     * Note: the previous statement mentioned it must contain "either Unicode or...". In HTML, all attributes value strings much contain Unicode values, so that statement is not a restriction. In fact, [not all Unicode values are legal,](https://www.w3.org/TR/2012/WD-html-markup-20120329/syntax.html#syntax-text) so it isn't even correct.
> 
> 
>     1. If the value of aria-brailleroledescription does not contain Unicode Braille Pattern characters, translate the value according to the user's preferred translation table.
> 
> 
>     * Note: The above is the first clause. I think the second clause would be better written as "Otherwise, pass the value to the user without translation." I don't think it is helpful to write a condition on the second clause as there is no alternative third clause (i.e., AT should do one of those two things).

pkra:
@NSoiffer thanks for those. I'll make an update to address this. 

The only detail I don't plan to address right now is the part in (1) where your version would forbid the use of Dot-0 entirely. That's only because I didn't see anything similar for the "regular" role description, i.e., it doesn't forbid whitespace entirely.

NSoiffer:
Oops, you are correct that I blew it for excluding Dot-0. I was too aggressive in shortening the language.  Hopefully you can keep the gist of the wording change that uses the word "character" where appropriate (i.e., when referring to values in a string).

pkra:
@NSoiffer thanks for confirming.

pkra:
I've pushed two commits

* addressing @NSoiffer's editorial comments 
* addd @mcking65's comment from last telco that it should not duplicate role or role-description
* removed problematic line identified by @cookiecrook 

For the last part, my reasoning was: 

a) we noticed it didn't match role-description (but I've filed #1131 as promised), 
b) we weren't really sure what it should have said (e.g., "a WAI-ARIA role" seems like it should have been "its WAI-ARIA role") and 
c) it would add burden on UAs 
d) we now have an authoring SHOULD NOT.

So I think, this is ready.

sinabahram:
I'm excited to get this finalized, but I do have the following questions and observations.

1. For the intro summary, we have "Defines a human-readable, author-localized abbreviated description for the role of an element, which is intended to be converted into Braille."

Should we say presented in Braille? Conversion seems loaded with questions around translation VS not, which are handled later on in the definition, so should we avoid ambiguity here?

2. We have "Authors MUST NOT use aria-brailleroledescription without providing a more specific aria-roledescription."

I understand the point, but can we clarify more specific than what? Right now, it kind of reads like the aria-roledescription must be more specific than aria-brailleroledescription, which is most definitely not the point of that statement.

Maybe we just want to drop the words "more specific" or at least "more".

3. This sentence really raised eyebrows for me. "In general, aria-brailleroledescription should only be used in rare cases when a aria-roledescription is excessively verbose when rendered in Braille."

Yes, I agree that is a majority use case, but should we really restrict all possible uses? I know it says "should" not "must", but still, that seems restrictive. Do we know for sure there are not cases, for example in educational contexts, where a particular Braille role is simply more appropriate independent of its length?

What about in a simple case where I wish to explore the most appropriate UI principles for two different user groups, speech users and Braille users. So, I construct an experiment to measure the efficacy of particular roles expressed in speech and others in Braille, and length is not the only variable I care to explore? For example, I could argue that speech generally should follow semantic prioritization, or putting important stuff up front, which is why most screen readers announce roles first in a virtual cursor context but not in a focus traversal context. But, do we actually have data on this for Braille? For all we know the geometric advantages of edge based detection would imply that a longer label, but with consistent spacing offers efficiencies in certain working contexts.

It just seems a bit too restrictive without offering evidence, but please don't misunderstand me. I know how valuable and prime a principle it is to be efficient on a Braille display, so I think it is a healthy default position, but I just wish to make sure we feel very certain that length should really surpass all else.

4. I humbly submit that the cognitive load to parse this sentence is exponentially higher than anything else in the description. "The value of aria-brailleroledescription does not contain any characters in Unicode Braille Patterns (U+2800..U+28FF) or consists of only characters in Unicode Braille Patterns (U+2800..U+28FF) while not only containing Braille Pattern dots-0 (U+2800)."

It is crystal clear, but I feel it requires thinking like an engineer, technical writer, or attorney. I feel like we can do a little bit better with those three requirements by explicitly enumerating them. In addition to enumerating them, perhaps we can offer an example to illustrate the two possibilities, one with unicode Braille and one without, along with it must not only consist of 2800. Even a four row good/bad table could go a long way?

5. Just a quick grammar thing here. "Instead, aria-brailleroledescription is meant be used only when aria-roledescription cannot provide an adequate braille representation, "

I believe we meant "meant to be"

6. In the code example, the image has an empty alt with aria-label labeling the planet. Is there a reason why this is done instead of the alt being set to Jupiter and reducing the use of aria attributes by 1? Does accessible name calc result in different results in those two cases? If not, then shouldn't the approach that always minimizes aria usage be preferred?

7. The code, when viewed as HTML, exposes escape characters, like so:


I didn't quote the above since it's code. Notice the backslash character seemingly escaping the equal sign.

8. We have the following sentence. "In the previous example, a braille display may display "Jupiter, pln" in Braille rather than the verbose "Jupiter, planet"."

Wouldn't a Braille display actually show those reversed e.g. "pln Jupiter"?

Thanks for entertaining this long comment.

pkra:
Thanks, @sinabahram. I don't have a strong opinion on any of these but some editorial comments below.

Regarding point 2., I probably copied the "more specific" from aria-roledescription (where it refers to being more specific than the role). I agree "more specific" can be confusing; "valid" is probably also sufficient (unless that's considered extraneous/implied in a spec context).

Points 5-8, I think, are simply mistakes that I'll be happy to fix.

pkra:
Note to self: address https://github.com/w3c/aria/pull/1142#issuecomment-574983257 (prohibit on role=generic)

pkra:
Rebased on w3c/aria master without issues (double checked).

pkra:
Just to recap today's call:

Re https://github.com/w3c/aria/pull/1097#issuecomment-574486030, I will fix point 2) as well since "more specific" was copies from roledescription.

Since points 1, 3, and 4 have been revised several times to find consensus, I'd suggest to leave them as is. I suspect we'll work out more details down the road.

sinabahram:
As promised on today's call, I reviewed my previous feedback. Yes I'm willing to live with 1, 3, and 4 not being addressed, so I'm going to wait for a commit addressing 2, and then hopefully that means 2, 5, 6, 7, and 8 are all addressed. At which point, I'm happy to review/approve this.

pkra:
@sinabahram thank you for the review! I'll work those in for next week.

pkra:
@sinabahram I've just pushed a fix for point 2), removing the "more specific" part as discussed. I hope this addresses everything. As with the braillelabel, I'd like to first merge before addressing the complex sentence about braille patterns.

sinabahram:
Does this fix also address points 5 through 8 that were mentioned as being able to be handled before? I still see empty alt text in the code snippet; for example, which makes me think I’m either not looking at the most up to date version or maybe this was not handled in the latest  update?

pkra:
Thanks. @sinabahram commit https://github.com/w3c/aria/pull/1097/commits/39af2b360f4f549166f9ca7189ddcb8ec191f7f7 adressed those points.

sinabahram:
Clearly, I’m doing something wrong, as when I go here:

https://raw.githack.com/w3c/aria/39af2b360f4f549166f9ca7189ddcb8ec191f7f7/index.html#aria-brailleroledescription

 

I see the same code snippet from before with that Jupiter button having the empty alt on the button, which is my way of checking if I’m looking at the right version or not.

 

I went to the PR. I selected to show all the changes, not just one from a single commit, and I then selected raw, and simply changed the domain to raw.githack.com to view it in the  browser. Not sure what I’m missing here ….

carmacleod:
@sinabahram We use pr-preview on this repo, so the easiest way to see a current preview is to look for a link called "Preview" and a link called "Diff" that are automatically generated (and kept up-to-date) by a friendly bot and inserted at the end of the initial comment. HTH!

pkra:
@carmacleod thanks for helping out! @sinabahram sorry, I misunderstood your earlier question. Let me know if there's anything else I can do to make review easier.

sinabahram:
That is super helpful RE diff and preview links, thank you.

 

Here is the link I am now looking at:

https://pr-preview.s3.amazonaws.com/pkra/aria/pull/1097.html#aria-brailleroledescription

 

here is the source code I see in that Jupiter example:

<button aria-roledescription="planet" aria-brailleroledescription="pln" id="jupiter" aria-label="jupiter">



</button>

 

As I now have mentioned several times, that still has ‘alt=””’. Am I just completely failing at GitHub at this point, or is that still there, even though it’s supposed to be filled in and aria-label removed as has been confirmed by @pkra patiently multiple times now?

 

I also note that I see more commits listed on the PR than I do in that comment, and furthermore, the last edit by PR bot was back in January 16, not when @pkra’s last comment was.

 

This has surpassed frustrating into silly, so I apologize if I’m failing at the most basic aspect of review *sheepish smile*.

 

I welcome any advice on what I’m doing wrong.

pkra:
@sinabahram thanks for taking the time to go through it again. I'm afraid I misread an earlier comment of yours.

Originally, your review had pointed out extra backslashes in the code sample, in particular the alt text. I had addressed this by removing the backslashes. Now I see that later on you pointed out empty alt-text which I completely misread, thinking of the backslash issue.

Perhaps I can pretend to have an excuse since the sample is based on [your playground repo](https://github.com/sinabahram/aria-playground/blob/71e7c3617ac15228e51fe3dbef82ec229f4c9b6c/CustomBrailleOutput.html#L79-L81). In any case, the button has an aria-label so it seems to me the img might have an empty alt. But I don't mind changing it.

Let me know what you'd prefer.

sinabahram:
Well, it seems like we should change it, as that’s the right thing to do, from what I know from graceful degradation or progressive enhancement, whichever way we wish to look at it. I think not relying on ARIA wherever possible is accepted as the defacto approach, but I’m happy to be educated on this.

 

And RE that code from the workshop, I think we actually had to do that to make up for an NVDA bug or something, can’t remember, but it would definitely be good not to propagate that workaround, and I may just be misremembering it.

 

Also, everything else looks fine, since I’ve now read it like 5 times, 😊.

pkra:
@sinabahram I've updated the code sample (and re-applied a commit that was on the wrong branch on my end).

pkra:
Oh, posting just as you approved. Thanks!

sinabahram:
I've approved the changes. We are good to go on this one.

pkra:
Just to link this here as well: https://github.com/w3c/aria/issues/1205 has been opened to improve the note.

PR #1100, opened 2019-10-17
Clarify wording in "Including Elements in the Accessibility Tree" section

### LINKED ISSUE #841: Tree Inclusion focus/activedescendant requirements need clarification

carmacleod:
Resolves #841

Starting fresh with only the relevant changes from PR https://github.com/w3c/aria/pull/950 to ensure that nothing bogus gets merged.

So... "take two": @cookiecrook @jnurthen @joanmarie @mcking65 

Please re-review these 2 sections. Preview links are:
[Excluding Elements from the Accessibility Tree](https://raw.githack.com/w3c/aria/issue841/index.html#tree_exclusion)
[Including Elements in the Accessibility Tree](https://raw.githack.com/w3c/aria/issue841/index.html#tree_inclusion)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1100.html" title="Last updated on Apr 22, 2020, 2:51 AM UTC (a1beec0)">Preview</a> (<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1100.html#tree_exclusion" title="#tree_exclusion">#tree_exclusion</a>) (<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1100.html#tree_inclusion" title="#tree_inclusion">#tree_inclusion</a>) | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1100/09d27e5...a1beec0.html" title="Last updated on Apr 22, 2020, 2:51 AM UTC (a1beec0)">Diff</a>


index.html
```diff
@@ -12763,14 +12763,19 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
...
-            <li>Elements that are focusable, or have an ID <a>attribute</a> and an ancestor with the <pref>aria-activedescendant</pref> attribute that matches the implicit or explicit <a>semantics</a> of the required context role. In either case, the element may receive focus and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event.</li>
+            <li>Elements that are not <a>hidden</a> and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>:
+              <ul>
+                <li>Elements that are focusable may receive focus and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event.</li>
+                <li>Elements that have an ID <a>attribute</a>, and are <a>owned</a> by, controlled by, or have an ancestor with the <pref>aria-activedescendant</pref> attribute, and that match the implicit or explicit <a>semantics</a> of the required context role, may receive focus and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event.</li>
```
mcking65: This text doesn't quite capture the requirements. The element with the ID is not the element that is controlled; its container is controlled, and that is only in the case where the container is not focusable, i.e., a textbox or combobox is focusable and controls the container.
And, the role of the element with the ID must match a required context role of the container.
Additionally, the container must support aria-active descendant.

Per above, we should also remove the reference to the focus event.

I suggest the following.

```suggestion
                <li>Elements that have an ID <a>attribute</a>, and are contained in or <a>owned</a> by an element that supports the <pref>aria-activedescendant</pref> property, and have an implicit or explicit role that matches the required context role of that containing element, and that containing element is either focusable and has the <pref>aria-activedescendant</pref> property or is controled bby an element that is focusable and has the <pref>aria-activedescendant<pref> property.</li>
```

index.html
```diff
@@ -12763,14 +12763,19 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
...
+                <li>Any element that receives focus and fires an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event, even if the element or one of its ancestor elements has its <sref>aria-hidden</sref> attribute set to <code>true</code>.</li>
```
mcking65: If you accept the suggestion for the first item in this list, this item can be removed.

index.html
carmacleod: @mcking65 
Wow, that sentence has a lot of words. :)
Would it be any clearer to use specific roles and split it into 2 separate points as follows?

- Elements that have an ID attribute, and are contained in or owned by a focusable composite, group or application that has the aria-activedescendant property, and have an implicit or explicit role that matches the required context role of their containing element.

- Elements that have an ID attribute, and are contained in or owned by a listbox, tree, grid or dialog that is controlled by a focusable textbox or combobox that has the aria-activedescendant property, and have an implicit or explicit role that matches the required context role of their containing element.

---

Alternatively, if a single, more generic, sentence is preferred, then perhaps:

- Elements that have an ID attribute, and are contained in or owned by an element that supports and has the aria-activedescendant property, and have an implicit or explicit role that matches the required context role of their containing element, and the containing element is focusable or is controlled by an element that is focusable.

jnurthen:
@mcking65 please re-review

index.html
carmacleod: @mcking65 For now, I have committed the "single sentence" option (2nd alternative, above).
Please let me know if you prefer any of the other alternatives above.

carmacleod:
@jnurthen @joanmarie @mcking65 Please review this. Other recent issues have been quoting the including/excluding sections that this PR covers, and I'm worried that pieces of the old sections may be reworded, and then we will have complicated conflicting words and thoughts to merge. It would be way easier to merge this first, before people start layering on new content.

carmacleod:
@jnurthen @joanmarie @mcking65 - ping.

Note that this PR will now close https://github.com/w3c/aria/pull/1172 as well (from the wide review).

JAWS-test:
I'm not sure, if the following at role=presentation is clear enough:

> In addition, the element's descendants and text content are generally included

Because not all descendants are included, but only those that are not required for the current role with role=presentation

carmacleod:
@JAWS-test 
> I'm not sure, if the following at role=presentation is clear enough:
> > In addition, the element's descendants and text content are generally included

> Because not all descendants are included, but only those that are not required for the current role with role=presentation

The information is there... but unfortunately finding it requires reading the whole [presentation role](https://w3c.github.io/aria/#presentation) section, and not just the conflict resolution subsection.

Specifically, these 3 paragraphs, when taken together, explain fairly clearly:

> The presentation role is used on an element that has implicit native semantics, meaning that there is a default accessibility API role for the element. Some elements are only complete when additional descendant elements are provided. For example, in HTML, table elements (matching the table role) require tr descendants (the row role), which in turn require th or td children (the cell, columnheader, rowheader roles). Similarly, lists require list item children. The descendant elements that complete the semantics of an element are described in WAI-ARIA as required owned elements.

> When an explicit or inherited role of presentation is applied to an element with the implicit semantic of a WAI-ARIA role that has required owned elements, in addition to the element with the explicit role of presentation, the user agent MUST apply an inherited role of presentation to any owned elements that do not have an explicit role defined. Also, when an explicit or inherited role of presentation is applied to a host language element which has required children as defined by the host language specification, in addition to the element with the explicit role of presentation, the user agent MUST apply an inherited role of presentation to any required children that do not have an explicit role defined.

> For any element with a role of presentation and which is not focusable, the user agent MUST NOT expose the implicit native semantics of the element (the role and its states and properties) to accessibility APIs. However, the user agent MUST expose content and descendant elements that do not have an explicit or inherited role of presentation. Thus, the presentation role causes a given element to be treated as having no role or to be removed from the accessibility tree, but does not cause the content contained within the element to be removed from the accessibility tree.

So I will keep the sentence you pointed out:
> In addition, the element's descendants and text content are generally included. 

but I will change the sentence right after that to point to the presentation role section instead of the conflict resolution subsection:
> These exceptions and conditions are documented in the [presentation (role)](https://w3c.github.io/aria/#presentation) section.

carmacleod:
@jnurthen Build is passing, there are no conflicts, and even though pr-preview is failing, the 2 links in the opening comment still show an up-to-date preview:
[Excluding Elements from the Accessibility Tree](https://raw.githack.com/w3c/aria/issue841/index.html#tree_exclusion)
[Including Elements in the Accessibility Tree](https://raw.githack.com/w3c/aria/issue841/index.html#tree_inclusion)

So, this PR is ready for final review.

carmacleod:
@cookiecrook @jnurthen @joanmarie @mcking65 Please review.

The preview links are good. Note that both the Excluding elements and the Including elements sections have changes. The diff is broken - it's only showing deletions for some reason - so you'll have to look at the code to see what changed. 😄 

Please note: this PR and it's (closed) parent PR have been waiting for almost a year (April 11, 2019). All previous review comments have been addressed, but there were enough changes that this should be reviewed again by all concerned parties. :)

Also please note that this contains changes that address wide review comments for 1.2.

It would be great if this could go in really soon, because I've started forgetting what's in here and accidentally making [some of the same changes](https://github.com/w3c/aria/pull/1218) again. 😅

jnurthen:
@cookiecrook @mcking65 @joanmarie please re-review. I need to merge this.

index.html
```diff
@@ -13189,22 +13189,22 @@ <h1>Accessibility Tree</h1>
...
 	  <ul>
-	    <li>Elements with <rref>none</rref> or <rref>presentation</rref> as the first role in the role attribute. However, their exclusion is conditional and depends on other factors. In addition, the element's descendants and text content are generally included. These exceptions and conditions are documented in the <a href="#conflict_resolution_presentation_none">Presentational Roles Conflict Resolution</a> section.</li>
 	    <li>Elements, including their descendent elements, that have host language semantics specifying that the element is not displayed, such as CSS <code>display:none</code>, <code>visibility:hidden</code>, or the HTML <code>hidden</code> attribute.</li>
-          </ul>
+	    <li>Elements with <rref>none</rref> or <rref>presentation</rref> as the first role in the role attribute. However, their exclusion is conditional and depends on other factors. In addition, the element's descendants and text content are generally included. These exceptions and conditions are documented in the <a href="#presentation">presentation (role)</a> section.</li>
```
cookiecrook, line 13193: Editorial nit: 
> their exclusion is conditional and depends on other factors. In addition, the element's descendants…

could be:
> their exclusion is conditional. The element's descendants…

index.html
```diff
@@ -13219,16 +13219,20 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
...
           <ul>
-            <li>Text nodes</li>
-            <li>Elements that may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>.</li>
-            <li>Elements that are focusable, or have an ID <a>attribute</a> and an ancestor with the <pref>aria-activedescendant</pref> attribute that matches the implicit or explicit <a>semantics</a> of the required context role. In either case, the element may receive focus and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event.</li>
+            <li>Elements that are not <a>hidden</a> and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>:
+              <ul>
+                <li>Elements that are focusable even if the element or one of its ancestor elements has its <sref>aria-hidden</sref> attribute set to <code>true</code>.</li>
```
cookiecrook, line 13226: Wow. I filed almost the same issue in #1234. There I suggested clarifying the difference between "user focusable" (e.g. tab focusable) and elements that "become focused" by other means like scripted focus.

index.html
```diff
@@ -13219,16 +13219,20 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
...
+                <li>Elements that have an ID <a>attribute</a>, and are contained in or <a>owned</a> by an element that supports and has the <pref>aria-activedescendant</pref> property, and have an implicit or explicit role that matches the required context role of their containing element, and their containing element is either focusable or is controlled by an element that is focusable.</li>
```
cookiecrook, line 13227: This is much more clear.

index.html
```diff
@@ -13219,16 +13219,20 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
...
+              </ul>
+	        </li>
             <li>Elements that have an explicit role or a global <abbr title="Accessible Rich Internet Application">WAI-ARIA</abbr> attribute and do not have <sref>aria-hidden</sref> set to <code>true</code>. (See <a href="#tree_exclusion">Excluding Elements in the Accessibility Tree</a> for additional guidance on <sref>aria-hidden</sref>.)</li>
             <li>Elements that have an ID which is referenced by another element via a <abbr title="Accessible Rich Internet Application">WAI-ARIA</abbr> property and are not <a>hidden</a>.
             	<p class="note">Text equivalents for hidden referenced objects may still be used in the <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd">name and description calculation</a> [[ACCNAME-1.1]] even when not included in the accessibility tree.</p>
             </li>
           </ul>
+          <p>User agents MAY provide an <a>accessible object</a> in the <a>accessibility tree</a> for <a>Text node</a>s.</p>
```
cookiecrook, line 13235: Is this necessary? Seems like it could be omitted.

index.html
cookiecrook: Above comment is referring to the last line: 13227. I think 13226 needs more work related to the comments in #1234

mcking65, line 13227: >Elements that have an ID attribute, and are contained in or owned by an element that supports and has the aria-activedescendant  property,

An option in a listbox controlled by a combobox does not meet this condition. The aria-activedescendant property is on the combobox, not on the owning listbox.

Looking at how we have worded the requirements for processing conflict resolution, I don't think we need all the details here. Instead, I suggest the following:

```suggestion
                <li>Elements that have an ID <a>attribute</a> and that have a role and context of a valid target of the <pref>aria-activedescendant</pref> property.</li>
```

index.html
carmacleod: Agree. Done.

index.html
carmacleod: @cookiecrook 
The history of that line is in https://github.com/w3c/aria/pull/950#discussion_r286189188.
Summary:
- The [spec currently says that text nodes MUST be in the a11y tree](https://w3c.github.io/aria/#tree_inclusion).
- @joanmarie said that on some platforms, the text is moved up into the parent node, and the text node itself is not included in the tree, so need to remove the MUST.
- I changed it to MAY.

I'm not sure whether the line should be omitted, given that it used to say MUST... perhaps we should leave it in with MAY so that implementers know what happened to it? @joanmarie, do you have any comment about this?

index.html
carmacleod: @cookiecrook You said in https://github.com/w3c/aria/pull/1100#discussion_r405919477 that this sentence was "much more clear". Matt has suggested that we reduce the sentence to its essential point, to "avoid duplicating the aria-activedescendant specification". Personally, I find the pared-down sentence easier to understand, but it would be great if you could take a quick look before I go ahead with the change.

I would reword it a tiny bit as follows:

> Elements that have an ID attribute, and have the role and context of a valid target of the aria-activedescendant property.

index.html
carmacleod: Need to define the term "focusable". Needs to include "user focusable" and "script focusable".

index.html
joanmarie: As discussed during the ARIA call, I'm good with removing it.

index.html
carmacleod: @mcking65 Regarding "a role" vs. "the role", I think it's ok as "the role" because it has the role and context of "a" valid target, i.e. any of the possible valid targets.

index.html
carmacleod: Added definition of focusable over in https://github.com/w3c/aria-common/pull/39.

carmacleod:
@cookiecrook @joanmarie @mcking65 All comments addressed. Ready for re-re-review. :)

Note that I had to add the definition of focusable over in https://github.com/w3c/aria-common/pull/39. Here's the prose:

> An element is focusable if it can be given the input focus by either the user (for example, using mouse or keyboard) or by script.

I was afraid to open cans of worms if I mentioned tabindex or activedescendant. Keep it simple...

Let me know if any further changes are needed.

index.html
```diff
@@ -13219,11 +13219,14 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
...
           <ul>
-            <li>Text nodes</li>
-            <li>Elements that may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>.</li>
-            <li>Elements that are focusable, or have an ID <a>attribute</a> and an ancestor with the <pref>aria-activedescendant</pref> attribute that matches the implicit or explicit <a>semantics</a> of the required context role. In either case, the element may receive focus and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> focus event.</li>
+            <li>Elements that are not <a>hidden</a> and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>:
+              <ul>
+                <li>Elements that are <a>focusable</a> even if the element or one of its ancestor elements has its <sref>aria-hidden</sref> attribute set to <code>true</code>.</li>
```
cookiecrook, line 13226: @aleventhal Are you good with this? Does it need to be more specific about whether or not it is exposed to the tree at the time that the element focused?

index.html
```diff
@@ -13219,11 +13219,14 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
...
+                <li>Elements that have an ID <a>attribute</a>, and have the role and context of a valid target of the <pref>aria-activedescendant</pref> property.</li>
```
cookiecrook, line 13227: CC @alice. 

@carmacleod I think this one as written requires all potential activedescendants to be exposed to the accessibility tree at all times. Perhaps:

> Elements that are a valid target of <pref>aria-activedescendant</pref>.

Note:  I specifically did not call refer to attributes (`id` or `aria-*`) because element reflection—when complete—would likely make the attribute-focused prose obsolete.

carmacleod: @cookiecrook I love the brevity of this, and agree that we've already defined "a valid target of aria-activedescendant" in other places in the spec. (e.g. [Managing Focus](https://w3c.github.io/aria/#managingfocus), [aria-activedescendant](https://w3c.github.io/aria/#aria-activedescendant), etc).

Just to double-check, you're saying that potential activedescendants don't all have to be exposed to the accessibility tree at all times... so will user agents always be able to tell when to fire appropriate events?

Good point about `id`, but - yikes! I don't want to look at how many other places in the spec will need rewriting when element reflection is done... I believe we use `id` or `IDREF` all over the place (including in supporting docs/specs like APG, ACCNAME, etc).

Cross that bridge when we come to it? (Don't get me wrong - I'm looking forward to element reflection, just not looking forward to rewriting the specs... 😄 )

cookiecrook: > you're saying that potential activedescendants don't all have to be exposed to the accessibility tree at all times

As an example, most of the activedescendants in this example are hidden at most times. Even when the field is focused. For example, type "te" and all US states but Tennessee and Texas are hidden from the tree.

https://webkit.org/blog-files/aria1.0/combobox_with_live_region_status.html

carmacleod: > As an example, most of the activedescendants in this example are hidden at most times.

I don't see `aria-activedescendant` in that example because it's using "owned descendants", but it's similar enough that I believe you. 😄 

I'm going with your suggestion:
> Elements that are a valid target of aria-activedescendant.

cookiecrook: > I don't see aria-activedescendant in that example because it's using "owned descendants", but it's similar enough that I believe you. 😄

Ack! Sorry for pointing you to the wrong example, but I'm thankful the point came across anyway.

carmacleod:
Just waiting on 2 things:
1. need PR https://github.com/w3c/aria-common/pull/39 (definition of focusable) reviewed and merged first: 
  > Focusable
  > An element is focusable if it can be given the input focus by either the user (for example, using mouse or keyboard) or by script.
2. waiting on @aleventhal to answer @cookiecrook's question in https://github.com/w3c/aria/pull/1100#discussion_r407750951 about this point in the [tree inclusion section](https://raw.githack.com/w3c/aria/issue841/index.html#tree_inclusion):
  > Elements that are focusable even if the element or one of its ancestor elements has its aria-hidden attribute set to true.

cookiecrook:
@carmacleod wrote:
> 1. waiting on @aleventhal to answer @cookiecrook's question in [#1100 (comment)](https://github.com/w3c/aria/pull/1100#discussion_r407750951) about this point in the [tree inclusion section](https://raw.githack.com/w3c/aria/issue841/index.html#tree_inclusion):

We can address that in a follow-up issue. I don't think it needs to delay this one... Marking PR as Reviewed.

carmacleod:
Only waiting on 1 thing:

@jamesn @cookiecrook Please review PR w3c/aria-common#39 - you may not be receiving notifications from that repo. All that PR does is add a definition for focusable. For convenience, here are the words:

> Focusable
> An element is focusable if it can be given the input focus by either the user (for example, using mouse or keyboard) or by script.

@jnurthen Please merge this (#1100) as soon as possible. Today would be good. People are still trying to make conflicting changes in these 2 sections because this PR (and its parent) have been sitting here for so long. I don't think you need to wait for @joanmarie and @mcking65 to re-re-review. They have seen it many times.

PR #1127, opened 2019-12-05
Clean up use of "may"

zcorpan:
"May" is an RFC 2119 keyword. Change instances that formally allow something to uppercase MAY, and rephrase others to "can" or "might".

Also tweak surrounding wording in some places. "MAY want to do X" allows the conformance class to want something, which is a weird conformance requirement.

***

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Faria%2F0cabe3d0940344869c54a849b4ca0c2b6aaae245%2Findex.html%3FisPreview%3Dtrue)

```
Navigation timeout of 28987 ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231127.)._
</details>


index.html
```diff
@@ -343,7 +343,7 @@ <h3>Information for Authors</h3>
 	    <h3>Information for User Agents</h3>
 	    <p>The user agent MUST do the following to implement <pref>aria-activedescendant</pref>:</p>
 	    <ol>
-	      <li>Implement the host language method for keyboard navigation so that widgets that support <code>aria-activedescendant</code> may be included in the tab order.</li>
+	      <li>Implement the host language method for keyboard navigation so that widgets that support <code>aria-activedescendant</code> MAY be included in the tab order.</li>
```
carmacleod: I think this one should be "can"?

carmacleod:
I think (?) that in sections marked with "This section is non-normative." it is ok to use "may"?
For example, all of section 1 is non-normative. We should probably discuss this on the call.

zcorpan:
> I think (?) that in sections marked with "This section is non-normative." it is ok to use "may"/"should"?

I disagree. It is confusing for readers to know what is normative, what is intended to be normative, what is intended to non-normative, when using RFC 2119 keywords in non-normative sections.

[Bikeshed](https://github.com/tabatkins/bikeshed) has [a lint check for only using the keywords in normative sections](https://github.com/tabatkins/bikeshed/issues/958) for this reason. ARIA doesn't use bikeshed, but still, the rationale applies the same.

index.html
zcorpan:
```suggestion
	      <li>Implement the host language method for keyboard navigation so that widgets that support <code>aria-activedescendant</code> can be included in the tab order.</li>
```

index.html
```diff
@@ -12894,8 +12894,8 @@ <h2>Conflicts with Host Language Semantics</h2>
...
+		<p>Certain ARIA features are critical to building a complete model in the accessibility API. Such features are not expected to conflict with native host language semantics (though they might complement them). Therefore, host languages MUST NOT declare strong native semantics that prevent use of the following ARIA features:</p>
```
scottaohara: i'd say "can" over "might" here

The original 'may' here was more of "is allowed to complement" (can) rather than a "could complement" as "might" implies.

jnurthen:
I can certainly understand bikeshed doing this as RFC2119 words do not appear in uppercase in bikeshed leading to ambiguities. However, in respec documents they do. I'm not convinced that replacing all instances of may with other words is worthwhile especially as RFC2119 allows these usages.

index.html
```diff
@@ -3219,7 +3219,7 @@ <h2>Definition of Roles</h2>
 			<rdef>form</rdef>
 			<div class="role-description">
 				<p>A <rref>landmark</rref> region that contains a collection of items and objects that, as a whole, combine to create a form. See related <rref>search</rref>.</p>
-				<p>A form may contain a mix of host language form controls, scripted controls, and hyperlinks. Authors are reminded to use native host language semantics to create form controls whenever possible. If the purpose of a form is to submit search criteria, authors SHOULD use the <rref>search</rref> role instead of the generic <code>form</code> role.</p>
+				<p>A form might contain a mix of host language form controls, scripted controls, and hyperlinks. Authors are reminded to use native host language semantics to create form controls whenever possible. If the purpose of a form is to submit search criteria, authors SHOULD use the <rref>search</rref> role instead of the generic <code>form</code> role.</p>
```
carmacleod: I think I would use "can" here (instead of might)

index.html
```diff
@@ -4670,7 +4670,7 @@ <h2>Definition of Roles</h2>
 		<div class="role" id="log">
 			<rdef>log</rdef>
 			<div class="role-description">
-				<p>A type of <a>live region</a> where new information is added in meaningful order and old information may disappear. See related <rref>marquee</rref>.</p>
+				<p>A type of <a>live region</a> where new information is added in meaningful order and old information might disappear. See related <rref>marquee</rref>.</p>
```
carmacleod: Consider using "can" here.

index.html
scottaohara: fwiw, either seems appropriate here.

carmacleod:
Interesting. Apparently [RFC 2119](https://tools.ietf.org/html/rfc2119) has an update called [RFC 8174](https://tools.ietf.org/html/rfc8174) that says:
> RFC 2119 specifies common key words that may be used in protocol specifications.  This document [RFC 8174] aims to reduce the ambiguity by clarifying that only UPPERCASE usage of the key words have the defined special meanings.

Anyhow, I agree that it's possible that authors coming from the HTML spec to the ARIA spec could find lowercase "may" confusing. So, as long as no sentences have changed meaning or sound awkward, and as long as no uppercase MAY is used in non-normative sections, I would be ok with this going in (after review comments are addressed).

Also, I noticed that _in normative sections_, some of these changes really do remove any potential ambiguity without changing meaning. There are also some really useful changes, like removing "choose to" from "An author MAY choose to", and changing "MUST never" to "MUST NOT". (I need to finish reviewing - I've only read through about half).

@jnurthen Are you ok with this reasoning so far?

index.html
```diff
@@ -235,22 +235,22 @@ <h2>User Agent Support</h2>
...
+			<p>Many of the requirements in the definitions of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> <a>roles</a>, <a>states</a>, and <a>properties</a> can be checked automatically during the development process, similar to other quality control processes used for validating code. To assist authors who are creating custom widgets, authoring tools MAY compare widget roles, states, and properties to those supported in <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> as well as those supported in related and cross-referenced roles, states, and properties. Authoring tools MAY notify authors of errors in widget design patterns, and MAY also prompt developers for information that cannot be determined from context alone. For example, a scripting library can determine the labels for the tree items in a tree view, but would need to prompt the author to label the entire tree. To help authors visualize a logical accessibility structure, an authoring environment might provide an outline view of a web resource based on the <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> markup.</p>
```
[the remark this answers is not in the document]
zcorpan:
```suggestion
			<p>Many of the requirements in the definitions of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> <a>roles</a>, <a>states</a>, and <a>properties</a> can be checked automatically during the development process, similar to other quality control processes used for validating code. To assist authors who are creating custom widgets, authoring tools can compare widget roles, states, and properties to those supported in <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> as well as those supported in related and cross-referenced roles, states, and properties. Authoring tools MAY notify authors of errors in widget design patterns, and can also prompt developers for information that cannot be determined from context alone. For example, a scripting library can determine the labels for the tree items in a tree view, but would need to prompt the author to label the entire tree. To help authors visualize a logical accessibility structure, an authoring environment might provide an outline view of a web resource based on the <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> markup.</p>
```

index.html
```diff
@@ -235,22 +235,22 @@ <h2>User Agent Support</h2>
...
+			<p>In both HTML and SVG, <code>tabindex</code> is an important way browsers support keyboard <a href="#host_general_focus">focus navigation</a> for implementations of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr>; authoring and debugging tools MAY check to make sure <code>tabindex</code> values are properly set. For example, error conditions can include cases where more than one treeitem in a tree has a <code>tabindex</code> value greater than or equal to 0, where <code>tabindex</code> is not set on any treeitem, or where <pref>aria-activedescendant</pref> is not defined when the element with the role tree has a <code>tabindex</code> value of greater than or equal to 0.</p>
```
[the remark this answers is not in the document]
zcorpan:
```suggestion
			<p>In both HTML and SVG, <code>tabindex</code> is an important way browsers support keyboard <a href="#host_general_focus">focus navigation</a> for implementations of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr>; authoring and debugging tools can check to make sure <code>tabindex</code> values are properly set. For example, error conditions can include cases where more than one treeitem in a tree has a <code>tabindex</code> value greater than or equal to 0, where <code>tabindex</code> is not set on any treeitem, or where <pref>aria-activedescendant</pref> is not defined when the element with the role tree has a <code>tabindex</code> value of greater than or equal to 0.</p>
```

index.html
```diff
@@ -290,7 +290,7 @@ <h2><abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> Roles</h
...
+		<p><abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> provides a collection of accessibility <a>states</a> and <a>properties</a> which are used to support platform <a>accessibility <abbr title="Application Programing Interfaces">APIs</abbr></a> on various operating system platforms. <a>Assistive technologies</a> MAY access this information through an exposed <a>user agent</a> DOM or through a mapping to the platform accessibility <abbr title="Application Programing Interfaces">API</abbr>. When combined with <a>roles</a>, the user agent can supply the assistive technologies with user interface information to convey to the user at any time. Changes in states or properties will result in a notification to assistive technologies, which could alert the user that a change has occurred.</p>
```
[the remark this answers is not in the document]
zcorpan:
```suggestion
		<p><abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> provides a collection of accessibility <a>states</a> and <a>properties</a> which are used to support platform <a>accessibility <abbr title="Application Programming Interfaces">APIs</abbr></a> on various operating system platforms. <a>Assistive technologies</a> can access this information through an exposed <a>user agent</a> DOM or through a mapping to the platform accessibility <abbr title="Application Programming Interfaces">API</abbr>. When combined with <a>roles</a>, the user agent can supply the assistive technologies with user interface information to convey to the user at any time. Changes in states or properties will result in a notification to assistive technologies, which could alert the user that a change has occurred.</p>
```

Also fixed typo "Programing" with one m.

index.html
zcorpan:
```suggestion
		<p>Certain ARIA features are critical to building a complete model in the accessibility API. Such features are not expected to conflict with native host language semantics (though they can complement them). Therefore, host languages MUST NOT declare strong native semantics that prevent use of the following ARIA features:</p>
```

index.html
zcorpan:
```suggestion
				<p>A form can contain a mix of host language form controls, scripted controls, and hyperlinks. Authors are reminded to use native host language semantics to create form controls whenever possible. If the purpose of a form is to submit search criteria, authors SHOULD use the <rref>search</rref> role instead of the generic <code>form</code> role.</p>
```

index.html
zcorpan:
```suggestion
				<p>A type of <a>live region</a> where new information is added in meaningful order and old information can disappear. See related <rref>marquee</rref>.</p>
```

zcorpan:
@jnurthen do you think readers will know which spec generator was used for each spec, and base their interpretation thereafter?

In my opinion, having a strict hygiene with the normative keywords makes the spec clearer for readers, regardless of the decision of using uppercase or not. This PR is a start; it would also be useful to do the same with the other keywords, and have a linter to catch any use of the keywords where they shouldn't (to minimize confusion) be used.

index.html
```diff
@@ -249,8 +249,8 @@ <h2>Co-Evolution of <abbr title="Accessible Rich Internet Applications">WAI-ARIA
...
+			<p>Many of the requirements in the definitions of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> <a>roles</a>, <a>states</a>, and <a>properties</a> can be checked automatically during the development process, similar to other quality control processes used for validating code. To assist authors who are creating custom widgets, authoring tools can compare widget roles, states, and properties to those supported in <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> as well as those supported in related and cross-referenced roles, states, and properties. Authoring tools MAY notify authors of errors in widget design patterns, and can also prompt developers for information that cannot be determined from context alone. For example, a scripting library can determine the labels for the tree items in a tree view, but would need to prompt the author to label the entire tree. To help authors visualize a logical accessibility structure, an authoring environment might provide an outline view of a web resource based on the <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> markup.</p>
```
scottaohara: still one more "MAY" here.
>Authoring tools MAY notify authors of errors

jnurthen:
> @jnurthen do you think readers will know which spec generator was used for each spec, and base their interpretation thereafter?

I'd love bikeshed based specs to change to use RFC2119 as it is written to eliminate confusion. Recognizing that this is not going to happen we already ensure we don't use must or should in a non-normative way. 

I see diminishing returns in changing other RFC2119 terms when used in a non-normative context. If someone wants to go through the process fine - but I'm not going to be merging this in the 1.2 timeframe. 

If we are going to do this we might as well be complete and look for instances of recommended, optional and required (note - required is going to be really hard to do as we have a required property - no idea how we are going to deal with that one)

tobie:
The WHATWG infra spec clarifies this in its [conformance section](https://infra.spec.whatwg.org/#conformance), fwiw (emphasis mine):

> All assertions, diagrams, examples, and notes are non-normative, as are all sections explicitly marked non-normative. Everything else is normative.
>
> The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" are to be interpreted as described in RFC 2119. [RFC2119]
>
> These keywords have _**equivalent meaning when written in lowercase**_ and _**cannot appear in non-normative content.**_
>
> _**This is a willful violation of RFC 8174**_, motivated by legibility and a desire to preserve long-standing practice in many non-IETF-published pre-RFC 8174 documents. [RFC8174]
>
> All of the above is applicable to both this standard and any document that uses this standard. Documents using this standard are encouraged to limit themselves to "must", "must not", "should", and "may", and to use these in their lowercase form as that is generally considered to be more readable.
>
> _**For non-normative content "strongly encouraged", "strongly discouraged", "encouraged", "discouraged", "can", "cannot", "could", "could not", "might", and "might not" can be used instead.**_

zcorpan:
> I'd love bikeshed based specs to change to use RFC2119 as it is written to eliminate confusion.

The confusion that this PR tries to address is use of normative keywords in non-normative contexts -- something that we know from experience is confusing, which is why the Infra Standard now prohibits that.

Even if all specs were changed to use uppercase (which would regress readability), I believe the above is still true.

> If we are going to do this we might as well be complete and look for instances of recommended, optional and required

Sure.

> (note - required is going to be really hard to do as we have a required property - no idea how we are going to deal with that one)

`aria-required` (marked up as code) is clearly different from required. This is not an issue.

index.html
```diff
@@ -235,22 +235,22 @@ <h2>User Agent Support</h2>
...
```
scottaohara: cannot have a MAY here.  Section 1 is non-normative.

zcorpan:
```suggestion
			<p>Many of the requirements in the definitions of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> <a>roles</a>, <a>states</a>, and <a>properties</a> can be checked automatically during the development process, similar to other quality control processes used for validating code. To assist authors who are creating custom widgets, authoring tools can compare widget roles, states, and properties to those supported in <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> as well as those supported in related and cross-referenced roles, states, and properties. Authoring tools can notify authors of errors in widget design patterns, and can also prompt developers for information that cannot be determined from context alone. For example, a scripting library can determine the labels for the tree items in a tree view, but would need to prompt the author to label the entire tree. To help authors visualize a logical accessibility structure, an authoring environment might provide an outline view of a web resource based on the <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> markup.</p>
```

zcorpan:
Thanks, @scottaohara ! Fixed, and I've also rebased this on latest master.

The diff is easier to review (for me, at least) using [`--word-diff=color`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---word-diffltmodegt)

spectranaut:
We decided in a meeting we should definitely do this change because it is difficult to distinguish normative/capitalized words from non-normative words when you are using a screen reader. Unfortunately our process for keeping track of the editor's draft and the next ARIA version will make merging this difficult, so I will pick PR up after we defined a new process at TPAC in September.

It was decided to make sure no normative words (not just may) will be used in a non-normative context.

spectranaut:
thanks for this PR @zcorpan. I opened a new one with your fixes, rebased, and fixes for all the addition non-normative "may"s added since you opened this: https://github.com/w3c/aria/pull/1901

Issue #508, opened 2017-01-05
Figure out <mark> element


michael-n-cooper:
[HTML 5 mark](https://www.w3.org/TR/html51/textlevel-semantics.html#the-mark-element) needs handling. Could relate to drag and drop. This could have multiple ARIA features.

MichielBijl:
For those interested: `<mark>` has default styling in Chrome, Firefox, Safari, and Edge. It applies a yellow background to the element (regardless of text colour I might add).

aleventhal:
https://github.com/w3c/aria/pull/1133
Examples and more context in ARIA annotations explainer.

pkra:
Closing this as resolved via #1133. I've opened https://github.com/w3c/html-aam/issues/316 which seems like a detail that's missing. Please re-open if anything else is missing.

PR #1133, opened 2019-12-11
Add mark role

aleventhal:
Closes issue #508

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1133.html" title="Last updated on Jan 10, 2020, 8:21 PM UTC (27eb18a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1133/f519fca...27eb18a.html" title="Last updated on Jan 10, 2020, 8:21 PM UTC (27eb18a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4836,8 +4836,13 @@
 		<div class="role" id="mark">
 			<rdef>mark</rdef>
 			<div class="role-description">
-				<p>The mark element represents a run of text in one document marked or highlighted for reference purposes, due to its relevance in another context.</p>
-				<p>When used in a quotation or other block of text referred to from the prose, it indicates a highlight that was not originally present but which has been added to bring the reader's attention to a part of the text that might not have been considered important by the original author when the block was originally written, but which is now under previously unexpected scrutiny. When used in the main prose of a document, it indicates a part of the document that has been highlighted due to its likely relevance to the user's current activity.</p>
+				<p>Content which is marked or highlighted for reference or notation purposes, due to the content's relevance in the enclosing context.</p>
+				<p>Example uses for <rref>mark</rref> include:</p>
+				<ul>
+					<li>Highlighting text in a quotation which is of special interest but is not marked in the original source material, comparable using a highlighter pen to mark passages of a print article.</li>
+					<li>Indicating portions of the content that are relevant to the user's current activity, such as highlighting text matches found by an page-implemented search feature.</li>
+				</ul>
+				<p>Don't use <rref>mark</rref> for purely decorative styling such as syntax highlighting.</p>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -4832,6 +4832,88 @@
 				</tbody>
 			</table>
 		</div>
+    <div class="role" id="mark">
+        <rdef>mark</rdef>
+        <div class="role-description">
+            <p>A mark contains a run of content, marked or highlighted to indicate that other content refers to it, or to indicate special relevance to the user's current activity.</p>
+        </div>
+        <table class="role-features">
+            <caption>Characteristics:</caption>
+            <thead>
+                <tr>
+                    <th scope="col">Characteristic</th>
+                    <th scope="col">Value</th>
+                </tr>
+            </thead>
+            <tbody>
+                <tr>
+                    <th class="role-abstract-head" scope="row">Is Abstract:</th>
+                    <td class="role-abstract"> </td>
+                </tr>
+                <tr>
+                    <th class="role-parent-head" scope="row">Superclass Role:</th>
+                    <td class="role-parent"><rref>section</rref></td>
+                </tr>
+                <tr>
+                    <th class="role-base-head" scope="row">Base Concept:</th>
+                    <td class="role-base"> </td>
+                </tr>
+                <tr>
+                    <th class="role-related-head" scope="row">Related Concepts:</th>
+                    <td class="role-related"><code>&lt;mark&gt;</code> in [[HTML]]</td>
+                </tr>
+                <tr>
+                    <th class="role-scope-head" scope="row">Required Context Role:</th>
+                    <td class="role-scope"> </td>
+                </tr>
+                <tr>
+                    <th class="role-mustcontain-head" scope="row">Required Owned Elements:</th>
+                    <td class="role-mustcontain"> </td>
+                </tr>
+                <tr>
+                    <th class="role-required-properties-head">Required States and Properties:</th>
+                    <td class="role-required-properties"> </td>
+                </tr>
+                <tr>
+                    <th class="role-properties-head" scope="row">Supported States and Properties:</th>
+                    <td class="role-properties"> </td>
+                </tr>
+                <tr>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4832,6 +4832,88 @@ <h2>Definition of Roles</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="mark">
+			<rdef>mark</rdef>
+			<div class="role-description">
+				<p>A mark contains a run of content, marked or highlighted to indicate that other content refers to it, or to indicate special relevance to the user's current activity.</p>
```
joanmarie: I'm ok with this. However, with most of the other role-parity* roles, we've incorporated the language from the HTML spec for the corresponding element. Do you think it would be worthwhile to do so here? https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-mark-element

*I realize your needs are not role-parity related. But given that there is a `mark`, the new `mark` role will also be addressing role parity.

aleventhal:
@joanmarie @jnurthen I've updated the CL to match the HTML spec text and add mark to the role categorization section.

index.html
joanmarie: This works for me. Thanks!

We should raise this at the upcoming ARIA meeting to see if everyone agrees this can land in master.

index.html
```diff
@@ -4832,6 +4833,90 @@ <h2>Definition of Roles</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="mark">
+			<rdef>mark</rdef>
+			<div class="role-description">
+				<p>
+					The mark element represents a run of text in one document marked or highlighted for reference purposes, due to its relevance in another context. When used in a quotation or other block of text referred to from the prose, it indicates a highlight that was not originally present but which has been added to bring the reader's attention to a part of the text that might not have been considered important by the original author when the block was originally written, but which is now under previously unexpected scrutiny. When used in the main prose of a document, it indicates a part of the document that has been highlighted due to its likely relevance to the user's current activity.
```
joanmarie: Oops, one more nit/problem: If you look at https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1133.html#role_definitions, you'll notice that there is no short entry. We usually have the first paragraph of any role be a sentence or two max and then elaborate in subsequent paragraphs.

To solve this problem, you could potentially just close the first paragraph after "context."

Sorry for not catching that immediately!!

carmacleod:
The HTML text for mark element is pretty obscure without examples. (i.e. why "one document"? And the "previously unexpected scrutiny" sentence is really long...):

> The mark element represents a run of text in one document marked or highlighted for reference purposes, due to its relevance in another context.
>
> When used in a quotation or other block of text referred to from the prose, it indicates a highlight that was not originally present but which has been added to bring the reader's attention to a part of the text that might not have been considered important by the original author when the block was originally written, but which is now under previously unexpected scrutiny. When used in the main prose of a document, it indicates a part of the document that has been highlighted due to its likely relevance to the user's current activity.

I kinda preferred the original description:
> A mark contains a run of content, marked or highlighted to indicate that other content refers to it, or to indicate special relevance to the user's current activity.

Or the "similar to HTML but more clearly written" [MDN text for mark element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark):
> The HTML `<mark>` element represents text which is marked or highlighted for reference or notation purposes, due to the marked passage's relevance or importance in the enclosing context.
>
> Typical use cases for `<mark>` include:
>
> - When used in a quotation (`<q>`) or block quote (`<blockquote>`), it generally indicates text which is of special interest but is not marked in the original source material, or material which needs special scrutiny even though the original author didn't think it was of particular importance. Think of this like using a highlighter pen in a book to mark passages that you find of interest.
> - Otherwise, `<mark>` indicates a portion of the document's content which is likely to be relevant to the user's current activity. This might be used, for example, to indicate the words that matched a search operation.

If we keep the HTML description, it really needs to be clarified with an example or two.

carmacleod:
> Not gonna lie, I thought my text was shorter and more to the point.

Yep. :)

> But it's probably clearer overall to consistently copy over the exact HTML spec text

Not sure... and what if it changes? Maintenance hassle...

> I was caught off guard by the stuff about in a quotation vs prose.

I was caught off guard by "a run of text in one document", which would naturally be followed by
"... in another document", but instead is followed by "... in another context".
I would change "document" to "context".

I also liked the MDN idea of breaking up that dense HTML paragraph into a list of "Typical use cases".

And I found the analogy, "Think of this like using a highlighter pen in a book to mark passages that you find of interest.", quite helpful.

joanmarie:
I personally like the MDN text. What do others think?

aleventhal:
@joanmarie @jnurthen @carmacleod How about this? It is quite similar to the MDN text.

index.html
```diff
@@ -4832,6 +4833,94 @@ <h2>Definition of Roles</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="mark">
+			<rdef>mark</rdef>
+			<div class="role-description">
+				<p>Content which is marked or highlighted for reference or notation purposes, due to the content's relevance in the enclosing context.</p>
+				<p>Example uses for <rref>mark</rref> include:</p>
+				<ul>
+					<li>Highlighting text in a quotation which is of special interest but is not marked in the original source material, comparable using a highlighter pen to mark passages of a print article.</li>
```
```suggestion
					<li>Highlighting text in a quotation which is of special interest but is not marked in the original source material, comparable to using a highlighter pen to mark passages of a print article.</li>
```

index.html
```diff
@@ -4832,6 +4833,94 @@ <h2>Definition of Roles</h2>
...
+					<li>Indicating portions of the content that are relevant to the user's current activity, such as highlighting text matches found by an page-implemented search feature.</li>
```
```suggestion
					<li>Indicating portions of the content that are relevant to the user's current activity, such as highlighting text matches found by a search feature.</li>
```

carmacleod:
Nice! I like it. Only a couple of nits, mentioned above.

aleventhal:
@carmacleod Thanks, good catches. What do you think now?

index.html
```diff
@@ -4832,6 +4833,94 @@ <h2>Definition of Roles</h2>
...
+		<div class="role" id="mark">
+			<rdef>mark</rdef>
+			<div class="role-description">
+				<p>Content which is marked or highlighted for reference or notation purposes, due to the content's relevance in the enclosing context.</p>
+				<p>Example uses for <rref>mark</rref> include:</p>
+				<ul>
+					<li>Highlighting text in a quotation which is of special interest but is not marked in the original source material, comparable to using a highlighter pen to mark passages of a print article.</li>
+					<li>Indicating portions of the content that are relevant to the user's current activity, such as highlighting text matches found by a search feature.</li>
+				</ul>
+				<p>Don't use <rref>mark</rref> for purely decorative styling such as syntax highlighting.</p>
```
joanmarie: What abaout "Authors SHOULD  not use mark...."?

Nit: In addition, please use `<code>` tags instead of `<rref>` when referring to the thing being defined in the section. The reason why is the first instance of an `<rref>` in any section will be a link to the associated content. I personally don't think there's any reason why we want a link in the `mark` section to the `mark` section.

aleventhal: Good suggestions. Done.

Issue #749, opened 2018-05-09
Annotations


jnurthen:
from https://www.w3.org/WAI/ARIA/track/issues/655

> Proposal to support commenting, inline error highlighting, etc. in documents.
> 
> Office wants to be able to make these types of annotations more accessible. Commenting on a document is a pretty common scenario, and this would be useful to many web app developers.
> 
> We'd like to explore creating this as an extension spec. I have a person from the Office team who is willing to do the heavy lifting.
> http://www.w3.org/2014/04/annotation/submissions/Microsoft_Position_Paper_on_Annotations.pdf

jnurthen:
Also https://www.w3.org/WAI/ARIA/track/issues/670

> See Response from Doug Schepers to me reqarding the formation of an Annotation Working Group pertaining to remote annoation requirements:
> http://lists.w3.org/Archives/Public/public-digipub-ig/2014Jul/0030.html

jnurthen:
https://www.w3.org/WAI/ARIA/track/actions/1454

> dd aria-hasannotation attribute with values: comment, footnote, endnote, reference, insertion, deletion, modification, true, false, more where aria-hasannotation has only one value at a time

jnurthen:
https://www.w3.org/WAI/ARIA/track/actions/1455

> Add aria-annotatedby where it takes idrefs as a value and must only be applied when that same element has aria-hasannotation

jnurthen:
https://www.w3.org/WAI/ARIA/track/actions/1456

aleventhal:
I've put an updated draft proposal on the Wiki:
https://github.com/w3c/aria/wiki/ARIA-annotations-draft-proposal
This draft incorporates comments from the F2F in Lyon.

azaroth42:
The J Paul Getty Trust would formally object to this unless it is at least aligned with existing W3C standards for Annotation.

joanmarie:
Hi @azaroth42. Thanks for your review and feedback. Could you please let us know what specifically you find is not in alignment and/or what you find objectionable? The ARIA Working Group would find this information helpful before proceeding. Thanks!!

BigBlueHat:
[...]

PR #1134, opened 2019-12-11
Add suggestion role

aleventhal:
Relates to ARIA Annotations, issue #749

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1134.html#suggestion" title="Last updated on Feb 13, 2020, 9:27 PM UTC (6e36557)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1134/f519fca...6e36557.html" title="Last updated on Feb 13, 2020, 9:27 PM UTC (6e36557)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -8220,11 +8220,13 @@
 		<div class="role" id="suggestion">
 			<rdef>suggestion</rdef>
 			<div class="role-description">
-					<p>A suggestion contains a single proposed change to content that can be approved or rejected.
-					For example, in an editing system that supports multiple authors, one author may suggest a change, and another author would be responsible
+					<p>A single proposed change to content that can be approved or rejected.</p>
+					<p>For example, in an editing system that supports multiple authors, one author may suggest a change, and another author would be responsible
 					for accepting or rejecting the suggestion.
-					The suggested change must be present as child content, as an insertion, a deletion, or one of each.</p>
-					<p>When a suggestion is accepted, the suggestion role should be removed.</p>
+					The suggested change must be present as child content, as an <rref>insertion</rref>, a <rref>deletion</rref>, or one of each.</p>
+					<p>When a suggestion is accepted, the <code>suggestion</code> role should no longer be used. This can be done by removing or replacing the role, or
+						removing the element. In addition, child <rref>insertion</rref> and <rref>deletion</rref> elements can be retained or transformed into the final accepted
+					  content.</p>
 			</div>
 			<table class="role-features">
 					<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -8224,6 +8224,15 @@
 					<p>For example, in an editing system that supports multiple authors, one author may suggest a change, and another author would be responsible
 					for accepting or rejecting the suggestion.
 					The suggested change must be present as child content, as an <rref>insertion</rref>, a <rref>deletion</rref>, or one of each.</p>
+					<pre class="example highlight">
+						&lt;p&gt;
+						  The best pet is a
+						  &lt;span role="suggestion"&gt;
+						    &lt;span role="deletion"&gt;cat&lt;/span&gt;
+						    &lt;span role="insertion"&gt;dog&lt;/span&gt;
+						  &lt;/span&gt;
+						&lt;/p&gt;
+					</pre>
 					<p>When a suggestion is accepted, the <code>suggestion</code> role should no longer be used. This can be done by removing or replacing the role, or
 						removing the element. In addition, child <rref>insertion</rref> and <rref>deletion</rref> elements can be retained or transformed into the final accepted
 					  content.</p>
--- a/index.html
+++ b/index.html
@@ -8235,7 +8235,7 @@
 						&lt;/p&gt;
 					</pre>
 					<p>When a suggestion is accepted, authors SHOULD remove the <code>suggestion</code> role, indicating that the proposed revision has been made.
-						In addition, child <rref>insertion</rref> and <rref>deletion</rref> elements can be retained or transformed into the final accepted content.
+						After the <code>suggestion</code> role is removed, child <rref>insertion</rref> and <rref>deletion</rref> elements can either be retained to document the revision or replaced with the revised content.
 					  </p>
 			</div>
 			<table class="role-features">
--- a/index.html
+++ b/index.html
@@ -8223,7 +8223,7 @@
 					<p>A single proposed change to content that can be approved or rejected.</p>
 					<p>For example, in an editing system that supports multiple authors, one author may suggest a change, and another author would be responsible
 					for accepting or rejecting the suggestion.
-					The suggested change MUST be present as child content, as an <rref>insertion</rref>, a <rref>deletion</rref>, or one of each. A suggestion MUST NOT contain
+					Authors MUST ensure that a <code>suggestion</code> contains either one <rref>insertion</rref> child or one <rref>deletion</rref> child or ensure that it contains two children where one is an <rref>insertion</rref> and the other is a <rref>deletion</rref>.
 				  any other children. Additional information such as related comments, authoring info and time stamps MAY be provided via <pref>aria-description</pref> or <pref>aria-details</pref>.</p>
 					<pre class="example highlight">
 						&lt;p&gt;
--- a/index.html
+++ b/index.html
@@ -8222,10 +8222,9 @@
 			<div class="role-description">
 					<p>A single proposed change to content.</p>
 					<p>For example, in an editing system that supports multiple users, one user may suggest a change, and user would be responsible
-					for accepting or rejecting the suggestion.
-					Authors MUST ensure that a <code>suggestion</code> contains either one <rref>insertion</rref> child or one <rref>deletion</rref> child or ensure that it contains two children where one is an <rref>insertion</rref> and the other is a <rref>deletion</rref>.
-					Authors MUST ensure a <code>suggestion</code> does not contain any other children.
-					Authors MAY use <pref>aria-details</pref> or <pref>aria-description</pref> to associate the <code>suggestion</code> with information,such as related comments, authoring info, and time stamps.</p>
+					for accepting or rejecting the suggestion.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -8216,6 +8216,95 @@ <h5>Presentational Roles Conflict Resolution</h5>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="suggestion">
+			<rdef>suggestion</rdef>
+			<div class="role-description">
+					<p>A suggestion contains a proposed change to content that can be approved or rejected.
```
joanmarie: In the explainer, it states:

> role="suggestion" -- used to group proposed changes in a document (role "deletion" and "insertion" children).

So I was envisioning (potentially) multiple "deletion" and/or "insertion" children. But from the wording above, my impression is that a given suggestion is limited to a single change. Which is it? And can we make the spec language more explicit so that there is no question?

index.html
```diff
@@ -8216,6 +8216,95 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
+					For example, in an editing system that supports multiple authors, one author may suggest a change, and another author would be responsible
+					for accepting or rejecting the suggestion.
+					The suggested change must be present as child content, as an insertion, a deletion, or one of each.</p>
+					<p>When a suggestion is accepted, the suggestion role should be removed.</p>
```
joanmarie: Does this mean the *role value* of `suggestion` goes away, but the element and its `insertion` and/or `deletion` child(ren) remain? In which case, what is the new role for the element formerly roled `suggestion`?

OR does it mean that the *element* which has the `suggestion` role goes away? If so, I assume the removal includes all the `insertion` and/or `deletion` children.

I'm guessing the text means the latter, but it would be good to clarify things. Also mind putting `code` tags around "suggestion" when "suggestion" refers to the role?

index.html
aleventhal: It's for a single change, so the possibilities are:
- 1 deletion child
- 1 insertion child
- both a deletion and insertion child

aleventhal: I changed the first sentence of the definition to use "single proposed change"

index.html
aleventhal: I've reworded it, PTAL.

I think it's mostly important that "suggestion" goes away -- whether or not the element is kept, a new role is used, or change markup is kept is up to the author.

joanmarie:
Looks good to me overall, but as I'm sitting here implementing support for this role in Orca, I have a question: Your text says:

> The suggested change must be present as child content, as an `insertion`, a `deletion`, or one of each.

Can there be other children, like a paragraph explaining why the proposed insertion and/or deletion should be made? Or a time stamp? Or am I guaranteed to _only_ find `insertion` and/or `deletion` children?

Arguably Orca should handle both scenarios because authors. But it would be nice to know (and perhaps see text in the spec stating) what the *right* usage looks like. Also, if we do want to prevent authors from putting non-`insertion`/`deletion` children in a `suggestion`, adding this statement to the spec should then cause it to be checked by validators.

aleventhal:
@joanmarie  PTAL, I updated the text to make this more clear.

> Looks good to me overall, but as I'm sitting here implementing support for this role in Orca, I have a question: Your text says:
> 
> > The suggested change must be present as child content, as an `insertion`, a `deletion`, or one of each.
> 
> Can there be other children, like a paragraph explaining why the proposed insertion and/or deletion should be made? Or a time stamp? Or am I guaranteed to _only_ find `insertion` and/or `deletion` children?
> 
> Arguably Orca should handle both scenarios because authors. But it would be nice to know (and perhaps see text in the spec stating) what the _right_ usage looks like. Also, if we do want to prevent authors from putting non-`insertion`/`deletion` children in a `suggestion`, adding this statement to the spec should then cause it to be checked by validators.

index.html
```diff
@@ -8216,6 +8217,107 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
 			</table>
 		</div>
+		<div class="role" id="suggestion">
+			<rdef>suggestion</rdef>
+			<div class="role-description">
+					<p>A single proposed change to content that can be approved or rejected.</p>
+					<p>For example, in an editing system that supports multiple authors, one author may suggest a change, and another author would be responsible
+					for accepting or rejecting the suggestion.
+					The suggested change MUST be present as child content, as an <rref>insertion</rref>, a <rref>deletion</rref>, or one of each. A suggestion MUST NOT contain
+				  any other children. Additional information such as related comments, authoring info and time stamps MAY be provided via aria-description or aria-details.</p>
```
```suggestion
				  any other children. Additional information such as related comments, authoring info and time stamps MAY be provided via <pref>aria-description</pref> or <pref>aria-details</pref>.</p>
```

carmacleod: Can't decide if we would need to have ` on the element with role <rref>suggestion</rref>` at the end of the sentence with aria-details, or if it's obvious from context and that would just be too many words.

index.html
```diff
@@ -8216,6 +8217,107 @@ <h5>Presentational Roles Conflict Resolution</h5>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="suggestion">
+			<rdef>suggestion</rdef>
+			<div class="role-description">
+					<p>A single proposed change to content that can be approved or rejected.</p>
+					<p>For example, in an editing system that supports multiple authors, one author may suggest a change, and another author would be responsible
+					for accepting or rejecting the suggestion.
+					The suggested change MUST be present as child content, as an <rref>insertion</rref>, a <rref>deletion</rref>, or one of each. A suggestion MUST NOT contain
```
mcking65: In most role definitions, we present these types of structural requirements as normative author MUST statements. The spec is not consistent in this regard; for example, compare option to listbox or listitem to list. I think making them normative author MUST statements is the better editorial practice. It makes the spec more clear.

```suggestion
					Authors MUST ensure that a <code>suggestion</code> contains either one <rref>insertion</rref> child or one <rref>deletion</rref> child or ensure that it contains two children where one is an <rref>insertion</rref> and the other is a <rref>deletion</rref>.
```

index.html
```diff
@@ -8216,6 +8217,107 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
+				  any other children. Additional information such as related comments, authoring info and time stamps MAY be provided via <pref>aria-description</pref> or <pref>aria-details</pref>.</p>
```
```suggestion
					Authors MUST ensure a <code>suggestion</code> does not contain any other children.
					Authors MAY use <pref>aria-details</pref> or <pref>aria-description</pref> to associate the <code>suggestion</code> with information,such as related comments, authoring info, and time stamps.</p>
```

index.html
```diff
@@ -8216,6 +8217,107 @@ <h5>Presentational Roles Conflict Resolution</h5>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="suggestion">
+			<rdef>suggestion</rdef>
+			<div class="role-description">
+					<p>A single proposed change to content that can be approved or rejected.</p>
```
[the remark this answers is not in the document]
aleventhal: How about, "that can be approved or rejected by a document author". The reason I like it is that it just makes it that much clearer what we're talking about.

index.html
```diff
@@ -8216,6 +8217,107 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
+					<p>For example, in an editing system that supports multiple authors, one author may suggest a change, and another author would be responsible
+					for accepting or rejecting the suggestion.
+					The suggested change MUST be present as child content, as an <rref>insertion</rref>, a <rref>deletion</rref>, or one of each. A suggestion MUST NOT contain
+				  any other children. Additional information such as related comments, authoring info and time stamps MAY be provided via <pref>aria-description</pref> or <pref>aria-details</pref>.</p>
+					<pre class="example highlight">
+						&lt;p&gt;
+						  The best pet is a
+						  &lt;span role="suggestion"&gt;
+						    &lt;span role="deletion"&gt;cat&lt;/span&gt;
+						    &lt;span role="insertion"&gt;dog&lt;/span&gt;
```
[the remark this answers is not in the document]
aleventhal: Haha. The author can still reject the suggestion :)

index.html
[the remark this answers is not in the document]
aleventhal: Nevermind, I see the next paragraph does what we want. Approving this suggestion :)

aleventhal:
@mcking65 I think I got everything, but I needed to do some polish after as there was an incomplete sentence snippet that remained. Can you take a look?

index.html
```diff
@@ -8216,6 +8217,107 @@ <h5>Presentational Roles Conflict Resolution</h5>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="suggestion">
+			<rdef>suggestion</rdef>
+			<div class="role-description">
+					<p>A single proposed change to content.</p>
+					<p>For example, in an editing system that supports multiple users, one user may suggest a change, and user would be responsible
```
```suggestion
					<p>For example, in an editing system that supports multiple users, one user may suggest a change, and another user would be responsible
```

index.html
```diff
@@ -8216,6 +8217,107 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
+						&lt;p&gt;
+						  The best pet is a
+						  &lt;span role="suggestion"&gt;
+						    &lt;span role="deletion"&gt;cat&lt;/span&gt;
+						    &lt;span role="insertion"&gt;dog&lt;/span&gt;
+						  &lt;/span&gt;
+						&lt;/p&gt;
+					</pre>
+					<p>When a suggestion is accepted, authors SHOULD remove the <code>suggestion</code> role, indicating that the proposed revision has been made.
+						In addition, child <rref>insertion</rref> and <rref>deletion</rref> elements can be retained or transformed into the final accepted content.
```
mcking65: The words "in addition" kind of imply this is an extension of the SHOULD statement. I don't think it is. I think it is just informative authoring guidance. I recommend the following to improve clarity:

```suggestion
						After the <code>suggestion</code> role is removed, child <rref>insertion</rref> and <rref>deletion</rref> elements can either be retained to document the revision or replaced with the revised content.
```

PR #1135, opened 2019-12-11
Add comment role

aleventhal:
Refers to issue #749

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1135.html#comment" title="Last updated on Feb 7, 2020, 8:07 PM UTC (75a6436)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1135/7193e40...75a6436.html" title="Last updated on Feb 7, 2020, 8:07 PM UTC (75a6436)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2384,7 +2384,7 @@
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
-						<td class="role-properties"> </td>
+						<td class="role-properties"><pref>aria-level</pref></td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
--- a/index.html
+++ b/index.html
@@ -2344,7 +2344,7 @@
 					<li>Otherwise, if the comment relates to other content in the page:
 						<ul>
 							<li>Provide <pref>aria-details</pref> on the element containing the commented content with a value refering to the element with role <code>comment</code>.</li>
-							<li>If there are many comments relating to the same commented content, either identify all of them individually via <pref>aria-details</pref>, or identify a parent container of the comments, and assign a role of <rref>group</rref> or <rref>region</rref> to the container.</li>
+							<li>If there are many comments relating to the same commented content, either identify all of them individually via <pref>aria-details</pref>, or identify a parent container of the comments. If <pref>aria-details</pref> is used to indentify to a parent container for the comments, authors SHOULD assign a role of <rref>group</rref> or <rref>region</rref> to the container.</li>
 						</ul>
 					</li>
 				</ol>
--- a/index.html
+++ b/index.html
@@ -2336,7 +2336,7 @@
 				<ol>
 					<li>If the comment is a reply to another <code>comment</code>:
 						<ul>
-							<li>If all ancestor comments are available in the DOM, make the reply <code>comment</code> a semantic descendant of the parent <code>comment</code> being replied to, either as a DOM descendant element, or via <pref>aria-owns</pref>.</li>
+							<li>If all ancestor comments are available in the DOM, make each reply <code>comment</code> a semantic descendant of the <code>comment</code> to which it is replying, either by making it a DOM descendant element or by using <pref>aria-owns</pref>.</li>
 							<li>Alternatively, if all ancestor comments are not in the DOM, such as when comments are paginated, the hierarchical
 								level MAY be indicated via <pref>aria-level</pref>. Additional group positional information MAY be indicated via <pref>aria-posinset</pref> and <pref>aria-setsize</pref>.</li>
 						</ul>
--- a/index.html
+++ b/index.html
@@ -2334,7 +2334,7 @@
 				<p>A comment contains a remark expressing a reaction to other content within the page.</p>
 				<p>The relationship of comments to other content SHOULD be exposed as follows:</p>
 				<ol>
-					<li>When the comment refers to other content within the page, the source content SHOULD use <pref>aria-details</pref> to point to the related comment.</li>
+					<li>If the comment refers to other content within the page, the source content SHOULD use <pref>aria-details</pref> to point to the related comment.</li>
 					<li>If the comment is a reply to another comment:
 						<ul>
 							<li>If all ancestors are available in the DOM, the reply comment SHOULD be a semantic descendant of the comment it relates to, either as a DOM descendant element, or via <pref>aria-owns</pref>. In this case, user agents MUST automatically compute the level and expose it to assistive technologies. This is the preferred method for defining
--- a/index.html
+++ b/index.html
@@ -2397,15 +2397,6 @@
 						<td class="role-inherited">Placeholder</td>
 					</tr>
 					<tr>
-						<th class="role-disallowed-head" scope="row">Prohibited States and Properties:</th>
-						<td class="role-disallowed">
-							<ul>
-								<li><pref>aria-label</pref></li>
-								<li><pref>aria-labelledby</pref></li>
-							</ul>
-						</td>
-					</tr>
-					<tr>
 						<th class="role-namefrom-head" scope="row">Name From:</th>
 						<td class="role-namefrom">
 							<ul>
--- a/index.html
+++ b/index.html
@@ -2344,7 +2344,7 @@
 					<li>Otherwise, if the comment relates to other content in the page:
 						<ul>
 							<li>Provide <pref>aria-details</pref> on the element containing the commented content with a value refering to the element with role <code>comment</code>.</li>
-							<li>If there are many comments relating to the same commented content, either identify all of them individually via <pref>aria-details</pref>, or identify a parent container of the comments. If <pref>aria-details</pref> is used to indentify to a parent container for the comments, authors SHOULD assign a role of <rref>group</rref> or <rref>region</rref> to the container.</li>
+							<li>If there are multiple comments related to the same commented content, either provide a value for <pref>aria-details</pref> on the commented content that refers to each individual comment, or use <pref>aria-details</pref> to refer to a parent container of the comments. If <pref>aria-details</pref> refers to an element containing comments rather than <code>comment</code> elements, authors SHOULD assign a role of <rref>group</rref> or <rref>region</rref> to the referenced container.</li>
 						</ul>
 					</li>
 				</ol>
--- a/index.html
+++ b/index.html
@@ -2331,7 +2331,7 @@
 		<div class="role" id="comment">
 			<rdef>comment</rdef>
 			<div class="role-description">
-				<p>A comment contains a remark expressing a reaction to other content within the page.</p>
+				<p>A comment contains content expressing reaction to other content within the page.</p>
 				<p>The relationship of comments to other content SHOULD be exposed as follows:</p>
 				<ol>
 					<li>If the comment refers to other content within the page, the source content SHOULD use <pref>aria-details</pref> to point to the related comment.</li>
--- a/index.html
+++ b/index.html
@@ -2332,7 +2332,7 @@
 			<rdef>comment</rdef>
 			<div class="role-description">
 				<p>A comment contains a remark expressing a reaction to other content within the page.</p>
-				<p>The relationship of comments to other content SHOULD be exposed:</p>
+				<p>The relationship of comments to other content SHOULD be exposed as follows:</p>
 				<ol>
 					<li>When the comment refers to other content within the page, the source content SHOULD use <pref>aria-details</pref> to point to the related comment.</li>
 					<li>When the comment is a reply to another comment:
--- a/index.html
+++ b/index.html
@@ -2335,7 +2335,7 @@
 				<p>The relationship of comments to other content SHOULD be exposed as follows:</p>
 				<ol>
 					<li>When the comment refers to other content within the page, the source content SHOULD use <pref>aria-details</pref> to point to the related comment.</li>
-					<li>When the comment is a reply to another comment:
+					<li>If the comment is a reply to another comment:
 						<ul>
 							<li>If all ancestors are available in the DOM, the reply comment SHOULD be a semantic descendant of the comment it relates to, either as a DOM descendant element, or via <pref>aria-owns</pref>. In this case, user agents MUST automatically compute the level and expose it to assistive technologies. This is the preferred method for defining
 								the comment hierarchy, as it creates a single root comment that can be referred to by commented content via <pref>aria-details</pref>.</li>
--- a/index.html
+++ b/index.html
@@ -2337,9 +2337,9 @@
 					<li>When the comment refers to other content within the page, the source content SHOULD use <pref>aria-details</pref> to point to the related comment.</li>
 					<li>When the comment is a reply to another comment:
 						<ul>
-							<li>If possible, the reply comment SHOULD be a semantic descendant of the comment it relates to, either as a DOM descendant element, or via <pref>aria-owns</pref>. In this case, user agents MUST automatically compute the level and expose it to assistive technologies. This is the preferred method for defining
+							<li>If all ancestors are available in the DOM, the reply comment SHOULD be a semantic descendant of the comment it relates to, either as a DOM descendant element, or via <pref>aria-owns</pref>. In this case, user agents MUST automatically compute the level and expose it to assistive technologies. This is the preferred method for defining
 								the comment hierarchy, as it creates a single root comment that can be referred to by commented content via <pref>aria-details</pref>.</li>
-							<li>Alternatively, when the parent comment is not rendered in the page, such as when comments are paginated, the hierarchical
+							<li>Alternatively, if all ancestor comments are not in the DOM, such as when comments are paginated, the hierarchical
 								level MAY be indicated via <pref>aria-level</pref>.</li>
 						</ul>
 					</li>
--- a/index.html
+++ b/index.html
@@ -2343,7 +2343,7 @@
 					</li>
 					<li>Otherwise, if the comment relates to other content in the page:
 						<ul>
-							<li>Provide <pref>aria-details</pref> on the commented content, identifying the commented element.</li>
+							<li>Provide <pref>aria-details</pref> on the element containing the commented content with a value refering to the element with role <code>comment</code>.</li>
 							<li>If there are many comments relating to the same commented content, either identify all of them individually via <pref>aria-details</pref>, or identify a parent container of the comments, and assign a role of <rref>complementary</rref> to the container.</li>
 						</ul>
 					</li>
--- a/index.html
+++ b/index.html
@@ -2344,7 +2344,7 @@
 					<li>Otherwise, if the comment relates to other content in the page:
 						<ul>
 							<li>Provide <pref>aria-details</pref> on the element containing the commented content with a value refering to the element with role <code>comment</code>.</li>
-							<li>If there are many comments relating to the same commented content, either identify all of them individually via <pref>aria-details</pref>, or identify a parent container of the comments, and assign a role of <rref>complementary</rref> to the container.</li>
+							<li>If there are many comments relating to the same commented content, either identify all of them individually via <pref>aria-details</pref>, or identify a parent container of the comments, and assign a role of <rref>group</rref> or <rref>region</rref> to the container.</li>
 						</ul>
 					</li>
 				</ol>
--- a/index.html
+++ b/index.html
@@ -640,6 +640,7 @@
 				<li><rref>caption</rref></li>
 				<li><rref>cell</rref></li>
 				<li><rref>columnheader</rref></li>
+				<li><rref>comment</rref></li>
 				<li><rref>definition</rref></li>
 				<li><rref>deletion</rref></li>
 				<li><rref>directory</rref></li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2326,6 +2327,101 @@ <h2>Definition of Roles</h2>
...
+						<td class="role-mustcontain">
+						</td>
+					</tr>
+					<tr>
+						<th class="role-required-properties-head">Required States and Properties:</th>
+						<td class="role-required-properties"> </td>
+					</tr>
+					<tr>
+						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
+						<td class="role-properties"> </td>
```
```suggestion
						<td class="role-properties"><pref>aria-level</pref></td>
```

index.html
```diff
@@ -2326,6 +2327,101 @@ <h2>Definition of Roles</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="comment">
+			<rdef>comment</rdef>
+			<div class="role-description">
+				<p>A comment contains a remark expressing a reaction to other content within the page.</p>
+				<p>The relationship of comments to other content SHOULD be exposed:</p>
```
```suggestion
				<p>The relationship of comments to other content SHOULD be exposed as follows:</p>
```

(Just sort of feels like the "as follows" is missing).

index.html
```diff
@@ -2326,6 +2327,101 @@ <h2>Definition of Roles</h2>
...
+				<ol>
+					<li>When the comment refers to other content within the page, the source content SHOULD use <pref>aria-details</pref> to point to the related comment.</li>
+					<li>When the comment is a reply to another comment:
+						<ul>
+							<li>If possible, the reply comment SHOULD be a semantic descendant of the comment it relates to, either as a DOM descendant element, or via <pref>aria-owns</pref>. In this case, user agents MUST automatically compute the level and expose it to assistive technologies. This is the preferred method for defining
+								the comment hierarchy, as it creates a single root comment that can be referred to by commented content via <pref>aria-details</pref>.</li>
+							<li>Alternatively, when the parent comment is not rendered in the page, such as when comments are paginated, the hierarchical
```
```suggestion
							<li>Alternatively, if the parent comment is not rendered in the page, such as when comments are paginated, the hierarchical
```

index.html
```diff
@@ -2326,6 +2327,101 @@ <h2>Definition of Roles</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="comment">
+			<rdef>comment</rdef>
+			<div class="role-description">
+				<p>A comment contains a remark expressing a reaction to other content within the page.</p>
+				<p>The relationship of comments to other content SHOULD be exposed:</p>
+				<ol>
+					<li>When the comment refers to other content within the page, the source content SHOULD use <pref>aria-details</pref> to point to the related comment.</li>
```
```suggestion
					<li>If the comment refers to other content within the page, the source content SHOULD use <pref>aria-details</pref> to point to the related comment.</li>
```

(Just a suggestion - I don't know why I prefer "if" over "when" here).

index.html
```diff
@@ -2326,6 +2327,101 @@ <h2>Definition of Roles</h2>
...
+					<li>When the comment is a reply to another comment:
```
```suggestion
					<li>If the comment is a reply to another comment:
```

(As above - just a suggestion).

index.html
```diff
@@ -2326,6 +2327,101 @@ <h2>Definition of Roles</h2>
...
+						<ul>
+							<li>If possible, the reply comment SHOULD be a semantic descendant of the comment it relates to, either as a DOM descendant element, or via <pref>aria-owns</pref>. In this case, user agents MUST automatically compute the level and expose it to assistive technologies. This is the preferred method for defining
```
```suggestion
							<li>The reply comment SHOULD be a semantic descendant of the comment it relates to, either as a DOM descendant element, or via <pref>aria-owns</pref>. In this case, user agents MUST automatically compute the level and expose it to assistive technologies. This is the preferred method for defining
```

(I think the "If possible" just waters down the SHOULD. After all, if it's not possible, then authors will have to ignore the SHOULD).

index.html
```diff
@@ -2326,6 +2327,101 @@ <h2>Definition of Roles</h2>
...
+						<td class="role-related"> </td>
+					</tr>
+					<tr>
+						<th class="role-scope-head" scope="row">Required Context Role:</th>
+						<td class="role-scope"> </td>
+					</tr>
+					<tr>
+						<th class="role-mustcontain-head" scope="row">Required Owned Elements:</th>
+						<td class="role-mustcontain">
+						</td>
```
carmacleod: The `td` needs to be all on one line in order to be ignored by the script, i.e.
`<td class="role-mustcontain"> </td>`

(This will hide the empty "Required Owned Elements:" characteristic row - which is currently visible).

index.html
aleventhal: I just think it's clearer in the original, because even though that's technically correct, it lets me know that there can be times where it's not possible. Specifically, in the paginated comment use case. 

It could be changed to, "If all ancestors are available in the DOM, the reply comment SHOULD ..."

index.html
aleventhal: Thanks!

index.html
carmacleod: I like your suggested text, with "If all ancestors are available in the DOM, ..."

aleventhal: I changed to "If all ancestor comments are available in the DOM"

carmacleod: Ah, already committed. :)
I'll mark this resolved.

carmacleod:
Need to add aria-setsize and aria-posinset as supported attributes.

index.html
```diff
@@ -2326,6 +2327,102 @@ <h2>Definition of Roles</h2>
...
+			<div class="role-description">
+				<p>A comment contains a remark expressing a reaction to other content within the page.</p>
+				<p>The relationship of comments to other content SHOULD be exposed as follows:</p>
+				<ol>
+					<li>If the comment refers to other content within the page, the source content SHOULD use <pref>aria-details</pref> to point to the related comment.</li>
+					<li>If the comment is a reply to another comment:
+						<ul>
+							<li>If all ancestor comments are available in the DOM, the reply comment SHOULD be a semantic descendant of the comment it relates to, either as a DOM descendant element, or via <pref>aria-owns</pref>. In this case, user agents MUST automatically compute the level and expose it to assistive technologies. This is the preferred method for defining
+								the comment hierarchy, as it creates a single root comment that can be referred to by commented content via <pref>aria-details</pref>.</li>
+							<li>Alternatively, if all ancestor comments are not in the DOM, such as when comments are paginated, the hierarchical
```
mcking65: This alternative is unclear because it implies that aria-owns or DOM hierarchy is not used to relate the comments and replies that are present in the DOM. So, if DOM hierarchy is not being used in this alternative, and if there is a reply to a reply present in the DOM, then how are those replies related to one another?

Separately, this should mention aria-posinset and aria-setsize. I've added that. This suggestion, however, does not address the structural ambiguity.

```suggestion
							<li>Alternatively, if all ancestor comments are not in the DOM, such as when comments are paginated, specify the hierarchical level with <pref>aria-level</pref>, the position of the comment with in that level with <pref>aria-posinset</pref>, and the number of comments at the same level with <pref>aria-setsize</pref>.</li>
```

index.html
```diff
@@ -2326,6 +2327,102 @@ <h2>Definition of Roles</h2>
...
+								level MAY be indicated via <pref>aria-level</pref>.</li>
```
mcking65: Closed the `<li>` with the previous suggestion, so will suggest closing the `<ul>` here.

```suggestion
							</ul>
```

index.html
```diff
@@ -2326,6 +2327,102 @@ <h2>Definition of Roles</h2>
...
+						</ul>
```
mcking65: Suggested closing the `<ul>` on prior line so closing the `<li>` here.

```suggestion
					</li>
```

index.html
```diff
@@ -2326,6 +2327,102 @@ <h2>Definition of Roles</h2>
...
+					</li>
```
mcking65: Suggested closing the `<li>` on prior line, so close the `<ol>` here.

```suggestion
				</ol>
```

index.html
```diff
@@ -2326,6 +2327,102 @@ <h2>Definition of Roles</h2>
...
+				</ol>
```
mcking65: Add a paragraph with user agent requirements about aria-level, aria-posinset, and aria-setsize here. I suppose this is not the place to specify how the computation should be done is it? I am wondering about the situations where the author has specified it on some but not others? What should the user agent do? Is it sufficient for the author to declare aria-level on the first comment element in the DOM? would the user agent then be able to compute subsequent values based on that value? BTW, we do not have normative user agent MUST statements for aria-level, aria-posinset, or aria-setsize for any other element that supports them. I think we should. but, we do not.

```suggestion
				<p>If the author has not explicitly declared <pref>aria-level</pref>, <pref>aria-posinset</pref>, or <pref>aria-setsize</pref> for a <code>comment</code> element, user agents MUST automatically compute the values and expose them to assistive technologies.</p>
```

index.html
```diff
@@ -2326,6 +2327,102 @@ <h2>Definition of Roles</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="comment">
+			<rdef>comment</rdef>
+			<div class="role-description">
+				<p>A comment contains a remark expressing a reaction to other content within the page.</p>
+				<p>The relationship of comments to other content SHOULD be exposed as follows:</p>
+				<ol>
+					<li>If the comment refers to other content within the page, the source content SHOULD use <pref>aria-details</pref> to point to the related comment.</li>
```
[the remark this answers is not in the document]
carmacleod: Regarding the "many comments for a blog" scenario, why we would prefer to have `aria-details` point to a comment-container element rather than having it point to the list of root comments?
- Is it because it simplifies the code for authors (they wouldn't have to come up with a bunch of unique ids)?
- Or is it possible that the list of ids could get too long for UA to maintain?
- Or are we trying to encourage authors to use a landmark in this use case?

Also (sorry to bring this up if it is already decided) if we are suggesting pointing to a container for the blog comments use case, why aren't we suggesting pointing to a container (even if just a div) for every use case that has more than one comment? (in which case we could consider going back to aria-details having only one id)?

Anyhow, if we allow choosing between a list of ids or a container, the we need to add some words that explain _why_ to use one method or the other, to help authors decide which way to code their particular use case.

index.html
```diff
@@ -2326,6 +2327,102 @@ <h2>Definition of Roles</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="comment">
+			<rdef>comment</rdef>
+			<div class="role-description">
+				<p>A comment contains a remark expressing a reaction to other content within the page.</p>
```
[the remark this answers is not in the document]
aleventhal: I also removed the "in page" part, because I agree that it's nice to be less restrictive on that. A comment could relate to content in another page, e.g. "View next 20 comments" might take the user to a new page. However, it would usually be in the same page.

index.html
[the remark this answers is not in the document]
mcking65: > Regarding the "many comments for a blog" scenario, why we would prefer to have `aria-details` point to a comment-container element rather than having it point to the list of root comments?

The language is meant to be neutral and not suggest that  there is preference. I think we might work that out in practices.

> * Is it because it simplifies the code for authors (they wouldn't have to come up with a bunch of unique ids)?

Simplify code yes -- Modifying the article attributes  every time a new comment is posted could be problematic in some applications.

> * Or is it possible that the list of ids could get too long for UA to maintain?

That is a good question. I have no idea what the limits are there. We would certainly want to consider this question when writing practices.

> * Or are we trying to encourage authors to use a landmark in this use case?

Not necessarily strong encouragement, but certainly suggesting it for consideration.

> Also (sorry to bring this up if it is already decided) if we are suggesting pointing to a container for the blog comments use case, why aren't we suggesting pointing to a container (even if just a div) for every use case that has more than one comment? (in which case we could consider going back to aria-details having only one id)?

There are certainly going to be designs where multiple comments or other types of annotations for an element will not all be together in the same container. So, we definitely need to support multiple IDs for aria-details.

If there are multiple comments but all in a single thread, there is no need for a container. In that case, only the root comment is referenced by aria-details. However, if the thread is paged, I am not yet clear on how that would work.

> Anyhow, if we allow choosing between a list of ids or a container, the we need to add some words that explain _why_ to use one method or the other, to help authors decide which way to code their particular use case.

I don't think we need that in the spec.

I'm still thinking about the container case. At one time, @aleventhal had proposed a commentsection role. I didn't and still do not think that is the approach that best addresses both the needs of users and AT devs. I am especially not a fan of yet one more role that has such a special purpose semantic. The more roles of that nature we have, the less useful they become to users. Sometimes you want the comment collection in a landmark region, and sometimes you do not. So, making commentsection a region would be a bad idea because it could force region proliferation into some apps. On the other hand, if it is its own special thing, that means locating it is yet one more very specialized function, of which we already have too many by at least one order of magnitude, and approaching 2.

There is definite advantage to knowing what is at the other end of a details relationship. That enables screen readers in particular to provide the user with information the lets them know whether or not the details relationship is something they need to exercise.

Perhaps one way would be to have a property called aria-containsannotations that would be supported on group, list, region, complementary, table, grid, and treegrid. The property could support tokens like comment, suggestion, endnote. Then, if aria-details points to that container, the screen reader would know to look inside and maybe even count the number of elements with each of the types suggested by the value of the property. Now, the screen reader gets to know what to tell the user. The engineer only needs one ID relationship, and the designer of the accessible experience can decide what the best type of container is for the annotations.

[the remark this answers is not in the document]
aleventhal: @carmacleod, we're trying to keep things simple in the common case, but make difficult/rare things possible. It was a compromise but we decided that by having only "comment", it would be simpler and cleaner for most cases. It didn't seem like comment section was worth having.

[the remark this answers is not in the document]
aleventhal: @mcking65  I'm trying some things with your wording to IMO make it simpler. For one thing, I think just saying one case is for replies, and another is for non replies makes it easier to follow, and feels less convoluted, so I'm putting the emphasis more on that.

index.html
aleventhal: It's like tree view. In this case, the structure, posinset and setsize are all implied by the aria-level. The author can use aria-posinset and aria-setsize, but they don't have to. If they only use aria-level, the posinset and setsize can be computed by the user agent. It's clear when the set changes because the level goes up. This works in Chrome and Firefox for tree items, and in Chrome for comment.

Example:
Level 1  (posinset = 1, setsize = 1) 
Level 2 (posinset = 1, setsize = 2)
Level 3 (posinset = 1, setsize = 1)
Level 2 (posinset = 2, setsize = 2)

aleventhal: From the user agent perspective, tje only reason to point to a container element of comments is for the aria-level case. You can use as many ids in aria-details as you want.

It might be more convenient to point to a containing element though, because if comments are being constantly added and removed, it's no longer necessary to update aria-details. That's a separate issue though, from what is being commented (e.g. a blog article). I think this new text confuses all of that. I'll try to fix.

index.html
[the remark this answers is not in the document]
carmacleod: I came back to this issue to strike out my mention of aria-details always pointing to a single container (perfectly happy to let that drop ;) and I saw that you replied already. Thanks for the quick response!

> There are certainly going to be designs where multiple comments or other types of annotations for an element will not all be together in the same container. So, we definitely need to support multiple IDs for aria-details.

I had just come to the same conclusion after re-reading aria-details. The detail elements could conceivably be in multiple places. So it doesn't make sense to require authors to always use a container.

I also realized that comments aren't the only type of annotation that could have more than one instance. Content could easily have footnotes, a term could have several definitions, etc.

And of course, you have pointed out that aria-details can point to any mix of annotations - e.g. a comment and a footnote. In fact, aria-details can point to elements with any role, or mix of roles.

The aria-containsannotations attribute idea sounds interesting. I agree that something like that is better than adding special-purpose roles like commentsection.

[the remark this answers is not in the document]
aleventhal: I don't think we need an actual property aria-containsannotations. If we want, we could compute that infromation automatically for the target of an aria-details, when it is a landmark or generic role.

index.html
aleventhal: Hi all, I posted a new version. It is inspired by @mcking65's version but still fairly different in a few place. I think I addressed concerns but improved readability. I also made sure the container case wasn't related to what the commented content was. Please take a look.

index.html
```diff
@@ -2326,6 +2327,107 @@ <h2>Definition of Roles</h2>
...
 		</div>
+		<div class="role" id="comment">
+			<rdef>comment</rdef>
+			<div class="role-description">
+				<p>A comment contains content expressing reaction to other content.</p>
+				<p>Comments can annotate any visible content, from small spans of text, to other comments, to entire articles. Authors SHOULD identify the relationships between comments and the commented content, as follows:</p>
+				<ol>
+					<li>If the comment is a reply to another <code>comment</code>:
+						<ul>
+							<li>If all ancestor comments are available in the DOM, make the reply <code>comment</code> a semantic descendant of the parent <code>comment</code> being replied to, either as a DOM descendant element, or via <pref>aria-owns</pref>.</li>
```
mcking65: If the reply has a parent, it is already a descendant. Recommend removing the word "parent" and also fix dangling preposition:

```suggestion
							<li>If all ancestor comments are available in the DOM, make each reply <code>comment</code> a semantic descendant of the <code>comment</code> to which it is replying, either by making it a DOM descendant element or by using <pref>aria-owns</pref>.</li>
```

mcking65:
I wonder if screen readers would really want to probe every use of aria-details referring to an element that is not a specific annotation type to find out whether it is a general description or a container of specific annotation types. If they have no problem with that, then there would be no use for a property like aria-containsannotations. So, whether it would add value is really a question for screen reader devs.

carmacleod:
> I wonder if screen readers would really want to probe every use of aria-details referring to an element that is not a specific annotation type to find out whether it is a general description or a container of specific annotation types.

Good question. If the container had a label, then the user might benefit more from hearing the container's label than from hearing about detailed annotation types within the container?

For example, if a complementary region was the target of an aria-details, and the author added a nice visible label saying _24 comments on "Dogs and their Owners"_ then you wouldn't want to hear just "contains 24 comments", and you wouldn't want to hear, "24 comments on "Dogs and their Owners", contains 24 comments".

```
<article aria-details="comment-section">
  <h1>Dogs and their Owners</h1>
  <p>This is all about dogs. Or is it all about their owners?</p>
</article>
<aside id="comment-section" aria-labelledby="comment-section-label">
  <h2 id="comment-section-label">24 comments on "Dogs and their Owners"<h2>
  ... 24 sibling comment elements go here ...
</aside>
```

index.html
```diff
@@ -2326,6 +2327,107 @@ <h2>Definition of Roles</h2>
...
+							<li>Alternatively, if all ancestor comments are not in the DOM, such as when comments are paginated, the hierarchical
+								level MAY be indicated via <pref>aria-level</pref>. Additional group positional information MAY be indicated via <pref>aria-posinset</pref> and <pref>aria-setsize</pref>.</li>
+						</ul>
+					</li>
+					<li>Otherwise, if the comment relates to other content in the page:
+						<ul>
+							<li>Provide <pref>aria-details</pref> on the commented content, identifying the commented element.</li>
```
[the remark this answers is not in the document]
aleventhal: To me it's more wordy and says the same thing, but since @carmacleod also likes it.

index.html
```diff
@@ -2326,6 +2327,107 @@ <h2>Definition of Roles</h2>
...
+							<li>If there are many comments relating to the same commented content, either identify all of them individually via <pref>aria-details</pref>, or identify a parent container of the comments, and assign a role of <rref>complementary</rref> to the container.</li>
```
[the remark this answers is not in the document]
aleventhal: Here's why I put it: we're going to provide a computed hint to the screen reader, it would be good to know what role(s) we need to provide that hint on. I think requiring role="complementary" with a computed hint would help have a compromise between having role="commentsection" and not. We essentially still get the benefits of comment section. It's more work for the browser, but have one fewer role.

index.html
aleventhal: I like having the word parent in there, it makes the sentence easier to read, since we're talking about 2 different comments here. I'll take the rest.

aleventhal: Ok, I see, you are using "to which it is replying" instead of "parent comment". That means the same thing to me but is longer. For example, when using something like "reddit" there are buttons that say things like "go to parent comment".

aleventhal: Wasn't sure I liked the wordiness, but, I went with this in the end.

aleventhal:
Matt, I took your suggestions except for complementary. I think there is value in requiring a specific role for a grouping element for comments. This will make it easier for the user agent to know when to bother computing a "hascomments" object attribute hint.

index.html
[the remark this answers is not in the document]
aleventhal: Actually a complementary suggest content that can stand alone, which this isn't.
Perhaps we should limit authors to a named region (if a landmark) or a group (if not a landmark).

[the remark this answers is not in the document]
aleventhal: As discussed on the call, I went with group or region.

aleventhal:
@mcking65 @jnurthen I've made the requested change to use group or region. I've made it a SHOULD. LMK what you think.

jnurthen:
@mcking65 can you take a look. I need to merge this one before I merge aria-details.

index.html
```diff
@@ -2326,6 +2327,107 @@ <h2>Definition of Roles</h2>
...
+							<li>Alternatively, if all ancestor comments are not in the DOM, such as when comments are paginated, the hierarchical
+								level MAY be indicated via <pref>aria-level</pref>. Additional group positional information MAY be indicated via <pref>aria-posinset</pref> and <pref>aria-setsize</pref>.</li>
+						</ul>
+					</li>
+					<li>Otherwise, if the comment relates to other content in the page:
+						<ul>
+							<li>Provide <pref>aria-details</pref> on the element containing the commented content with a value refering to the element with role <code>comment</code>.</li>
+							<li>If there are many comments relating to the same commented content, either identify all of them individually via <pref>aria-details</pref>, or identify a parent container of the comments. If <pref>aria-details</pref> is used to indentify to a parent container for the comments, authors SHOULD assign a role of <rref>group</rref> or <rref>region</rref> to the container.</li>
```
mcking65: We should be more specific about the use of aria-details. I also find "many" vague. This statement is true any time it is more than one.

```suggestion
							<li>If there are multiple comments related to the same commented content, either provide a value for <pref>aria-details</pref> on the commented content that refers to each individual comment, or use <pref>aria-details</pref> to refer to a parent container of the comments. If <pref>aria-details</pref> refers to an element containing comments rather than <code>comment</code> elements, authors SHOULD assign a role of <rref>group</rref> or <rref>region</rref> to the referenced container.</li>
```

PR #1136, opened 2019-12-11
Update aria-details

aleventhal:
1) Support multiple elements.
2) Support types of information/annotations that are not descriptions.

Refers to issue #749.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1136.html" title="Last updated on Feb 7, 2020, 9:05 PM UTC (698cd15)">Preview</a> (<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1136.html#aria-details" title="#aria-details">#aria-details</a>) (<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1136.html#AriaAttributes" title="#AriaAttributes">#AriaAttributes</a>) | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1136/7193e40...698cd15.html" title="Last updated on Feb 7, 2020, 9:05 PM UTC (698cd15)">Diff</a>


index.html
```diff
@@ -10711,12 +10711,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p>Unlike elements referenced by <code>aria-describedby</code>, the element referenced by <code>aria-details</code> is not used in either the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_name">Name Computation</a> or the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_description">Description Computation</a> as defined in the Accessible Name and Description specification [[ACCNAME-1.1]]. Thus, the content of an element referenced by <code>aria-details</code> is not flattened to a string when presented to assistive technology users. This makes <code>aria-details</code> particularly useful when converting the information to a string would cause a loss of information or make the extended description more difficult to understand.</p>
+				<p>Identifies one or more elements that provide additional, extended information about the <a>object</a>.</p>
```
jnurthen:
```suggestion
				<p>Identifies the <a>element</a> (or elements) that provide additional, extended information about the <a>object</a>.</p>
```
For consistency with other definitions

index.html
```diff
@@ -2567,7 +2567,7 @@ <h2>Definition of Roles</h2>
 			<rdef>definition</rdef>
 			<div class="role-description">
 				<p>A definition of a term or concept. See related <rref>term</rref>.</p>
-				<p>Authors SHOULD identify the <a>element</a> being defined by giving that element a role of <rref>term</rref> and referencing it with the <pref>aria-labelledby</pref> <a>attribute</a> or by making the element with role <rref>term</rref> a descendant of the element with role <code>definition</code>.</p>
+				<p>Authors SHOULD identify the <a>element</a> being defined by giving that element a role of <rref>term</rref> and referencing it with the <pref>aria-details</pref> <a>attribute</a> or by making the element with role <rref>term</rref> a descendant of the element with role <code>definition</code>.</p>
```
jnurthen: Seems ancillary to this pull request. @joanmarie  would this need any corresponding CORE-AAM or implementation changes? If so I suggest moving this to a separate PR.

index.html
```diff
@@ -10711,12 +10711,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p>Unlike elements referenced by <code>aria-describedby</code>, the element referenced by <code>aria-details</code> is not used in either the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_name">Name Computation</a> or the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_description">Description Computation</a> as defined in the Accessible Name and Description specification [[ACCNAME-1.1]]. Thus, the content of an element referenced by <code>aria-details</code> is not flattened to a string when presented to assistive technology users. This makes <code>aria-details</code> particularly useful when converting the information to a string would cause a loss of information or make the extended description more difficult to understand.</p>
+				<p>Identifies one or more elements that provide additional, extended information about the <a>object</a>.</p>
+				<p>The additional information can be:</p>
+				<ul>
+					<li>A description or general annotation (this is the default). See related <pref>aria-describedby</pref>.</li>
```
jnurthen: As discussed at https://www.w3.org/2019/09/16-aria-minutes.html#item02 I think we need to stop using the word description in aria-details prose https://www.w3.org/2019/09/16-aria-minutes.html#item02

index.html
```diff
@@ -10711,12 +10711,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Unlike elements referenced by <code>aria-describedby</code>, elements referenced by <code>aria-details</code> are not used in either the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_name">Name Computation</a> or the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_description">Description Computation</a> as defined in the Accessible Name and Description specification [[ACCNAME-1.1]]. Thus, the content of elements referenced by <code>aria-details</code> are not flattened to a string when presented to assistive technology users. This makes <code>aria-details</code> particularly useful when converting the information to a string would cause a loss of information or make the extended description more difficult to understand.</p>
```
jnurthen:
```suggestion
				<p>Unlike elements referenced by <code>aria-describedby</code>, elements referenced by <code>aria-details</code> are not used in either the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_name">Name Computation</a> or the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_description">Description Computation</a> as defined in the Accessible Name and Description specification [[ACCNAME-1.1]]. Thus, the content of elements referenced by <code>aria-details</code> are not flattened to a string when presented to assistive technology users. This makes <code>aria-details</code> particularly useful when converting the information to a string would cause a loss of information or make the extended information more difficult to understand.</p>
```

index.html
```diff
@@ -13125,7 +13134,7 @@ <h2>Interface Mixin <dfn>AriaAttributes</dfn></h2>
 				attribute FrozenArray&lt;Element&gt;? ariaControlsElements;
 				attribute DOMString? ariaCurrent;
 				attribute FrozenArray&lt;Element&gt;? ariaDescribedByElements;
-				attribute Element? ariaDetailsElement;
+				attribute Element? ariaDetailsElements;
```
jnurthen: This would need to be a FrozenArray like ariaDescribedByElements

index.html
aleventhal: Doesn't require implementation changes in browser, or any CORE-AAM changes that I would be aware of.

I can put in separate change, but it did seem related to having a more consistent understanding of what aria-details does.

index.html
aleventhal: What if we were clear that a description is only one of many potential types of aria-details/annotations? If we avoid saying description at all, how would authors know that they can use aria-details for a description that can be navigated to?

index.html
aleventhal: Thanks.

index.html
joanmarie: I agree that there is no need to change Core-AAM or user agent implementations.

But I tend to agree with James that this change does seem related, but not part of the aria-details revamp.

index.html
```diff
@@ -10711,12 +10711,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p>Unlike elements referenced by <code>aria-describedby</code>, the element referenced by <code>aria-details</code> is not used in either the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_name">Name Computation</a> or the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_description">Description Computation</a> as defined in the Accessible Name and Description specification [[ACCNAME-1.1]]. Thus, the content of an element referenced by <code>aria-details</code> is not flattened to a string when presented to assistive technology users. This makes <code>aria-details</code> particularly useful when converting the information to a string would cause a loss of information or make the extended description more difficult to understand.</p>
+				<p>Identifies the <a>element</a> (or elements) that provide additional, extended information about the <a>object</a>.</p>
```
joanmarie: Nit: I'd lose the parens around "or elements".

joanmarie: Also: Why did you move the "see related"?

index.html
joanmarie: I think it's clear that a description is only one of many potential types from the fact that it is followed by comment and definition and footnote and endnote. 

I like "general annotation." 

As I alluded to above, I don't see the benefit of moving the "See related" here. Please explain.

joanmarie: Do we want to put an authors SHOULD and a role for the general annotation case?

Context: In my overall comment/review, I asked for the requirement that authors name the object which has the details. This makes it possible for screen readers to say things like "has details in the requirements list." Rather than just "has details" or "has details in list." Then I subsequently realized that we have plenty of objects/roles which cannot be named....

index.html
aleventhal: Ok I will separate.

index.html
jnurthen: All the others use "element (or elements)" so if we change it we should change all the others....

index.html
aleventhal: Created https://github.com/w3c/aria/pull/1147

index.html
aleventhal: What would the role be if the aria-details is a general annotation or description? I don't think there is always an appropriate role for a general annotation.

Also, while I think the author should use a role if there is an appropriate one, but that's a general ARIA truth anyway and is just adding more verbage. 

I move the "See related" back to where it was, but I actually think it was better where I moved it. Because aria-describedby is only related and only makes potential sense for a general description or annotation, not for a comment, footnote, etc.

index.html
aleventhal: Done.

aleventhal: Ah, I missed James' comment. So do we need to add them back and remove the parens in a different PR?

jnurthen: I think we should leave them in all.

joanmarie: I think we should take them all out :stuck_out_tongue: BUT, if Aaron's text is consistent with what we do elsewhere, then I withdraw my suggestion for this PR. Sorry for the noise!

index.html
joanmarie: Given we agreed that we wouldn't be pushing authors to name the objects, my suggestion to push the role is obsolete.

As for the see related, let's see what others think. But for the most part I believe our "see related" instances are in the first paragraph/short definition of the related thing.

aleventhal: Is this good as it stands now?

index.html
```diff
@@ -10711,12 +10711,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Unlike elements referenced by <code>aria-describedby</code>, elements referenced by <code>aria-details</code> are not used in either the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_name">Name Computation</a> or the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_description">Description Computation</a> as defined in the Accessible Name and Description specification [[ACCNAME-1.1]]. Thus, the content of elements referenced by <code>aria-details</code> are not flattened to a string when presented to assistive technology users. This makes <code>aria-details</code> particularly useful when converting the information to a string would cause a loss of information or make the extended information more difficult to understand.</p>
```
aleventhal: I copied this line, but are we sure aria-describedby is used in the accessible name computation? Obviously used in accessible description, but name too?

carmacleod:
@aleventhal What is the use case for multiple elements? Does it merit an example?

aleventhal:
> @aleventhal What is the use case for multiple elements? Does it merit an example?

Example is multiple comments on the same content. These may not be part of the same thread. For example, Google Docs allows this.

carmacleod:
> Example is multiple comments on the same content. 

Ah, ok - I see. Thanks.

index.html
scottaohara: re: meeting comment - remove mention of accessible name.

index.html
```diff
@@ -10711,13 +10711,22 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Unlike elements referenced by <code>aria-describedby</code>, elements referenced by <code>aria-details</code> are not used in the Accessible <a href="https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_description">Description Computation</a> as defined in the Accessible Name and Description specification [[ACCNAME-1.1]]. Thus, the content of elements referenced by <code>aria-details</code> are not flattened to a string when presented to assistive technology users. This makes <code>aria-details</code> particularly useful when converting the information to a string would cause a loss of information or make the extended information more difficult to understand.</p>
+				<p>It is valid for an element to contain both <code>aria-details</code> and a description via <code>aria-describedby</code> or <code>aria-description</code>. In this case, user agents should minimally expose the <code>aria-details</code> relation and the computed flat description string. In user agents that cannot expose multiple descriptive relations, the first <code>aria-details</code> takes precedence.</p>
```
mcking65: This paragraph is implying normative UA requirements and offering new information about aria-details as well as making comparisons with other properties. I think the present wording could be hard to follow for some who are not intimately familiar with the differences among aria-details, aria-describedby, and aria-details. I also think the UA requirements should be more explicit. I suggest the following.

```suggestion
				<p>The <code>aria-details</code> property supports referring to multiple elements. For example, a paragraph in a document editor may reference multiple comments that are not related to each other. If a user agent relies on an accessibility API that does not support exposing multiple descriptive relations, the user agent SHOULD expose the relationship to the first element referenced by <code>aria-details</code>.</p>
				<p>It is valid for an element to have both <code>aria-details</code> and a description specified with either <code>aria-describedby</code> or <code>aria-description</code>. If a user agent relies on an accessibility API that does not support exposing multiple descriptive relations, and if an element has both <code>aria-details</code> and <rref>aria-describedby</rref>, the user agent SHOULD expose the <code>aria-details</code> relation and the description string computed from the <rref>aria-describedby</rref> relationship.</p>
```

index.html
```diff
@@ -10711,13 +10711,23 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<ul>
+					<li>Description or general annotation: <code>aria-details</code> refers to an element with any other role.</li>
+					<li>Comment: <code>aria-details</code> refers to an element with role <rref>comment</rref>.</li>
+					<li>Definition: <code>aria-details</code> is applied to an element with role <rref>term</rref> and refers to an element with role <rref>definition</rref>.</li>
+					<li>Footnote: <code>aria-details</code> refers to an element with role <code>doc-footnote</code>. This role is defined in [[DPUB-ARIA]].</li>
+					<li>Endnote: <code>aria-details</code> refers to an element with role <code>doc-endnote</code>. This role is defined in [[DPUB-ARIA]].</li>
+				</ul>
+
+				<p>The <code>aria-details</code> attribute references elements that provide more detailed information than would normally be provided by <pref>aria-describedby</pref>. It enables <a>assistive technologies</a> to make users aware of the availability of extended information as well as navigate to it. Authors SHOULD ensure elements referenced by <code>aria-details</code> are visible to all users.</p>
```
```suggestion
				<p>The <code>aria-details</code> attribute references elements that provide more detailed information than would normally be provided by <pref>aria-describedby</pref>. It enables <a>assistive technologies</a> to make users aware of the availability of extended information as well as navigate to it. Authors SHOULD ensure that elements referenced by <code>aria-details</code> are visible to all users.</p>
```

index.html
```diff
@@ -13125,7 +13135,7 @@ <h2>Interface Mixin <dfn>AriaAttributes</dfn></h2>
 				attribute FrozenArray&lt;Element&gt;? ariaControlsElements;
 				attribute DOMString? ariaCurrent;
 				attribute FrozenArray&lt;Element&gt;? ariaDescribedByElements;
-				attribute Element? ariaDetailsElement;
+				attribute FrozenArray? ariaDetailsElements;
```
carmacleod: I believe this should be `FrozenArray<Element>` like the other FrozenArrays of Elements:
```suggestion
				attribute FrozenArray&lt;Element&gt;? ariaDetailsElements;
```

aleventhal: Wow good catch!

index.html
```diff
@@ -10711,13 +10711,23 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Identifies the <a>element</a> (or elements) that provide additional information related to the <a>object</a>. See related <pref>aria-describedby</pref>.</p>
+				<p>Assistive technologies MAY use the role of elements referenced by the <code>aria-details</code> property to help users understand the types of information associated with the element. Authors MAY convey the type of details associated with an element as follows:</p>
+				<ul>
+					<li>Description or general annotation: <code>aria-details</code> refers to an element with any other role.</li>
+					<li>Comment: <code>aria-details</code> refers to an element with role <rref>comment</rref>.</li>
+					<li>Definition: <code>aria-details</code> is applied to an element with role <rref>term</rref> and refers to an element with role <rref>definition</rref>.</li>
+					<li>Footnote: <code>aria-details</code> refers to an element with role <code>doc-footnote</code>. This role is defined in [[DPUB-ARIA]].</li>
```
carmacleod: The DPUB-ARIA reference is currently listed as:
> [DPUB-ARIA] 
>  Reference not found.

I'm pretty sure it needs to be [[dpub-aria-1.0]], but you'll have to try it to make sure that ReSpec has that string in its database. It should, because that's the string the [ARIA in HTML spec](https://w3c.github.io/html-aria/#normative-references) uses.

```suggestion
					<li>Footnote: <code>aria-details</code> refers to an element with role <code>doc-footnote</code>. This role is defined in [[dpub-aria-1.0]].</li>
```

index.html
```diff
@@ -10711,13 +10711,23 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+					<li>Endnote: <code>aria-details</code> refers to an element with role <code>doc-endnote</code>. This role is defined in [[DPUB-ARIA]].</li>
```
carmacleod: As above, I think (?) you need [[dpub-aria-1.0]].

```suggestion
					<li>Endnote: <code>aria-details</code> refers to an element with role <code>doc-endnote</code>. This role is defined in [[dpub-aria-1.0]].</li>
```

index.html
aleventhal: It works both in all caps or lower case if you do DPUB-ARIA-1.0. Why would it be lowercase?

aleventhal:
This branch has conflicts, but if I rebase to master and push, will that make the review harder? I think it will be harder to see changes between versions.

index.html
carmacleod: Sorry, I have no clue. 😄 
If it works with all caps, that's better, because most of [the refs in the ARIA spec](https://w3c.github.io/aria/#normative-references) use all caps.

carmacleod:
> if I rebase to master and push, will that make the review harder?

We're so close - maybe wait? @jnurthen ?

index.html
```diff
@@ -10711,13 +10711,23 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p>In some user agents, multiple reference relationships for descriptive information are not supported by the accessibility API. In such cases, if both <pref>aria-describedby</pref> and <code>aria-details</code> are provided on an element, <code>aria-details</code> takes precedence.</p>
+				<p>Identifies the <a>element</a> (or elements) that provide additional information related to the <a>object</a>. See related <pref>aria-describedby</pref>.</p>
+				<p>Assistive technologies MAY use the role of elements referenced by the <code>aria-details</code> property to help users understand the types of information associated with the element. Authors MAY convey the type of details associated with an element as follows:</p>
+				<ul>
+					<li>Description or general annotation: <code>aria-details</code> refers to an element with any other role.</li>
```
mcking65: The wording I suggested for this in the previous review assumed this item would be come last in the list. I think it will be far more clear if you move it to the end of the list.

index.html
mcking65: It seems like this should be the first paragraph after the definition. The above description of authoring techniques would be more understandable if this were read first. It would also help make the aria-describedby distinction and visibility requirements more prominent.

```suggestion
				<p>The <code>aria-details</code> attribute is for referencing elements that provide more detailed information than would normally be provided via <pref>aria-describedby</pref>. The <code>aria-details</code> property enables <a>assistive technologies</a> to make users aware of the availability of extended information and navigate to it. Authors SHOULD ensure elements referenced by <code>aria-details</code> are visible to all users.</p>
```

index.html
aleventhal: My mistake, thought I had made that change. Sorry! :)

index.html
aleventhal: Yes, I like that.

aleventhal: I moved it to the second paragraph and kept most of your wording. Please take a look.

mcking65:
Looking great now! Thank you for the updates!!

aleventhal:
Anyone else want to approve so we can get this in? 
@carmacleod @jnurthen @joanmarie

carmacleod:
@aleventhal 
Would you be able to add one teeny tiny completely inconsequential change "while you're there"?
In [example 18](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1136.html#example-18), the word "ramp" is missing from:
> In this example you will notice a skateboard **ramp** with a base and vertical board ...

I know... it's the equivalent of Lorem ipsum and nobody's going to read it, and you didn't even write that example... but I Can't. Let. Go... 😄

aleventhal:
> @aleventhal
> Would you be able to add one teeny tiny completely inconsequential change "while you're there"?

Done.

joanmarie:
Something just occurred to me: Are we introducing any normative dependencies on the dpub-aria spec by mentioning `doc-footnote` and `doc-endnote`?

I personally think footnote and/or endnote make sense as general ARIA roles. They show up on sites like Wikipedia.

So.... At the risk of adding just. one. more. thing! Do we want to create footnote and/or endnote roles and then use them in aria-details? (Sorry!!)

aleventhal:
If we define it in ARIA, can we just define it with exactly the same name including the "doc-" prefix? Otherwise we need to update all the browser and AT implementations to add "footnote" and "endnote" mappings. JAWS will now need to be updated because they read the xml-roles object attribute. Requiring a JAWS update sets us back quite a lot, because not all users upgrade. So I think making this "footnote" creates a real issue.

I also don't like adding the non-prefixed versions because it's just redundant and confusing.

Personally I don't see a problem with the dependency. How does it negatively affect end users or authors? If we add new role names that just adds to confusion and redundant code.

joanmarie:
> If we define it in ARIA, can we just define it with exactly the same name including the "doc-" prefix?

I think that's a discussion we might need to have with the group.

> Otherwise we need to update all the browser and AT implementations to add "footnote" and "endnote" mappings.

I agree that sucks. Though some platforms already have platform-specific roles for footnote.

> JAWS will now need to be updated because they read the xml-roles object attribute. Requiring a JAWS update sets us back quite a lot, because not all users upgrade. So I think making this "footnote" creates a real issue.

Aren't we going to need a JAWS update for `aria-details` supporting multiple targets, including those with the roles in your bulleted list? (Not saying this isn't a problem; just questioning whether or not the JAWS update problem goes away by our including `doc-footnote` and `doc-endnote` in the ARIA spec.)
 
> I also don't like adding the non-prefixed versions because it's just redundant and confusing.

In terms of redundant, if we added `endnote` and `footnote` (prefixed or not), I believe DPub would deprecate their roles. @TzviyaSiegman?

Related aside: We do have some issues around `doc-endnote` in the dpub-aria spec that we haven't yet resolved, namely that `doc-endnote` must be contained in a `list`, but an ARIA `list` must contain only `listitem` children.

> Personally I don't see a problem with the dependency. How does it negatively affect end users or authors? If we add new role names that just adds to confusion and redundant code.

I'll let @michael-n-cooper weigh in on that. But my understanding is that we try to avoid normative dependencies on other specs.

TzviyaSiegman:
> In terms of redundant, if we added endnote and footnote (prefixed or not), I believe DPub would deprecate their roles. @TzviyaSiegman?

Yes, I think we would. We need to confirm with Publishing WG that this will not wreak too much havoc.

aleventhal:
True that footnote is not as useful without the aria-details work going on, but it is still useful. As the user enters the footnote in JAWS today, they will hear that they're entering a footnote.

I think if we just call it footnote, we should consider how to keep current versions of JAWS relevant.

aleventhal:
Maybe I'm making too big of deal out of it. I don't know :/

Still, there are a lot of useful roles in DPUB-ARIA, like doc-pagebreak, doc-toc. I happen to like that spec :)

michael-n-cooper:
> Related aside: We do have some issues around |doc-endnote| in the 
> dpub-aria spec that we haven't yet resolved, namely that |doc-endnote| 
> must be contained in a |list|, but an ARIA |list| must contain only 
> |listitem| children.
>
>     Personally I don't see a problem with the dependency. How does it
>     negatively affect end users or authors? If we add new role names
>     that just adds to confusion and redundant code.
>
> I'll let @michael-n-cooper <https://github.com/michael-n-cooper> weigh 
> in on that. But my understanding is that we try to avoid normative 
> dependencies on other specs.
>
As far as the question of normative dependencies, it's fairly normal for 
specs to have them. The only issue is that the spec you're depending on 
has to be at least as mature as your own spec, so we don't wind up with 
a Recommendation that normatively depends on something that isn't a Rec 
and therefore could change, changing the normative requires of the Rec 
in the process.

This example is slightly odd because the vision, I think, was that 
DPub-ARIA would depend on ARIA, but not the other way around. For ARIA 
to depend on DPub-ARIA just in order to be able to allow roles as owned 
elements seems odd. I think we didn't think through that situation when 
we decided to allow ARIA modules. I'd like to have a more elegant way of 
handling this in the ARIA structure. But probably doing so introduces 
more complicatedness than it's worth for just a couple instances. So the 
solution of depending on DPub-ARIA may be the most practical.

I'm not too concerned about the possibility of DPub-ARIA being rescinded 
or something. We either control that or are closely involved with it, so 
would know that's happening and be able to adjust accordingly.

jcsteh:
I'm adding my vote (for whatever it's worth) for *not* creating new names
(in whatever spec) for existing roles; i.e. we already have doc-footnote,
so let's not add footnote. There's already enough messiness in the a11y
landscape. I don't think we want to create more mess without a really
compelling reason, and IMO, this isn't sufficiently compelling.

Issue #891, opened 2019-01-24
Request for aria-description to set accessible description


backwardok:
For the accessible name/label of an element, there is [aria-label](https://www.w3.org/TR/wai-aria/#aria-label) to set the label directly and [aria-labelledby](https://www.w3.org/TR/wai-aria/#aria-labelledby) to set it via reference to another node. However, for the description of an element, there is only [aria-describedby](https://www.w3.org/TR/wai-aria/#aria-describedby). This requires providing extra descriptive content in a node elsewhere in the DOM to be referenced by `id`, even if its only purpose is for the description of another node. 

In developing, I've found that there are occasionally extra pieces of information that are visually understood but don't have a complete textual representation that need to be provided to AT users in order for it to make more sense. In order to provide that context, visually hidden nodes get added to the markup to expose them with `aria-describedby`. However, unless the nodes purely for description are also hidden from AT via `aria-hidden` or applying hidden CSS like `display: none`, an AT user will end up coming across that descriptive text separately, and potentially out of context.

An example I've seen of this is on Twitter, where each tweet has a button for reply, retweet, and like. The button also includes the number of replies/retweets/likes there are. In this context, it's important that the button label matches the actual action the button has (reply/retweet/like), but it's also important to expose the information that indicates how many existing interactions there were (replies/retweets/likes). Twitter currently uses `aria-describedby` with a set of nodes that are all at the beginning of the action section and it confusingly gets read out as text prior to reaching the related buttons. 

Some of the benefits I see to having an `aria-description` attribute:
- Simplification/reduction of the DOM tree (and accessibility tree to some extent)
- Removal of need to create `id`s on several nodes
- Easier association of the actual description to the node itself
- Relates to an existing ARIA attribute and would therefore be less difficult to support, as a description is already supported for accessibility nodes

accdc:
Hi, as an FYI in the short term, the title attribute will set the accessible description property, which might be useful in some cases where supplementary information is needed. Of course this will also result in a tooltip for sighted mouse users too though.

carmacleod:
There's [a long thread from 2011 that includes a request for aria-description](https://lists.w3.org/Archives/Public/wai-xtech/2011Apr/0007.html).
I didn't follow the thread all the way through, but it seems to contain some useful discussion.

stes-acc:
ACCDC wrote: " Of course this will also result in a tooltip for sighted mouse users too though."

[...]

PR #1137, opened 2019-12-11
Add aria-description

aleventhal:
Closes issue #891

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1137.html" title="Last updated on Jan 16, 2020, 5:14 AM UTC (73f119d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1137/f519fca...73f119d.html" title="Last updated on Jan 16, 2020, 5:14 AM UTC (73f119d)">Diff</a>


index.html
```diff
@@ -10711,7 +10711,47 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				</tbody>
 			</table>
 		</div>
-                <div class="property" id="aria-details">
+		<div class="property" id="aria-description">
+			<pdef>aria-description</pdef>
+			<div class="property-description">
+				<p>Defines a string value that describes or annotates the current element. See related <pref>aria-describedby</pref>.</p>
+				<p>The <pref>aria-description</pref> attribute is similar to the <pref>aria-label</pref> in that both profide a flat string to associate with the element, but a label should be concise, where a description is intended to provide more verbose information.</p>
```
backwardok: small typo:

```suggestion
				<p>The <pref>aria-description</pref> attribute is similar to the <pref>aria-label</pref> in that both provide a flat string to associate with the element, but a label should be concise, where a description is intended to provide more verbose information.</p>
```

index.html
```diff
@@ -10711,7 +10711,47 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>The purpose of <pref>aria-description</pref> is the same as that of <pref>aria-describedby</pref>. It provides the user with additional descriptive text for the object. The most common <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a> mapping for a description is the <a>accessible description</a> property.</p>
+				<p>If the description text is visible on screen, authors SHOULD NOT use <pref>aria-description</pref>.</p>
+				<ul>
+					<li>Authors SHOULD use <pref>aria-details</pref> when the description or annotation is large, complex or has useful semantics or structure. Using <pref>aria-details</pref> will allow assistive technology users to visit the structured content and provide additional navigation commands within the content.</li>
+					<li>Authors SHOULD use <pref>aria-describedby</pref> when the desription or annotation is very small and simple, and is best experienced as a flat string.</li>
```
backwardok: Should this be `aria-description` instead of `aria-describedby`? Or should the information also include that the information is visible? Otherwise it doesn't seem clear enough why you would use `aria-description`.

aleventhal: Hi, this bulleted list is trying to say what to use if the description is visible on screen. In that case, it doesn't make to use aria-description. They should use aria-details or aria-describedy in that case. I've changed the paragraph before to try to make it clearer that the list below is what to use instead, when the text is available in the page.

backwardok: Ah gotcha - update is a lot clearer now, thanks!

index.html
```diff
@@ -10711,7 +10711,47 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Defines a string value that describes or annotates the current element. See related <pref>aria-describedby</pref>.</p>
+				<p>The <pref>aria-description</pref> attribute is similar to the <pref>aria-label</pref> in that both provide a flat string to associate with the element, but a label should be concise, where a description is intended to provide more verbose information.</p>
+				<p>The purpose of <pref>aria-description</pref> is the same as that of <pref>aria-describedby</pref>. It provides the user with additional descriptive text for the object. The most common <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a> mapping for a description is the <a>accessible description</a> property.</p>
+				<p>If the description text is available in the content, authors SHOULD NOT use <pref>aria-description</pref>, but should use one of the following instead:</p>
+				<ul>
+					<li>Authors SHOULD use <pref>aria-describedby</pref> when the related description or annotation elements are hidden, or contain a simple, small description that is best experienced as a flat string, rather than by having the user navigate to them.</li>
```
joanmarie: TL;DR: I think we want to remove "are hidden, or". This is my conclusion after typing out the following:

I find this somewhat confusing for a few reasons:

1. "available in the content" (in the paragraph above) tends to suggest "rendered" -- at least to me. In contrast "when the related description or annotation elements are hidden" tends to suggest not rendered and (in my mind) NOT "available in the content." Perhaps we can tweak the language a tad?

2. assuming "are hidden" means "not rendered", isn't there an excellent chance the element containing the description will be pruned from the accessibility tree? In which case, at least some platform accessibility APIs will wind up with an accessible described-by relationship pointing to a null target. Relationship targets should ideally not be null.

    Granted, this winds up being not a problem for most platforms because user agents are expected to flatten the element referenced by `aria-describedby` and expose it via the platform's accessible description property. As long as ATs prefer that property, there's no need to use the platform relationship. So I guess my main objection here is one of aesthetics. :grinning:

    That said, why would an author want to create hidden / not-rendered elements and then reference them via an ARIA property for the purpose of causing user agents to piece the description together as a flat string? *IF* the description is hidden / not-rendered, wouldn't it be a lot simpler for all parties to just use the new `aria-description`?

3. assuming "are hidden" means "not rendered", then doesn't the text here contradict what we state for `aria-label`/`aria-labelledby`? The spec text states:

    > If the interface is such that it is not possible to have a visible label on the screen, authors SHOULD use aria-label and SHOULD NOT use aria-labelledby.

Taking the above three observations into account, I would think that authors SHOULD use the new `aria-description` property and NOT `aria-describedby` "when the related description or annotation elements are hidden."

Or am I missing something?

index.html
```diff
@@ -11375,7 +11415,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				<tbody>
 					<tr>
 						<th class="property-related-head" scope="row">Related Concepts:</th>
-						<td class="property-related"><code>title</code> attribute in [[HTML]]</td>
+						<td class="property-related"> </td>
```
joanmarie: Good catch!

index.html
aleventhal: I still think this is good text, perhaps instead of "available in the content", we say "available in the DOM".
However, I think the first option of these 2 is slightly preferable:
1.
```
<p id="desc">The quick brown fox jumps over the lazy dog abcdef ..... 100 more words ..."</p>
<input type="text" aria-describedby="desc">
```
2.
```
<p>The quick brown fox jumps over the lazy dog abcdef ..... 100 more words ..."</p>
<input type="text" aria-description=">The quick brown fox jumps over the lazy dog abcdef ..... 100 more words ..."">
```

Benefits of version 2:
- Shorter markup
- No redundant text that could accidentally diverge
- In the hidden case, some user agents will expose a relationship, some will expose no relationship (as opposed to a null one, I haven't seen that happen). Note that Microsoft has been working on bringing hidden nodes into the exposed hierarchy for Chrome.

Interesting point about sounding like it conflicts with the aria-label/aria-labelledby text. The intention doesn't contradict. It isn't saying to put a hidden description in the DOM instead of using aria-description. It's saying that if a description is in the DOM, may as well use it.

Or?

aleventhal: I opted to remove "are hidden, or", rather than copy the text from aria-label. The reason is that in this case there are 3 properties to chose from, and I like the way the bulleted list works.

carmacleod:
Should aria-description be prohibited on generic role?

index.html
```diff
@@ -10711,7 +10711,47 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				</tbody>
 			</table>
 		</div>
-                <div class="property" id="aria-details">
+		<div class="property" id="aria-description">
+			<pdef>aria-description</pdef>
+			<div class="property-description">
+				<p>Defines a string value that describes or annotates the current element. See related <pref>aria-describedby</pref>.</p>
+				<p>The <pref>aria-description</pref> attribute is similar to the <pref>aria-label</pref> in that both provide a flat string to associate with the element, but a label should be concise, where a description is intended to provide more verbose information.</p>
```
```suggestion
				<p>The <pref>aria-description</pref> attribute is similar to the <pref>aria-label</pref> in that both provide a flat string to associate with the element, but a label should be concise, whereas a description is intended to provide more verbose information.</p>
```

Consider using whereas instead of where.

scottaohara:
> Should aria-description be prohibited on generic role?

if so, then likely so should `aria-describedby` and `aria-details` but that's not the case now.  I think this warrants discussion though, but presently I'd think it's reasonable to continue to allow it.

aleventhal:
> Should aria-description be prohibited on generic role?

I'm leaning against it as an unnecessary restriction and I'm not sure what the benefit would be Consider that not everything that has an annotation will be necessarily highlighted, and thus won't have a mark role. There is the possibility that something will have an annotation on a passage of text that only becomes obvious to visual users after you click on it. IMO it could be on a span, etc.

With aria-label/name etc. I think the benefit of prohibiting it on some roles was that it would be possible to accidentally hide the inner contents from the AT. But that's not an issue here.

carmacleod:
Other than "similar to aria-label", nothing really says "this is invisible". Perhaps that should be explicit.

carmacleod:
After merging this, need to update https://w3c.github.io/accname/ to say that aria-describedby overrides aria-description.

index.html
```diff
@@ -10711,7 +10711,47 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				</tbody>
 			</table>
 		</div>
-                <div class="property" id="aria-details">
+		<div class="property" id="aria-description">
+			<pdef>aria-description</pdef>
+			<div class="property-description">
+				<p>Defines a string value that describes or annotates the current element. See related <pref>aria-describedby</pref>.</p>
+				<p>The <pref>aria-description</pref> attribute is similar to <pref>aria-label</pref> in that both provide a flat string to associate with the element, but a label should be concise, whereas a description is intended to provide more verbose information.</p>
+				<p>The purpose of <pref>aria-description</pref> is the same as that of <pref>aria-describedby</pref>. It provides the user with additional descriptive text for the object. The most common <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a> mapping for a description is the <a>accessible description</a> property. User agents must give precedence to aria-describedby over aria-description when computing the accessible description property.</p>
```
```suggestion
				<p>The purpose of <pref>aria-description</pref> is the same as that of <pref>aria-describedby</pref>. It provides the user with additional descriptive text for the object. The most common <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a> mapping for a description is the <a>accessible description</a> property. User agents MUST give precedence to aria-describedby over aria-description when computing the accessible description property.</p>
```

This should probably be a normative MUST?

carmacleod:
@aleventhal Sorry for the back-and-forth, but I think your original phrase before https://github.com/w3c/aria/pull/1137/commits/c7113b67951bfa52f2e2d26ce9936f652db5ea06 was better, i.e:
> If the description text is available in the DOM, ...

(In fact, I think I'll open an issue for aria-label to say "available in the DOM" as well... :)

What I had in mind for visible was adding something like the following (paraphrased from [aria-label](https://w3c.github.io/aria/#aria-label)):
> In cases where providing a visible description is not the desired user experience, authors MAY set the accessible description of the element using aria-description.

----

Here are 2 possible places where that sentence could fit in:
`1.`
> The purpose of aria-description is the same as that of aria-describedby. It provides the user with additional descriptive text for the object. **In cases where providing a visible description is not the desired user experience, authors MAY set the accessible description of the element using aria-description.** The most common accessibility API mapping for a description is the accessible description property. User agents MUST give precedence to aria-describedby over aria-description when computing the accessible description property.
>
> If the description text is available in the DOM, authors SHOULD NOT use aria-description, but should use one of the following instead:

----

`2.`
> The purpose of aria-description is the same as that of aria-describedby. It provides the user with additional descriptive text for the object. The most common accessibility API mapping for a description is the accessible description property. User agents MUST give precedence to aria-describedby over aria-description when computing the accessible description property.
>
> **In cases where providing a visible description is not the desired user experience, authors MAY set the accessible description of the element using aria-description.** If the description text is available in the DOM, authors SHOULD NOT use aria-description, but should use one of the following instead:

aleventhal:
Ok, makes sense. I used the second option, but I added a "However, " to help things flow. What do you think?

aleventhal:
> In addition to my inline comments, I think it would be good to add a statement indicating which takes precedence when both `aria-describedby` and `aria-description` are used. 
This is done:
"User agents MUST give precedence to <pref>aria-describedby</pref> over <pref>aria-description</pref> when computing the accessible description property."

index.html
```diff
@@ -10711,7 +10711,47 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>The <pref>aria-description</pref> attribute is similar to <pref>aria-label</pref> in that both provide a flat string to associate with the element, but a label should be concise, whereas a description is intended to provide more verbose information.</p>
+				<p>The purpose of <pref>aria-description</pref> is the same as that of <pref>aria-describedby</pref>. It provides the user with additional descriptive text for the object. The most common <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a> mapping for a description is the <a>accessible description</a> property. User agents MUST give precedence to <pref>aria-describedby</pref> over <pref>aria-description</pref> when computing the accessible description property.</p>
+				<p>In cases where providing a visible description is not the desired user experience, authors MAY set the accessible description of the element using aria-description. However, if the description text is available in the DOM, authors SHOULD NOT use <pref>aria-description</pref>, but should use one of the following instead:</p>
```
```suggestion
				<p>In cases where providing a visible description is not the desired user experience, authors MAY set the accessible description of the element using <pref>aria-description</pref>. However, if the description text is available in the DOM, authors SHOULD NOT use <pref>aria-description</pref>, but should use one of the following instead:</p>
```

carmacleod:
> I used the second option, but I added a "However, " to help things flow. What do you think?

Looks good!

PR #1142, opened 2019-12-21
prohibit aria-roledescription from role=generic

### LINKED ISSUE #1140: aria-roledescription should not be allowed on role generic
On the call today we decided it was an oversight that `role=generic` will now be allowed to have `aria-roledescription`.

`aria-roledescription` should be prohibited from being used on `role=generic`

scottaohara:
closes #1140, pending answering [comment about note](https://github.com/w3c/aria/issues/1140#issuecomment-568189345)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1142.html" title="Last updated on Mar 4, 2020, 12:48 AM UTC (5c5bd92)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1142/db5a8c5...5c5bd92.html" title="Last updated on Mar 4, 2020, 12:48 AM UTC (5c5bd92)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3360,9 +3360,10 @@
 							<ul>
 								<li><pref>aria-label</pref></li>
 								<li><pref>aria-labelledby</pref></li>
-							</ul>					
+								<li><pref>aria-roledescription</pref></li>
+							</ul>
 						</td>
-					</tr>						
+					</tr>
 					<tr>
 						<th class="role-namefrom-head" scope="row">Name From:</th>
 						<td class="role-namefrom">prohibited</td>
--- a/index.html
+++ b/index.html
@@ -12137,6 +12137,7 @@
 				<p>User agents MUST NOT expose the <code>aria-roledescription</code> property if any of the following conditions exist:</p>
 				<ol>
 					<li>The element to which <code>aria-roledescription</code> is applied does not have a valid WAI-ARIA role or does not have an implicit WAI-ARIA role semantic.</li>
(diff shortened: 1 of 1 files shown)


mtrootyy:
`aria-brailleroledescription` also should be prohibited on role `generic` for the same reason.

pkra:
> `aria-brailleroledescription` also should be prohibited on role `generic` for the same reason.

:+1: (But it's less of a problem since the spec says that UAs must not expose aria-brailleroledescription if there's no valid aria-roledescription.)

scottaohara:
agreed, though still not a bad idea to be explicit :)  making a commit to add this now.

jnurthen:
@scottaohara Can you please remove the commit prohibiting brailleroledescription on generic from this PR. I can't merge that to the 1.2 branch (which is the desire for this) as we won't have brailleroledescription there. We can ask @pkra to add that to #1097

scottaohara:
@jnurthen removed it.

carmacleod:
Or maybe:
> The element to which aria-roledescription is applied has an explicit or implicit WAI-ARIA role that prohibits aria-roledescription.

...and make it the 2nd point?

mcking65:
I think it is important to have this UA restriction language in roledescription before merging:

>The element to which aria-roledescription is applied has an explicit or implicit WAI-ARIA role that prohibits aria-roledescription.

jnurthen:
@scottaohara pushed a change - can you take a look

scottaohara:
i think that looks good, thank you @jnurthen, sorry i didn't get to adding it sooner.

Issue #1117, opened 2019-11-06
Consider prohibiting accessible name for listitem, rowgroup, term, time


zcorpan:
See https://github.com/w3c/aria-practices/pull/1231 , the following roles are categorized as "Do Not Name" in APG:

role | Necessity of Naming | Guidance
-- | -- | --
listitem | Do Not Name | Naming is not supported by assistive technologies; it is necessary to include relevant content within the list item.
rowgroup | Do Not Name | Naming is not supported by assistive technologies.
term | Do Not Name | Since a term is usually the name for the role="definition" element, it could be confusing if the term itself also has a name.
time | Do Not Name | Naming is not supported by assistive technologies.

Although the rationale provided for several of these is that naming isn't supported by ATs, it's not clear if they *should* be supported.

For `time` in particular, I think it's in the same situation as `paragraph`, where naming is prohibited.

> I'm not sure if naming a `time` is ever useful. A `time`, from what I understand, isn't typically focusable, but is more like a normal word within a paragraph. To understand the context, you'd read the full paragraph.

 https://github.com/w3c/aria-practices/pull/1231#discussion_r340288166

I think `listitem` is like `paragraph`. It can have a marker like a bullet or a number, but that's not an accessible name. Are there reasons to name list items that are somehow different from paragraphs?

For `rowgroup` I could imagine it being theoretically helpful to have a name if you navigate between rowgroups in a table or grid, but I'm not sure how ATs can navigate these structures.

`term` seems like `time`, it's usually a word or a phrase within a paragraph.

scottaohara:
definitely agree on `term` and `time`.

I personally think prohibiting naming of `rowgroup` makes sense too, but feel like there could be a use case there I can't presently think of if used within a `grid`.

`listitem` is the one I think would need consensus on, especially since it was specifically _not_ prohibited during 1.2.  See #833 for at least one reference to that.

aleventhal:
Related: `term` -- I know what the spec says, but shouldn't a definition be linked to a term via `aria-describedby`, rather than `aria-labelledby`?

carmacleod:
Note that `term` and `definition` are now covered in https://github.com/w3c/aria/pull/1147.
In that PR, naming is prohibited on both roles.

I opened APG https://github.com/w3c/aria-practices/issues/1299 to change Necessity of Naming to Prohibited for both `term` and `definition`.

joanmarie:
[...]

PR #1147, opened 2020-01-07
term/definition should use aria-details instead of aria-labelledby

aleventhal:
Relates to issue #749 

A definition should not change the accessible name of a term, and thus aria-labelledby is inappropriate. Rather, a definition is an annotation and should use aria-details.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1147.html" title="Last updated on Feb 11, 2020, 11:15 PM UTC (6187596)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1147/f519fca...6187596.html" title="Last updated on Feb 11, 2020, 11:15 PM UTC (6187596)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2567,7 +2567,7 @@
       <rdef>definition</rdef>
       <div class="role-description">
         <p>A definition of a term or concept. See related <rref>term</rref>.</p>
-        <p>Authors SHOULD identify the <a>element</a> being defined and assign that element a role of <rref>term</rref>. In addition, authors SHOULD set <pref>aria-details</pref> on the <rref>term</rref>, pointing to the relevant <code>definition</code>.</p>
+        <p>Authors SHOULD identify the <a>element</a> being defined and assign that element a role of <rref>term</rref>. In addition, authors SHOULD set <pref>aria-details</pref> on the <rref>term</rref>, pointing to the related <code>definition</code>.</p>
       </div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -2567,7 +2567,7 @@
       <rdef>definition</rdef>
       <div class="role-description">
         <p>A definition of a term or concept. See related <rref>term</rref>.</p>
-        <p>Authors SHOULD identify the <a>element</a> being defined and assign that element a role of <rref>term</rref>. In addition, authors SHOULD set <pref>aria-details</pref> on the <rref>term</rref>, pointing to the related <code>definition</code>.</p>
+        <p>Authors MUST identify the <a>element</a> being defined and assign that element a role of <rref>term</rref>.</p>
       </div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
@@ -2615,8 +2615,17 @@
 						<td class="role-inherited">Placeholder</td>
 					</tr>
 					<tr>
+						<th class="role-disallowed-head" scope="row">Prohibited States and Properties:</th>
+						<td class="role-disallowed">
+							<ul>
+								<li><pref>aria-label</pref></li>
+								<li><pref>aria-labelledby</pref></li>
+							</ul>
+					 </td>
+					</tr>
+					<tr>
 						<th class="role-namefrom-head" scope="row">Name From:</th>
-						<td class="role-namefrom">author</td>
+						<td class="role-namefrom">prohibited</td>
 					</tr>
 					<tr>
 						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
@@ -8772,7 +8781,7 @@
 			<rdef>term</rdef>
 			<div class="role-description">
 				<p>A word or phrase with a corresponding definition. See related <rref>definition</rref>.</p>
-				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user. Where a matching <rref>definition</rref> exists, authors SHOULD set <pref>aria-details</pref> on the <code>term</code>, pointing to the <rref>definition</rref>.</p>
+				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user. Where a matching <rref>definition</rref> exists, authors MUST set <pref>aria-details</pref> on the <code>term</code>, pointing to the <rref>definition</rref>.</p>
 				<p>Authors SHOULD NOT use the <code>term</code> role on interactive elements such as links because doing so could prevent users of <a>assistive technologies</a> from interacting with those elements.</p>
 			</div>
 			<table class="role-features">
--- a/index.html
+++ b/index.html
@@ -2567,7 +2567,7 @@
       <rdef>definition</rdef>
       <div class="role-description">
         <p>A definition of a term or concept. See related <rref>term</rref>.</p>
-        <p>Authors SHOULD identify the <a>element</a> being defined and assign that element a role of <rref>term</rref>. In addition, authors should set <pref>aria-details</pref> on the <rref>term</rref>, pointing to the relevant <code>definition</code>.</p></p>
+        <p>Authors SHOULD identify the <a>element</a> being defined and assign that element a role of <rref>term</rref>. In addition, authors SHOULD set <pref>aria-details</pref> on the <rref>term</rref>, pointing to the relevant <code>definition</code>.</p></p>
       </div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
@@ -8772,7 +8772,7 @@
 			<rdef>term</rdef>
 			<div class="role-description">
 				<p>A word or phrase with a corresponding definition. See related <rref>definition</rref>.</p>
-				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user. Where a matching <rref>definition</rref> exists, authors should set <pref>aria-details</pref> on the <code>term</code>, pointing to the <rref>definition</rref>.</p>
+				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user. Where a matching <rref>definition</rref> exists, authors SHOULD set <pref>aria-details</pref> on the <code>term</code>, pointing to the <rref>definition</rref>.</p>
 				<p>Authors SHOULD NOT use the <code>term</code> role on interactive elements such as links because doing so could prevent users of <a>assistive technologies</a> from interacting with those elements.</p>
 			</div>
 			<table class="role-features">
--- a/index.html
+++ b/index.html
@@ -8825,8 +8825,17 @@
 						<td class="role-inherited">Placeholder</td>
 					</tr>
 					<tr>
+						<th class="role-disallowed-head" scope="row">Prohibited States and Properties:</th>
+						<td class="role-disallowed">
+							<ul>
+								<li><pref>aria-label</pref></li>
+								<li><pref>aria-labelledby</pref></li>
+							</ul>
+						</td>
+					</tr>
+					<tr>
 						<th class="role-namefrom-head" scope="row">Name From:</th>
-						<td class="role-namefrom">author</td>
+						<td class="role-namefrom">prohibited</td>
 					</tr>
 					<tr>
 						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
--- a/index.html
+++ b/index.html
@@ -8780,8 +8780,8 @@
 		<div class="role" id="term">
 			<rdef>term</rdef>
 			<div class="role-description">
-				<p>A word or phrase with a corresponding definition. See related <rref>definition</rref>.</p>
-				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user. Where a matching <rref>definition</rref> exists, authors MUST set <pref>aria-details</pref> on the <code>term</code>, pointing to the <rref>definition</rref>.</p>
+				<p>A word or phrase with an optional corresponding definition. See related <rref>definition</rref>.</p>
+				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user. Where a matching <rref>definition</rref> exists, authors MUST set <pref>aria-details</pref> on the <code>term</code>, pointing to the <rref>definition</rref>. Otherwie, <pref>aria-details</pref> MUST point to a form or control that take user input that provides the definition.</p>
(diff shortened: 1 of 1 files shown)


scottaohara:
with this change should we also update [definition to be prohibited from having aria-label / labelledby?](https://github.com/w3c/aria/issues/1117#issuecomment-557564506).

aleventhal:
> with this change should we also update [definition to be prohibited from having aria-label / labelledby?](https://github.com/w3c/aria/issues/1117#issuecomment-557564506).

Yes, @jnurthen , do you think I should do that in this PR?

jnurthen:
> Yes, @jnurthen , do you think I should do that in this PR?

Sounds good to me.

aleventhal:
> with this change should we also update [definition to be prohibited from having aria-label / labelledby?](https://github.com/w3c/aria/issues/1117#issuecomment-557564506).

Done, good catch.

index.html
```diff
@@ -2564,11 +2564,11 @@ <h2>Definition of Roles</h2>
...
 		<div class="role" id="definition">
-			<rdef>definition</rdef>
-			<div class="role-description">
-				<p>A definition of a term or concept. See related <rref>term</rref>.</p>
-				<p>Authors SHOULD identify the <a>element</a> being defined by giving that element a role of <rref>term</rref> and referencing it with the <pref>aria-labelledby</pref> <a>attribute</a> or by making the element with role <rref>term</rref> a descendant of the element with role <code>definition</code>.</p>
-			</div>
+      <rdef>definition</rdef>
+      <div class="role-description">
+        <p>A definition of a term or concept. See related <rref>term</rref>.</p>
+        <p>Authors SHOULD identify the <a>element</a> being defined by giving that element a role of <rref>term</rref> and referencing it with the <pref>aria-details</pref> <a>attribute</a> or by making the element with role <rref>term</rref> a descendant of the element with role <code>definition</code>.</p>
```
scottaohara: I worry this sentence could be interpreted to say authors should put the `aria-details` on the `role=definition` element, rather than on the `role=term` element and have it point to the definition's `id`.

would something like this text work (adjust further as necessary), and do we want to add a quick code snippet?

>Authors SHOULD identify the element being defined by giving it a role of `term`. The `term` should then reference the `definition` with the `aria-details` attribute.  Alternatively, a `term` may be a descendant of the element with role `definition`, to convey its association.

```html
<span role=term aria-details=def>some word</span> 
<span role=definition id=def>some definition...</span>

<!-- or  -->

<span role=definition>
  <span role=term>some word</span> some definition...
</span>
```

aleventhal: I think you're right. I'm updating the spec text for both term and definition. In addition to clarifying, I'm removing the option to use the descendant approach. It really doesn't make any sense to have one be the descendant of the other, and won't lead to a good or consistent AT experience.

index.html
```diff
@@ -2564,11 +2564,11 @@ <h2>Definition of Roles</h2>
...
 		<div class="role" id="definition">
-			<rdef>definition</rdef>
-			<div class="role-description">
-				<p>A definition of a term or concept. See related <rref>term</rref>.</p>
-				<p>Authors SHOULD identify the <a>element</a> being defined by giving that element a role of <rref>term</rref> and referencing it with the <pref>aria-labelledby</pref> <a>attribute</a> or by making the element with role <rref>term</rref> a descendant of the element with role <code>definition</code>.</p>
-			</div>
+      <rdef>definition</rdef>
+      <div class="role-description">
+        <p>A definition of a term or concept. See related <rref>term</rref>.</p>
+        <p>Authors SHOULD identify the <a>element</a> being defined and assign that element a role of <rref>term</rref>. In addition, authors should set <pref>aria-details</pref> on the <rref>term</rref>, pointing to the relevant <code>definition</code>.</p></p>
```
scottaohara: capitalize 'should' in "In addition, authors SHOULD set..."

index.html
```diff
@@ -8772,7 +8772,7 @@ <h5>Presentational Roles Conflict Resolution</h5>
 			<rdef>term</rdef>
 			<div class="role-description">
 				<p>A word or phrase with a corresponding definition. See related <rref>definition</rref>.</p>
-				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user.</p>
+				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user. Where a matching <rref>definition</rref> exists, authors should set <pref>aria-details</pref> on the <code>term</code>, pointing to the <rref>definition</rref>.</p>
```
scottaohara: capitalize 'should' in "...exists, authors SHOULD set..."

aleventhal:
@scottaohara Thanks for catching that. How's it look now?

index.html
```diff
@@ -2564,11 +2564,11 @@ <h2>Definition of Roles</h2>
...
 		<div class="role" id="definition">
-			<rdef>definition</rdef>
-			<div class="role-description">
-				<p>A definition of a term or concept. See related <rref>term</rref>.</p>
-				<p>Authors SHOULD identify the <a>element</a> being defined by giving that element a role of <rref>term</rref> and referencing it with the <pref>aria-labelledby</pref> <a>attribute</a> or by making the element with role <rref>term</rref> a descendant of the element with role <code>definition</code>.</p>
-			</div>
+      <rdef>definition</rdef>
+      <div class="role-description">
+        <p>A definition of a term or concept. See related <rref>term</rref>.</p>
+        <p>Authors SHOULD identify the <a>element</a> being defined and assign that element a role of <rref>term</rref>. In addition, authors SHOULD set <pref>aria-details</pref> on the <rref>term</rref>, pointing to the relevant <code>definition</code>.</p>
```
```suggestion
        <p>Authors SHOULD identify the <a>element</a> being defined and assign that element a role of <rref>term</rref>. In addition, authors SHOULD set <pref>aria-details</pref> on the <rref>term</rref>, pointing to the related <code>definition</code>.</p>
```

Consider using "related" here instead of "relevant", because `aria-details` is a "Relationship attribute" which makes the `term` and `definition` "related". :)

aleventhal:
@carmacleod, that makes sense to me. Let's bring it up at next week's meeting? How about we get this in for now and think about what to do about that?

carmacleod:
> How about we get this in for now and think about what to do about that?

Yep - agreed.

scottaohara:
oh, @aleventhal I actually thought you had updated `definition` prohibit `aria-label` and `aria-labelledby` per our [previous comments](https://github.com/w3c/aria/pull/1147#issuecomment-571715152).  completely glazed over that, and see it's just `term` here.

i think we should prohibit the naming of `definition` here, it was one of the last things we talked about on the call this week.

index.html
```diff
@@ -2564,11 +2564,11 @@ <h2>Definition of Roles</h2>
...
 		<div class="role" id="definition">
-			<rdef>definition</rdef>
-			<div class="role-description">
-				<p>A definition of a term or concept. See related <rref>term</rref>.</p>
-				<p>Authors SHOULD identify the <a>element</a> being defined by giving that element a role of <rref>term</rref> and referencing it with the <pref>aria-labelledby</pref> <a>attribute</a> or by making the element with role <rref>term</rref> a descendant of the element with role <code>definition</code>.</p>
-			</div>
+      <rdef>definition</rdef>
+      <div class="role-description">
+        <p>A definition of a term or concept. See related <rref>term</rref>.</p>
+        <p>Authors SHOULD identify the <a>element</a> being defined and assign that element a role of <rref>term</rref>. In addition, authors SHOULD set <pref>aria-details</pref> on the <rref>term</rref>, pointing to the related <code>definition</code>.</p>
```
jnurthen: This reads like something that should be in the `term` definition not the `definition` definition. Can a `definition` exist without a matching term? I'm unsure from this definition and I think we should make this clear one way or the other

aleventhal: I don't know? I'd say a definition requires a term, although I believe a term could exist without a definition. 

Also, I thought of mentioning the aria-details here just to make sure people saw that and got it right. I was afraid they'd see it needs a term but not go look at that, haha.

Want me to remove the part about aria-details?

aleventhal: Ok:
- I removed the sentence about aria-details and only mention it in term.
- Now using MUST to say that a definition needs a term
- Now using MUST in the description of term to say if it has a matching definition, it MUST use aria-details to point to it.

How's that?

carmacleod: Now the spec for term sounds confused. The first sentence says it has "a corresponding definition". The 2nd sentence says "a definition has been provided by the author or is expected to be provided by the user". Then the 3rd sentence says "Where a matching definition exists, ...".

I'm not sure whether term can/should be standalone? Technically, html dfn can - kinda - be standalone, but it's usually within a paragraph or maybe a dl that contains its definition, and I'm not sure what it would mean semantically to have an undefined term?

aleventhal: I don't think it's more confusing than what we have, which already says there is a definition OR one is expected from the user. But, I do think it's confusing. I've pushed a change that may help clear it up. LMK what you think.

index.html
```diff
@@ -8771,8 +8780,8 @@ <h5>Presentational Roles Conflict Resolution</h5>
 		<div class="role" id="term">
 			<rdef>term</rdef>
 			<div class="role-description">
-				<p>A word or phrase with a corresponding definition. See related <rref>definition</rref>.</p>
-				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user.</p>
+				<p>A word or phrase with an optional corresponding definition. See related <rref>definition</rref>.</p>
+				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user. Where a matching <rref>definition</rref> exists, authors MUST set <pref>aria-details</pref> on the <code>term</code>, pointing to the <rref>definition</rref>. Otherwise, <pref>aria-details</pref> MUST point to a form or control that take user input that provides the definition.</p>
```
scottaohara: regarding:
>MUST point to a form or control that take user input that provides the definition

* that "takes" user input
* could this not also point to a link that takes users to the definition?

i like this idea though.

aleventhal: It could take the user to a link with the definition, but I'm not sure we really have to say that.

```suggestion
				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user. Where a matching <rref>definition</rref> exists, authors MUST set <pref>aria-details</pref> on the <code>term</code>, pointing to the <rref>definition</rref>. Otherwise, <pref>aria-details</pref> MUST point to a form or control that takes user input that provides the definition.</p>
```

aleventhal: I committed something similar, but didn't like repeating the word "that" -- how about "that takes user input to provide the definition", or "that takes user input providing a definition"

index.html
```diff
@@ -8771,8 +8780,8 @@ <h5>Presentational Roles Conflict Resolution</h5>
 		<div class="role" id="term">
 			<rdef>term</rdef>
 			<div class="role-description">
-				<p>A word or phrase with a corresponding definition. See related <rref>definition</rref>.</p>
-				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user.</p>
+				<p>A word or phrase with an optional corresponding definition. See related <rref>definition</rref>.</p>
+				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user. Where a matching <rref>definition</rref> exists, authors MUST set <pref>aria-details</pref> on the <code>term</code>, pointing to the <rref>definition</rref>. Otherwise, <pref>aria-details</pref> MUST point to a form or control that takes user input to provide the definiton</p>
```
```suggestion
				<p>The <code>term</code> role is used to explicitly identify a word or phrase for which a <rref>definition</rref> has been provided by the author or is expected to be provided by the user. Where a matching <rref>definition</rref> exists, authors MUST set <pref>aria-details</pref> on the <code>term</code>, pointing to the <rref>definition</rref>. Otherwise, <pref>aria-details</pref> MUST point to a form or control that takes user input to provide the definition.</p>
```

carmacleod:
@aleventhal 
> Where a matching definition exists, authors MUST set aria-details on the term, pointing to the definition. Otherwise, aria-details MUST point to a form or control that takes user input to provide the definition.

If `aria-details` is required on a `term`, then it should be listed under "Required States and Properties:" in the characteristics table.

I guess that would mean that a `term` cannot be "standalone", i.e. we couldn't have a teacher say, "Here's the list of terms for tomorrow's test.", if those definitions are expected to be written on paper. :)
(contrived example... thanks for your patience as I slowly think this through... ;)

aleventhal:
True, maybe we should leave that as a possibility. What do others think?

aleventhal:
Thinking about this more. What's the user benefit from having term marked up, when there is no definition? Screen readers don't have any special navigation keystrokes, and the fact that it's a term is going to be obvious from context. I would have thought that the benefits of using term come when you can tie the definition to it.

scottaohara:
a role parity argument could be made.  since `dfn` should be exposed as a `term`, there'd be no requirement that they be associated with a `definition`.

that said, i don't think it's a bad idea to strongly suggest that authors do create these associations...but it's not going to be possible to force a consistent pairing of terms and definitions for anyone using the `dfn` element

aleventhal:
So, we change it to SHOULD or MAY?

carmacleod:
SHOULD?

carmacleod:
Actually, I think I like that. Only a couple of nits:
- add the missing `<rref>'s` and `<pref>'s`
- change "is visible" back to "exists"
- use "form control" instead of "form"
- consider changing "an optional corresponding definition" back to "a corresponding definition"
(it seems wrong now to say "optional", since `aria-details` is required. i.e. there's always going to be either an element with `role="definition"`, or a "virtual, user-entered definition" element)

Question: Should we also add a note to `definition` so that authors don't wipe out the semantics of form controls (if any)?
> Authors SHOULD NOT use the `definition` role on interactive elements such as form controls because doing so could prevent users of assistive technologies from interacting with those elements.

aleventhal:
> Actually, I think I like that. Only a couple of nits:
> 
> * add the missing `<rref>'s` and `<pref>'s`

Done, except for the one where the user might enter a definition, since the definition doesn't actually exist yet.

> * change "is visible" back to "exists"

Basically did that, but it was word to have , exists, by itself, so I rearranged the sentence.

> * use "form control" instead of "form"

See if you like what I did for that.

> * consider changing "an optional corresponding definition" back to "a corresponding definition"

I think we've decided it's optional for role parity.

>   (i.e. it seems wrong now to say "optional", since `aria-details` is required. i.e. there SHOULD always be either an element with `role="definition"`, or a "virtual, user-entered definition" element)

I don't think we are requiring aria-details, just saying you should use it if the optional thing is there, or a form at least.

> Question: Should we also add a note to `definition` so that authors don't wipe out the semantics of the form control (if any)?

Done.

carmacleod:
Ok, looks good to me now!
I already officially "approved"... before @jnurthen made us run around the track a couple more times. 😄

PR #1149, opened 2020-01-09
use preferred wording introduced in aria-description PR

carmacleod:
When saying what determines when to use aria-labelledby over aria-label, use:
> If the label text is available in the DOM,

instead of:
> If the label text is visible on screen,

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1149.html" title="Last updated on Jan 16, 2020, 7:58 PM UTC (216449b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1149/1e27601...216449b.html" title="Last updated on Jan 16, 2020, 7:58 PM UTC (216449b)">Diff</a>


index.html
```diff
@@ -11362,7 +11362,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>If the label text is available in the DOM,  authors SHOULD use <pref>aria-labelledby</pref> and SHOULD NOT use <pref>aria-label</pref>. There may be instances where the name of an element cannot be determined programmatically from the content of the element, and there are cases where providing a visible label is not the desired user experience. Most host languages provide an attribute that could be used to name the element (e.g., the <code>title</code> attribute in [[HTML]]), yet this could present a browser tooltip. In the cases where a visible label or visible tooltip is undesirable, authors MAY set the accessible name of the element using <pref>aria-label</pref>. As required by the <a href="#textalternativecomputation">accessible name and description computation</a>, user agents give precedence to <pref>aria-labelledby</pref> over <pref>aria-label</pref> when computing the accessible name property.</p>
```
scottaohara: nit: extra spaces between "DOM," and "authors"
>DOM,  authors

carmacleod: oops - fixed :)

jongund:
Seems to mix "in the DOM" and "visible label: references:

Here is a suggested edit:
If the label text is available in the DOM,  authors SHOULD use <pref>aria-labelledby</pref> and SHOULD NOT use <pref>aria-label</pref>. There may be instances where the name of an element cannot be determined programmatically from the DOM, and there are cases where referencing DOM content is not the desired user experience. Most host languages provide an attribute that could be used to name the element (e.g., the <code>title</code> attribute in [[HTML]]), yet this could present a browser tooltip. In the cases where DOM content or a tooltip is undesirable, authors MAY set the accessible name of the element using <pref>aria-label</pref>. As required by the <a href="#textalternativecomputation">accessible name and description computation</a>, user agents give precedence to <pref>aria-labelledby</pref> over <pref>aria-label</pref> when computing the accessible name property.

carmacleod:
Good catch, @jongund. Thanks for reading the whole paragraph. I've committed your suggested text.

PR #1156, opened 2020-01-16
Add explanation of aria-live usage for aria-errormessage example

### LINKED ISSUE #1139: aria-live says only use assertive when "interruption is imperative". Why does aria-errormessage example use it rather than polite
Aria specification  states
>  Because an interruption may disorient users or cause them to not complete their current task, authors SHOULD NOT use the assertive value unless the interruption is **_imperative_**.

 https://www.w3.org/TR/wai-aria-1.2/#aria-live

However the example code given in the aria-errormessage property uses aria-live=assertive https://www.w3.org/TR/wai-aria-1.2/#aria-errormessage
```<span id="msgID" aria-live="assertive"><span style="visibility:visible">Invalid time:...>```
If aria-live=assertive is "imperative" on an error message, then we should document why.

SOLUTION 1 
Add the sentence after the example:
> ```aria-live="assertive"``` is used in this example instead of ```polite``` because the error message should be read before the user finishes the field and tabs to next field where the previous error message may be confused with the next field's message?

SOLUTION 2
Change the example from ```aria-live=assertive``` to ```polite```

schne324:
resolves #1139

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1156.html" title="Last updated on Jan 16, 2020, 6:21 PM UTC (f7ce26e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1156/18522e3...f7ce26e.html" title="Last updated on Jan 16, 2020, 6:21 PM UTC (f7ce26e)">Diff</a>


schne324:
Ooops! my editor seems to trim trailing whitespace. I will update this pr with only the change described in #1139

schne324:
closing this in favor of my new PR

PR #1157, opened 2020-01-16
Add explanation of aria-live usage for aria-errormessage example

### LINKED ISSUE #1139: aria-live says only use assertive when "interruption is imperative". Why does aria-errormessage example use it rather than polite

schne324:
resolves #1139

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1157.html" title="Last updated on Jan 30, 2020, 10:55 PM UTC (5146bf1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1157/18522e3...5146bf1.html" title="Last updated on Jan 30, 2020, 10:55 PM UTC (5146bf1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -11037,6 +11037,7 @@
 				&lt;input id="startTime" type="text" aria-errormessage="msgID" aria-invalid="true" value="11:30 PM" &gt;
 				&lt;span id="msgID" aria-live="assertive"&gt;&lt;span style="visibility:visible"&gt;Invalid time: the time must be between 9:00 AM and 5:00 PM&lt;/span&gt;&lt;/span&gt;
 				</pre>
+				<p><code>aria-live="assertive"</code> is used in this example in favor of <code>aria-live="polite"</code> to ensure the error message is read before the user is done interacting with the field. Otherwise, the user may confuse the message with that of the next field.</p>
       </div>
 			<table class="property-features">
 				<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -11037,7 +11037,7 @@
 				&lt;input id="startTime" type="text" aria-errormessage="msgID" aria-invalid="true" value="11:30 PM" &gt;
 				&lt;span id="msgID" aria-live="assertive"&gt;&lt;span style="visibility:visible"&gt;Invalid time: the time must be between 9:00 AM and 5:00 PM&lt;/span&gt;&lt;/span&gt;
 				</pre>
-				<p class="note"><code>aria-live="assertive"</code> is used in this example in favor of <code>aria-live="polite"</code> to ensure the error message is read before the user is done interacting with the text input. Otherwise, the user may confuse the message with that of the next text input.</p>
+				<p class="note">This example uses <code>aria-live="assertive"</code> to indicate that assistive technologies should immediately announce the error message rather than completing other queued announcements before announcing the error message. This increases the likelihood that users are aware of the error message before they move focus out of the input.</p>
       </div>
 			<table class="property-features">
 				<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -11037,7 +11037,7 @@
 				&lt;input id="startTime" type="text" aria-errormessage="msgID" aria-invalid="true" value="11:30 PM" &gt;
 				&lt;span id="msgID" aria-live="assertive"&gt;&lt;span style="visibility:visible"&gt;Invalid time: the time must be between 9:00 AM and 5:00 PM&lt;/span&gt;&lt;/span&gt;
 				</pre>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -11037,6 +11037,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				&lt;input id="startTime" type="text" aria-errormessage="msgID" aria-invalid="true" value="11:30 PM" &gt;
 				&lt;span id="msgID" aria-live="assertive"&gt;&lt;span style="visibility:visible"&gt;Invalid time: the time must be between 9:00 AM and 5:00 PM&lt;/span&gt;&lt;/span&gt;
 				</pre>
+				<p><code>aria-live="assertive"</code> is used in this example in favor of <code>aria-live="polite"</code> to ensure the error message is read before the user is done interacting with the field. Otherwise, the user may confuse the message with that of the next field.</p>
```
scottaohara: the content that precedes the code example refers to the form controls as "text inputs", but this new content refers to them as "fields".

I suggest we stick with the previous use of "text input"

jnurthen:
> Aside from my previous comment, this content reads a bit like it should be marked as a note to me. Which is fine imo, but i'm curious if we need to mark it as such?
> 
> If others disagree, then please disregard :)

+1 to a note

index.html
```diff
@@ -11037,6 +11037,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				&lt;input id="startTime" type="text" aria-errormessage="msgID" aria-invalid="true" value="11:30 PM" &gt;
 				&lt;span id="msgID" aria-live="assertive"&gt;&lt;span style="visibility:visible"&gt;Invalid time: the time must be between 9:00 AM and 5:00 PM&lt;/span&gt;&lt;/span&gt;
 				</pre>
+				<p class="note"><code>aria-live="assertive"</code> is used in this example in favor of <code>aria-live="polite"</code> to ensure the error message is read before the user is done interacting with the text input. Otherwise, the user may confuse the message with that of the next text input.</p>
```
mcking65: The phrasing "in favor of" makes it sound like polite is actually favored.

I don't think it is possible to guess so precisely what might happen if polite were used instead of assertive. I would prefer we offer a more general, and potentially easier to understand, rationale for using assertive.

```suggestion
				<p class="note">This example uses <code>aria-live="assertive"</code> to indicate that assistive technologies should immediately announce the error message rather than completing other queued announcements before announcing the error message. This increases the likelihood that users are aware of the error message before they move focus out of the input.</p>
```

index.html
```diff
@@ -11037,6 +11037,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				&lt;input id="startTime" type="text" aria-errormessage="msgID" aria-invalid="true" value="11:30 PM" &gt;
 				&lt;span id="msgID" aria-live="assertive"&gt;&lt;span style="visibility:visible"&gt;Invalid time: the time must be between 9:00 AM and 5:00 PM&lt;/span&gt;&lt;/span&gt;
 				</pre>
+				<p class="note">This example uses <code>aria-live="assertive"</code> to indicate that assistive technologies should immediately announce the error message rather than completing other queued announcements before announcing the error message. This increases the likelihood that users are aware of the error message before they move focus out of the input.</p>
```
carmacleod: The first sentence contains 3 instances of "the error message" and 3 instances of either "announce", "announcing", or "announcements", all of which which gives it a bit of a "tongue-twistery" feel.

May I suggest removing 1 instance of each, as follows? I think it would still have the same meaning:

```suggestion
				<p class="note">This example uses <code>aria-live="assertive"</code> to indicate that assistive technologies should immediately announce the error message rather than completing other queued announcements first. This increases the likelihood that users are aware of the error message before they move focus out of the input.</p>
```

jnurthen:
@schne324 I think this is ready to merge once review comments are resolved

Issue #1033, opened 2019-08-14
Clarify "required owned element"


WilcoFiers:
[5.2.5 Required Owned Elements](https://www.w3.org/TR/wai-aria-1.2/#mustContain):

> Any element that will be owned by the element with this role. For example, an element with the role list will own at least one element with the role group or listitem.

As a non-native English speaker, this first sentence doesn't make sense to me. It doesn't seem to say anything, rather it seems like an unfinished sentence. As though what it should have been is "Any element that will be owned by the element with this role MUST be one of the required owned elements", but the sentence got clipped.

Then in the second sentence it says "will own at least", even though that is followed by a paragraph explaining how there are cases where a list can be empty. Which strongly implies that something like this would be fine:

```html
<div role="list">
  <div role="listitem">List item</div>
  <div> not a list item</div>
</div>
```

This "at least one" is especially confusing. Should this have been all owned elements SHOULD have a role from the required owned elements list?

JAWS-test:
In this context, it should be clarified whether all child elements that do not belong to the required owned elements automatically have role=presenation implicitly or not. If this is always the case, this should be added to the specification when defining required owned. If this is not always the case, it should be explicitly specified for each role whether it is the case or not.

Currently, it is handled differently by the browsers depending on the role and element:

```
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="utf-8">
		<title>required owned element</title>
	</head>
	<body>
		<h1>List</h1>
		<div role=list>
			<input type=checkbox aria-label="checkbox in role=list">
			<div role=listitem>li1</div>
			<div role=listitem>li2</div>
			<h3>heading in role=list</h3>
		</div>

		<ul>
			<input type=checkbox aria-label="checkbox in ul">
			<li>li3</li>
			<li>li4</li>
			<h3>h3 in ul</h3>
		</ul>	

		<h1>Table</h1>
		<div role=table>
			<input type=checkbox aria-label="checkbox in role=table">
			<div role=row><span role=columnheader>ch1</span><span role=columnheader>ch2</span></div>
			<div role=row><span role=cell>c3</span><span role=cell>c4</span></div>
			<h3>heading in role=table</h3>
		</div>

		<table>
			<input type=checkbox aria-label="checkbox in table">
			<tr><th>th1</th><th>th2</th></tr>
			<tr><td>td3</td><td>td4</td></tr>
			<h3>h3 in table</h3>
		</table>	

		<h1>Listbox</h1>
[...]

PR #1162, opened 2020-01-21
Clarify the intent of "required owned elements"

WilcoFiers:
Change is motivated by issue https://github.com/w3c/aria/issues/1033

***

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2FWilcoFiers%2Faria%2Fd936e0121b295b8fc1a67a31b8e4acff168989d6%2Findex.html%3FisPreview%3Dtrue)

```
Navigation timeout of 27838 ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231162.)._
</details>


jongund:
Here is a suggested edit to the change:

The role required for all owned children of a an element with a container role.  For example, an element with the role `list` will own at least one child element with the role `listitem`. Authors MUST ensure at least one of the required roles is either a child element or referenced by aria-owns from the container element.

index.html
```diff
@@ -470,7 +470,7 @@ <h3>Prohibited States and Properties</h3>
 			</section>
 		<section id="mustContain">
 			<h3>Required Owned Elements</h3>
-			<p>Any <a>element</a> that will be <a>owned</a> by the element with this <a>role</a>. For example, an element with the role <rref>list</rref> will own at least one element with the role <rref>listitem</rref>.</p>
+			<p>The <a>role</a> required for all <a>owned</a> children. For example, an element with the role <rref>list</rref> will own at least one element with the role <rref>listitem</rref>. Authors MUST ensure all <a>owned</a> child nodes, and all elements referenced by <pref>aria-owns</pref> from the active element, have at least one of the roles specified in the <em>required owned element</em> for the role of the active element.</p>
```
```suggestion
			<p>The <a>role</a> or roles required for all <a>owned elements</a> of an element with this role. For example, an element with the role <rref>list</rref> will own at least one element with the role <rref>listitem</rref>. Authors MUST ensure that all <a>owned elements</a> have one of the roles specified in the <em>required owned element</em> for this role.</p>
```

Notes:
Other definitions in [5.2 Characteristics of Roles](https://w3c.github.io/aria/#Properties) 
use "this role", "the role" or "a role" to refer to the thing the definition applies to.

We can take advantage of the definition of "owned element" (so that we don't need as many words):
> [Owned Element](https://w3c.github.io/aria/#dfn-owned-element)
> An 'owned element' is any DOM descendant of the element, any element specified as a child via aria-owns, or any DOM descendant of the owned child.

Here's what the above suggested change looks like in prose:
> The role or roles required for all owned elements of an element with this role. For example, an element with the role list will own at least one child element with the role listitem. Authors MUST ensure that all owned elements have one of the roles specified in the required owned elements for this role.

JAWS-test:
I think the new definition for "required owned elements" does not work, because it now requires that all descendants of the element must have one of the specified roles, which is not correct. 

Example:
* for table the required owned elements are row and rowgroup. but these are only the required child elements. cell, columnheader and rowheader are also allowed as descendants. 

My suggestion would be that no reference is made to "owned elements", but only to the child elements (or to the first descendant element with explicit or implicit role, because child elements with role=presentation should be allowed as descendants)

carmacleod:
@JAWS-test Good point. The clause about "or any DOM descendant of the owned child" in the definition of "owned element" creates problems. I will try to come up with another definition of "required owned elements". Thanks for pointing this out.

index.html
carmacleod: Not using this suggestion.

carmacleod:
I think @jongund's suggestion works:
> The role required for all owned children of an element with a container role. For example, an element with the role list will own at least one child element with the role listitem. Authors MUST ensure at least one of the required roles is either a child element or referenced by aria-owns from the container element.

It just needs the `<a>...</a>`, etc. Put back in. I will create a github suggestion using those words.

index.html
carmacleod: Using @jongund's words, with added markup:

```suggestion
			<p>The <a>role</a> required for all <a>owned</a> children of an element with a container role. For example, an element with the role <rref>list</rref> will own at least one child element with the role <rref>listitem</rref>. Authors MUST ensure at least one of the required roles is either a child element or referenced by <pref>aria-owns</pref> from the container element.</p>
```

JAWS-test:
My question would be whether it is possible to be a little more precise at this point:
* Are child elements those of the DOM or the Accessibility Tree? 
In the following example, option is not a DOM child element
```
<div role=listbox>
  <div role=presentation>
    <div role=option>
    </div>
  </div>
</div>
```

* Must all child elements have one of the required owned roles (as with listbox) or may other child elements also be present (as with table)?

index.html
scottaohara: Do we need to define "container role"? Do we need this term? 

This update includes both "container role" and then ends with "container element". 

i'm counting 5 instances of "container role" in the spec now, where those instances could be revised to "container element with role..." or even dropping the word "container".

jnurthen:
this is now overcome by #1454

PR #1163, opened 2020-01-21
Clarify the role of default values

WilcoFiers:
We worked on this in the [ACT rules community group](https://act-rules.github.io/rules/4e8ab6), where we noticed there were some inconsistencies on how AT handled the defaults. It would help to be more explicit about the role of default values.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 27, 2020, 9:03 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2FWilcoFiers%2Faria%2Ff413b6bbdcb4b8ab4f7d1348422f94f95f191ad3%2Findex.html%3FisPreview%3Dtrue)

```
waiting for function failed: timeout 30000ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231163.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -450,7 +450,7 @@
 		</section>
 		<section id="requiredState">
 			<h3>Required States and Properties</h3>
-			<p><a>States</a> and <a>properties</a> specifically required for the <a>role</a> and subclass roles. Content authors MUST provide a non-empty <span>value</span> for required states and properties, except if current value is the default for that state or property for the role. Content authors MUST NOT use the value <code>undefined</code> for required states and properties, unless <code>undefined</code> is an explicitly-supported value of that state or property.</p>
+			<p><a>States</a> and <a>properties</a> specifically required for the <a>role</a> and subclass roles. Content authors MUST provide a non-empty <span>value</span> for required states and properties. Content authors MUST NOT use the value <code>undefined</code> for required states and properties, unless <code>undefined</code> is an explicitly-supported value of that state or property.</p>
 			<p>When an <a>object</a> inherits from multiple ancestors and one ancestor indicates that property is supported while another ancestor indicates that it is required, the property is required in the inheriting object.</p>
 			<p class="note">A host language attribute with the appropriate <a href="#implicit_semantics">implicit WAI-ARIA semantic</a> fulfills this requirement.</p>
 		</section>
--- a/index.html
+++ b/index.html
@@ -450,13 +450,13 @@
 		</section>
 		<section id="requiredState">
 			<h3>Required States and Properties</h3>
-			<p><a>States</a> and <a>properties</a> specifically required for the <a>role</a> and subclass roles. Content authors MUST provide a non-empty <span>value</span> for required states and properties. Content authors MUST NOT use the value <code>undefined</code> for required states and properties, unless <code>undefined</code> is an explicitly-supported value of that state or property.</p>
+			<p><a>States</a> and <a>properties</a> specifically required for the <a>role</a> and subclass roles. Content authors MUST provide a non-empty <span>value</span> for required states and properties, except if current value is the default for that state or property for the role. Content authors MUST NOT use the value <code>undefined</code> for required states and properties, unless <code>undefined</code> is an explicitly-supported value of that state or property.</p>
 			<p>When an <a>object</a> inherits from multiple ancestors and one ancestor indicates that property is supported while another ancestor indicates that it is required, the property is required in the inheriting object.</p>
 			<p class="note">A host language attribute with the appropriate <a href="#implicit_semantics">implicit WAI-ARIA semantic</a> fulfills this requirement.</p>
 		</section>
 		<section id="supportedState">
 			<h3>Supported States and Properties</h3>
-			<p><a>States</a> and <a>properties</a> specifically applicable to the <a>role</a> and child roles. <a>User agents</a> MUST map all supported states and properties for the role to an accessibility API. Content authors MAY provide <span>values</span> for supported states and properties, but need not in some cases where default values are sufficient.</p>
+			<p><a>States</a> and <a>properties</a> specifically applicable to the <a>role</a> and child roles. <a>User agents</a> MUST map all supported states and properties for the role to an accessibility API. If the state or property is undefined, and it has a default value for the role, <a>User agents</a> MUST use the default value. Content authors MAY provide <span>values</span> for supported states and properties, but need not in some cases where default values are sufficient.</p>
 			<p class="note">A host language attribute with the appropriate <a href="#implicit_semantics">implicit WAI-ARIA semantic</a> fulfills this requirement.</p>
 		</section>
 		<section id="inheritedattributes">
--- a/index.html
+++ b/index.html
@@ -456,7 +456,7 @@
 		</section>
 		<section id="supportedState">
 			<h3>Supported States and Properties</h3>
-			<p><a>States</a> and <a>properties</a> specifically applicable to the <a>role</a> and child roles. <a>User agents</a> MUST map all supported states and properties for the role to an accessibility API. If the state or property is undefined, and it has a default value for the role, <a>User agents</a> MUST use the default value. Content authors MAY provide <span>values</span> for supported states and properties, but need not in some cases where default values are sufficient.</p>
+			<p><a>States</a> and <a>properties</a> specifically applicable to the <a>role</a> and child roles. <a>User agents</a> MUST map all supported states and properties for the role to an accessibility API. If the state or property is undefined, and it has a default value for the role, <a>user agents</a> MUST use the default value. Content authors MAY provide <span>values</span> for supported states and properties, but need not in some cases where default values are sufficient.</p>
 			<p class="note">A host language attribute with the appropriate <a href="#implicit_semantics">implicit WAI-ARIA semantic</a> fulfills this requirement.</p>
 		</section>
 		<section id="inheritedattributes">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -450,13 +450,13 @@ <h3>Abstract Roles</h3>
 		</section>
 		<section id="requiredState">
 			<h3>Required States and Properties</h3>
-			<p><a>States</a> and <a>properties</a> specifically required for the <a>role</a> and subclass roles. Content authors MUST provide a non-empty <span>value</span> for required states and properties. Content authors MUST NOT use the value <code>undefined</code> for required states and properties, unless <code>undefined</code> is an explicitly-supported value of that state or property.</p>
+			<p><a>States</a> and <a>properties</a> specifically required for the <a>role</a> and subclass roles. Content authors MUST provide a non-empty <span>value</span> for required states and properties, except if current value is the default for that state or property for the role. Content authors MUST NOT use the value <code>undefined</code> for required states and properties, unless <code>undefined</code> is an explicitly-supported value of that state or property.</p>
```
jnurthen: There should be no required states and properties with default values so we don't need this addition. (although saying that I see we missed one in our sweep of these which needs fixing https://github.com/w3c/aria/issues/700 )

index.html
```diff
@@ -450,13 +450,13 @@ <h3>Abstract Roles</h3>
...
 			<p class="note">A host language attribute with the appropriate <a href="#implicit_semantics">implicit WAI-ARIA semantic</a> fulfills this requirement.</p>
 		</section>
 		<section id="supportedState">
 			<h3>Supported States and Properties</h3>
-			<p><a>States</a> and <a>properties</a> specifically applicable to the <a>role</a> and child roles. <a>User agents</a> MUST map all supported states and properties for the role to an accessibility API. Content authors MAY provide <span>values</span> for supported states and properties, but need not in some cases where default values are sufficient.</p>
+			<p><a>States</a> and <a>properties</a> specifically applicable to the <a>role</a> and child roles. <a>User agents</a> MUST map all supported states and properties for the role to an accessibility API. If the state or property is undefined, and it has a default value for the role, <a>User agents</a> MUST use the default value. Content authors MAY provide <span>values</span> for supported states and properties, but need not in some cases where default values are sufficient.</p>
```
jnurthen: This sentence needs combining with the subsequent sentence. I'm not sure how to read the 2 together.

carmacleod: Perhaps the last sentence belongs after the definition clause? i.e.
```
<p><a>States</a> and <a>properties</a> specifically applicable to the <a>role</a> and child roles. Content authors MAY provide <span>values</span> for supported states and properties, but need not in some cases where default values are sufficient. <a>User agents</a> MUST map all supported states and properties for the role to an accessibility API. If the state or property is undefined, and it has a default value for the role, <a>User agents</a> MUST use the default value.</p>
```

carmacleod: Easier to read like this:
> States and properties specifically applicable to the role and child roles. Content authors MAY provide values for supported states and properties, but need not in some cases where default values are sufficient. User agents MUST map all supported states and properties for the role to an accessibility API. If the state or property is undefined, and it has a default value for the role, User agents MUST use the default value.

charmarkk: Should the second instance of "user agents", the one after "role [comma]", be lowercase? That aside, this looks good to me too.

```suggestion
			<p><a>States</a> and <a>properties</a> specifically applicable to the <a>role</a> and child roles. <a>User agents</a> MUST map all supported states and properties for the role to an accessibility API. If the state or property is undefined, and it has a default value for the role, <a>user agents</a> MUST use the default value. Content authors MAY provide <span>values</span> for supported states and properties, but need not in some cases where default values are sufficient.</p>
```

index.html
```diff
@@ -450,13 +450,13 @@ <h3>Abstract Roles</h3>
...
 			<p class="note">A host language attribute with the appropriate <a href="#implicit_semantics">implicit WAI-ARIA semantic</a> fulfills this requirement.</p>
 		</section>
 		<section id="supportedState">
 			<h3>Supported States and Properties</h3>
-			<p><a>States</a> and <a>properties</a> specifically applicable to the <a>role</a> and child roles. <a>User agents</a> MUST map all supported states and properties for the role to an accessibility API. Content authors MAY provide <span>values</span> for supported states and properties, but need not in some cases where default values are sufficient.</p>
+			<p><a>States</a> and <a>properties</a> specifically applicable to the <a>role</a> and child roles. <a>User agents</a> MUST map all supported states and properties for the role to an accessibility API. If the state or property is undefined, and it has a default value for the role, <a>user agents</a> MUST use the default value. Content authors MAY provide <span>values</span> for supported states and properties, but need not in some cases where default values are sufficient.</p>
```
joanmarie: Editorial nit: "use the default value" perhaps should be "expose the default value" or something that clarifies what "use" means.

Non-editorial issue: This introduces a normative change and because it's a MUST, we will need tests and at least two passing implementations -- possibly for all cases where a state or property has a default value....

Possible solution: Changing "MUST" to "SHOULD" means we're accomplishing the "clarify" part by stating what our expectations are. If we truly want it to be a MUST, then perhaps do that in 1.3?

index.html
carmacleod:
```suggestion
			<p><a>States</a> and <a>properties</a> specifically required for the <a>role</a> and subclass roles. Content authors MUST provide a non-empty <span>value</span> for required states and properties. Content authors MUST NOT use the value <code>undefined</code> for required states and properties, unless <code>undefined</code> is an explicitly-supported value of that state or property.</p>
```

index.html
carmacleod: This reorders the sentences to (hopefully) address @jnurthen's comment in https://github.com/w3c/aria/pull/1163#discussion_r369301194 so that:
1. _first_, authors are told that they may set a value,
2. _then_ user agents are told that (in addition to mapping the attr to an api), if the authors don't set a value (i.e. it's undefined), they should expose the default value (if there is one for the role)

Also, addresses @joanmarie's concern about MUST by using SHOULD. (We can discuss whether or not to make this a normative requirement in 1.3.) Also, this changes "use the default value" to "expose the default value" per joanie's request.

I also deleted "some" from "in some cases" because I like deleting words. ;)

```suggestion
			<p><a>States</a> and <a>properties</a> specifically applicable to the <a>role</a> and child roles. Content authors MAY provide <span>values</span> for supported states and properties, but need not in cases where default values are sufficient. <a>User agents</a> MUST map all supported states and properties for the role to an accessibility API. If the state or property is undefined and it has a default value for the role, <a>user agents</a> SHOULD expose the default value.</p>
```

carmacleod:
@jnurthen This PR is ready for review, but I have no idea why the build is failing (500 Internal Server Error). So, since pr-preview won't run, here is a [hacked-together preview link of the Supported States and Properties role characteristics paragraph](http://htmlpreview.github.io/?https://github.com/w3c/aria/blob/f413b6bbdcb4b8ab4f7d1348422f94f95f191ad3/index.html#supportedState) that doesn't have the respec stuff added, but should still be sufficient for reviewing.

carmacleod:
build failure may be due to w3c/spec-generator#284

carmacleod:
@jnurthen Build passed. PR-preview is still failing, but this is a small change that can be reviewed just by looking at the code: https://github.com/w3c/aria/pull/1163/files

@joanmarie's review comments were addressed (see [discussion in resolved section](https://github.com/w3c/aria/pull/1163#discussion_r392252590)).

So this should be good to go now!

jnurthen:
> @jnurthen Build passed. PR-preview is still failing, but this is a small change that can be reviewed just by looking at the code: https://github.com/w3c/aria/pull/1163/files

Pr-preview actually seems to work... no idea why there is still an error
https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fw3c%2Faria%2Fpull%2F1163%2F18522e3.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2FWilcoFiers%2Faria%2Fpull%2F1163.html

carmacleod:
That seems to be just the "Diff" link, not the plain "Preview" link.

index.html
joanmarie: I think the "User agents MUST map all supported states and properties" is not necessary and might not be desirable:

* User agents already need to expose supported states and properties as stated in the Core-AAM. Thus saying in the ARIA spec that they need to do so doesn't seem to buy us anything.
* There are instances in the Core-AAM where the mapping is "not mapped" because the correct "mapping" is the absence of exposing anything. For instance if `aria-haspopup` is set to `false`, you don't map on most platforms because platforms don't have a property or state which means "this thing doesn't have a popup." So In this particular case, what are we telling user agents to do when we say they "MUST map all supported states and properties?"
* If you add a "user agents MUST", I believe we need to add tests. And this particular "user agents MUST" seems pretty broad. We have a lot of roles and a lot of states and properties. And now we're talking about the combination of "all supported states and properties" for (I assume) all roles which have supported states and properties.

That said, you would not have included it if you didn't think it accomplished something. So please explain.

joanmarie:
New inline comment and I unresolved the associated comment. I'm happy we changed the user agents MUST I pointed out to a SHOULD. But I question whether or not we want to add the new user agents MUST you subsequently added.

Sorry!

PR #1180, opened 2020-01-29
clarify that modal requires an accessible name

WilcoFiers:
The use of the word "label" is confusing, and dialogs require an accessible name.
That is a MUST, not a SHOULD.

***
<a href="https://pr-preview.s3.amazonaws.com/WilcoFiers/aria/pull/1180.html" title="Last updated on Jan 29, 2020, 8:24 AM UTC (d6efd55)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1180/6ec58d5...WilcoFiers:d6efd55.html" title="Last updated on Jan 29, 2020, 8:24 AM UTC (d6efd55)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2726,7 +2726,7 @@
 			<div class="role-description">
 				<p>A dialog is a descendant window of the primary window of a web application. For <abbr title="Hypertext Markup Language">HTML</abbr> pages, the primary application window is the entire web document, i.e., the <code>body</code> element.</p>
 				<p>Dialogs are most often used to prompt the user to enter or respond to information. A dialog that is designed to interrupt workflow is usually modal. See related <rref>alertdialog</rref>.</p>
-<p>Authors SHOULD provide a dialog label, which can be done with the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attribute.</p>
+				<p>Authors MUST provide an accessible name for a dialog, which can be done with the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attribute.</p>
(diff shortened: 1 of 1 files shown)


jnurthen:
Agree. We should either make this change or remove this sentence entirely as I don't believe it adds anything over and above the accessible name required: true and name from: author. 

Authoring guidance like this is better in the Authoring Practices (where this is already listed)

carmacleod:
I like this change, and I would prefer that the sentence stay in the spec prose for dialog.

I agree that it doesn't specify anything over and above accessible name required: true and name from: author, however it's surprising how many authors just don't get this.

I would almost go one further and add another sentence that says, "If the dialog contains visible content that serves as a description, authors SHOULD use `aria-describedby` on the dialog element to point to that content."

carmacleod:
Come to think of it, it might even be a good idea to have a sentence pointing out `aria-modal`, because it is buried in the list of inherited properties and there's no indication in the prose that it just might be really important... perhaps something like:
> If the dialog is modal, then authors _MUST_ set the dialog's `aria-modal` to `true`.

scottaohara:
I'm a bit wary of those additional suggestions @carmacleod.  IMO those definitely belong over in the guidance, as I can think of situations where I very well wouldn't agree to either of those.

carmacleod:
> I can think of situations where I very well wouldn't agree to either of those.

Oh! ... what are the situations?

carmacleod:
This PR is good to go.
+1

I'll open another issue to consider adding more about aria-describedby and aria-modal to the dialog prose. :)

PR #1188, opened 2020-02-03
Add Aria-repeated

### LINKED ISSUE #1044: Create property for repeated content
This is related to https://github.com/w3c/dpub-aria/issues/10 and is an idea which @jnurthen mentioned to me. Filing it here so we don't lose it.

The basic idea is this: There may be content which is repeated for visual purposes which a screen reader user may or may not wish to have presented every time it is encountered. One example is running headers and footers in documents with a print-layout presentation. But there are undoubtedly others.

If we create a boolean property a la `aria-repeatedcontent`, then screen readers could add an option to present (or not) repeated content.

aleventhal:
closes  #1044

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1188.html#aria-repeated" title="Last updated on Feb 7, 2020, 2:43 AM UTC (2dd3073)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1188/6ec58d5...2dd3073.html" title="Last updated on Feb 7, 2020, 2:43 AM UTC (2dd3073)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -12191,14 +12191,15 @@
 			<div class="property-description">
 				<p>Indicates content is repeated for visual purposes.</p>
 				<p>The <code>aria-repeated</code> helps indicate that content is repeated, and that presenting the repeated content to assistive technology users could produce underesirable redundancy in the output. Unlike <pref>aria-hidden</pref>, <code>aria-repeated</code> content may be presented by screen readers to users when requested by the user.</p>
-				<p>Authors SHOULD NOT use <code>aria-repeated</code> on the very first instance of the repeated content, and SHOULD use <code>aria-repeated</code> on subsequent redundant instances of the same content. Setting <code>aria-repeated</code> on a subtree indicates the entire subtree is redundant, and it cannot be reversed within that subtree by setting <code>aria-repeated="false"</code> on a descendant.</p>
+				<p>Authors SHOULD only set <code>aria-repeated</code> to <code>true</code> on redundant instances of content because assistive technologies MAY skip
+			  over these instances during presentation to the user. Setting <code>aria-repeated</code> on a subtree indicates the entire subtree is redundant. If <code>aria-repeated</code> is true for a subtree, it cannot be reversed within that subtree by setting <code>aria-repeated</code> to false on a descendant.</p>
  				<p>Here are some examples of repeated content:</p>
 				<ul>
 					<li>Running headers and footer sections, repeated on each page</li>
 					<li>Repeated table header rows</li>
 					<li>Pull quote</li>
 				</ul>
-				<p>Because the repeated content may be important in some scenarios, e.g. during proofreading, assistive technologies MAY present options so that screen readers users can choose whether or not to filter <code>aria-repeated</code> content out of the current view.</p>
+				<p>Because the repeated content may be important in some scenarios, e.g. during proofreading, assistive technologies MAY provide options so that screen readers users can choose whether or not to filter <code>aria-repeated</code> content out of the current view.</p>
 			</div>
 			<table class="property-features">
 				<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -11318,7 +11318,7 @@
 				<tbody>
 					<tr>
 						<th class="state-related-head" scope="row">Related Concepts:</th>
-						<td class="state-related"> </td>
+						<td class="state-related">aria-repeated</td>
 					</tr>
 					<tr>
 						<th class="state-applicability-head" scope="row">Used in Roles:</th>
@@ -12190,7 +12190,7 @@
 			<pdef>aria-repeated</pdef>
 			<div class="property-description">
 				<p>Indicates content is repeated for visual purposes.</p>
-				<p>The <code>aria-repeated</code> helps indicate that content is repeated for visual purposes, and that presenting the repeated content to assistive technology users could produce underesirable redundancy in the output.</p>
+				<p>The <code>aria-repeated</code> helps indicate that content is repeated, and that presenting the repeated content to assistive technology users could produce underesirable redundancy in the output. Unlike <pref>aria-hidden</pref>, <code>aria-repeated</code> content may be presented by screen readers to users when requested by the user.</p>
 				<p>Authors SHOULD NOT use <code>aria-repeated</code> on the very first instance of the repeated content, and SHOULD use <code>aria-repeated</code> on subsequent redundant instances of the same content. Setting <code>aria-repeated</code> on a subtree indicates the entire subtree is redundant, and it cannot be reversed within that subtree by setting <code>aria-repeated="false"</code> on a descendant.</p>
  				<p>Here are some examples of repeated content:</p>
 				<ul>
@@ -12211,7 +12211,7 @@
 				<tbody>
 					<tr>
 						<th class="property-related-head" scope="row">Related Concepts:</th>
-						<td class="property-related"> </td>
+						<td class="property-related">aria-hidden</td>
 					</tr>
 					<tr>
 						<th class="property-applicability-head" scope="row">Used in Roles:</th>
--- a/index.html
+++ b/index.html
@@ -3742,7 +3742,7 @@
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
-						<td class="role-properties"> </td>
+            <td class="role-properties"><pref>aria-repeated</pref></td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
@@ -5845,7 +5845,7 @@
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
-						<td class="role-properties"> </td>
+            <td class="role-properties"><pref>aria-repeated</pref></td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
@@ -6735,7 +6735,7 @@
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
-						<td class="role-properties"> </td>
+            <td class="role-properties"><pref>aria-repeated</pref></td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
@@ -6917,6 +6917,7 @@
 								<li><sref>aria-expanded</sref></li>
 								<li><pref>aria-level</pref></li>
 								<li><pref>aria-posinset</pref></li>
+								<li><pref>aria-repeated</pref></li>
 								<li><pref>aria-rowindex</pref></li>
 								<li><pref>aria-rowindextext</pref></li>
 								<li><pref>aria-setsize</pref></li>
@@ -7014,7 +7015,7 @@
 					</tr>
 					<tr>
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
-						<td class="role-properties"> </td>
+            <td class="role-properties"><pref>aria-repeated</pref></td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -13471,6 +13535,7 @@ <h2>Substantive changes targetted for the 1.3 release</h2>
 		<h2>Substantive changes since the last public working draft</h2>
 		<ul>
 			<!-- EdNote: After each WD publish, move contents of this list into the next one below. -->
+			<li>03-Feb-2020: Add <pref>aria-repeated</pref>.</li>
```
joanmarie: Thanks for adding the ChangeLog entry! Based on experience, this tends to result in merge conflicts because things don't land right away and/or in the expected order. Therefore, James and I typically do the ChangeLog after. (I'm not saying don't include it; merely giving you the option.)

joanmarie:
Looks like my other comments got lost :( I'll rewrite.

index.html
```diff
@@ -12185,6 +12186,67 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="property" id="aria-repeated">
+			<pdef>aria-repeated</pdef>
+			<div class="property-description">
+				<p>Indicates content is repeated for visual purposes.</p>
+				<p>The <code>aria-repeated</code> helps indicate that content is repeated, and that presenting the repeated content to assistive technology users could produce underesirable redundancy in the output. Unlike <pref>aria-hidden</pref>, <code>aria-repeated</code> content may be presented by screen readers to users when requested by the user.</p>
+				<p>Authors SHOULD NOT use <code>aria-repeated</code> on the very first instance of the repeated content, and SHOULD use <code>aria-repeated</code> on subsequent redundant instances of the same content. Setting <code>aria-repeated</code> on a subtree indicates the entire subtree is redundant, and it cannot be reversed within that subtree by setting <code>aria-repeated="false"</code> on a descendant.</p>
```
joanmarie: What if the very first instance of the repeated content is a pull quote which, for aesthetic reasons, appears first?

In addition, I assume that authors MAY use <code>aria-repeated</code> on the very first instance of the repeated content -- IF they set it to `false`.

Therefore what about something in the ballpark of: 

      Authors SHOULD only set <code>aria-repeated</code> to <code>true</code>
      on redundant instances of content because assistive technologies MAY skip
      over these instances during presentation to the user.

index.html
```diff
@@ -12185,6 +12186,67 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+ 				<p>Here are some examples of repeated content:</p>
+				<ul>
+					<li>Running headers and footer sections, repeated on each page</li>
+					<li>Repeated table header rows</li>
+					<li>Pull quote</li>
+				</ul>
+				<p>Because the repeated content may be important in some scenarios, e.g. during proofreading, assistive technologies MAY present options so that screen readers users can choose whether or not to filter <code>aria-repeated</code> content out of the current view.</p>
```
joanmarie: nit: s/present options/provide options/
nit: s/screen readers users/screen reader users/

aleventhal: Will change

index.html
aleventhal: Ok, I took this but wanted to keep the part about aria-repeated=false.
Could we actually disallow the false value completely? I don't see how it's useful.

index.html
```diff
@@ -12185,6 +12186,68 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="property" id="aria-repeated">
+			<pdef>aria-repeated</pdef>
+			<div class="property-description">
+				<p>Indicates content is repeated for visual purposes.</p>
+				<p>The <code>aria-repeated</code> helps indicate that content is repeated, and that presenting the repeated content to assistive technology users could produce underesirable redundancy in the output. Unlike <pref>aria-hidden</pref>, <code>aria-repeated</code> content may be presented by screen readers to users when requested by the user.</p>
```
joanmarie: Nit: "The aria-repeated *property* helps..." (Or delete the "The")

index.html
```diff
@@ -12185,6 +12186,68 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Authors SHOULD only set <code>aria-repeated</code> to <code>true</code> on redundant instances of content because assistive technologies MAY skip
+			  over these instances during presentation to the user. Setting <code>aria-repeated</code> on a subtree indicates the entire subtree is redundant. If <code>aria-repeated</code> is true for a subtree, it cannot be reversed within that subtree by setting <code>aria-repeated</code> to false on a descendant.</p>
```
joanmarie: Nit: `code` tags for "true" and "false" since you're using them as values.

index.html
```diff
@@ -12185,6 +12186,68 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+ 				<p>Here are some examples of repeated content:</p>
+				<ul>
+					<li>Running headers and footer sections, repeated on each page</li>
+					<li>Repeated table header rows</li>
+					<li>Pull quote</li>
+				</ul>
+				<p>Because the repeated content may be important in some scenarios, e.g. during proofreading, assistive technologies MAY provide options so that screen readers users can choose whether or not to filter <code>aria-repeated</code> content out of the current view.</p>
```
joanmarie: s/screen readers users/screen reader users/

index.html
joanmarie: I'm good with this change, but I'm not resolving the conversation yet because you ask about disallowing the `false` value completely. That might be a discussion worth having with the group.

I personally think disallowing `false` on a boolean would be weird. So either `false` is the default value that no one uses because why bother? Or `aria-repeated` shouldn't be a boolean. IMHO.

JAWS-test: If I have a submit button at the beginning and one at the end of a form, wouldn't it be better to use `aria-repeated=true` on the first submit button?

JAWS-test:
What if area with `aria-repeated=true` contains interactive elements? These should always be output correctly when navigating with the tab key, right? Skipping the areas should only be done during linear reading.  This may need to be added.

index.html
carmacleod: > content **may** be presented by screen readers 

If it's a normative "may" (which is what I think you meant?) then it needs to be MAY.
If not, then please change it to something like:
> content can be presented by screen readers

index.html
carmacleod: > repeated content **may** be important 

This instance of lowercase 'may' isn't normative, but we have recently decided to try not to use lowercase 'may', 'must' or 'should' in the prose just in case it confuses anyone. (There's usually a decent alternative word that doesn't mess up the meaning). So the above could be rewritten as either:
> repeated content might be important 

or:
> repeated content can be important

index.html
```suggestion
				<p>The <code>aria-repeated</code> helps indicate that content is repeated, and that presenting the repeated content to assistive technology users could produce undesirable redundancy in the output. Unlike <pref>aria-hidden</pref>, <code>aria-repeated</code> content may be presented by screen readers to users when requested by the user.</p>
```

index.html
```diff
@@ -12185,6 +12186,68 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+		<div class="property" id="aria-repeated">
+			<pdef>aria-repeated</pdef>
+			<div class="property-description">
+				<p>Indicates content is repeated for visual purposes.</p>
+				<p>The <code>aria-repeated</code> helps indicate that content is repeated, and that presenting the repeated content to assistive technology users could produce underesirable redundancy in the output. Unlike <pref>aria-hidden</pref>, <code>aria-repeated</code> content may be presented by screen readers to users when requested by the user.</p>
+				<p>Authors SHOULD only set <code>aria-repeated</code> to <code>true</code> on redundant instances of content because assistive technologies MAY skip
+			  over these instances during presentation to the user. Setting <code>aria-repeated</code> on a subtree indicates the entire subtree is redundant. If <code>aria-repeated</code> is true for a subtree, it cannot be reversed within that subtree by setting <code>aria-repeated</code> to false on a descendant.</p>
+ 				<p>Here are some examples of repeated content:</p>
+				<ul>
+					<li>Running headers and footer sections, repeated on each page</li>
```
```suggestion
					<li>Running header and footer sections, repeated on each page</li>
```

carmacleod:
Should `complementary` role support `aria-repeated` as well?
The [doc-pullquote example in DPUB-ARIA](https://w3c.github.io/dpub-aria/#doc-pullquote) uses an html `aside` to contain the repeated quote.

index.html
mcking65: Probably should tell authors not to apply it to elements that contain interactive content. I would not want my screen reader to hide either submit button regardless of what repeated is set to.

mcking65:
I have quite a few editorial comments, but it is hard to figure out what they ought to be with all the unresolved suggestions. So, I'll wait on the word-smithing.

Other concerns:
1. I'm concerned about it being applied to interactive content. Clearly, a link in a footnote is not a concern, but buttons in forms, etc. could be problematic. We need some kind of author SHOULD NOT.
2. Which roles should support it? Seems like group and generic would be naturals, but not region. If regions are being repeated, then I'd suggest that the region role is not being used correctly.

carmacleod:
> If regions are being repeated, then I'd suggest that the region role is not being used correctly.

Good point @mcking65. Ignore my previous question about whether complementary role should support aria-repeated - it shouldn't. If a user doesn't want to hear repeated content, then they don't want it to be a landmark, either.

I guess the [DPUB-ARIA doc-pullquote example](https://w3c.github.io/dpub-aria/#example-39) should probably use a div instead of an aside.

Hmmm, that example currently uses aria-hidden to hide the repeated aside from AT. I wonder what it means to use aria-hidden on a landmark? They seem to be in conflict. Do AT respect the landmark, or the aria-hidden?

mcking65:
@carmacleod wrote:

> Hmmm, that example currently uses aria-hidden to hide the repeated aside from AT. I wonder what it means to use aria-hidden on a landmark? They seem to be in conflict. Do AT respect the landmark, or the aria-hidden?

Nothing overrides aria-hidden. However, there are the unspecified author error recovery behaviors where focusability overrides naming, not readability, in some browsers that we discussed today.

mcking65:
I wonder if we have the right name on this feature. Is this really limited to completely redundant content?

The use cases all seem related to enabling some kind of reader mode. The objective seems to be to indicate that certain content should be ignored when in that kind of mode.

What if it were ... 

>aria-skipwhenreading: Indicates content that assistive technologies may ignore when in a continuous or simplified reading mode.

Some assistive technologies have modes that enable users to read continuously without having to continuously interact with their device. Similarly, some browsers have reader views that focus the user on the primary content. This property enables authors to indicate which content is safe to omit when in such a mode or view. Examples of such content are:
1. Page headers and footers in a book
2. Pull quotes
3 ...

jongund:
Maybe we need something more like `aria-sameas` with a reference to what it is the same as.   More likely that the author would not abuse and screen readers would have more information.

jnurthen:
<joke>
I still like `aria-hidden='🤷‍♀️'` 
</joke>

aleventhal:
I want to wait on resolving suggestions until we all talk about it during our weekly meeting.

aleventhal:
> Maybe we need something more like `aria-sameas` with a reference to what it is the same as. More likely that the author would not abuse and screen readers would have more information.

This is interesting

aleventhal:
> I wonder if we have the right name on this feature. Is this really limited to completely redundant content?

This might be my favorite.

jongund:
I think we want to be a little bit careful about giving authors the ability to tell ATs to skip  content.  I think ARIA should be more descriptive rather than prescriptive.  Let's provide information about the content and give the ATs good information to improve user experience.

aleventhal:
> I think we want to be a little bit careful about giving authors the ability to tell ATs to skip content. I think ARIA should be more descriptive rather than prescriptive. Let's provide information about the content and give the ATs good information to improve user experience.

I like that sentiment. I'm not sure if requiring us to be able to point to the id of the redundant content will really help:
1) In docs, the entire document is not always loaded, so the original running header may not be available to point to
2) Authors could just use a fake invisible offscreen object to point to
3) Authors can always use aria-hidden if they really want to have things not be read

At least with this proposed property (whatever we call it), the screen reader will be able to provide the user an option to read it anyway.

jnurthen:
@aleventhal is this still needed? If not please close it.

aleventhal:
I don't need it now that we have doc-pageheader, doc-pagefooter and doc-pagenum.
There were some use cases for this, but there were also some concerns about author misuse.

PR #1198, opened 2020-02-13
Deprecate globals

### LINKED ISSUE #1164: Deprecate global attributes instead of removing them
ARIA 1.2 is proposing to remove the following global attributes:

- aria-disabled
- aria-errormessage
- aria-haspopup
- aria-invalid

While I understand the need for such a change, I do not think removing them is the right way to go here. The web is a slow developing platforms. Removing features from a specification isn't going to stop anyone from using them. Browsers, assistive tech, authoring tools and validators will have to support these attributes for years to come. Taking attributes out, especially global attributes, is too abrupt a change in my opinion.

I think these attributes need to be deprecated instead or removed. Deprecating them makes for much greater transparency. I noticed this was done for the "directory" role, I think the same can be done here.

jnurthen:
Very Draft PR to close #1164

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1198.html" title="Last updated on May 6, 2020, 10:55 PM UTC (ed1ddce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1198/757b6b3...ed1ddce.html" title="Last updated on May 6, 2020, 10:55 PM UTC (ed1ddce)">Diff</a>


index.html
```diff
@@ -10895,7 +10895,8 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-                <p class="note" title="Usage on columnheader, rowheader and row">While <sref>aria-disabled</sref> is currently supported on <rref>columnheader</rref>, <rref>rowheader</rref>, and <rref>row</rref>, in a future version the working group plans to prohibit its use on elements with any of those three roles except when they are in the context of a <rref>grid</rref> or <rref>treegrid</rref>.</p>
+				<p class="note" title="Usage on columnheader, rowheader and row">While <sref>aria-disabled</sref> is currently supported on <rref>columnheader</rref>, <rref>rowheader</rref>, and <rref>row</rref>, in a future version the working group plans to prohibit its use on elements with any of those three roles except when they are in the context of a <rref>grid</rref> or <rref>treegrid</rref>.</p>
+				<p class="note">This state is being deprecated as a global state in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically allowed.</p>
```
```suggestion
				<p class="note">This state is being deprecated as a global state in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically supported.</p>
```

index.html
```diff
@@ -11068,6 +11069,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				<p>Indicates whether a grouping element owned or controlled by this element is expanded or collapsed.</p>
 				<p>The <sref>aria-expanded</sref> attribute is applied to a focusable, interactive element that toggles visibility of content in another element. For example, it is applied to a parent <rref>treeitem</rref> to indicate whether its child branch of the tree is shown. Similarly, it can be applied to a <rref>button</rref> that controls visibility of a section of page content.</p>
 				<p>If a grouping container that can be expanded or collapsed is not <a>owned</a> by the element that has the <sref>aria-expanded</sref> attribute, the author SHOULD identify the controlling relationship by referencing the container from the element that has <sref>aria-expanded</sref> with the <pref>aria-controls</pref> property.</p>
+				<p class="note">This state is being deprecated as a global state in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically allowed.</p>
```
```suggestion
				<p class="note">This state is being deprecated as a global state in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically supported.</p>
```

index.html
```diff
@@ -11224,7 +11226,9 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
         <p>The <code>aria-haspopup</code> property is an enumerated type. <a>User agents</a> MUST treat any value of <code>aria-haspopup</code> that is not included in the list of allowed values, including an empty string, as if the value <code>false</code> had been provided. To provide backward compatibility with ARIA 1.0 content, user agents MUST treat an <code>aria-haspopup</code> value of <code>true</code> as equivalent to a value of <code>menu</code>.</p>
         <p><a>Assistive technologies</a> SHOULD NOT expose the <code>aria-haspopup</code> property if it has a value of <code>false</code>.</p>
-        <p class="note">A <rref>tooltip</rref> is not considered to be a popup in this context.</p>
+		<p class="note">A <rref>tooltip</rref> is not considered to be a popup in this context.</p>
+		<p class="note">This property is being deprecated as a global property in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically allowed.</p>
```
```suggestion
		<p class="note">This property is being deprecated as a global property in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically supported.</p>
```

index.html
```diff
@@ -11364,6 +11368,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p>When the user attempts to submit data involving a field for which <pref>aria-required</pref> is <code>true</code>, authors MAY use the <sref>aria-invalid</sref> attribute to signal there is an error. However, if the user has not attempted to submit the form, authors SHOULD NOT set the <sref>aria-invalid</sref> attribute on required <a>widgets</a> simply because the user has not yet entered data.</p>
 				<p>For future expansion, the <sref>aria-invalid</sref> attribute is an enumerated type. Any value not recognized in the list of allowed <span>values</span> MUST be treated by user agents as if the value <code>true</code> had been provided. If the attribute is not present, or its value is <code>false</code>, or its value is an empty string, the default value of <code>false</code> applies.</p>
+				<p class="note">This state is being deprecated as a global state in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically allowed.</p>
```
```suggestion
				<p class="note">This state is being deprecated as a global state in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically supported.</p>
```

common/script/aria.js
```diff
@@ -425,15 +432,19 @@ require(["core/pubsubhub"], function( respecEvents ) {
                             for (var k = 0; k < sortedList.length; k++) {
                                 var role = sortedList[k];
                                 var req = "";
+                                var dep = "";
                                 if (role.required) {
                                     req = " <strong>(required)</strong>";
                                 }
+                                if (role.deprecated) {
+                                    dep = " <strong>(deprecated in ARIA 1.2)</strong>"
```
carmacleod: I can't help thinking that authors could easily misunderstand this, and think that the attribute itself has been deprecated (not just its use as a global). Suggest adding "as global" to hopefully help mitigate any confusion.

```suggestion
                                    dep = " <strong>(deprecated as global in ARIA 1.2)</strong>"
```

jnurthen:
Note: this can't be merged as it changes files in aria-common too - but please review. I will split it out once it is reviewed

mcking65:
There are some cases with multiple inherritance where the script isn't doing quite what we want.

The columnheader role is a subclass of gridcell. The gridcell role supports aria-disabled, aria-expanded, aria-haspopup, and aria-invalid. But, the script is incorrectly deprecating these on columnheader. Note that we have a note in the columnheader prose saying that in a future version we plan to limit use of aria-disabled to columnheaders in a grid context; that note does not mention aria-expanded, aria-haspopup, or aria-invalid. I am not sure why; I don't remember the history of that note. rowheader has similar problems.

In the last working draft, aria-haspopup was not deprecated on menuitemcheckbox and menuitemradio. It makes sense not to support it on these two roles. However, I don't understand why the script is deprecating on these given that one of their superclasses is menuitem, and menuitem supports aria-haspopup. Perhaps this is similar to the columnheader and rowheader issue. One of the superclasses of menuitemcheckbox is checkbox, and it doesn't support aria-haspopup.

I don't recall discussion of why aria-haspopup is not supported on the option role. We have it on treeitem and tab; seems like it should be on option.

The script is incorrectly deprecating aria-disabled on treeitem, perhaps because of its listitem superclass.

This is kind of a nit, but the wording for abstract roles, e.g., command, should be something like "use on concrete subclass roles of this role prohibited in ARIA 1.2"

jnurthen:
@mcking65 I think I have fixed all those issues. Can you please take another look.

index.html
```diff
@@ -11068,6 +11069,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				<p>Indicates whether a grouping element owned or controlled by this element is expanded or collapsed.</p>
 				<p>The <sref>aria-expanded</sref> attribute is applied to a focusable, interactive element that toggles visibility of content in another element. For example, it is applied to a parent <rref>treeitem</rref> to indicate whether its child branch of the tree is shown. Similarly, it can be applied to a <rref>button</rref> that controls visibility of a section of page content.</p>
 				<p>If a grouping container that can be expanded or collapsed is not <a>owned</a> by the element that has the <sref>aria-expanded</sref> attribute, the author SHOULD identify the controlling relationship by referencing the container from the element that has <sref>aria-expanded</sref> with the <pref>aria-controls</pref> property.</p>
+				<p class="note">This state is being deprecated as a global state in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically supported.</p>
```
carmacleod, line 11072: This note needs to be moved to `aria-errormessage`.

index.html
```diff
@@ -11084,7 +11086,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 					</tr>
 					<tr>
 						<th class="state-applicability-head" scope="row">Used in Roles:</th>
-						<td class="state-applicability">Placeholder</td>
+						<td class="state-applicability">Use as a global deprecated in ARIA 1.2</td>
```
carmacleod, line 11089: This `state-applicability` row needs to be moved to `aria-errormessage`.

jnurthen:
@carmacleod I don't think we have discussed it but I believe it is consistent with the rest of the specification if it works this way.

PR #1207, opened 2020-02-27
Add info about layout and bounds to generic

### LINKED ISSUE #1182: Concerning the "generic" role
I have a few reservations with distinguishing the role "generic" from role "presentation". ARIA 1.1 isn't explicit about this, but example 8 (under `presentation`) indicates that `span` elements implicitly have no role:

> For example, according to an accessibility API, the following markup elements would appear to have identical role semantics (no role) and identical content.

```html
<!-- 1. [role="presentation"] negates the implicit 'heading' role semantics but does not affect the contents. -->
<h1 role="presentation"> Sample Content </h1>

<!-- 2. There is no implicit role for span, so only the contents are exposed. -->
<span> Sample Content </span>
...
```

It is not at all unreasonable to assume the same is true for `div`, which is how most browsers have interpreted this. In all Chromium / Webkit browsers (as best I can tell), a `div` is treated as having no role, the same way putting `role="presentation"` on other elements. The only major browser that seems to differ from this is Firefox, which treats `div` elements as containers.

Where this matters is in things like set size counts. The following code snippet is counted in Chrome as having 3 items, but in Firefox it is counted as having 2:

```html
<span role="menu">
  <a href="" role="menuitem">Home</a>
  <span>
    <a href="" role="menuitem">About</a>
    <a href="" role="menuitem">Contact us</a>
  </span>
</span>
```

I think Chromium / Webkit do the right thing here in excluding these `span` elements from the accessibility tree. That is definitely consistent with what ARIA 1.1 says about `span` elements in example 8.

This type of pattern is very common. You can see it a lot in component architecture. It produces "div soup". Parent-child relationships quickly get disrupted by wrapping components in deeper abstraction layers. This isn't a big issue in Chromium because it just won't put the `div` in the accessibility tree, and that `menuitem` will still be counted as a child of `menu`. 

Introducing the role `generic` seems like it will make some fairly common patterns more difficult to make accessible. I am not all too sure why anyone would want an element with a generic role in their page at all. When using `div`, you'd either want it ignored, or for it to have some type of role, like `group`. The only explanation there is in ARIA 1.2 is this:

> A generic can provide a limited number of accessible states and properties for its descendants, such as aria-live attributes. This differentiates it from the presentation role.

But the problem with this is that the `generic` role has the exact same allowed attributes as `presentation` does. You can use `aria-live` on an element with `presentation`. I'm not sure it works, but if not, that's something browsers need to fix.

In short, can the `generic` role be removed?

carmacleod:
Resolves #1182.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1207.html#generic" title="Last updated on Mar 18, 2020, 8:05 PM UTC (b758229)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1207/f41f707...b758229.html" title="Last updated on Mar 18, 2020, 8:05 PM UTC (b758229)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3404,8 +3404,8 @@
 			<rdef>generic</rdef>
 			<div class="role-description">
 				<p>A nameless container <a>element</a> that has no semantic meaning on its own.</p>
-				<p>The <code>generic</code> role is intended for implementors of User Agents. Authors SHOULD NOT use this role in content. Authors MAY use <rref>presentation</rref> or <rref>none</rref> to remove implicit accessibility semantics, or a semantic container role, such as <rref>group</rref>, to semantically group descendants in a named container.</p>
-				<p>Like an element with role <rref>presentation</rref>, a <code>generic</code> can provide a limited number of accessible states and properties for its descendants, such as <pref>aria-live</pref> attributes. However, unlike an element with role <rref>presentation</rref>, <code>generic</code> elements are exposed in <a>accessibility APIs</a> so that assistive technologies can gather certain properties such as layout and bounds.</p>
+				<p>The <code>generic</code> role is intended for use as the implicit role of generic elements in host languages (such as <abbr title="Hypertext Markup Language">HTML</abbr> <code>div</code> or <code>span</code>), so is primarily for implementors of user agents. Authors SHOULD NOT use this role in content. Authors MAY use <rref>presentation</rref> or <rref>none</rref> to remove implicit accessibility semantics, or a semantic container role such as <rref>group</rref> to semantically group descendants in a named container.</p>
+				<p>Like an element with role <rref>presentation</rref>, an element with role <code>generic</code> can provide a limited number of accessible states and properties for its descendants, such as <pref>aria-live</pref> attributes. However, unlike elements with role <rref>presentation</rref>, <code>generic</code> elements are exposed in <a>accessibility APIs</a> so that assistive technologies can gather certain properties such as layout and bounds.</p>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -3404,9 +3404,8 @@
 			<rdef>generic</rdef>
 			<div class="role-description">
 				<p>A nameless container <a>element</a> that has no semantic meaning on its own.</p>
-				<p>Contrast with <rref>group</rref>, which semantically groups its descendants in a named container.</p>
-				<p>A <code>generic</code> can provide a limited number of accessible states and properties for its descendants, such as <pref>aria-live</pref> attributes. In addition, <code>generic</code> containers are exposed to the API so that assistive technologies can gather certain properties such as layout and bounds. This differentiates it from the <rref>presentation</rref> role.</p>
-				<p>The <code>generic</code> role is intended for implementors of User Agents. Authors SHOULD NOT use this role in content.</p>
+        <p>The <code>generic</code> role is intended for implementors of User Agents. Authors SHOULD NOT use this role in content. Authors MAY use <rref>presentation</rref> to remove implicit accessibility semantics or a variety of other named containers, such as  <rref>group</rref>, to semantically group descendants in a named container.</p>
+				<p>Like an element with role <rref>presentation</rref>, a <code>generic</code> can provide a limited number of accessible states and properties for its descendants, such as <pref>aria-live</pref> attributes. However, unlike an element with role <rref>presentation</rref>, <code>generic</code> container elements are exposed in <a>accessibility APIs</a> so assistive technologies can gather certain properties such as layout and bounds.</p>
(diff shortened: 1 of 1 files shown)


carmacleod:
Very nice, @mcking65 !
I like having the most important info (i.e. "authors should not use generic") up front.
Also, "like presentation" and "unlike presentation" helps keep the generic vs presentation statements separate.
This is much more coherent now.

I made a few tiny editorial changes to your changes:
- added "or none" after the first instance of "presentation" (felt like it needed to be there)
- changed "a variety of other named containers such as group" to "a semantic container role such as group" because "named containers" and "named container" were in the same sentence and it felt redundant
- added "that" in "so that assistive technologies can..."

@jnurthen We should probably merge https://github.com/w3c/aria/pull/1142 before merging this? It deletes the editor's note about triage of global properties, and makes `aria-roledescription` prohibited on `generic` (further differentiating `generic` from `presentation`).

carmacleod:
@jnurthen @joanmarie @mcking65 

This PR is ready for final review. The only changes are to the prose for generic role, and I have already incorporated Matt's editorial suggestions.

Here's the [preview](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1207.html#generic).

index.html
```diff
@@ -3404,9 +3404,8 @@ <h2>Definition of Roles</h2>
 			<rdef>generic</rdef>
 			<div class="role-description">
 				<p>A nameless container <a>element</a> that has no semantic meaning on its own.</p>
-				<p>Contrast with <rref>group</rref>, which semantically groups its descendants in a named container.</p>
-				<p>A <code>generic</code> can provide a limited number of accessible states and properties for its descendants, such as <pref>aria-live</pref> attributes. This differentiates it from the <rref>presentation</rref> role.</p>
-				<p>The <code>generic</code> role is intended for implementors of User Agents. Authors SHOULD NOT use this role in content.</p>
+				<p>The <code>generic</code> role is intended for implementors of User Agents. Authors SHOULD NOT use this role in content. Authors MAY use <rref>presentation</rref> or <rref>none</rref> to remove implicit accessibility semantics, or a semantic container role, such as <rref>group</rref>, to semantically group descendants in a named container.</p>
```
cookiecrook: The change LGTM as-is, but perhaps also change this clause: "intended for implementors of User Agents."

I suggest: "intended for use as the default computed role of generic elements in host languages (e.g. `<span> or `<div>`)." 

This is partially for UA implementors, but for compatibility testing purposes, including web author tests using WebDriver... Not only implementors. If you prefer, we could file a separate issue to address this minor prose change.

joanmarie: Nit: don't capitalize "User Agents"

Nit: The "Authors MAY" sentence seems rather commaful. Are the commas needed in between "role" and "such" and after the "</rref>"?

index.html
```diff
@@ -3404,9 +3404,8 @@ <h2>Definition of Roles</h2>
...
+				<p>Like an element with role <rref>presentation</rref>, a <code>generic</code> can provide a limited number of accessible states and properties for its descendants, such as <pref>aria-live</pref> attributes. However, unlike an element with role <rref>presentation</rref>, <code>generic</code> elements are exposed in <a>accessibility APIs</a> so that assistive technologies can gather certain properties such as layout and bounds.</p>
```
joanmarie: Nit: I personally don't think we want to treat "generic" as a noun (i.e. in the first sentence).

index.html
carmacleod: @cookiecrook I added your new "intended for use as..." words (changed "default computed role" to "implicit role" because that seems to be the standard lingo). I wanted to keep the "implementors of user agents" to reinforce to authors that they really have no business setting role=generic on anything...

I will paste the full 3 paragraphs below - please have a look if you have a moment.

@joanmarie "user agents" it is, and I removed 2 commas that I agree were excessive.

index.html
carmacleod: changed "a generic" to "an element with role generic"

carmacleod:
@jnurthen Ready for final review.

pr-preview is timing out (I ran it a few times - will try again later), but here is the updated generic role prose:
> A nameless container element that has no semantic meaning on its own.
> 
> The generic role is intended for use as the implicit role of generic elements in host languages (such as HTML div or span), so is primarily for implementors of user agents. Authors SHOULD NOT use this role in content. Authors MAY use presentation or none to remove implicit accessibility semantics, or a semantic container role such as group to semantically group descendants in a named container.
> 
> Like an element with role presentation, an element with role generic can provide a limited number of accessible states and properties for its descendants, such as aria-live attributes. However, unlike elements with role presentation, generic elements are exposed in accessibility APIs so that assistive technologies can gather certain properties such as layout and bounds.

carmacleod:
@jnurthen ping - build is passing and pr-preview is good.
Quick! Review this before something else breaks! 😄

Issue #748, opened 2018-05-08
Definition of "owned" is not strict enough; should actually be direct children, not any descendant


jnurthen:
From - https://www.w3.org/Bugs/Public/show_bug.cgi?id=26307

The definition of "owned" in the ARIA spec includes any descendant (or any descendant of a node that's aria-owned).

That means that according to the spec, it's okay to have a list with listitems that are descendants but not direct children, or a grid with rows that are descendants but not direct children, e.g.:

```
<div role="list">
  <div>
    <div role="listitem">One</div>
  </div>
  <div>
    <div role="listitem">Two</div>
  </div>
</div>

<div role="grid">
  <div>
    <div role="row">
      <div>
        <div role="gridcell">1, 1</div>
      </div>
      <div>
        <div role="gridcell">2, 1</div>
      </div>
    </div>
  </div>
</div>
```

In practice, this doesn't seem to work with *any* browser or screen reader, unless you use aria-owns or role=presentation.

It occasionally works in some circumstances when the browser skips over extra unneeded container elements - but it definitely doesn't work in general.

Either the spec needs to better match how browsers and assistive technology have implemented their support for owned relationships, or browsers need to start doing a lot of extra work to search all descendants for required owned elements.

jnurthen:
See related info in auto-wcag https://github.com/auto-wcag/auto-wcag/blob/c031c8987712478e95932c022b78ebf01d22b44f/pages/glossary/owner-element.md

jnurthen:
@WilcoFiers - does this cover it?

jnurthen:
Proposal: 
Fix owned element language to restrict owned elements to 
- Direct Children
- Elements referenced with aria-owns
- Descendant elements where each ancestor node between the parent and the element has a role of none or generic

Should we also allow
- Descendant elements of the element referenced with aria-owns where each ancestor node between the element referenced with aria-owns and the owned element has a role of none or generic

(need to make this language completely unambiguous - this is a first proposal and needs work)

accdc:
Clarification is good, but this is a thorny one, because this restriction has the potential to break ARIA menus and trees and listboxes where additional markup is almost always used in the wild for styling purposes.

E.G what if the intervening markup includes LI elements that are not presentational? 

I agree this has never worked reliably before, but browsers should be capable of at the very least to process explicit role associations and identify in which order interactive constructs include their parent child role associations. I know this is doable, because I already had to do this as part of the AccName algorithm code to get it to work right.

[...]

PR #1213, opened 2020-03-13
revise Owned Element term & Required Context Role

### LINKED ISSUE #1161: Inconsistent use of "owned by"
Reading through the document, there seem to be cases where "owned by" is meant to be either all descendants (as defined) but there are other places where it is strongly implied "owned by" is a parent/child relationship. Firstly, the definition:

> [Owned Element](https://www.w3.org/TR/wai-aria-1.2/#dfn-owned-element)
> An 'owned element' is any DOM descendant of the element, any element specified as a child via aria-owns, or any DOM descendant of the owned child.

But then if we read `menuitem`, it says the following:

> [menuitem (role)](https://www.w3.org/TR/wai-aria-1.2/#menuitem)
> In order to identify that they are related widgets, authors MUST ensure that menu items are owned by an element with role menu or menubar, or by a role group which itself is owned by an element with role menu or menubar.

This last part "or by a role group which ..." doesn't make sense if "owned by" applies to all descendants. If a menu item owns all descendants, it automatically owns a menuitem owned by a group that's a child of the menu. The `menuitem` even owns it, if it's in a `list` which is in a `grid` which is in a `header` which is in a `menu`.

And that last part is where I'm starting to have proper concerns. The following seems to be conform to ARIA 1.2:

```html
<div role="menu">
  <div role="list">
    <div role="header">
      <div role="menuitem">
        <div role="listitem">Tower of oddly nested roles</div>
      </div>
    </div>
  </div>
</div>
```

Even more explicitly, 9.2 basically says that elements are only supposed to have one owned element: 

> [9.2 States and Properties](https://www.w3.org/TR/wai-aria-1.2/#document-handling_author-errors_states-properties)
> 2. Correct usage with regard to DOM tree structure, such as an element being owned by more than one other element.

On the other hand, there are places where it matters that "owned elements" aren't just the child elements, but all descendants (or at least, up to a certain level):

> If aria-readonly is set on an element with role grid, user agents MUST propagate the value to all gridcell elements owned by the grid and expose the value in the accessibility API.

In this case, it matters that `aria-readonly` propagates to the `cell` of a `grid`. Although even here it could be argued that `aria-readonly` shouldn't propagate down into nested `grid` elements. That propagation is far more pronounced for the `presentation` role, and here it **really** matters if "owned elements" is limited to children, or all descendants:

> presentation (role)
> When an explicit or inherited role of presentation is applied to an element with the implicit semantic of a WAI-ARIA role that has required owned elements, in addition to the element with the explicit role of presentation, **the user agent MUST apply an inherited role of presentation to any owned elements that do not have an explicit role defined**. 

As far as I can tell, browser vendors have taken this to mean all children, and everything directly referenced through `aria-owns`. That's very sensible, otherwise a single misplaced `role="presentation"` on the `body` element could wrack the semantics of the entire page.

## Proposal

I think the simplest solution here would be to split the "owned elements" definition, into "owned children" and "owned descendants". That seems to address most of these issues. Additionally, you'd have to do something about "presentation" and "none", make it clear those elements can't "own" anything. 

We'd also get an answer on whether or not "generic" should be included in the accessibility tree or not. Browsers don't seem to agree on that.

A final little bonus thing would be to make it explicit that element referenced with "aria-owns" are owned children of the first element with that attribute, and that other elements can not own them.

scottaohara:
This PR aims to close #1161 for ARIA 1.2.

It expands the definition of "owned element" to mention the concept of specific required context roles.  

Required Context Role was revised to indicate there are "...no other intermixed containing elements with non-presentational roles" that separate the role from its required context role.

Valid and invalid examples were added to this section to further demonstrate.

Other issues/PRs such as #1162 will help further clarify how owned elements may need to be setup. 

This update touches on the topic raised in issue #748, which is marked for 1.3.... though I would say this PR does not "solve" that issue... it at least might help clarify things a bit prior to additional work needing to be done for 1.3.

***

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Faria%2Fe6baf7cf97595fe515947e69bf7dc806caeb1372%2Findex.html%3FisPreview%3Dtrue)

```
Navigation timeout of 27999 ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231213.)._
</details>

### DIFF
--- a/common/terms.html
+++ b/common/terms.html
@@ -136,11 +136,11 @@
     </dd>
     <dt><dfn data-lt="owned element|owned|owned element's|owned elements">Owned Element</dfn></dt>
     <dd>
-      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child. 'Owned elements' can have specific <a href="#scope">required context roles</a> in relation to their <a class="termref" href="#dfn-owning">owning elements</a>.</p>
+      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child. 'Owned elements' can have specific <a href="#scope">required context roles</a> in relation to their <a class="termref" href="#dfn-owning-element">owning elements</a>.</p>
     </dd>
     <dt><dfn data-lt="owning element|owning">Owning Element</dfn></dt>
     <dd>
-      <p>An 'owning element' is any <abbr title="Document Object Model">DOM</abbr> ancestor of the <a>element</a>, or any element with an <pref>aria-owns</pref> attribute which references the ID of the   element. </p>
+      <p>An 'owning element' is any <abbr title="Document Object Model">DOM</abbr> ancestor of the <a>element</a>, or any element with an <pref>aria-owns</pref> attribute which references the ID of the element. 'Owning elements' can have specific <a href="#mustContain">required owned roles</a> in relation to their <a class="termref" href="#dfn-owned-element">owned elements</a>.</p>
     </dd>
     <dt><dfn>Perceivable</dfn></dt>
     <dd>
--- a/common/terms.html
+++ b/common/terms.html
@@ -136,7 +136,7 @@
     </dd>
     <dt><dfn data-lt="owned element|owned|owned element's|owned elements">Owned Element</dfn></dt>
     <dd>
-      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child.</p>
+      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child. An 'owned element' can require specific <a href="#scope">required context roles</a> in relation to their <a class="termref" href="#dfn-owning">owning elements</a>.</p>
     </dd>
     <dt><dfn data-lt="owning element|owning">Owning Element</dfn></dt>
     <dd>
--- a/common/terms.html
+++ b/common/terms.html
@@ -136,7 +136,7 @@
     </dd>
     <dt><dfn data-lt="owned element|owned|owned element's|owned elements">Owned Element</dfn></dt>
     <dd>
-      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child. An 'owned element' can require specific <a href="#scope">required context roles</a> in relation to their <a class="termref" href="#dfn-owning">owning elements</a>.</p>
+      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child. 'Owned elements' can have specific <a href="#scope">required context roles</a> in relation to their <a class="termref" href="#dfn-owning">owning elements</a>.</p>
     </dd>
     <dt><dfn data-lt="owning element|owning">Owning Element</dfn></dt>
     <dd>
--- a/common/terms.html
+++ b/common/terms.html
@@ -140,7 +140,7 @@
     </dd>
     <dt><dfn data-lt="owning element|owning">Owning Element</dfn></dt>
     <dd>
-      <p>An 'owning element' is any <a>element</a> with an <pref>aria-owns</pref> attribute that references the ID of the element or any of its <abbr title="Document Object Model">DOM</abbr> ancestors, or if the element does not have an ancestor that is referenced by an <pref>aria-owns</pref> attribute then any <abbr title="Document Object Model">DOM</abbr> ancestor of the <a>element</a>. 'Owning elements' can have specific <a href="#mustContain">required owned roles</a> in relation to their <a class="termref" href="#dfn-owned-element">owned elements</a>.</p>
+      <p>An 'owning element' is any <a>element</a> with an <pref>aria-owns</pref> attribute that references either the ID of the element or any of its <abbr title="Document Object Model">DOM</abbr> ancestors, or if the element does not have an ancestor that is referenced by an <pref>aria-owns</pref> attribute then any <abbr title="Document Object Model">DOM</abbr> ancestor of the <a>element</a>. 'Owning elements' can have specific <a href="#mustContain">required owned roles</a> in relation to their <a class="termref" href="#dfn-owned-element">owned elements</a>.</p>
     </dd>
     <dt><dfn>Perceivable</dfn></dt>
     <dd>
--- a/common/terms.html
+++ b/common/terms.html
@@ -136,11 +136,11 @@
     </dd>
     <dt><dfn data-lt="owned element|owned|owned element's|owned elements">Owned Element</dfn></dt>
     <dd>
-      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a> unless the <abbr title="Document Object Model">DOM</abbr> descendant is the target of an <pref>aria-owns</pref> relation, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child. 'Owned elements' can have specific <a href="#scope">required context roles</a> in relation to their <a class="termref" href="#dfn-owning-element">owning elements</a>.</p>
+      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a> unless the <abbr title="Document Object Model">DOM</abbr> descendant or any of its ancestors is the target of an <pref>aria-owns</pref> relation, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child. 'Owned elements' can have specific <a href="#scope">required context roles</a> in relation to their <a class="termref" href="#dfn-owning-element">owning elements</a>.</p>
     </dd>
     <dt><dfn data-lt="owning element|owning">Owning Element</dfn></dt>
     <dd>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -477,7 +477,31 @@ <h3>Required Owned Elements</h3>
 		</section>
 		<section id="scope">
 			<h3>Required Context Role</h3>
-			<p>The required context role defines the owning container where this <a>role</a> is allowed. If a role has a required context, authors MUST ensure that an element with the role is contained inside (or <a>owned</a> by) an element with the required context role. For example, an element with role <code>listitem</code> is only meaningful when contained inside (or owned by) an element with role <code>list</code>.</p>
+			<p>The required context role defines the owning container where this <a>role</a> is allowed. If a role has a required context, authors MUST ensure that an element with the role is contained inside (or <a>owned</a> by) an element with the required context role, with no other intermixed containing elements with non-presentational roles.</p>
```
WilcoFiers: What exactly are "elements with non-presentational roles"? This term is only ever used as "explicit non-presentational role", which is clear. But for the implicit case this could be misunderstood. It can either mean it has no non-presentional role in ARIA, or it has no non-presentational role in the accessibility API.

The `canvas` element is one that comes to mind. There is no ARIA role that maps to `canvas`, so in that sense it would not count for the required context. But if you consider what role it is [mapped to in the accessibility API](https://w3c.github.io/html-aam/#el-canvas), canvas does have a non-presentational role.

scottaohara: Yes.  That does need more clarity if it's still unclear that one shouldn't be inserting improper markup and roles between the element/role and its required context role.

Essentially needs to be revised to call out `role=presentation` and `none` are fine, anything else is not.

jnurthen:
build seems to be failing due to https://github.com/w3c/spec-generator/issues/284

index.html
scottaohara: I've made further updates to this text to address your comments, @WilcoFiers 

preview still doesn't seem to be working, but not getting build failures anymore at least.

common/terms.html
```diff
@@ -136,7 +136,7 @@
     </dd>
     <dt><dfn data-lt="owned element|owned|owned element's|owned elements">Owned Element</dfn></dt>
     <dd>
-      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child.</p>
+      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child. An 'owned element' can require specific <a href="#scope">required context roles</a> in relation to their <a class="termref" href="#dfn-owning">owning elements</a>.</p>
```
carmacleod: Suggestions:
- change 'require' to 'have' because 'require' is implied by 'required context roles'
- make singular/plural consistent in new sentence (the suggestion below changes to "all plural", but if preferred could change to "all singular", i.e. "An 'owned element' can have a specific required context role in relation to its owning element.")

```suggestion
      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child. 'Owned elements' can have specific <a href="#scope">required context roles</a> in relation to their <a class="termref" href="#dfn-owning">owning elements</a>.</p>
```

scottaohara: thanks!

index.html
```diff
@@ -477,7 +477,31 @@ <h3>Required Owned Elements</h3>
 		</section>
 		<section id="scope">
 			<h3>Required Context Role</h3>
-			<p>The required context role defines the owning container where this <a>role</a> is allowed. If a role has a required context, authors MUST ensure that an element with the role is contained inside (or <a>owned</a> by) an element with the required context role. For example, an element with role <code>listitem</code> is only meaningful when contained inside (or owned by) an element with role <code>list</code>.</p>
+			<p>The required context role defines the owning container where this <a>role</a> is allowed. If a role has a required context, authors MUST ensure that the element is contained inside (or <a>owned</a> by) an element with the required context role. There can be no other intermixed containing elements unless those elements are explicitly set as presentational, e.g. with a <rref>none</rref> role.</p>
+			<p>For example, an element with role <code>listitem</code> is only meaningful when contained inside (or owned by) an element with role <code>list</code>, as demonstrated by the following code snippet.</p>
+			<div class="example">
```
jnurthen: Needs to be a `<pre>` or an `<aside>` in order to get all the example text from respec. See https://github.com/w3c/respec/wiki/example

index.html
```diff
@@ -477,7 +477,31 @@ <h3>Required Owned Elements</h3>
...
+            &lt;div role="listitem">Direct child of the list.&lt;/div>
+            &lt;div role="none">
+              &lt;div role="listitem">Descendant contained by a presentational element.&lt;/div>
+            &lt;/div>
+          &lt;/div>
+          &lt;div role="listitem" id="owned_item">List item that is "owned" by its sibling "list".&lt;/div>
+        </pre>
+      </div>
+
+      <p>The following example shows <code>listitem</code> elements that do not meet the required context role expectations.</p>
```
jnurthen: I don't like anti-patterns as examples. Can we put these 2 examples together and add comments to them to explain why they don't meet the requirements

common/terms.html
```diff
@@ -136,7 +136,7 @@
     </dd>
     <dt><dfn data-lt="owned element|owned|owned element's|owned elements">Owned Element</dfn></dt>
     <dd>
-      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child.</p>
+      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child. 'Owned elements' can have specific <a href="#scope">required context roles</a> in relation to their <a class="termref" href="#dfn-owning">owning elements</a>.</p>
```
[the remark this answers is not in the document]
aleventhal: > any DOM descendant of the element

This is false. An elements's ownership chain cannot split into two. If an element or DOM ancestor of it is already owned by another element via aria-owns, then it has been captured, at the point of the aria-owns. The next DOM parent up loses it as a child.

For example:

```
<div id="main-container">
  <div id="foo-owner" aria-owns="foo"></div>
  <div id="original-dom-foo-parent">
    <div id="foo">
      <div id="foo-child"></div>
    </div>
  </div>
</div>
```

The accessible object hierarchy becomes:
id=main container
.id=foo-owner
..id=foo
...id=foo-child
.original-dom-parent

You can follow up the accessible object hierarchy to see what owns what. The following 2 items contradict , which contradicts "any DOM descendant of the element":
- foo is owned by foo-owner and main-container, but not owned by original-dom-foo-parent
- foo-child is owned by foo, foo-owner and main-container, but not owned by original-dom-foo-parent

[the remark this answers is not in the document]
carmacleod: @jnurthen 
> - Remove from this PR and create a new PR against aria-common

Understood. Let's keep the definitions here until the PR is fully reviewed, and then move them over to aria-common before merging.

> - Verify that any other references to owned element in any other spec fit this definition.

The definition of Owned element is identical to what was there before (it just adds a nice link to "required owned elements").

@aleventhal 
>> any DOM descendant of the element

> This is false.

True, however this definition of Owned element goes back to [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element). That said, perhaps we can add a new clause. Not sure if we can do it without making it confusing, but here goes:
> An 'owned element' is any DOM descendant of the element unless that DOM descendant is the target of an aria-owns relation, any element specified as a child via aria-owns, or any DOM descendant of the owned child.

index.html
carmacleod: Fixed

index.html
carmacleod: Fixed. Please see https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1213.html#example-5

common/terms.html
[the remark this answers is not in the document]
carmacleod: @aleventhal
I decided to go ahead and add the clause mentioned above (changed 'that' to 'the').
So the first sentence in the definition of Owned element would be:

> An 'owned element' is any DOM descendant of the element unless the DOM descendant is the target of an aria-owns relation, any element specified as a child via aria-owns, or any DOM descendant of the owned child.

@jnurthen, I'm marking this conversation as resolved, with the understanding that the definitions need to be moved to a separate PR in aria-common.

carmacleod:
@aleventhal @jnurthen @curtbellew @WilcoFiers
I have marked this as Draft so that it does not get accidentally merged before moving the definitions over to aria-common, however I believe this PR is ready for final review.

@aleventhal I would particularly like your review to ensure that I addressed your concern about split ownership... without making the definitions too convoluted to read... :)

Preview links:
1. Definition of [Owned element](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1213.html#dfn-owned-element)
    - this definition has changed slightly to explicitly disallow split ownership
    - a new sentence points to the definition of the required context role characteristic
2. New definition for [Owning element](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1213.html#dfn-owning-element)
    - points to
3. Updated definition of [Required Context Role](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1213.html#scope)

common/terms.html
```diff
@@ -136,11 +136,11 @@
     </dd>
     <dt><dfn data-lt="owned element|owned|owned element's|owned elements">Owned Element</dfn></dt>
     <dd>
-      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child.</p>
+      <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a> unless the <abbr title="Document Object Model">DOM</abbr> descendant or any of its ancestors is the target of an <pref>aria-owns</pref> relation, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child. 'Owned elements' can have specific <a href="#scope">required context roles</a> in relation to their <a class="termref" href="#dfn-owning-element">owning elements</a>.</p>
```
pkra, line 139: >  or any of its ancestors is the target of an <pref>aria-owns</pref> relation

I don't understand this restriction. Why is it problematic if an "aria-owned" element's parent is already "aria-owned"?

>  or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child

Are elements owned by the owned child owned?

carmacleod:
@pkra 
>> or any of its ancestors is the target of an aria-owns relation

> I don't understand this restriction. Why is it problematic if an "aria-owned" element's parent is already "aria-owned"?

That was my attempt to wordsmith @aleventhal's comment https://github.com/w3c/aria/pull/1213#discussion_r426840747.
If you can think of a clearer way to say it, please let me know.

>> or any DOM descendant of the owned child

> Are elements owned by the owned child owned?

Heh, I bet you can't say that 5 times fast. ;)
Anyhow, I think the answer is 'yes'. :)

scottaohara:
i'm going to close (but not delete) this PR, since i think Sarah's work makes this irrelevant

PR #1223, opened 2020-03-26
Combobox: Revise UA, AT, and checker guidance per agreed text in issue 1178

### LINKED ISSUE #1178: Add UA requirements for ARIA 1.1 under combobox
Under combobox it says: 

> The ARIA 1.0 specification describes a combobox pattern where the input element with the combobox role references the popup element with aria-owns instead of aria-controls. User agents, assistive technologies, and conformance checkers SHOULD continue to support the ARIA 1.0 pattern so that existing implementations of the ARIA 1.0 pattern remain functional. The ARIA 1.1 specification describes a combobox pattern where the element with role combobox is a composite container instead of a focusable input. User agents and assistive technologies did not adequately support the ARIA 1.1 pattern, so the ARIA 1.2 pattern supersedes the ARIA 1.1 pattern, which authors SHOULD abandon.

In combobox, it is mentioned that UA should continue to support the ARIA 1.0 pattern. I think the same needs to happen for the ARIA 1.1 pattern. Yes, authors should move away from it, but until they have, browsers have to maintain support for this pattern. This could very easilly take a decade. I think ARIA 1.2 needs to be clear on this.

mcking65:
Fixes #1178

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1223.html" title="Last updated on Apr 23, 2020, 3:35 PM UTC (404cc6e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1223/5f19d17...404cc6e.html" title="Last updated on Apr 23, 2020, 3:35 PM UTC (404cc6e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2154,18 +2154,14 @@
  		  <p class="ednote" title="Validity changes combobox for ARIA 1.2">
 			   Please review the following carefully. As a result of these changes a combobox following the ARIA 1.1 combobox specification will no longer conform with the ARIA specification.
 		  </p>
-        <p>
-          The ARIA 1.0 specification describes a <code>combobox</code> pattern where the input element with the <code>combobox</code> role references the popup element with <pref>aria-owns</pref> instead of <pref>aria-controls</pref>.
-          <a>User agents</a> and <a>assistive technologies</a> SHOULD continue to support the ARIA 1.0 pattern so that existing implementations of the ARIA 1.0 pattern remain functional.
-          However, conformance checkers SHOULD recommend that authors convert to the ARIA 1.2 pattern by replacing <pref>aria-owns</pref> with <pref>aria-controls</pref>.
-        </p>
-        <p>
-          The ARIA 1.1 specification describes a <code>combobox</code> pattern where the element with role <code>combobox</code> is a composite container instead of a focusable input.
-          <a>User agents</a> and <a>assistive technologies</a> did not adequately support the ARIA 1.1 pattern, so the ARIA 1.2 pattern supersedes the ARIA 1.1 pattern.
-          Authors SHOULD convert implementations of the ARIA 1.1 pattern to the ARIA 1.2 pattern.
-          Conformance checkers SHOULD warn authors that the ARIA 1.1 pattern is no longer valid.
-          User agents and assistive technologies MAY phase out support for the ARIA 1.1 pattern at their own discretion.
-        </p>
+        <div class="note">
+          <p>The structural requirements  for <code>combobox</code> described in this version of the specification are different from the requirements described in ARIA 1.0 and ARIA 1.1:</p>
+          <ul>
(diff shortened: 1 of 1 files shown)


cookiecrook:
> I'd rather this paragraph be deleted entirely from the spec.

Alternately, remove *all* the RFC requirements from this prose and make the whole section an informative Note to explain what happened.

mcking65:
cookiecrook commented:

> Trying to clarify the comments @mcking65 made on the ARIA call this morning.
> 
> I think this would result in very little change to comboboxes that use the role on native text input elements, like this:
> 
> ```
> <input type="text" role="combobox" aria-owns"my_listbox">
> ```

This PR does not change anything for implementations like this that use aria-owns, which is the 1.0 pattern. First, the combobox is made from a native input, which has a value attribute. So, the requirement is to give the combobox that value. In addition, however, this PR changes the spec to specify that UA calculate value from content only for combobox elements that implement the 1.2 pattern. If The UA were to calculate value from content for the 1.0 style implementation, that would be an error.

The 1.2 pattern changes the above by replacing aria-owns with aria-controls so there is no child content. And, since the input itself has a value, that value is the value of the combobox.

> But it could be different on this type of variant, which we still see a lot of web authors using.
> 
> ```
> <div role="combobox" tabindex="0">  
> 	<input type="text" role="textbox"> 
> 	… listbox too...
> </div>
> ```

This PR should not change value calculation for this code, which is invalid BTW. ARIA has never allowed a focusable parent. 1.1 allows a parent container, but it is not allowed to be focusable.

Again, this UA requirement is only for implementations of the 1.2 pattern. So, if the UA were to calculate a value from content for the above, it would be a UA error.

The 1.2 combobox is an input, not a composite. It is not permitted to have interactive descendants. So, the descendants of the above code should not be included in a value calculation.

> The way I read the PR is that the text `input` element’s value will be reflected to the parent `div` node in the accessibility APIs? Is that what you intended?

I didn't have any intention that would happen, and I am happy to make whatever change necessary to clarify. 

My intention is that The UA should only calculate value from content if the combobox is not an HTML text input with a value attribute  and if it meets the requirements of the 1.2 specification, ie, it is a focusable element with the combobox role, it does not have any interactive or focusable descendants, and it controls a popup element with role listbox, tree, grid, or dialog. If the element with role combobox is a native input with a value attribute, that value should still be used as the combobox value.

In other words, the requirements in the 1.2 version of the spec only apply to authors implementing ARIA 1.2. If the author is using ARIA 1.1, they should only use requirements in ARIA 1.1. And, if the author is using 1.0, the Author should be following ARIA 1.0 requirements.

Similarly, the user agent should implement combobox requirements in the 1.2 specification only for combobox elements that conform to the the 1.2 specification. This is a new requirement in 1.2. I'd be happy to add some clarifying language to either the paragraph that describes this requirement or to the notes at the end after we come to consensus on #1178. Personally, I think it would be better at the end.

mcking65:
@cookiecrook, I have revised the information about previous versions of combobox to be entirely informative.

Issue #711, opened 2018-04-05
aria-valuetext would be useful on many more roles


jnurthen:
There are some places in the APG where aria-labelledby is being used to provide values. I would like to see the ability to support aria-valuetext on more roles to provide a readable value-like attribute for the item.

An example is https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html Where the button has <button aria-haspopup="listbox" aria-labelledby="exp_elem exp_button" id="exp_button">Neptunium</button>

Another place where this would be essential in order to be able to mimic HTML functionality would be creating a labelled textbox. As the spec currently is written there is no way to provide both an accessible name and a value for something with role="textbox" for example.

JAWS-test:
In my opinion, aria-labelledby should only be used for labels and never be used for the value. Label and value are different things. It is important that assistive technology can distinguish between them to enable correct output. I.e. for textbox and searchbox only the text content can be used as value at the moment. But this is not defined in the specification. Should therefore be supplemented.

It should also be checked whether aria-valuetext will be allowed as an attribute for textbox and searchbox in the future. The reason is that not all HTML elements can have text content, e.g. <code>img</code>. So what would be the value of <code>img role=textbox</code>?

ZoeBijl:
Please be aware that the [`textbox`-role is not allowed on the `img`-element](https://www.w3.org/TR/html-aria/#img-alt).

JAWS-test:
Why? If the img has an alt attribute it is allowed:

> Any role except presentation or none

ZoeBijl:
Ah you’re right, missed that one. I’m very curious as to why that’s allowed. Not entirely sure in which scenario you’d want to make an image a textbox. 

@stevefaulkner do you know more about that?

JAWS-test:
@ZoeBijl 
I don't want to turn a graphic into a text box. I wouldn't ever want to make a textbox out of any HTML element, because there is the input element and that's good enough. The problem is: Others want this and I have to deal with their results. So what I want: A clear rule in the ARIA specification where a textbox or searchbox gets its value from.

ZoeBijl:
There should be an ARIA equivalent of HTML’s `value`-attribute. We’ll have to see how to best work that into the spec without too many changes.

JAWS-test:
[...]

PR #1225, opened 2020-03-26
Require user agents to expose a value for combobox elements

mcking65:
This resolves the simplest use case identified by issue #711 by requiring user agents to expose a value for elements with the combobox role. This will enable combobox to be used to create a custom select.

This PR requires user agents to expose a value to assistive technologies. If the combobox element is anHTML element that has a value, user agents must expose that value. Otherwise, they expose a string that represents descendant content using methods described in accname.

The idea of using valuetext doesn't make sense in the simplest use case because it would just force authors to copy innertext into an attribute. In more complex cases, an aria-valuedby attribute could be use to specified the element or elements that provide the value. However, we can save that for ARIA 1.3; it is not essential for making a custom select.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 15, 2020, 2:00 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Faria%2Fa1fc4d77fdbc25f9ab2d8af60ee406f3607e59d8%2Findex.html%3FisPreview%3Dtrue)

```
Navigation timeout of 30000 ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231225.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2134,10 +2134,13 @@
           When a descendant of the popup element is active, authors MAY set <pref>aria-activedescendant</pref> on the <code>combobox</code> to a value that refers to the active element within the popup while focus remains on the <code>combobox</code> element.
         </p>
         <p>
-          User agents MUST compute a value for elements with role <code>combobox</code> and expose the value to <a>assistive technologies</a>.
-          If the <code>combobox</code> element is also a host language element that provides a value, such as an HTML <code>input</code> element, the value of the combobox is the value of that element.
-          Otherwise, the value of the <code>combobox</code> is computed from descendant elements of the <code>combobox</code> using the same algorithm that is used to compute the name of a <rref>button</rref> from its descendant content.
+          User agents MUST expose the value of elements with role <code>combobox</code> to <a>assistive technologies</a>.
+          The value of a <code>combobox</code> is represented by one of the following:
         </p>
+        <ul>
+          <li>If the <code>combobox</code> element is a host language element that provides a value, such as an HTML <code>input</code> element, the value of the combobox is the value of that element.</li>
+          <li> Otherwise, the value of the <code>combobox</code> is represented by its descendant elements and can be determined using the same method used to compute the name of a <rref>button</rref> from its descendant content.</li>
+        </ul>
 				<pre class="example highlight">
           &lt;label for="tag_combo">Tag&lt;/label>
 		      &lt;input type="text" id="tag_combo"
--- a/index.html
+++ b/index.html
@@ -2114,7 +2114,7 @@
         </p>
 				<p>
           Typically, the initial state of a <code>combobox</code> is collapsed.
-          In the collapsed state, only the <code>combobox</code> element and an optional popup control <rref>button</rref> are visible.
+          In the collapsed state, only the <code>combobox</code> element and a separate, optional popup control <rref>button</rref> are visible.
           A <code>combobox</code> is said to be expanded when both the <code>combobox</code> element showing its current value and its associated popup element are visible.
           Authors MUST set <sref>aria-expanded</sref> to <code>true</code> on an element with role <code>combobox</code> when it is expanded and <code>false</code> when it is collapsed.
         </p>
@@ -2127,16 +2127,16 @@
           If the <code>combobox</code> popup element has a role other than <rref>listbox</rref>, authors MUST specify a value for <pref>aria-haspopup</pref> that corresponds to the role of its popup.
         </p>
 				<p>
-          If the user interface includes an additional icon that allows the visibility of the popup to be controlled via pointer and touch events, authors SHOULD ensure that element has role <rref>button</rref> and that it is focusable but not included in the page <kbd>Tab</kbd> sequence.
+          If the user interface includes an additional icon that allows the visibility of the popup to be controlled via pointer and touch events, authors SHOULD ensure that element has role <rref>button</rref>, that it is focusable but not included in the page <kbd>Tab</kbd> sequence, and that it is not a descendant of the element with role <code>combobox</code>.
           In addition, to be keyboard accessible, authors SHOULD provide keyboard mechanisms for moving focus between the <code>combobox</code> element and elements contained in the popup.
           For example, one common convention is that <kbd>Down Arrow</kbd> moves focus from the input to the first focusable descendant of the popup element.
           If the popup element supports <pref>aria-activedescendant</pref>, in lieu of moving focus, such keyboard mechanisms can control the value of <pref>aria-activedescendant</pref> on the <code>combobox</code> element.
           When a descendant of the popup element is active, authors MAY set <pref>aria-activedescendant</pref> on the <code>combobox</code> to a value that refers to the active element within the popup while focus remains on the <code>combobox</code> element.
         </p>
         <p>
-          User agents MUST compute a value for the <code>combobox</code>.
-          If the <code>combobox</code> element is also a host language element that provides a value, such as an HTML <code>input</code> element, user agents MUST set the value of the combobox to that value.
-          Otherwise, user agents must compute the value of the <code>combobox</code> from the descendant content of the <code>combobox</code> element.
+          User agents MUST compute a value for elements with role <code>combobox</code> and expose it to <a>assistive technologies</a>.
+          If the <code>combobox</code> element is also a host language element that provides a value, such as an HTML <code>input</code> element, the value of the combobox is the value of that element.
+          Otherwise, the value of the <code>combobox</code> is computed from descendant elements of the <code>combobox</code> using the same algorithm that is used to compute the name of a <rref>button</rref> from its descendant content.
         </p>
 				<pre class="example highlight">
           &lt;label for="tag_combo">Tag&lt;/label>
--- a/index.html
+++ b/index.html
@@ -2133,6 +2133,11 @@
           If the popup element supports <pref>aria-activedescendant</pref>, in lieu of moving focus, such keyboard mechanisms can control the value of <pref>aria-activedescendant</pref> on the <code>combobox</code> element.
           When a descendant of the popup element is active, authors MAY set <pref>aria-activedescendant</pref> on the <code>combobox</code> to a value that refers to the active element within the popup while focus remains on the <code>combobox</code> element.
         </p>
+        <p>
+          User agents MUST compute a value for the <code>combobox</code>.
+          If the <code>combobox</code> element is also a host language element that provides a value, such as an HTML <code>input</code> element, user agents MUST set the value of the combobox to that value.
+          Otherwise, user agents must compute the value of the <code>combobox</code> from the descendant content of the <code>combobox</code> element.
+        </p>
 				<pre class="example highlight">
           &lt;label for="tag_combo">Tag&lt;/label>
 		      &lt;input type="text" id="tag_combo"
--- a/index.html
+++ b/index.html
@@ -2139,7 +2139,7 @@
         </p>
         <ul>
           <li>If the <code>combobox</code> element is a host language element that provides a value, such as an HTML <code>input</code> element, the value of the combobox is the value of that element.</li>
-          <li> Otherwise, the value of the <code>combobox</code> is represented by its descendant elements and can be determined using the same method used to compute the name of a <rref>button</rref> from its descendant content.</li>
+          <li>Otherwise, the value of the <code>combobox</code> is represented by its descendant elements and can be determined using the same method used to compute the name of a <rref>button</rref> from its descendant content.</li>
         </ul>
 				<pre class="example highlight">
           &lt;label for="tag_combo">Tag&lt;/label>
--- a/index.html
+++ b/index.html
@@ -2134,7 +2134,7 @@
           When a descendant of the popup element is active, authors MAY set <pref>aria-activedescendant</pref> on the <code>combobox</code> to a value that refers to the active element within the popup while focus remains on the <code>combobox</code> element.
         </p>
         <p>
-          User agents MUST compute a value for elements with role <code>combobox</code> and expose it to <a>assistive technologies</a>.
+          User agents MUST compute a value for elements with role <code>combobox</code> and expose the value to <a>assistive technologies</a>.
           If the <code>combobox</code> element is also a host language element that provides a value, such as an HTML <code>input</code> element, the value of the combobox is the value of that element.
           Otherwise, the value of the <code>combobox</code> is computed from descendant elements of the <code>combobox</code> using the same algorithm that is used to compute the name of a <rref>button</rref> from its descendant content.
         </p>


curtbellew:
I think this makes sense.  I do wonder how this would be impacted by comboboxes that include a drop down arrow icon.  In our example https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html#ex3_label the drop down icon is a child of the combobox and it does have an appropriate aria-label of "Show vegetable options".  In this case would we run the risk of the value being calculated to be something like "Apple Show vegetable options"?

JAWS-test:
Would it make sense to insert an analogue sentence also for textbox and searchbox?

mcking65:
@curtbellew commented:

> I think this makes sense. I do wonder how this would be impacted by comboboxes that include a drop down arrow icon. In our example https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html#ex3_label the drop down icon is a child of the combobox and it does have an appropriate aria-label of "Show vegetable options". In this case would we run the risk of the value being calculated to be something like "Apple Show vegetable options"?

Perfectly fine risk because ARIA 1.2 makes that a deprecated pattern that checkers will call invalid and that AT don't generally support anyway.

If browsers or AT wanted to do so, they could easily detect the 1.1 implementation and calculate the value strictly based on the value of the textbox, which is what they do now. The revised UA support statement in PR #1223 allows for this.

mcking65:
@JAWS-test commented:

> Would it make sense to insert an analogue sentence also for textbox and searchbox?

Yes. I thought about doing that in this PR but opted for a minimal approach given where we are with ARIA 1.2. I think that can wait till ARIA 1.3 when we take on the rest of #711. 

However, since one of the goals of fixing combobox is to make custom selects possible, i.e., a robust equivalent of HTML:select@size=1, we really need this for ARIA 1.2.

In practice, ARIA textbox implementations are usually contenteditables, and they wouldn't be helped by this.

If the browser implementers said to go ahead and add it to textbox and searchbox now b/c it is trivial to do at the same time, then I'd be in favor.

curtbellew:
> @curtbellew commented:
> 
> > I think this makes sense. I do wonder how this would be impacted by comboboxes that include a drop down arrow icon. In our example https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html#ex3_label the drop down icon is a child of the combobox and it does have an appropriate aria-label of "Show vegetable options". In this case would we run the risk of the value being calculated to be something like "Apple Show vegetable options"?
> 
> Perfectly fine risk because ARIA 1.2 makes that a deprecated pattern that checkers will call invalid and that AT don't generally support anyway.
> 
> If browsers or AT wanted to do so, they could easily detect the 1.1 implementation and calculate the value strictly based on the value of the textbox, which is what they do now. The revised UA support statement in PR #1223 allows for this.

Thanks for that explanation, Matt.  Makes a lot of sense.  I think I'm confused by the 1.2 ARIA document though.  I was thinking that the document refers to the drop down icon in the following quote.
"Typically, the initial state of a combobox is collapsed. In the collapsed state, only the combobox element and an optional popup control button are visible."

mcking65:
@curtbellew commented:

> Thanks for that explanation, Matt. Makes a lot of sense. I think I'm confused by the 1.2 ARIA document though. I was thinking that the document refers to the drop down icon in the following quote.
> "Typically, the initial state of a combobox is collapsed. In the collapsed state, only the combobox element and an optional popup control button are visible."

Yes, the optional popup control button would be the dropdown icon. It is not part of the combobox element; it is a sibling element.

If there is something ambiguous about that text, I'd be happy to make an editorial PR to clarify.

cookiecrook:
Looking into this. It may be okay, but it would be a change in the way WebKit exposes these to the accessibility APIs on different platforms, so I need to double-check with the API owners.

cookiecrook:
Trying to clarify the comments @mcking65 made on the ARIA call this morning.

I think this would result in very little change to comboboxes that use the role on native text input elements, like this:

```html
<input type="text" role="combobox" aria-owns"my_listbox">
```

But it could be different on this type of variant, which we still see a lot of web authors using.

```html
<div role="combobox" tabindex="0"> <!-- focusable parent node uses the role --> 
	<input type="text" role="textbox"> <!-- contains a textbox -->
	… listbox too...
</div>
```

The way I read the PR is that the text `input` element’s value will be reflected to the parent `div` node in the accessibility APIs? Is that what you intended? 

Thanks for clarifying.

cookiecrook:
@aleventhal wrote:
> Are we sure this won't result in double speech in some screen readers?

Not sure at all. I think it's possible this change would result in double-speak for more multiple assistive technologies. 

@mcking65 wrote:
> Perfectly fine risk because ARIA 1.2 makes that a deprecated pattern that checkers will call invalid and that AT don't generally support anyway.

I believe it's wishful thinking to forego this risk. The conformance checked statement is vague, and unlikely to be implemented, and even if it is implemented by some checkers, most authors don't use accessibility conformance checkers. 

I'm not sure the "UAs must compute value" statement resolves anything. Whether the UA does or the AT does is an implementation detail IMO, as long as the author markup is valid.

Could this issue be moved from the Spec to the ARIA-AAM?

jnurthen:
I think this should be deferred to 1.3 - it doesn't seem to fit with the scope of things that can go into 1.2 at this stage.

mcking65:
cookiecrook commented:

> Trying to clarify the comments @mcking65 made on the ARIA call this morning.
> 
> I think this would result in very little change to comboboxes that use the role on native text input elements, like this:
> 
> ```
> <input type="text" role="combobox" aria-owns"my_listbox">
> ```

That is correct. This PR does not change anything for implementations like this that use aria-owns, which is the 1.0 pattern. First, the combobox is made from a native input, which has a value attribute. So, the requirement is to give the combobox that value. In addition, however, this PR changes the spec to specify that UA calculate value from content only for combobox elements that implement the 1.2 pattern. If The UA were to calculate value from content for the 1.0 style implementation, that would be an error.

The 1.2 pattern changes the above by replacing aria-owns with aria-controls so there is no child content. And, since the input itself has a value, that value is the value of the combobox.

> But it could be different on this type of variant, which we still see a lot of web authors using.
> 
> ```
> <div role="combobox" tabindex="0">  
> 	<input type="text" role="textbox"> 
> 	… listbox too...
> </div>
> ```

This PR should not change value calculation for this code, which is invalid BTW. ARIA has never allowed a focusable parent. 1.1 allows a parent container, but it is not allowed to be focusable.

Again, this UA requirement is only for implementations of the 1.2 pattern. So, if the UA were to calculate a value from content for the above, it would be a UA error.

The 1.2 combobox is an input, not a composite. It is not permitted to have interactive descendants. So, the descendants of the above code should not be included in a value calculation.

> The way I read the PR is that the text `input` element’s value will be reflected to the parent `div` node in the accessibility APIs? Is that what you intended?

I didn't have any intention that would happen, and I am happy to make whatever change necessary to clarify. 

My intention is that The UA should only calculate value from content if the combobox is not an HTML text input with a value attribute  and if it meets the requirements of the 1.2 specification, ie, it is a focusable element with the combobox role, it does not have any interactive or focusable descendants, and it controls a popup element with role listbox, tree, grid, or dialog. If the element with role combobox is a native input with a value attribute, that value should still be used as the combobox value.

In other words, the requirements in the 1.2 version of the spec only apply to authors implementing ARIA 1.2. If the author is using ARIA 1.1, they should only use requirements in ARIA 1.1. And, if the author is using 1.0, the Author should be following ARIA 1.0 requirements.

Similarly, the user agent should implement combobox requirements in the 1.2 specification only for combobox elements that conform to the the 1.2 specification. This is a new requirement in 1.2. I'd be happy to add some clarifying language to either the paragraph that describes this requirement or to the notes at the end after we come to consensus on #1178. Personally, I think it would be better at the end.

mcking65:
@cookiecrook commented:

> @aleventhal wrote:
> 
> > Are we sure this won't result in double speech in some screen readers?
> 
> Not sure at all. I think it's possible this change would result in double-speak for more multiple assistive technologies.

If UA implement this only for 1.2 comboboxes as intended, there won't be any double speak. Check out this codepen by Sarah. (Ignore the second one with valuetext.):
https://codepen.io/smhigley/full/PoqyRdV

> I'm not sure the "UAs must compute value" statement resolves anything. Whether the UA does or the AT does is an implementation detail IMO, as long as the author markup is valid.

It resolves the issue that there is no specification of who should and how should the value of a combobox be coded and computed.

in 1.0, this was also ambiguous. However, in practice,  it was implicit that a combobox had to be made from an HTML text input. No one tried to do it any other way. So, browsers just picked up that value because they already did so for native html inputs.

With 1.1, combobox was required to contain a textbox, and again, that provided an implicit source of value even though the spec didn't directly address the issue of how the author should code or the user agent should compute the value.

In 1.2, a combobox does not have to be made from or contain a textbox. This finally aligns the ARIA spec with the way HTML:select@size=1 has been mapped on several platforms since the beginning of ARIA.

This alignment now makes it possible to build custom versions of HTML select with ARIA. However, we can no longer rely on accidental alignment of the stars around the way textboxes have worked in order for AT to get the value of a combobox that is not built from an HTML input. As we work on a custom select for APG, we have discovered that it is essential the spec be explicit about the source of the value.

mcking65:
@jnurthen wrote:

> I think this should be deferred to 1.3 - it doesn't seem to fit with the scope of things that can go into 1.2 at this stage.

This is a pressing issue, and I would like to urge continued effort to arrive at consensus.

This PR resolves a critical oversight in the prior combobox PR. We discovered this gap through the ARIA WG process, which includes developing practices implementations and testing them with browsers and AT.

There are many people out there who have been struggling to make custom selects that work with AT. This is the final change required to make that feasible. The value of arriving at common understanding of how to do this now is very significant.

sinabahram:
I also would like to see this resolved in 1.2

index.html
```diff
@@ -2132,6 +2132,11 @@ <h2>Definition of Roles</h2>
           For example, one common convention is that <kbd>Down Arrow</kbd> moves focus from the input to the first focusable descendant of the popup element.
           If the popup element supports <pref>aria-activedescendant</pref>, in lieu of moving focus, such keyboard mechanisms can control the value of <pref>aria-activedescendant</pref> on the <code>combobox</code> element.
           When a descendant of the popup element is active, authors MAY set <pref>aria-activedescendant</pref> on the <code>combobox</code> to a value that refers to the active element within the popup while focus remains on the <code>combobox</code> element.
+        </p>
+        <p>
+          User agents MUST compute a value for the <code>combobox</code>.
+          If the <code>combobox</code> element is also a host language element that provides a value, such as an HTML <code>input</code> element, user agents MUST set the value of the combobox to that value.
+          Otherwise, user agents must compute the value of the <code>combobox</code> from the descendant content of the <code>combobox</code> element.
```
carmacleod, line 2139:
```suggestion
          Otherwise, user agents MUST compute the value of the <code>combobox</code> from the descendant content of the <code>combobox</code> element.
```

carmacleod:
I think it's important that we resolve this in 1.2 because I can't imagine telling people that the combobox spec is different in 1.0, 1.1, 1.2 AND 1.3 - that would just be a nightmare.

I don't know whether it belongs in the spec or the AAM.
If we are adding this to the spec, then I am fine with @mcking65's words in this PR (only one nit to fix).
Yes, there are 3 instances of UA MUST, but if we switch them to SHOULD for easier 1.2 merging, with the intent of changing to MUST in 1.3, then we are just back to the nightmare mentioned above. (I believe we have 1 implementor already - see below - and maybe @aleventhal can ensure that Chrome will implement, which would make 2).

I gathered a bit of data using @smhigley's [select-only combobox example](https://codepen.io/smhigley/full/PoqyRdV) (the first one: "Choose a fruit, with text content") that @mcking65 mentioned.

- FF, Chrome, Edge don't set MSAA accValue or IA2 Value. UIA has ValuePattern, but Value is not set.
  - JAWS and NVDA don't say value in FF or Chrome
  - Narrator doesn't say value in Edge
  - VoiceOver doesn't say value in Chrome

- Safari doesn't show AXValue in devtools, but Accessibility Inspector **_does_** show AXValue set to combobox content.
  - VoiceOver **_does_** say value

@cookiecrook Can you confirm that Safari/WebKit is already setting the value for this combobox? (i.e. verify that Apple already conforms to the words in this PR).

@aleventhal You said:
> For example, NVDA already figures this situation out and reads the value
from the combobox's child, IIRC.

Can you confirm that NVDA is _not_ actually reading the value? I am running 2019.3.1, which I believe is the latest, and NVDA is not reading the value of this combobox in either FF or Chrome.

@joanmarie Sorry that I'm not set up to test Linux. Can you give us any data on whether or not Orca reads the value of this combobox?

joanmarie:
@carmacleod Here are my results:

Orca master + Firefox Nightly:
* "Choose a fruit, with text content combo box. Apple. opens listbox"
* "Choose a fruit, with valuetext combo box. Apple. opens listbox"

Orca master + Chromium (locally built):
* "Choose a fruit, with text content Apple combo box. Apple. opens listbox"
* "Choose a fruit, with valuetext Apple combo box. Apple. opens listbox"

Orca master + Epiphany (WebKitGtk) 3.34.4:
* "Choose a fruit, with text content combo box. Apple."
* "Choose a fruit, with valuetext combo box. Apple."

So in all cases Orca is speaking "Apple". I will now see why it's double-presenting it for Chromium.

Does that answer your question? If not, please let me know. And thanks for asking!

carmacleod:
@joanmarie Wow - is Orca being very smart and figuring the value out? Or are the nightly and locally built user agents actually making the value available? Or maybe Orca is being smart and Chromium has implemented very recently?

joanmarie:
> @joanmarie Wow - is Orca being very smart and figuring the value out? Or are the nightly and locally built user agents actually making the value available? Or maybe Orca is being smart and Chromium has implemented very recently?

Orca goes looking for the value/displayed text in a combobox. It always has. Because authors, differences in native toolkits, etc. 

I was multi-tasking at the time I responded to your comment and had the nightly/locally-built versions running. So that's what I reported to you. :smile: But I get the same results with Firefox stable (v74.0.1) as I do with Firefox nightly. And the same results with Chrome stable (80.0.3987.162) as I do with Chromium built today.

carmacleod:
> Orca goes looking for the value/displayed text in a combobox. It always has. Because authors, differences in native toolkits, etc.

Orca is a very wise screen reader. 😄

I assume that, were the browsers to provide a value through their accessibility API, Orca would stop looking and go with that? i.e. Orca would not actually need to make any changes if browsers implement the words in this PR? (Did you figure out the double-speak? I'm assuming it's just some nit, and does not actually indicate that Chromium is providing a value... since it doesn't provide one in Windows Chrome...)

joanmarie:
> I assume that, were the browsers to provide a value through their accessibility API, Orca would stop looking and go with that?

If that is the consensus, yes.

> i.e. Orca would not actually need to make any changes if browsers implement the words in this PR?

Not necessarily correct. The spec states that user agents need to calculate a value. But it doesn't state how that value should be exposed (does it?).

In fact, at the platform-independent / pure-ARIA level, where is the calculated value going to go?
* `aria-valuetext`
* name of the `listbox`?
* someplace else?

The answer to the above question determines what mappings will be needed (if any) in the Core-AAM.

Right now, Orca is not looking at AtspiObject attributes for any value (which is what it would need to start doing should `aria-valuetext` be the place where exposure is done).

Potentially-related aside, I initially only tested what happened when the combo box got focus; not what happened when you arrow in the expanded list box. I have just updated my earlier comment to reflect that when the "with valuetext" example is expanded, Orca does not present anything. Things work just fine for the "with text content" version.

> (Did you figure out the double-speak? I'm assuming it's just some nit, and does not actually indicate that Chromium is providing a value... since it doesn't provide one in Windows Chrome...)

Yeah, fixed that a while ago. (And had updated the comment above to reflect that. I figure that there should just be one place where current results should go.)

joanmarie:
When I up and down arrow in the "with valuetext" version, I am not getting any accessibility events. Are we missing an `aria-activedescendant` change?

carmacleod:
> When I up and down arrow in the "with valuetext" version, I am not getting any accessibility events. Are we missing an aria-activedescendant change?

We're only looking at the "with text content" one (the other one is an experiment) - sorry for the confusion!

Will get back to you about the other questions - just thinking things through...

aleventhal:
Can someone provide me with 1 or more reasonable examples that authors might actually do, and where the combobox value should be computed by the UA? I'm not sure we're all working on the same thing. For example, one of the examples above had 2 items in the tab order, which wouldn't be very usable. @joanmarie what markup were you testing with?

joanmarie:
> @joanmarie what markup were you testing with?

The example page @carmacleod mentioned above: https://github.com/w3c/aria/pull/1225#issuecomment-610482361

aleventhal:
@joanmarie @carmacleod In that example page, there is no descendant of role=combobox, in the DOM or AX tree, that has a value.

index.html
aleventhal: Computing the value from the descendant content -- how? At first I assume it was copying a value from a descendant.

aleventhal:
> @joanmarie @carmacleod In that example page, there is no descendant of role=combobox, in the DOM or AX tree, that has a value.

I understand now. The new spec text is not requesting that the UA copy the value from a descendant, but rather compute the value from the text contents. I misunderstood.

Should it apply the name calculation to the descendant contents, and then apply that to the value? Otherwise, what does the computation look like?

smhigley:
@aleventhal I'd pictured it working similar to a live region, computing name from contents for the combobox, and then applying that to the value.

@joanmarie the second combo's activedescendant should work now. I was copypasta-ing stuff to test, and hadn't bothered to change the `id`s 😁

joanmarie:
> @aleventhal I'd pictured it working similar to a live region, computing name from contents for the combobox, and then applying that to the value.

I'm still confused about what "value" means here? Comboboxes don't have any ARIA properties related to value, right? So once this calculation takes place what does the result get applied *to*?

And whatever the answer to my question is, I think that answer belongs as part of the spec text.

> @joanmarie the second combo's activedescendant should work now. I was copypasta-ing stuff to test, and hadn't bothered to change the `id`s

Thanks! Now the value is being presented by Orca, all without any need to change the spec text. :wink:

index.html
joanmarie: > Computing the value from the descendant content -- how? 

And exposing it where? I don't see how user agents can know what to implement based on the current text.

carmacleod:
@aleventhal I've been trying to figure out what the best way to expose this is. Could definitely use your help - been a while since I've looked at these APIs. See table below - I don't know if that's the right direction - please advise. I also don't know if we would need valuechanged events?

@joanmarie said:
> The spec states that user agents need to calculate a value. But it doesn't state how that value should be exposed (does it?).
> In fact, at the platform-independent / pure-ARIA level, where is the calculated value going to go?

I guess the idea was to expose it in the same way that browsers currently expose custom textbox values, i.e. compute value from text content and expose in appropriate "value" platform API, but not part of platform-independent ARIA API. We weren't trying to go full-fledged `aria-valuetext` at this time - we can think about that for 1.3.

Not sure if this could/would go in this "special" section in Core-AAM [Special Processing Requiring Additional Computation](https://w3c.github.io/core-aam/#mapping_additional). I actually can't find where the observed textbox "value" behavior is specified. It's not in the [Core-AAM table entry for textbox](https://w3c.github.io/core-aam/#details-id-87). It seems to be just done magically or by some unwritten consensus... ;)

I wrote a little [textbox example](https://carmacleod.github.io/playground/text-input.html) to compare native text inputs with custom textboxes. **NOTE:** The most interesting one is the last one, with label "Custom read-only textbox using div with aria-readonly:". It has text content, and Chrome/FF on Win expose it in some kind of "Value" platform location. JAWS and NVDA both read the value. We want that for combobox, too. (VoiceOver reads it too, but AFAICT Safari isn't providing it? For some reason, it doesn't think my custom read-only textbox is a valid textbox).

Just looking through the platform A11y APIs, I am guessing that we would need something like the following (but like I said, need help here from people who are currently using the platform A11y APIs):

Platform | A11y API
------------ | -------------
MSAA + IAccessible2 | [IAccessible::get_accValue](https://docs.microsoft.com/en-us/windows/win32/api/oleacc/nf-oleacc-iaccessible-get_accvalue)
UIA | [ValuePattern.Value](https://docs.microsoft.com/en-us/dotnet/api/system.windows.automation.valuepattern.valuepatterninformation.value?view=netframework-4.8#System_Windows_Automation_ValuePattern_ValuePatternInformation_Value)
ATK/AT-SPI | [atk-value-get-value-and-text](https://developer.gnome.org/atk/stable/AtkValue.html#atk-value-get-value-and-text) and/or a "value" object attribute?
AX API | AXValue?

Not sure if platform valuechanged event would be needed also, but maybe?

Platform | A11y Event API
------------ | -------------
MSAA + IAccessible2 event | EVENT_OBJECT_VALUECHANGE
UIA event | PropertyChangedEvent<br>Property: AriaProperties?? Value??
ATK/AT-SPI event | object:property-change:accessible-value?
AX API Notification | AXValueChanged

aleventhal:
@carmacleod that looks right. In Chrome all we need is to expose it with the internal value, and the right thing will happen on each platform, including value change events. It should end up looking like <select size=1>. I'll add automated tests to be sure, and we'll need to check it with various screen readers. Hopefully I'll get this into a Canary build soon and whoever wants to can test it.

joanmarie:
@carmacleod + @aleventhal: If you look on the AtkValue page you linked to, you will see:

> Description
> AtkValue should be implemented for components which either display a value from a bounded range, or which allow the user to specify a value from a bounded range, or both. 

It's for things like spinbuttons and sliders and progress bars and the like.

carmacleod:
@joanmarie Oh, right. Rats.

So does ATK/AT-SPI expose the text content of a custom read-only textbox like the one labelled "Custom read-only textbox using div with aria-readonly:" in my little [textbox example page](https://carmacleod.github.io/playground/text-input.html)? (i.e. the last textbox on the page).

If so, then how is it exposed? Is it using a "Value" object attribute? Something else?

If the custom textbox value is _not_ exposed, then how should it be exposed?
Combobox should do the same thing, whatever that is.

joanmarie:
@carmacleod Yes, your custom read-only text box has a value I can get: Namely it's text content which is exposed via AtkText/AtspiText. That's how Orca gets it. That's how it gets "Apple" for the combobox examples. And it has nothing to do with ARIA. AtspiText is how I get the text from divs and spans and headings and paragraphs and ....

And that's why I continue to be confused as to why this is a problem for Windows. Because for both Firefox and Chrome, the ATK and IA2 implementations are nearly identical. If I can get the text, I would think that NVDA can too. Unless I'm missing something.

And it seems macOS can get the same info already via AXValue.

And if all this is the case, why do user agents need to calculate anything?

aleventhal:
Hi Joanie, on Windows/MSAA/IA2, the value is used for more things, such as the current value of a combo box. Windows screen readers look for that and read changes to it in a combobox. If we want to avoid having Windows screen readers write extra code for this ARIA combobox case, we need to expose the value so the ARIA combobox looks like every other.

carmacleod:
@joanmarie Good question. I guess it's a philosophical issue... i.e. who should implement this thing - the UAs or the ATs. :) At the moment, for `combobox`, the UAs aren't calculating, and the ATs aren't looking.

From what I saw, the Windows UAs _do_ calculate this for `textbox`, and they put it in "a place called Value" that I can see in the Accessibility Tree. I believe AT can access this "Value" using `MSAA IAccessible::get_accValue` or `UIA Value.Value` because I see the same string in Inspect.exe for both of those when I look at my [Custom read-only textbox using div with aria-readonly](https://carmacleod.github.io/playground/text-input.html). Screen snapshots for custom read-only `textbox` below.

Screen snapshot 1 of 2 shows "Value" in Chrome's Accessibility Tree, and MSAA "Value" in Inspect.exe:

image

----
Screen snapshot 2 of 2 shows "Value" in Firefox's Accessibility Tree, and UIA "Value.Value" in Inspect.exe:

image

index.html
mcking65: I'm about to make a commit that rewords this so this will become obsolete thread. The commit answers Aaron's question but not Joanie's.

Perhaps core-aam will someday address how to map the value of elements that have a value. But, James Craig and Aaron have explained to me that the lack of specification in AAM for things like the value of textbox (and combobox) is just an artefact of the fact that AAM is known not to be complete.  The AAMs have been built just as much by documenting what user agents do as opposed to specifying everything they need to do.

Perhaps in ARIA 1.3 we could add a valuedby attribute and address the core-aam gap. And, at the same time, we could write some language in the naming spec that specifies how valuedby computes. In the meantime, we don't need that to complete this requirement; we can rely on browser implementers to map value to accessibility APIs the same way they do now for textbox elements.

mcking65:
The [preview](https://raw.githack.com/w3c/aria/issue711-combobox-value-from-content/index.html#combobox) now shows a slightly revised wording based on discussion this morning with @cookiecrook and @aleventhal.

This revision specifies how to compute the value. It also specifies that user agents need to expose the value to assistive technologies, aligning the wording with other similar normative requirements placed on user agents.

mcking65:
@joanmarie commented:

> And that's why I continue to be confused as to why this is a problem for Windows. Because for both Firefox and Chrome, the ATK and IA2 implementations are nearly identical. If I can get the text, I would think that NVDA can too. Unless I'm missing something.
> 
> And it seems macOS can get the same info already via AXValue.
> 
> And if all this is the case, why do user agents need to calculate anything?

Even if Orca is extracting that text and speaking it, Orca does not know that the text is a value. Many screen readers treat value separate from name for widgets that have a value. But, they can't do that if the value of the widget isn't where it is expected to be in the accessibility API. The screen reader should be able to render the widget as an element that has a value.

Also, browsers  need a value for step 2E of name calculation:

>E. Otherwise, if the current node  is a control embedded within the label (e.g. the label  element in HTML or any element directly referenced by aria-labelledby) for another widget, where the user can adjust the embedded control's value, then include the embedded control as part of the text alternative in the following manner:

It's important we specify what to do for that case. BTW, the combobox step under 2e is out-of-date. It says to use the selected option. However, starting with ARIA 1.1, a combobox can control a grid, tree, or dialog, so there isn't always an option element. So, that needs to be fixed when we update accname.

So, this is not a Windows-only gap in the combobox specification. Arguably, this is a gap that has existed since ARIA 1.0. But, since combobox elements have pretty much always been built using HTML input elements, this gap hasn't really mattered in practice. Now, as we attempt to build combobox elements that function like an HTML select, this gap is really critical.

People not only need to mimic HTML select but also need to make elements that function like it but popup elements other than an ordinary listbox, e.g., dialogs or grids.

index.html
```diff
@@ -2127,11 +2127,16 @@ <h2>Definition of Roles</h2>
...
           In addition, to be keyboard accessible, authors SHOULD provide keyboard mechanisms for moving focus between the <code>combobox</code> element and elements contained in the popup.
           For example, one common convention is that <kbd>Down Arrow</kbd> moves focus from the input to the first focusable descendant of the popup element.
           If the popup element supports <pref>aria-activedescendant</pref>, in lieu of moving focus, such keyboard mechanisms can control the value of <pref>aria-activedescendant</pref> on the <code>combobox</code> element.
           When a descendant of the popup element is active, authors MAY set <pref>aria-activedescendant</pref> on the <code>combobox</code> to a value that refers to the active element within the popup while focus remains on the <code>combobox</code> element.
+        </p>
+        <p>
+          User agents MUST compute a value for elements with role <code>combobox</code> and expose it to <a>assistive technologies</a>.
```
cookiecrook, line 2137: `it` is grammatically ambiguous here. Replace `it` with `the value` and you should be good.

index.html
```diff
@@ -2127,11 +2127,16 @@ <h2>Definition of Roles</h2>
...
+          If the <code>combobox</code> element is also a host language element that provides a value, such as an HTML <code>input</code> element, the value of the combobox is the value of that element.
+          Otherwise, the value of the <code>combobox</code> is computed from descendant elements of the <code>combobox</code> using the same algorithm that is used to compute the name of a <rref>button</rref> from its descendant content.
```
cookiecrook, line 2139: > using the same algorithm that is used to compute the name of a button

I can live with this change as-is, but are we postponing a probleme by not adding a "value computation" to ACC-NAME?

index.html
mcking65: Fixed in 5c24ca8.

index.html
mcking65: Since accname 1.2 is on an independent, later  time table, we need to write this based on accname 1.1.

I think we can draft a value calculation in accname 1.2 and then revise this wording in ARIA 1.3 after accname 1.2 is final.

We could base the accname 1.2 draft on what we would also need if we want to revise aria-valuetext to have a broader scope and get closer to attribute parity with HTML value. Such changes to accname would also provide the support needed for an aria-valuedby attribute if we decide we want that as well.

mcking65:
Carolyn, thank you for all the fantastic analysis!!

@carmacleod commented:

>I guess the idea was to expose it in the same way that browsers currently expose custom textbox values, i.e. compute value from text content and expose in an appropriate "value" platform API,

Yes, exactly.

>but not as part of the platform-independent ARIA API.

Precisely, not until we have a value algorithm in accname as suggested by @cookiecrook and then an appropriate attribute mapping, possibly via redefined aria-valuetext, mapped in core-aam.

>I actually can't find where the observed textbox "value" behavior is specified. It's not in the Core-AAM table entry for textbox. It seems to be just done magically or by some unwritten consensus... ;)

It is due to unwritten consensus. @cookiecrook and @aleventhal have described to me how core-aam is not even close to comprehensive. They described this as a rather minor omision compared to other user agent behaviors that are not documented.

But, if we are going to change aria-valuetext or add aria-valuedby, we would definitely want to close the related gap in core-aam.

aleventhal:
I forked the codepen to do 2 things:
1. Remove the use of aria-autocomplete, which I don't think makes sense on a select-only combobox
2. Get rid of the second combobox, which uses aria-valuetext, a technique that we aren't discussing supporting

https://codepen.io/aleventhal/pen/MWaWpbb

carmacleod:
@joanmarie 
> I'm still somewhat bothered by the fact that we're telling user agents to calculate a value without associating an ARIA property with it.

User agents are already doing this for `textbox` value, so it's not a new thing... I guess the only new bit about it is that we're telling them to do it. (We should tell them to do it for textbox also, just to close the gap. <g>).

index.html
carmacleod: @mcking65 Reminder that on the call today, @joanmarie stated a preference for not using the word compute or computed, and instead using something more along the lines of... "exposed using the same method [as for determining the label of a button]" or similar.

JAWS-test:
I also prefer to allow aria-valuetext for more ARIA roles. Reason: https://github.com/w3c/aria/issues/711

jnurthen:
@JAWS-test expanding the roles which support aria-valuetext is not under consideration in this issue. Please only add comments relevant to this issue here so we don’t get sidetracked. If you want to comment about aria-valuetext being allowed on more roles #711 would be a more appropriate place.

aleventhal:
Hi all, this morning's Canary (84.0.4110.0) has an implementation of this. I tested with JAWS, NVDA and VoiceOver and it seems to work on. Feel free to let me know of any issues you notice.
I tested with this codepen: https://codepen.io/aleventhal/full/MWaWpbb

carmacleod:
Confirmed that in @aleventhal 's Chrome Canary, with focus on the "Select-only" combobox:
- In DevTools Accessibility panel, Computed Properties for the combobox has Value: "Banana"
- In Inspect.exe for the combobox, MSAA has Value: "Banana"
- In Accessibility Insights for the combobox, the UIA ValuePattern's Value = "Banana"
- NVDA, JAWS, and Narrator all read the newly-available Value 🎉 

Here's the detailed screen reader output for typing tab key to give focus to combo, 3 down arrows to expand list and move down 2 items, enter key to choose item, close list and return focus to combo.
(Note: Narrator only expands on alt+down arrow. It closes on enter but does not choose option).

<details>
<summary>NVDA output</summary>
Choose a fruit  combo box  Banana  collapsed<br>
expanded<br>
list<br>
Apple  not selected  1 of 8<br>
Banana  2 of 8<br>
Cherry  not selected  3 of 8<br>
Durian  not selected  4 of 8<br>
Choose a fruit  combo box  Durian  collapsed<br>
</details>

<details>
<summary>JAWS output</summary>
Choose a fruit Combo box <br>
Banana<br>
To change the selection use the Arrow keys.<br>
List box<br>
Banana <br>
2 of 8<br>
To move to an item press the Arrow keys.<br>
Cherry <br>
3 of 8<br>
To move to an item press the Arrow keys.<br>
Durian <br>
4 of 8<br>
To move to an item press the Arrow keys.<br>
Choose a fruit Combo box <br>
Durian<br>
To change the selection use the Arrow keys.<br>
</details>

<details>
<summary>Narrator output</summary>
Choose a fruit Banana combo box collapsed<br>
Banana 2 of 8 selected<br>
3 of 8 Cherry<br>
4 of 8 Durion<br>
Command Not available<br>
Choose a fruit Banana combo box collapsed<br>
</details>

Will test Chrome+VO on Mac next.

joanmarie:
> Hi all, this morning's Canary (84.0.4110.0) has an implementation of this. I tested with JAWS, NVDA and VoiceOver and it seems to work on. Feel free to let me know of any issues you notice.
> I tested with this codepen: https://codepen.io/aleventhal/full/MWaWpbb

I just modified Orca to look for the `valuetext` object attribute for comboboxes and fall back on its normal figure-out-the-value handling. It works with Chromium built locally (which has Aaron's change) and Chrome Unstable (which lacks it). Linux apparently still isn't yet cool enough for Canary. :grinning:

carmacleod:
Nice, @joanmarie !

I can confirm that VoiceOver works (but only if you use VO key navigation to operate the combobox. Not sure if it we expect combobox to be operable with simple arrow keys or not.)

<details>
<summary>VoiceOver output</summary>
Banana, Choose a fruit, collapsed, button<br>
You are currently on a button, inside of a frame. To display a list of options, press Control-Option-Space. To exit this web area, press Control-Option-Shift-Up Arrow.<br>
<code>Control-Option-Space</code><br>
Banana, Choose a fruit, expanded, button<br>
<code>Control-Option-Right Arrow</code><br>
List box 1 item selected. Banana (2 of 8)<br>
You are currently on a list box, inside of a frame. To begin interacting with the items in this list, press Control-Option-Shift-Down Arrow. To exit this web area, press Control-Option-Shift-Up Arrow.<br>
<code>Control-Option-Shift-Down Arrow</code><br>
In list box 1 item selected. Banana (2 of 8) Banana (2 of 8) [says item twice for some reason]<br>
<code>Control-Option-Right Arrow</code><br>
Cherry, (3 of 8)<br>
You are currently on a text element, inside of a list box.<br>
<code>Control-Option-Right Arrow</code><br>
Durian, (4 of 8)<br>
You are currently on a text element, inside of a list box.<br>
<code>Control-Option-Space</code><br>
Durian, Choose a fruit, collapsed, button<br>
</details>

----
I can also confirm that Accessibility Inspector shows the Value (I'm guessing that's the AXValue I outlined in red in this screen snapshot? @cookiecrook ? It used to say "AXValue", but now it just says "Value".)

image

JAWS-test:
@jnurthen 
My comment refers to https://github.com/w3c/aria/pull/1225#pullrequestreview-390871113 and it concerns exactly the topic discussed here: How can the value of a combobox be determined. The idea is to use the text content. I don't think this is a good idea, because there are HTML elements without text content. The argument was already mentioned in the other issue and that's why I linked it.

jnurthen:
@Jaws-test please stop. #711 was already linked in the first comment. This is adding nothing but noise to this issue.

mcking65:
@joanmarie, @cookiecrook, I revised the language to remove the word "compute".

index.html
```diff
@@ -2127,12 +2127,20 @@ <h2>Definition of Roles</h2>
...
           If the popup element supports <pref>aria-activedescendant</pref>, in lieu of moving focus, such keyboard mechanisms can control the value of <pref>aria-activedescendant</pref> on the <code>combobox</code> element.
           When a descendant of the popup element is active, authors MAY set <pref>aria-activedescendant</pref> on the <code>combobox</code> to a value that refers to the active element within the popup while focus remains on the <code>combobox</code> element.
         </p>
+        <p>
+          User agents MUST expose the value of elements with role <code>combobox</code> to <a>assistive technologies</a>.
+          The value of a <code>combobox</code> is represented by one of the following:
+        </p>
+        <ul>
+          <li>If the <code>combobox</code> element is a host language element that provides a value, such as an HTML <code>input</code> element, the value of the combobox is the value of that element.</li>
+          <li> Otherwise, the value of the <code>combobox</code> is represented by its descendant elements and can be determined using the same method used to compute the name of a <rref>button</rref> from its descendant content.</li>
```
aleventhal, line 2142: Nit: extra space after <li>

jnurthen:
Previews failing due to spec-generator being down 
https://github.com/w3c/spec-generator/issues/292

PR #1227, opened 2020-03-29
Remove Ajax from definition of live region

### LINKED ISSUE #1160: Remove the term "Ajax" from the spec
From the definition of Live region:

> These regions are not always updated as a result of a user interaction. This practice has become commonplace with the growing use of Ajax

This is such a dated term, would be good to take it out.

carmacleod:
Closes #1160

### DIFF
--- a/common/terms.html
+++ b/common/terms.html
@@ -99,7 +99,7 @@
     </dd>
     <dt><dfn data-lt="live region|live regions">Live Region</dfn></dt>
     <dd>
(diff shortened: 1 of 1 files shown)


carmacleod:
Aha! So it _is_ getting overwritten! I should have gone with my first instinct, but git-blame said michael-n-cooper, so I had to believe it. Can we create some kind of bot account for making those automatic commits, so that there's a hint where to look when things go wrong? (also, if there's conflicts, maybe the copy can fail and ping somebody?)

Anyhow, I'll try to figure out how to retarget this pull request against aria-common... and if I fail to figure out the correct git incantation to do that, I'll create a new PR. :)

pkra:
@carmacleod try this: on this page, activate the edit button (before the h1 in the DOM but after the h1 visually). This lets you not only change the title but also the base branch to point the PR to.

carmacleod:
Thanks, @pkra, that's great to know for branches in a repository, but I can't see any way to retarget a commit across repositories. So I'm about to commit the change and create a new PR against https://github.com/w3c/aria-common - hmmm, hope I don't have to create a fork - will test that theory in a few seconds...

carmacleod:
Phew - it worked. So I'll close this PR and add all of the reviewers to the other PR.
Sorry! But if people don't mind reviewing again... one more time, with feeling! ;)

jnurthen:
already merged it :)

PR #1247, opened 2020-04-21
chore: remove common/script/utility.js - not used anywhere

marcoscaceres:
This JS file is not used anywhere?


jnurthen:
This is used in the AAMs - I checked core-aam and it is used there.

Everything in the common directory is maintained in aria-common and copied to the following repositories.
accname master
annotation-aria master
aria master
aria stable
aria-practices master
core-aam master
core-aam stable
css-aam master
dpub-aam master
dpub-aria master
graphics-aam master
graphics-aria master
math-aam master
math-aria master
mathml-aam master
svg-aam master

jnurthen:
This is used in the AAMs - I checked core-aam and it is used there.

Everything in the common directory is maintained in aria-common and copied to the following repositories.
accname master
annotation-aria master
aria master
aria stable
aria-practices master
core-aam master
core-aam stable
css-aam master
dpub-aam master
dpub-aria master
graphics-aam master
graphics-aria master
math-aam master
math-aria master
mathml-aam master
svg-aam master

marcoscaceres:
oh... why are things copied? that seems extremely fragile given you can just `<script src>` (or `npm install` or even use git sub-modules)?

jnurthen:
> oh... why are things copied? that seems extremely fragile given you can just `<script src>` (or `npm install` or even use git sub-modules)?

The answer is likely - that is the way we have done it for years and years. it certainly pre-dates my involvement with the WG. @michael-n-cooper any specific reasons you can think of?

marcoscaceres:
yeah, I guess we can chalk it down to legacy and that's ok. I guess as long as there is some kind of system in place to propagate changes then it should be fine.

marcoscaceres:
ok, so, let's close this for now, and we can then figure out if utility.js can be deleted, or added directly to a particular spec. It seems to have something to do with "Platform Mapping Maintainers"? Any idea what that might be referring to?

PR #1248, opened 2020-04-21
chore: prevent crash when respecConfig.definitionMap is missing

marcoscaceres:
This prevents the respecConfig.definitionMap crash.

However, we need to trim common/terms.html to the ones that are used in this spec. And then export those terms so other specs can use them.


michael-n-cooper:
This needed to be done in the aria-common repository, see w3c/aria-common@07da795ef2372414b54e05558deaa2bf21d68610

Issue #1110, opened 2019-10-31
AriaAttributes has many integer attributes with string types


annevk:
It would be nicer if these were reflected as numbers. E.g., aria-level.

cc @alice @domenic

domenic:
If we do this (we probably should), then a choice needs to be made for each one, between the different types of [reflection](https://html.spec.whatwg.org/#reflect):

- `long`
- `long` limited to only non-negative numbers
- `unsigned long`
- `unsigned long` limited to only non-negative numbers greater than zero
- `unsigned long` limited only to non-negative numbers greater than zero with fallback
- `unsigned long` clamped to the range [min, max]
- `double`
- `unrestricted double`
-` double` limited to numbers greater than zero
- `unrestricted double` limited to numbers greater than zero

Probably don't consider the `long` ones.

alice:
I have a vague memory of discussing this and concluding that string reflection was the way to go, but I don't remember any details. @cookiecrook may remember?

alice:
Listing out non-string types per the ARIA value types (not including token types or true/false/undefined):
- `aria-atomic` (true/false, i.e. empty string is invalid but missing value default is false)
- `aria-busy` (true/false)
- `aria-colcount` (integer >= -1, although zero would be weird)
- `aria-colindex` (integer >= 1)
- `aria-colspan` (integer >= 1 - this is weirdly inconsistent with aria-rowspan)
- `aria-disabled` (true/false)
- `aria-level` (integer >= 1)
- `aria-modal` (true/false)
- `aria-multiline` (true/false)
- `aria-multiselectable` (true/false)
- `aria-posinset` (integer >= 1)
- `aria-readonly` (true/false)
- `aria-required` (true/false)
- `aria-rowcount` (integer >= -1)
- `aria-rowindex` (integer >= 1)
- `aria-rowspan` (integer >= 0 - weirdly inconsistent with aria-colspan)
- `aria-setsize` (integer >= -1)
- `aria-valuemax` (integer)
- `aria-valuemin` (integer)
- `aria-valuenow` (integer)

The true/false ones are interesting, since they're almost but not quite boolean.

annevk:
From Domenic It sounded like there was a preference for true/false values to be enumerated values so they can be extended in the future. I kinda wish they would not have used true and false as values then, but so be it. So those probably have to remain strings, though limited to known values would be nice.

[...]

PR #996, opened 2019-06-13
Initial aria-textseparation (depends on generic PR being merged)

carmacleod:
Separated `aria-textseparation` attribute definition from `generic` role definition.
This PR depends on https://github.com/w3c/aria/pull/805 being merged first.
This markup contains inline comments describing what would need to be manually merged.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jan 20, 2021, 10:59 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Faria%2F3b0182354d6553c914fe1c1b853d7f6f9cfe940c%2Findex.html%3FisPreview%3Dtrue)

```
waiting for function failed: timeout 30000ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%23996.)._
</details>

PR #1261, opened 2020-05-07
resolves #1058, enumerated values for DOMString reflection

### LINKED ISSUE #1058: AriaAttributes cannot be nullable
At least as currently defined it seems to use HTML's reflect concept and that only works for normal strings, not nullable strings.

cookiecrook:
This prose addition should reconcile our use of DOMString reflection for enumerated values for #1058.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 22, 2020, 9:31 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Faria%2F62603125ff5ada131c69644faa40cd8d26632ec2%2Findex.html%3FisPreview%3Dtrue)

```

[36mSpecification: https://rawcdn.githack.com/w3c/aria/62603125ff5ada131c69644faa40cd8d26632ec2/index.html?isPreview=true&publishDate=2020-05-22[39m
[36mReSpec version: 25.6.0[39m
[36mFile a bug: https://github.com/w3c/respec/[39m
[36mError: Error: Evaluation failed: Timeout: document.respecIsReady didn't resolve in 5093ms.[39m
[36m    at ExecutionContext._evaluateInternal (/u/spec-generator/node_modules/puppeteer/lib/ExecutionContext.js:122:13)[39m
[36m    at runMicrotasks (<anonymous>)[39m
[36m    at processTicksAndRejections (internal/process/task_queues.js:97:5)[39m
[36m    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer/lib/ExecutionContext.js:48:12)[39m
[36m    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:128:12)[39m
[36m    at async fetchAndWrite (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:95:18)[39m
[36m    at async Object.generate [as respec] (/u/spec-generator/generators/respec.js:14:20)[39m
[36m    at async generate (/u/spec-generator/server.js:91:29)[39m
[36m  -- ASYNC --[39m
[36m    at ExecutionContext.<anonymous> (/u/spec-generator/node_modules/puppeteer/lib/helper.js:111:15)[39m
[36m    at DOMWorld.evaluate (/u/spec-generator/node_modules/puppeteer/lib/DOMWorld.js:112:20)[39m
[36m    at runMicrotasks (<anonymous>)[39m
[36m    at processTicksAndRejections (internal/process/task_queues.js:97:5)[39m
[36m  -- ASYNC --[39m
[36m    at Frame.<anonymous> (/u/spec-generator/node_modules/puppeteer/lib/helper.js:111:15)[39m
[36m    at Page.evaluate (/u/spec-generator/node_modules/puppeteer/lib/Page.js:860:43)[39m
[36m    at Page.<anonymous> (/u/spec-generator/node_modules/puppeteer/lib/helper.js:112:23)[39m
[36m    at generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:128:23)[39m
[36m    at runMicrotasks (<anonymous>)[39m
[36m    at processTicksAndRejections (internal/process/task_queues.js:97:5)[39m
[36m    at async fetchAndWrite (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:95:18)[39m
[36m    at async Object.generate [as respec] (/u/spec-generator/generators/respec.js:14:20)[39m
[36m    at async generate (/u/spec-generator/server.js:91:29)[39m
[36m[39m
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231261.)._
</details>


cookiecrook:
@alice @annevk @domenic

carmacleod:
You may want to link to https://w3c.github.io/aria/#propcharacteristic_value somewhere in that prose... perhaps link "allowed values" or just the word "values"?

domenic:
This definitely helps, thanks!

What about when the table does not notate one of the values as "(default)"?

It would also be good to link:

- "enumerated attribute" to https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute
- "invalid value default" to https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#invalid-value-default
- "missing value default" to https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#missing-value-default

cookiecrook:
> What about when the table does not notate one of the values as "(default)"?

Good spot. There is only one: `aria-relevant` Filed as #1265.

cookiecrook:
It occurred to be that we may need to mention the difference between enumerated attributes that take a single value (token, tristate, etc) versus those that take multiple values (token list), including `aria-relevant` whose default is the two values "additions text"…

domenic:
One additional thing, is that you should add a similar blanket statement making it clear what the states and their corresponding keywords are. Something like "Each value in the table is a keyword for the attribute, mapping to a state of the same name".

Stepping back a bit further, it would probably make sense to move this out of the "ARIA Attribute Reflection" section, since it has a broader impact on the processing model of these attributes. E.g. it implies that:

- User agents must treat their values as ASCII case-insensitive
- User agents must treat both missing and invalid values as the default (whereas before it wasn't clear what to do with invalid values). Note that unless you define otherwise somewhere, "" triggers the same default (since it's not in the table, and is thus invalid).

domenic:
> those that take multiple values (token list), including aria-relevant whose default is the two values "additions text"…

Hmm, yeah, this doesn't use the enumerated attribute framework in HTML then, and will need a different processing model and reflection support. A precedent in HTML is [`rel`](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-rel), which states ([later](https://html.spec.whatwg.org/multipage/links.html#linkTypes)):

> To determine which link types apply to a link, a, area, or form element, the element's rel attribute must be split on ASCII whitespace. The resulting tokens are the keywords for the link types that apply to that element.

and

> Keywords are always ASCII case-insensitive, and must be compared as such.

It sounds like you'd want to layer some additional processing on top of that, e.g. if the resulting set is empty, then set it to « "additions", "text" ».

How many such cases are there? Maybe we should make this the HTML editor's job, to create a common infrastructure (like enumerated attributes) which could be shared by the relevant ARIA attributes and `<link>`/`<a>`'s `rel=""` and probably some others.

cookiecrook:
> How many such cases (of token lists) are there? Maybe we should make this the HTML editor's job, to create a common infrastructure (like enumerated attributes) which could be shared by the relevant ARIA attributes and <link>/<a>'s rel="" and probably some others.

In ARIA 1.2 there are two attrs, but only one relevant in the IDL context:
- `aria-dropeffect` (deprecated in ARIA 1.1, so not included in IDL reflection)
- `aria-relevant`

cookiecrook:
> it would probably make sense to move this out of the "ARIA Attribute Reflection" section, since it has a broader impact on the processing model of these attributes. 

Yes. I think it makes sense to include this in the values section, rather than just linking to it from the IDL section. I'll think on this a bit more before updating the PR. Thanks for your help.

domenic:
Hmm OK. That's a very interesting attribute.

I see two options:

### IDL attribute directly manipulates and reflects the content attribute

This version would involve changing `ariaRelevant` to `[SameObject, PutForwards=value] readonly attribute DOMTokenList ariaRelevant;`, which will trigger the very last paragraph of https://html.spec.whatwg.org/#reflecting-content-attributes-in-idl-attributes and take care of all the IDL stuff for you. It will allow authors to manipulate the result using the convenient features of `DOMTokenList` and the `[PutForwards]` annotation, e.g.:

```js
el.ariaRelevant = "aDDitions removals"; // shortcut for el.ariaRelevant.value = "aDDitions removals"
el.ariaRelevant.remove("removals");
el.ariaRelevant.add("all");
console.assert(el.ariaRelevant.value === "aDDitions all");
```

If you do that, then the resulting `.ariaRelevant` property will not contain `"additions"` or `"text"` by default; it will reflect the value of the attribute exactly. (So e.g. `el.ariaRelevant.contains("additions")` will be false, if the attribute is not present.)

This follows a number of attributes in HTML, like `link.relList`, `link.sizes`, `iframe.sandbox`, etc.

### IDL attribute is customized to reflect the data model better

This version would not use HTML's attribute reflection framework at all, and would not use `DOMTokenList`. Instead it would work to produce something that reflects the underlying data model, by defining custom getters and setters.

The setter would probably just set the attribute to the given value.

The getter would figure out the normalized list of tokens, and return a canonicalized form. For example, it could normalize `"all"` to `"additions removals text"`, and would normalize invalid values by removing them. It could even sort them. So e.g. I'm imagining

```js
el.ariaRelevant = "all      asdf additions !";
console.assert(el.ariaRelevant === "additions removals text");

el.ariaRelevant = "additions text REMOvALS   ";
console.assert(el.ariaRelevant === "addition removals text");

el.ariaRelevant = "";
console.assert(el.ariaRelevant === "additions text";
```

This kind of "custom reflection" has some precedent in HTML in attributes like `el.dir`, `el.spellcheck`, and and `el.autocapitalize`.

cookiecrook:
Okay. I think I'm going to file `aria-relevant` reflection as a separate issue and remove it from the 1.2 milestone in this PR, along with moving the line of prose to the values section.

carmacleod:
Just thought I'd mention that there's one more token list attribute that did not make it into ARIA 1.2, that may (or may not) go into ARIA 1.3.

Here's the preview for [aria-textseparation](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/996.html#aria-textseparation) from PR https://github.com/w3c/aria/pull/996. The default value is a single token, `style`.

It's a particularly weird attribute because it can only take 1 or 2 tokens. :)

annevk:
Looking at this PR I'm not sure how it solves the nullable problem? Or does it help with #1058, but not fully resolve it?

Also, `role` also seems to be defined as a list value or is that not considered an ARIA attribute? Either way it seems to have the same problem.

carmacleod:
`role` is an ARIA attribute and does also need to be taken into account for attribute reflection. Thanks for the reminder. I have no idea about the history of why it's living separately in its own spec.

index.html
```diff
@@ -10183,6 +10183,7 @@ <h3>Value</h3>
 				<dd>A list of one or more tokens.</dd>
 			</dl>
 			<p>These are generic types for states and properties, but do not define specific representation. See <a href="#state_property_processing">State and Property Attribute Processing</a> for details on how these values are expressed and handled in host languages.</p>
+			<p>When the ARIA content attribute definition includes a table enumerating the attribute's allowed values, that attribute is an <a href="https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute">enumerated attribute</a>. When the values table notates one of the values as "(default)", that default value is the <a href="https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#missing-value-default">missing value default</a> and <a href="https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#invalid-value-default">invalid value default</a> for the attribute.</p>
```
asurkov, line 10186: > When the ARIA content attribute definition includes a table enumerating the attribute's allowed values, that attribute is an enumerated attribute.

It'd be nice to give a couple of examples here, especially for enumerated "boolean" attributes, because imo this is  one of the confusing parts of ARIA spec, where an enumerated attribute has true/false values unlike HTML boolean attributes (compare aria-hidden and HTML:hidden). For example, something like this might make a good addition:
"For example, aria-readonly attribute is an enumerated attribute which can take true/false values or  aria-autocomplete which takes one of values from an enumerated list".

> When the values table notates one of the values as "(default)", that default value is the missing value default and invalid value default for the attribute.

I might be wrong but "missing value default" and "invalid value default" are mutually exclusive, i.e. default value can be either "missing value default" or "invalid value default", if so then "and" should be replaced on "or". Also it'd be good to give a couple of examples of attributes having default value for each case.

domenic: Missing value default and invalid value default are not mutually exclusive; the same value can be both.

+1 to examples.

cookiecrook:
@annevk wrote:

> Looking at this PR I'm not sure how it solves the nullable problem? Or does it help with #1058, but not fully resolve it?

@domenic's [comments starting here in #1058](https://github.com/w3c/aria/issues/1058#issuecomment-548544675) led me to believe this was sufficient given the current limitations of what is definable in WebIDL syntax today. If it's insufficient, I may need your help determining why or suggesting a resolution.

> Also, `role` also seems to be defined as a list value or is that not considered an ARIA attribute? Either way it seems to have the same problem.

Issue #1058 refers directly to the [`AriaAttributes`](https://w3c.github.io/aria/#AriaAttributes) interface, which does not include `role`. That one is defined in the [`AccessibilityRole`](https://w3c.github.io/aria/#AccessibilityRole) interface. We could address issues with that one too, but please file a separate issue if possible. Thank you.

index.html
cookiecrook: @asurkov wrote:

> It'd be nice to give a couple of examples here, especially for enumerated "boolean" attributes 

Agreed. I'll add some.

cookiecrook:
Ugh... Please disregard until I figure how those other commits were included in that last push.

annevk:
@cookiecrook I guess it depends on whether all attributes with a nullable string are enumerated. I thought that some were not, such as `ariaValueNow`.

cookiecrook:
@annevk wrote:

> I guess it depends on whether all attributes with a nullable string are enumerated. I thought that some were not, such as `ariaValueNow`. 

Ah. Thanks. Yes, now that we have the prose, I think the original intention for for the IDL nullable `?` flag is superseded entirely now. Added a commit to remove those from the `DOMString` props. Note: I left the others for `Element` and `FrozenArray<Element>` because those are pulled out in another branch.

cookiecrook:
Recycling the review requests because this diff has grown substantially.

cookiecrook:
@carmacleod wrote: 

> Just a few typos (I'll just push them in).

Thanks !

> (just curious... what is the missing value default for `role`? Is it "undefined"? 

Empty string `""` 

> Any chance we can talk the browsers into returning the implicit role for `element.role` if no explicit role was assigned? That would be so useful! 😄 )

That's pretty difficult to do without spinning up the accessibility code in most browsers now. TagName isn't the only factor that affects role. For example, a `<td>` outside of a table, doesn't get exposed as rows. Neither does a cell in a layout table. From the implementation side, I don't think that would be achievable in a minor point release change.

The `computedRole` property in [WebDriver #1439](https://github.com/w3c/webdriver/issues/1439) may be a more feasible way to approach that use case in the meantime.

index.html
```diff
@@ -10178,16 +10178,72 @@ <h3>Value</h3>
...
+			</aside>
+			<aside class="example">
+
+<!-- ReSpec needs these examples to be unindented. -->
+<pre>// @aria-busy example
+// true/false ~ enumerated boolean-like string; defaults to "false"
+
+// Note: String assignment and return value.
+el.ariaBusy = "true";
+el.ariaBusy; // "true"
```
asurkov, line 10218: This is unfortunate side effect of being an enumerated attribute. It's not user friendly I'd say: what feels like boolean should be exposed as a boolean. I suppose people will continiously bump into if (el.ariaBusy) bugs.

I'm curious if it's possible to keep such attributes as enumerated attributes (which allows us to have nested aria-hidden="false"), but have them to prove their own boolean-like mapping. At least perhaps it's worth to file an issue to investigate it.

Otherwise, I like the examples/change. It makes the whole things much easier to understand.

index.html
```diff
@@ -10178,16 +10178,72 @@ <h3>Value</h3>
...
+		<section class="informative" id="enumeration-example">
+			<h3>Example Enumerated Attribute Usage</h3>
+			<p>As noted in <a href="#typemapping">Mapping <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> Value Types to Languages</a>, attributes are included in host languages, and the syntax for representation of enumerated value types is governed by the host language.</p>
+			<p>All enumerated attribute getters and setters use string values, including the boolean-like enumerated <a href="#valuetype_true-false">true/false</a> type.</p>
+			<aside class="example">
+
+<!-- ReSpec needs these examples to be unindented. -->
+<pre>// HTML @hidden example (not @aria-hidden)
```
domenic, line 10198: Nit: I find the `@` syntax confusing, and would suggest using `hidden=""` instead.

index.html
```diff
@@ -10178,16 +10178,72 @@ <h3>Value</h3>
...
+	<section id="enumerated-attribute-values">
+		<h2>Enumerated Attribute Values</h2>
+		<p>When the ARIA attribute definition includes a table enumerating the attribute's allowed <span>values</span>, that attribute is an <a href="https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute">enumerated attribute</a>. When the values table notates one of the values as "(default)", that default value is the <a href="https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#missing-value-default">missing value default</a> and <a href="https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#invalid-value-default">invalid value default</a> for the attribute.</p>
```
domenic, line 10190: I suggest adding the sentence "Each value in the table is a keyword for the attribute, mapping to a state of the same name" after the first one.

index.html
domenic: It's not really possible. In HTML `true` and `false` as boolean values are only returned for boolean attributes, i.e. attributes for which the value doesn't matter, only the presence/absence. ARIA attributes are not like those; their values matter, which makes them reflect as strings.

index.html
cookiecrook: To make sure I understand you, by "after the first one" do you mean "after the first sentence"? Please confirm (👍 is fine) or clarify further. Thanks.

index.html
alice: For `aria-atomic`, `aria-busy`, `aria-disabled`, `aria-modal`, `aria-multiline`, `aria-multiselectable`,  `aria-readonly` and `aria-required`, there really are only two options, so the value doesn't actually matter in these cases. 

The missing value default is "false" for each of them, so presence/absence already works in cases where the value is missing, or the value is "true".  (Note that this isn't true for `aria-hidden`, which has a missing default value of "undefined").

I think that means these would be candidates for true boolean attributes, if we can deduce that usage patterns are such that it wouldn't break too much existing code. However, this would probably require browser instrumentation to figure out, since generally these attributes are inherently dynamic and unlikely to show up on page load.

It seems like we dodged this discussion over on https://github.com/w3c/aria/issues/1110, so perhaps we should revisit it?

For the purposes of this change, however, I don't think it needs to be resolved. It would be a "breaking" change from the point of view of the content attribute, but the IDL attribute would have the same relationship to page behaviour, so code using the IDL attribute as defined here would continue working just the same.

index.html
cookiecrook: > "Each value in the table is a keyword for the attribute, mapping to a state of the same name"

I'm having trouble with this sentence, since it uses the term "state" in a different way than how "states and properties" is used elsewhere in the spec. I tried to rid the ARIA spec of most instances of those terms at one point, but they've snuck back in. "state" is the right term here IMO, but perhaps there is another synonym that would avoid the ambiguity altogether.

index.html
cookiecrook: Marking this sub-thread as resolved. Please correct me if I missed something actionable for this PR.

index.html
cookiecrook: Perhaps 'condition'? "Each value in the table is a keyword for the attribute, mapping to a condition of the same name." I don't love it, but could live with it if no one objects.

domenic: Enumerated attributes are defined by their states, so it doesn't work to use a different word. I recognize there's an unfortunate terminology collision here, but it's a term of art that goes along with enumerated attribute, so kind of unavoidable...

PR #1278, opened 2020-05-22
Remove multiple inheritance from menuitemcheckbox and menuitemradio

### LINKED ISSUE #1277: menuitemradio/menuitemcheckbox shouldn't support aria-expanded. Maybe radio should?
ARIA 1.2 issue https://github.com/w3c/aria/issues/681 and its PR https://github.com/w3c/aria/pull/972 revisited all roles that supported `aria-expanded`, and then:
- removed support for aria-expanded from grouping/container roles (i.e. the controlled element)
- added support for aria-expanded to checkbox (and switch, which inherits from checkbox)

However, a couple of things fell through the cracks.

#### Part 1. `menuitemcheckbox` and `menuitemradio` should not support `aria-expanded`
Currently menuitemcheckbox and menuitemradio inherit [aria-expanded](https://w3c.github.io/aria/#aria-expanded) from menuitem (and checkbox). 

We all agreed earlier that [menuitem needs to support aria-expanded](https://github.com/w3c/aria/issues/454) in order to allow for submenus, but menuitemcheckbox and menuitemradio are leaf nodes in a menu, so aria-expanded should be prohibited on them.

Looking at https://github.com/w3c/aria/issues/681#issuecomment-358731813 and https://github.com/w3c/aria/issues/681#issuecomment-488776046, it seems pretty clear that we intended to make this happen for ARIA 1.2.

#### Part 2. maybe `radio` should support `aria-expanded` now that `checkbox` does
Neither https://github.com/w3c/aria/issues/681 nor https://github.com/w3c/aria/pull/972 mention `radio`, and so even though it was decided that `checkbox` should support `aria-expanded`, it's not clear whether `radio` was considered as well.

People are using disclosures on radios (with or without aria-expanded).

The [gov.uk design system has a pattern with conditionally revealed content between radios](https://design-system.service.gov.uk/components/radios/#conditionally-revealing-content) ([example](https://design-system.service.gov.uk/components/radios/conditional-reveal/index.html)), and they give this advice:
> Using this component, you can add conditionally revealing content to stacked radios, so users only see content when it’s relevant to them.
> For example, you could reveal an email address input only when a user chooses to be contacted by email.

[PatternFly has this guideline](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms#use-progressive-disclosure):
> Use progressive disclosure to hide and show fields based on the users’ selection. This approach keeps the workflow short and allows the user to focus on only the fields relevant to them. It also avoids using color to indicate whether or not a field is available, making it suitable for visually impaired users.

In addition to forms, I've seen this used in print dialogs: (all pages, current page, custom + textbox) and in the controls for charts or transaction lists (last 5 days, last month, last 6 months, custom + datepicker).

The [gov.uk people did some testing of this with screen readers](https://github.com/alphagov/govuk-frontend/issues/979#issuecomment-528932826) (because it was flagged by a validator), and they discovered that there is enough AT support for aria-expanded on radios that they decided to keep using it in their forms, given that user testing did not find it problematic. In fact, they had some feedback that indicated that it is helpful to have expanded/collapsed announced.

Looking at the gov.uk test results, I see that NVDA/FF, VO/Safari and VO/iOS Safari all announce expanded/collapsed. I have just verified these, so it looks like we have our 2 implementations.

carmacleod:
**NOTE: June 13, 2020**
For some reason pr-preview did not run, so the Preview and Diff are old (tooltip says "Last updated on May 26, 2020"). Please just look at the code changes in Files changed - there are only 3 changes (2 deletions and an addition).

Closes #1277

- removes checkbox from menuitemcheckbox superclasses (so it only inherits from menuitem)
- removes radio from menuitemradio superclasses (so it only inherits from menuitemcheckbox)
- requires aria-checked on menuitemcheckbox (which menuitemradio inherits)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1278.html" title="Last updated on May 26, 2020, 12:27 PM UTC (3f845d5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1278/3c600ae...3f845d5.html" title="Last updated on May 26, 2020, 12:27 PM UTC (3f845d5)">Diff</a>

### DIFF
--- a/common/script/aria.js
+++ b/common/script/aria.js
@@ -339,8 +339,10 @@ require(["core/pubsubhub"], function( respecEvents ) {
                                 attrs.push( { is: type, name: name, required: req, disallowed: dis, deprecated: dep } );                                               
 
                                 // remember that the state or property is
-                                // referenced by this role
-                                propList[name].roles.push(title);
+                                // referenced by this role (unless disallowed)
+                                if (!dis) {
+                                    propList[name].roles.push(title);
+                                }
                             });
                         }
                     });
@@ -492,10 +494,18 @@ require(["core/pubsubhub"], function( respecEvents ) {
                         var output = "";
                         var section = document.querySelector("#" + item.name);
                         var placeholder = section.querySelector(".state-applicability, .property-applicability");
-                        if (placeholder && ((placeholder.textContent || placeholder.innerText) === "Placeholder") && item.roles.length) {
+                        var deprecatedAsGlobal = placeholder && ((placeholder.textContent || placeholder.innerText) === "Use as a global deprecated in ARIA 1.2");
+                        if ((placeholder && ((placeholder.textContent || placeholder.innerText) === "Placeholder") || deprecatedAsGlobal) && item.roles.length) {
                             // update the used in roles list
                             var sortedList = [];
                             sortedList = item.roles.sort();
+                            if (deprecatedAsGlobal) {
+                              // remove roletype from the sorted list
+                              const index = sortedList.indexOf('roletype');
+                              if (index > -1) {
+                                  sortedList.splice(index, 1);
+                              }
+                            }
                             for (var j = 0; j < sortedList.length; j++) {
                                 output += "<li><rref>" + sortedList[j] + "</rref></li>\n";
                             }
@@ -503,7 +513,7 @@ require(["core/pubsubhub"], function( respecEvents ) {
                                 output = "<ul>\n" + output + "</ul>\n";
                             }
                             placeholder.innerHTML = output;
-                            // also update any inherited roles
+                            // update the inherits into roles list
                             var myList = [];
                             $.each(item.roles, function(j, role) {
                                 var children = getAllSubRoles(role);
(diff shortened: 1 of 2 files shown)


jnurthen:
I'd prefer to solve this by changing the inheritance of menuitemcheckbox.
To be honest the other states and properties menuitemcheckbox (and eventually menuitemradio) inherit make no sense either:
aria-errormessage, aria-invalid, aria-readonly and aria-required

jnurthen:
I'm going to tag the underlying issue for the agenda and if we don't have agreement through github we can discuss on Thursday.

carmacleod:
@jnurthen 
> I'd prefer to solve this by changing the inheritance of menuitemcheckbox.
> To be honest the other states and properties menuitemcheckbox (and eventually menuitemradio) inherit make no sense either:
> aria-errormessage, aria-invalid, aria-readonly and aria-required

That's brilliant, and I agree completely, and I will remove `checkbox` from menuitemcheckbox's superclasses in this PR (and add `aria-checked` to the supported attrs). However, I think menuitemcheckbox should still inherit from `menuitem`, so it doesn't solve the `aria-expanded` problem.

(Also, hmmm... maybe `aria-haspopup` should be prohibited on menuitemcheckbox/menuitemradio...).

carmacleod:
I guess `radio` should be removed from `menuitemradio`'s superclasses to correspond to removing `checkbox` as a superclass of `menuitemcheckbox`.

carmacleod:
This PR:
- Removes checkbox from menuitemcheckbox superclasses (so it only inherits from menuitem)
- Removes radio from menuitemradio superclasses (so it only inherits from menuitemcheckbox)
- Adds aria-checked as required for menuitemcheckbox (which is then inherited by menuitemradio)
- Prohibits aria-expanded and aria-haspopup on menuitemcheckbox and menuitemradio
- Fixes up the javascript so that prohibited attributes (i.e. aria-expanded and aria-haspopup) don't say they are used in or inherit into roles they are prohibited on (i.e. menuitemcheckbox and menuitemradio)
- Adds support for aria-expanded to radio

Since menuitemcheckbox and menuitemradio no longer inherit from an input (i.e. checkbox or radio),
they no longer support aria-errormessage, aria-invalid (2 of the 4 deprecated globals), aria-readonly or aria-required.

carmacleod:
From the call:
- don't support aria-expanded on radio
- do support aria-expanded on menuitemradio and menuitemcheckbox
- do remove the multiple inheritance (i.e. remove checkbox from menuitemcheckbox superclasses, and remove radio from menuitemradio superclasses)

carmacleod:
Ok, I've made the changes that were discussed on the call, removed the WIP label, and renamed this PR to reflect what it contains now:
- removes the multiple inheritance from menuitemcheckbox and menuitemradio, i.e.
  - removes checkbox from menuitemcheckbox superclasses (so it only inherits from menuitem)
  - removes radio from menuitemradio superclasses (so it only inherits from menuitemcheckbox)
  - requires aria-checked on menuitemcheckbox (which menuitemradio inherits)

This has the desired side effect of removing support for aria-errormessage, aria-invalid, aria-readonly and aria-required from menuitemcheckbox and menuitemradio.

I also moved the javascript to support prohibited non-global attributes to https://github.com/w3c/aria-common/pull/46 to be used later if we need it.

This PR is ready for review.

PR #1292, opened 2020-06-18
re-add aria-haspopup on links & add note for usage

### LINKED ISSUE #1289: aria-haspopup should still be supported on link
#999 / #1023 deprecates use of `aria-haspopup` as a global property, but does not include `link` as one of the `role`s for which `aria-haspopup` is still valid. In my opinion, a popover or dialog is very much like an in-page link, in that expanding it moves focus to another location within the same page. 

One use case I recently encountered where `aria-haspopup` seems appropriate on a link is for progressive enhancement in documentation, where code examples for one component link to documentation for related components and class definitions. For this use case, on-site links display the related documentation in a popup dialog, while off-site links open the related documentation in a new window. I think it makes sense to distinguish between these two link behaviors, and it seems odd to have to add `role="button"` to the on-site links.

jnurthen:
resolves #1289

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1292.html" title="Last updated on Jun 25, 2020, 5:31 PM UTC (28675f8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1292/2212c5e...28675f8.html" title="Last updated on Jun 25, 2020, 5:31 PM UTC (28675f8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4460,6 +4460,7 @@
 							<ul>
 								<li><sref>aria-disabled</sref></li>
 								<li><sref>aria-expanded</sref></li>
+								<li><pref>aria-haspopup</pref></li>
 							</ul>
 						</td>
 					</tr>
@@ -11568,6 +11569,11 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
 				<p>The <code>aria-haspopup</code> property is an enumerated type. <a>User agents</a> MUST treat any value of <code>aria-haspopup</code> that is not included in the list of allowed values, including an empty string, as if the value <code>false</code> had been provided. To provide backward compatibility with ARIA 1.0 content, user agents MUST treat an <code>aria-haspopup</code> value of <code>true</code> as equivalent to a value of <code>menu</code>.</p>
 				<p><a>Assistive technologies</a> SHOULD NOT expose the <code>aria-haspopup</code> property if it has a value of <code>false</code>.</p>
 				<p class="note">A <rref>tooltip</rref> is not considered to be a popup in this context.</p>
+				<p class="note"><code>aria-haspopup</code> is most relevant to use when there is a visual indicator in the element that triggers the popup. 
+					For example, most menu buttons are styled with a downward pointing triangle or chevron, which indicates to sighted users that the button will display a menu when clicked. 
+					If some functional difference is relevant to display to a sighted user by means of a different visual style, that functional difference is usually relevant to convey to users of assistive technology.
+					If there is no visual indication that a element will trigger a popup, authors are advised to consider whether use of <code>aria-haspopup</code> is necessary, and avoid using it when it's not.
+				</p>
 				<p class="note">This property is being deprecated as a global property in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically supported.</p>
 			</div>
 			<table class="property-features">
--- a/index.html
+++ b/index.html
@@ -11570,7 +11570,7 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
 				<p><a>Assistive technologies</a> SHOULD NOT expose the <code>aria-haspopup</code> property if it has a value of <code>false</code>.</p>
 				<p class="note">A <rref>tooltip</rref> is not considered to be a popup in this context.</p>
 				<p class="note"><code>aria-haspopup</code> is most relevant to use when there is a visual indicator in the element that triggers the popup. 
-					For example, most menu buttons are styled with a downward pointing triangle or chevron, which indicates to sighted users that the button will display a menu when clicked. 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -11568,6 +11569,11 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p><a>Assistive technologies</a> SHOULD NOT expose the <code>aria-haspopup</code> property if it has a value of <code>false</code>.</p>
 				<p class="note">A <rref>tooltip</rref> is not considered to be a popup in this context.</p>
+				<p class="note"><code>aria-haspopup</code> is most relevant to use when there is a visual indicator in the element that triggers the popup. 
+					For example, most menu buttons are styled with a downward pointing triangle or chevron, which indicates to sighted users that the button will display a menu when clicked. 
+					If some functional difference is relevant to display to a sighted user by means of a different visual style, that functional difference is usually relevant to convey to users of assistive technology.
+					If there is no visual indication that a element will trigger a popup, authors are advised to consider whether use of <code>aria-haspopup</code> is necessary, and avoid using it when it's not.
```
scottaohara, line 11575:
```suggestion
					If there is no visual indication that an element will trigger a popup, authors are advised to consider whether use of <code>aria-haspopup</code> is necessary, and avoid using it when it's not.
```

index.html
```diff
@@ -11568,6 +11569,11 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				<p>The <code>aria-haspopup</code> property is an enumerated type. <a>User agents</a> MUST treat any value of <code>aria-haspopup</code> that is not included in the list of allowed values, including an empty string, as if the value <code>false</code> had been provided. To provide backward compatibility with ARIA 1.0 content, user agents MUST treat an <code>aria-haspopup</code> value of <code>true</code> as equivalent to a value of <code>menu</code>.</p>
 				<p><a>Assistive technologies</a> SHOULD NOT expose the <code>aria-haspopup</code> property if it has a value of <code>false</code>.</p>
 				<p class="note">A <rref>tooltip</rref> is not considered to be a popup in this context.</p>
+				<p class="note"><code>aria-haspopup</code> is most relevant to use when there is a visual indicator in the element that triggers the popup. 
+					For example, most menu buttons are styled with a downward pointing triangle or chevron, which indicates to sighted users that the button will display a menu when clicked.
```
scottaohara, line 11573:
```suggestion
					For example, many controls styled with a downward pointing triangle or chevron, or represented by three consecutive dots, have become standard visual indicators that a popup will display when the control is activated. 
```

carmacleod: I like your new wording @scottaohara - it gives many more examples with very few extra words. :)

Does it make sense to say "containing" three consecutive dots instead of "represented by" three consecutive dots? That would allow for such things as "Open..." instead of just the "..." of a horizontal or vertical kebab.

scottaohara: yeh, good point @carmacleod 

i removed the word "represented", as i think think that allows for both interpretations (containing or represented by) to be inferred.

carmacleod: So then does it become "styled by three consecutive dots"?

index.html
```diff
@@ -11568,6 +11569,11 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				<p>The <code>aria-haspopup</code> property is an enumerated type. <a>User agents</a> MUST treat any value of <code>aria-haspopup</code> that is not included in the list of allowed values, including an empty string, as if the value <code>false</code> had been provided. To provide backward compatibility with ARIA 1.0 content, user agents MUST treat an <code>aria-haspopup</code> value of <code>true</code> as equivalent to a value of <code>menu</code>.</p>
 				<p><a>Assistive technologies</a> SHOULD NOT expose the <code>aria-haspopup</code> property if it has a value of <code>false</code>.</p>
 				<p class="note">A <rref>tooltip</rref> is not considered to be a popup in this context.</p>
+				<p class="note"><code>aria-haspopup</code> is most relevant to use when there is a visual indicator in the element that triggers the popup. 
+					For example, many controls styled with a downward pointing triangle or chevron, or three consecutive dots, have become standard visual indicators that a popup will display when the control is activated.
```
cookiecrook, line 11573:
```suggestion
					For example, many controls styled with a downward pointing triangle, chevron, or ellipsis (three consecutive dots) have become standard visual indicators that a popup will display when the control is activated. 
```

cookiecrook: Would it be appropriate to link to this portion of the ellipsis Wikipedia entry? https://en.wikipedia.org/wiki/Ellipsis#Usage_in_computer_system_menus

From the page: 

> In computer menu functions or buttons, an ellipsis means that upon selection more options (sometimes in the form of a dialog box) will be displayed, where the user can or must make a choice.

cookiecrook: Surprisingly, the Wikipedia entry for chevron does not have a similar section.

jnurthen: prefer not to link to wikipedia - not sure we need the definition

PR #1298, opened 2020-07-09
Update caption role example

jongund:
Simplifying the `caption` role definition example in the document to include only a reference for defining an accessible name.
The `caption` element in HTML is typically only used associating an accessible name with a table, so I think we should limit the `caption` role example to providing an accessible name.  More complicated naming and describing examples can be developed and vetted in the ARIA APG.

***

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Faria%2F0e0f9c9443f4cbc2909dc7442d3ed64106b429cc%2Findex.html%3FisPreview%3Dtrue)

```
Navigation timeout of 28032 ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231298.)._
</details>


scottaohara (approved):
<s>per our conversation about this revised example, I'm fine with the change to make this more simple. Per the previous example trying to help mitigate against authors following the HTML's spec example of using overly long captions, the only thing I'm wondering here is if we want to add a note to dissuade people from doing that?</s>

<s>If others don't think a note is necessary, then I'm good with this as is.</s>

Edit: 1.5 years later, no longer agree with past self since present self has gone in the completely different direction per other updates to `figure` that no longer make the `caption` role always provide an accName.

scottaohara:
@pkra my updates from other PRs supersede this PR.  I don't think this is needed anymore

pkra:
> @pkra my updates from other PRs supersede this PR. I don't think this is needed anymore

Thanks, @scottaohara.

Let's close this then. @jongund please re-open if you disagree.

PR #1315, opened 2020-08-27
Assistive tech SHOULD provide landmark navigation... user agents MAY

### LINKED ISSUE #350: Landmarks - Assistive technologies SHOULD; user agents MAY...
The 3rd paragraph for region (role) says:
"Assistive technologies should enable users to quickly navigate to elements with role region. Mainstream user agents may enable users to quickly navigate to elements with role region."

but the 3rd paragraph for banner, complementary, contentinfo, form, main, navigation, and search say:
"User agents should treat elements with the role of banner as navigational landmarks."
....and they don't say anything about assistive technologies. Which seems really odd, because currently assistive technologies enable landmark navigation, but user agents don't do a thing with them... (I wish they did).

I think (?) all of the landmark roles should use the wording in region role.

(((Although if I could, I'd spec it so that user agents MUST enable users to quickly navigate to landmark roles... ;)  Just kidding. Sort of.)))

carmacleod:
Resolves #350.

Issue #350 noted some discrepancies in the normative language across all of the various landmark roles.
This PR ensures that the wording is uniform, including making sure there is a normative statement for both AT and UA.

~I was given permission by the ARIA WG to go part-way (SHOULD instead of MUST) towards my [earlier comment](https://github.com/w3c/aria/issues/350#issue-150943681):~
~> if I could, I'd spec it so that user agents MUST enable users to quickly navigate to landmark roles...~

~So this PR basically uses statements similar to the following for each of the landmark roles:~
~> Assistive technologies SHOULD enable users to quickly navigate to landmarks.~
~> User agents SHOULD enable users to quickly navigate to elements whose implicit role is a landmark.~

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1315.html" title="Last updated on Mar 19, 2021, 5:03 PM UTC (fb093f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1315/a6bc636...fb093f5.html" title="Last updated on Mar 19, 2021, 5:03 PM UTC (fb093f5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4871,7 +4871,7 @@
 			<rdef>main</rdef>
 			<div class="role-description">
 				<p>A <rref>landmark</rref> containing the main content of a document.</p>
-				<p>This marks the content that is directly related to or expands upon the central topic of the document. The <code>main</code> <a>role</a> is a non-obtrusive alternative for "skip to main content" links, where the navigation option to go to the main content (or other <a>landmarks</a>) is provided by the <a>user agent</a> through a keyboard shortcut (or <abbr title="User Interface">UI</abbr> feature such as a side panel or dialog) or by <a>assistive technologies</a>.</p>
+				<p>This marks the content that is directly related to or expands upon the central topic of the document. The <code>main</code> <a>role</a> is a non-obtrusive alternative for "skip to main content" links, where the navigation option to go to the main content (or other <a>landmarks</a>) is provided by <a>assistive technologies</a>, or by a <a>user agent</a> or browser extension, through a keyboard shortcut or <abbr title="User Interface">UI</abbr> feature such as a side panel or dialog.</p>
 				<p><a>Assistive technologies</a> SHOULD enable users to quickly navigate to elements with role <code>main</code>.
   				<a>User agents</a> SHOULD treat elements with role <code>main</code> as navigational <a>landmarks</a>.
   				<a>User agents</a> MAY enable users to quickly navigate to elements with role <code>main</code>.</p>
--- a/index.html
+++ b/index.html
@@ -1345,7 +1345,7 @@
 			<div class="role-description">
 				<p>A <rref>landmark</rref> that contains mostly site-oriented content, rather than page-specific content.</p>
 				<p>Site-oriented content typically includes things such as the logo or identity of the site sponsor, and a site-specific search tool. A banner usually appears at the top of the page and typically spans the full width.</p>
-				<p><a>Assistive technologies</a> SHOULD enable users to quickly navigate to elements with role <code>banner</code>. Mainstream <a>user agents</a> SHOULD enable users to quickly navigate to <a>elements</a> that have an implicit role of <code>banner</code>.</p>
+				<p><a>User agents</a> and <a>assistive technologies</a> SHOULD enable users to quickly navigate to elements with role <code>banner</code>.</p>
 				<!-- keep the following paragraphs synced with the similar paragraphs in #main and #contentinfo-->
 				<p>Within any <rref>document</rref> or <rref>application</rref>, the author SHOULD mark no more than one <a>element</a> with the <code>banner</code> <a>role</a>.</p>
 				<p class="note">Because <code>document</code> and <code>application</code> elements can be nested in the <abbr title="Document Object Model">DOM</abbr>, they may have multiple <code>banner</code> elements as <abbr title="Document Object Model">DOM</abbr> descendants, assuming each of those is associated with different document nodes, either by a DOM nesting (e.g., <rref>document</rref> within <rref>document</rref>) or by use of the <pref>aria-owns</pref> <a>attribute</a>.</p>
@@ -2437,7 +2437,7 @@
 			<div class="role-description">
 				<p>A <rref>landmark</rref> that is designed to be complementary to the main content at a similar level in the DOM hierarchy, but remaining meaningful when separated from the main content.</p>
 				<p>There are various types of content that would appropriately have this <a>role</a>. For example, in the case of a portal, this may include but not be limited to show times, current weather, related articles, or stocks to watch. The complementary role indicates that contained content is relevant to the main content. If the complementary content is completely separable from the main content, it may be appropriate to use a more general role.</p>
-				<p><a>Assistive technologies</a> SHOULD enable users to quickly navigate to elements with role <code>complementary</code>. Mainstream <a>user agents</a> SHOULD enable users to quickly navigate to <a>elements</a> that have an implicit role of <code>complementary</code>.</p>
+				<p><a>User agents</a> and <a>assistive technologies</a> SHOULD enable users to quickly navigate to elements with role <code>complementary</code>.</p>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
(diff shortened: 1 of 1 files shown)


carmacleod:
Broken link to "landmark roles" in region role definition. Will fix...

index.html
```diff
@@ -1345,7 +1345,7 @@ <h2>Definition of Roles</h2>
 			<div class="role-description">
 				<p>A <rref>landmark</rref> that contains mostly site-oriented content, rather than page-specific content.</p>
 				<p>Site-oriented content typically includes things such as the logo or identity of the site sponsor, and a site-specific search tool. A banner usually appears at the top of the page and typically spans the full width.</p>
-				<p>User agents SHOULD treat elements with the role of <code>banner</code> as navigational <a>landmarks</a>.</p>
+				<p><a>Assistive technologies</a> SHOULD enable users to quickly navigate to elements with role <code>banner</code>. Mainstream <a>user agents</a> SHOULD enable users to quickly navigate to <a>elements</a> that have an implicit role of <code>banner</code>.</p>
```
mcking65, line 1348: Why do we need the word "mainstream"? I also don't understand why a user agent would limit the capability to elements with an implicit role. I have a hard time believing there would be a significant difference in implementation difficulty.


```suggestion
				<p><a>User agents</a> and <a>assistive technologies</a> SHOULD enable users to quickly navigate to elements with role <code>banner</code>.</p>
```

index.html
```diff
@@ -1345,7 +1345,7 @@ <h2>Definition of Roles</h2>
...
 				<!-- keep the following paragraphs synced with the similar paragraphs in #main and #contentinfo-->
 				<p>Within any <rref>document</rref> or <rref>application</rref>, the author SHOULD mark no more than one <a>element</a> with the <code>banner</code> <a>role</a>.</p>
```
mcking65, line 1350: I know this is not related to this PR, but I can't resist comment ... Wow is this out of date language. I guess it's been a while since I read it. It does not jive with the definition of role application. And, we do not typically refer to an entire web page as a document in the rest of the spec.

```suggestion
				<p>Authors SHOULD mark no more than one <a>element</a> on a page with the <code>banner</code> <a>role</a>.</p>
```

mcking65 (commented):
I could be convinced to support as is ... but would prefer that changes that align with my suggestion were made.

index.html
carmacleod: I agree that "on a page" is easier to understand than "within any document (role) or application (role)" in that sentence.
However, I'm not sure whether or not the note box that follows that sentence (in banner, contentinfo, and main) needs to be reworded as well, so I've opened issue https://github.com/w3c/aria/issues/1320 to look into this.
> NOTE
Because document and application elements can be nested in the DOM, they may have multiple banner elements as DOM descendants, assuming each of those is associated with different document nodes, either by a DOM nesting (e.g., document within document) or by use of the aria-owns attribute.

carmacleod:
@mcking65 
> Why do we need the word "mainstream"?

I'm happy to take out the word "mainstream". I wondered about it myself, but left it in to keep the number of changes to a minimum, and because I saw it in several other places in the spec. I've opened issue https://github.com/w3c/aria/issues/1321 to review the other occurrences.

> I also don't understand why a user agent would limit the capability to elements with an implicit role. I have a hard time believing there would be a significant difference in implementation difficulty.

I don't think it's about implementation difficulty. It's actually a philosophical issue... the user agents do not want to provide any behavior that relies solely on ARIA markup. I'm quite sure that, as you say, there wouldn't be any significant difference in implementation difficulty for implicit vs. explicit landmarks, but, to quote a UA dev, "It's a slippery slope. If we provide behavior for this ARIA role, then people will ask why we don't provide behavior for this other ARIA role".

That said, there's nothing stopping the ARIA spec from saying that they SHOULD enable users to navigate to _all_ landmarks regardless of ARIA or implicit markup... so I've changed this PR to use the wording that you provided above. I prefer your wording because it's simpler, and because it's what I really wanted to say in the first place:
> User agents and assistive technologies SHOULD enable users to quickly navigate to elements with role [landmark role].

index.html
carmacleod: Agreed.

cookiecrook:
Removing my review request, since as a Member Rep I'm not in a position to endorse a precedence change this significant. I stopped short of rejecting the review, because I can relate to some extent.

As I mentioned in this morning's ARIA WG meeting, it's possible various stakeholders will take issue with ARIA requiring mainstream user interface changes like this. Playing devil's advocate on a couple of the potential objections below.

**1. Lack of Precedence.** I don't think there is any other precedent for this in ARIA. As @carmacleod pointed out, [ARIA §1.3 User Agent Support](https://w3c.github.io/aria/#ua-support) states "The WAI-ARIA specification neither requires nor forbids user agents from enhancing native presentation and interaction behaviors on the basis of WAI-ARIA markup." So this would be a pretty major precedence change IMO. You might get pushback based on the slippery slope argument. The same section of the spec gives a non-binding suggestion, "Mainstream user agents might expose WAI-ARIA navigational landmarks (for example, as a dialog box or through a keyboard command) with the intention to facilitate navigation for all users" but to my knowledge the ARIA spec has never before included any mainstream UI as an RFC _requirement_. 

**2. Ambiguity.** The prior language (~"treat as landmarks") was also vague, but was widely interpreted as "expose landmarks to the platform Accessibility APIs", which all user agents now do. The new language states that UAs should (RFC-2119 SHOULD) "enable users to quickly navigate to" landmarks. It doesn't state how to do so or how to test the implementations, presumably because ARIA has never before defined this level of mainstream interaction detail. As such, it seems more like a wishful advocacy/evangelism change, and I'm not convinced a Rec-Track spec is the right place to do that... 

**3. Priority.** I don't think a sufficient case has been made for why mainstream users need this and why they'd need it now. Sure, we can speculate that landmark navigation might be nice for power keyboard users, but wouldn't a browser extension or a keyboard-driven AT solve that need? The WG would need to make a better case that this benefits all users and that a significant amount would use it. 

In either case, HTML or WICG may be the better place to discuss new behaviors for mainstream user agents. A polyfill or browser extension might be a good way to prototype the interaction you're hoping to achieve.

In the case that this stays in ARIA and the edit is merged into the 1.3 branch, I assume the editors agree on the following: 
- It would have to be removed before CR unless there were two implementation commitments and tests available. 
- It would have to be removed before PR unless there were two completed implementations.

mcking65:
@cookiecrook wrote:
> In the case that this stays in ARIA and the edit is merged into the 1.3 branch, I assume the editors agree on the following:
> 
> * It would have to be removed before CR unless there were two implementation commitments and tests available.
> * It would have to be removed before PR unless there were two completed implementations.

I thought only MUST statements require implementations and tests?

Regardless, I do find your arguments for why the spec might not (MAY NOT) be the most appropriate mechanism for advocating for browser support to be thought provoking, close to compelling. However, I do see distinct lines between browser support for landmark navigation and other types of browser support that would muddy the contract with authors. I don't see any risk of distorting the author/browser contract with this type of requirement. I'm assuming that contract is the top concern when it comes to ARIA influencing browser behavior.

smhigley (commented):
(official Microsoft hat on, the following is the response from Edge)

Microsoft recognizes the value that native-browser-supported landmark navigation would have; however we are hesitant to see such a value-add feature promoted to a "should" requirement for all user agents to support. We would prefer the specification maintain a "may" status [RFC2119]. We have not observed a dramatic demand for this feature and note that a variety of browser extensions currently provide this capability for users who would like to have this feature today.

carmacleod:
Ok, I've changed the language to essentially say the following for each landmark role:
> Assistive technologies SHOULD enable users to quickly navigate to elements with role [landmark].
> User agents SHOULD treat elements with the role of [landmark] as navigational landmarks.
> User agents MAY enable users to quickly navigate to elements with role [landmark].

This is basically what is currently in the spec, except that the first 2 sentences are now consistent across landmark roles, and the 3rd sentence is new. The abstract `landmark` role only includes sentences 1 and 3.

The only other changes are a slight rewording of the existing sentence in `main` role about skip links, removal of the word "Mainstream" from `landmark` and `region` roles, and fixing a broken link to the list of landmark roles in `region`.

@cookiecrook @smhigley 
Please re-review, or ask your people to re-review. Thanks!

@aleventhal 
I think it's important for you to review this one, please. :)

cookiecrook (approved):
Fine to submit as-is (thanks) but you might also consider leaving the role-specific text the same, and instead add these two new normative statements one time, in the Landmarks section.

carmacleod:
Thanks, @cookiecrook !
For the record, none of the normative statements are new. :)
Two of them were already in the landmark and region sections.
The third was in all the rest.
It just felt really inconsistent.
Plus it seemed odd that most of the landmark roles mentioned UA, but not AT.

jnurthen (approved):

Issue #524, opened 2017-02-08
Allow aria-setsize on container elements


mcking65:
ARIA allows specifying aria-colcount and aria-rowcount on elements that provide the context for elements that support aria-colindex and aria-rowindex.
So, for example, aria-colcount can be set on a grid.

In contrast, the feed, list, listbox, menu, radiogroup, tablist, and tree roles do not support aria-setsize.

This is problematic when the setsize in a large set changes; every member of the set has to be touched to change the setsize value.

Make aria-setsize consistent with aria-colcount and aria-rowcount by allowing it to be specified on container elements.

jessebeach:
Consider a list of 500 elements in a browser. Changing an attribute on 500 nodes would block main JS thread for a significant amount of time and also thrash the render tree. On a site like Pinterest, Tumblr, Twitter, Facebook, etc it would not be unusual to amass hundreds of items in a set in the UI. We can't set a requirement for this feature that requires updating the set size on each item in the set.

jnurthen:
+1 to allowing this

cookiecrook:
If ARIA 1.2 is still limited to HTML Role Parity, this issue is not in scope.

mcking65:
@cookiecrook, 

I don't recall reaching an agreement to limit ARIA 1.2 scope to strictly html 5.1 role parity during our TPAC discussions. 

The outcome I recall is that other minor, non-controversial changes that are ready-to-go would be considered for inclusion. This particular proposal is closer to correcting an error than adding something new. It is not the kind of issue that would slow down 1.2.

cookiecrook:
> I don't recall reaching an agreement to limit ARIA 1.2 scope to strictly html 5.1 role parity during our TPAC discussions.

@michael-n-cooper mentioned there was some agreement to limit the scope for the benefit of focus.

> The outcome I recall is that other minor, non-controversial changes that are ready-to-go would be considered for inclusion.  This particular proposal is closer to correcting an error than adding something new. It is not the kind of issue that would slow down 1.2.

If the W3C has "famous last words," they are "This won't slow down the release." The WG leadership intended the same small scope for ARIA 1.1, but it's now going on 3 years.

cookiecrook:
Actually, it may be closer to 4 years. I recall branching the 1.1 draft well before the 1.0 TR.

mcking65:
I think it would not be that hard to set criteria for ready-to-ship that won't slow things down and then boot things out that don't meet those criteria.

[...]

PR #1332, opened 2020-10-06
Updated aria-setsize and aria-posinset to clarify usage for authors

### LINKED ISSUE #747: Add author MUST that if you define setsize you MUST also define posinset.
https://www.w3.org/Bugs/Public/show_bug.cgi?id=26047
 Also Define UA behavior if one is specified but the other is not.

MelSumner:
Resolves #747 by updating prose for `aria-setsize` and `aria-posinset`. 

Still unresolved- what should UA do if one is defined and not the other?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1332.html" title="Last updated on Mar 18, 2021, 4:12 PM UTC (b6c013c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1332/02fee23...b6c013c.html" title="Last updated on Mar 18, 2021, 4:12 PM UTC (b6c013c)">Diff</a>


index.html
```diff
@@ -12360,7 +12360,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
   &lt;li role="option" aria-setsize="16" <strong>aria-posinset="7"</strong>&gt; cantaloupes &lt;/li&gt;
   &lt;li role="option" aria-setsize="16" <strong>aria-posinset="8"</strong>&gt; dates &lt;/li&gt;
 &lt;/ul&gt;</pre>
-				<p>Authors MUST set the value for <pref>aria-posinset</pref> to an integer greater than or equal to 1, and less than or equal to the size of the set when that size is known. Authors SHOULD use <pref>aria-setsize</pref>.</p>
+				<p>Authors MUST set the value for <pref>aria-posinset</pref> to an integer greater than or equal to 1, and less than or equal to the size of the set when that size is known. Authors SHOULD use <pref>aria-setsize</pref>. If <pref>aria-setsize</pref> is specified, authors MUST specify a value for <pref>aria-posinset</pref>.</p>
```
carmacleod, line 12363: I think we should simplify this a bit and delete the "SHOULD" sentence entirely?
If we do that, would need to re-order the "If one then the other" sentence.

```suggestion
				<p>Authors MUST set the value for <pref>aria-posinset</pref> to an integer greater than or equal to 1, and less than or equal to the size of the set when that size is known. If <pref>aria-posinset</pref> is specified, authors MUST specify a value for <pref>aria-setsize</pref>.</p>
```

carmacleod:
> Still unresolved- what should UA do if one is defined and not the other?

Good question. Looking through the spec, I see that the section on [handling author errors in states and properties](https://w3c.github.io/aria/#document-handling_author-errors_states-properties) says:

> In general, user agents do not do much validation of WAI-ARIA properties. User agents MAY do some minor validation on request, such as [...] enforcing things like aria-posinset being within 1 and aria-setsize, inclusive. User agents are not responsible for logical validation, such as the following:
> ...
> 6. Determining implicit values of aria-setsize and aria-posinset when they are specified on some but not all the elements of the set.

So whatever we decide to ask User Agents to do regarding fallback for posinset without setsize (and vice-versa) probably needs to stay within the bounds of "MAY do some minor validation". I'm not sure that UA could do much here anyhow, although perhaps...

- if posinset is specified without setsize, then UA could fallback to aria-setsize=-1 (i.e. unknown size)
- if setsize is specified without posinset, then UA could fallback to aria-posinset=1 (which would result in AT saying, for example, 1 of 50, 1 of 50, 1 of 50... which is not useful, but it would give the user an idea that "something is messed up" and they would know to ignore the reported position... ;)

cookiecrook (approved):

index.html
```diff
@@ -12360,7 +12360,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
   &lt;li role="option" aria-setsize="16" <strong>aria-posinset="7"</strong>&gt; cantaloupes &lt;/li&gt;
   &lt;li role="option" aria-setsize="16" <strong>aria-posinset="8"</strong>&gt; dates &lt;/li&gt;
 &lt;/ul&gt;</pre>
-				<p>Authors MUST set the value for <pref>aria-posinset</pref> to an integer greater than or equal to 1, and less than or equal to the size of the set when that size is known. Authors SHOULD use <pref>aria-setsize</pref>.</p>
+				<p>Authors MUST set the value for <pref>aria-posinset</pref> to an integer greater than or equal to 1, and less than or equal to the size of the set when that size is known. If <pref>aria-posinset</pref> is specified, authors MUST specify a value for <pref>aria-setsize</pref>.</p>
```
cookiecrook, line 12363:
```suggestion
				<p>Authors MUST set the value for <pref>aria-posinset</pref> to an integer greater than or equal to 1, and less than or equal to the size of the set when that size is known. If <pref>aria-posinset</pref> is specified on an item, authors MUST specify a value for <pref>aria-setsize</pref> once on the container, or individually on all the items.</p>
```

index.html
```diff
@@ -13002,7 +13002,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p>If all items in a set are present in the document structure, it is not necessary to set this property, as the <a>user agent</a> can automatically calculate the set size and position for each item. However, if only a portion of the set is present in the document structure at a given moment (in order to reduce document size), this property is needed to provide an explicit indication of set size.</p>
-				<p>Authors MUST set the value of <code>aria-setsize</code> to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.</p>
+				<p>Authors MUST set the value of <code>aria-setsize</code> to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>. If <pref>aria-setsize</pref> is specified, authors MUST specify a value for <pref>aria-posinset</pref>.</p>
```
cookiecrook, line 13005: From the call today, it sounds like @mcking65 disagrees with this additional requirement.

cookiecrook (changes requested):

index.html
cookiecrook: @carmacleod reminded us in the call that the spec excluded this on the container. IIRC, that was in the context of multi-level lists, since the some of the sub-list containers (and their levels) may not be rendered in the DOM.

MelSumner: @cookiecrook so should I not accept the suggestion? Or is there more thinking to be done around the prose?

MelSumner:
@jnurthen can we add this one to the next agenda, to get complete clarification on the comments so we can ship this? Thanks!

index.html
MelSumner: Reviewing it again today, we will take this out and allow for implicit value calculation.

index.html
MelSumner: @joanmarie @aleventhal @jcsteh what do you think about this proposed addition/change?

index.html
MelSumner, line 13005:
```suggestion
				<p>Authors MUST set the value of <code>aria-setsize</code> to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.
```

index.html
mcking65: What is the issue with specifying setsize on a container in a multi-level list? Obviously, you can only do this if a container with the appropriate scope exists. That is not true in all implementations, but that should not prevent it when such a container does exist.

mcking65, line 12363: Because it is not always required, the normative requirement needs a conditional phrase at the beginning. Otherwise, it can be read as requiring the attribute. The requirement is really only about the range of allowed values.

Example:

```suggestion
				<p>When specifying <pref>aria-posinset</pref>, Authors MUST specify a value that is an integer greater than or equal to 1, and less than or equal to the size of the set when that size is known. If authors specify <pref>aria-posinset</pref>,  authors MUST also specify a value for <pref>aria-setsize</pref>.</p>
```

index.html
```diff
@@ -12360,7 +12360,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
   &lt;li role="option" aria-setsize="16" <strong>aria-posinset="7"</strong>&gt; cantaloupes &lt;/li&gt;
   &lt;li role="option" aria-setsize="16" <strong>aria-posinset="8"</strong>&gt; dates &lt;/li&gt;
 &lt;/ul&gt;</pre>
-				<p>Authors MUST set the value for <pref>aria-posinset</pref> to an integer greater than or equal to 1, and less than or equal to the size of the set when that size is known. Authors SHOULD use <pref>aria-setsize</pref>.</p>
+				<p>Authors MUST set the value for <pref>aria-posinset</pref> to an integer greater than or equal to 1, and less than or equal to the size of the set when that size is known. If <pref>aria-posinset</pref> is specified, authors MUST specify a value for <pref>aria-setsize</pref>.</p>
 				<p>When exposing <code>aria-posinset</code> on a <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, or <rref>menuitemradio</rref>, authors SHOULD set the value of <code>aria-posinset</code> with respect to the total number of items in the <rref>menu</rref>, excluding any separators.</p>
```
mcking65, line 12364: This may be slightly beyond the scope of the issue, but is simple to address. This wording of "exposing" is usually a verb we associate with AT, not authors. I recommend changing to following:

```suggestion
				<p>When specifying <code>aria-posinset</code> on a <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, or <rref>menuitemradio</rref>, authors SHOULD set the value of <code>aria-posinset</code> with respect to the total number of items in the <rref>menu</rref>, excluding any separators.</p>
```

mcking65 (changes requested):

index.html
carmacleod: @mcking65 
> What is the issue with specifying setsize on a container in a multi-level list?

aria-setsize is not supported on containers. Whether or not it should be can be discussed in https://github.com/w3c/aria/issues/524, so that this PR is simply scoped to clarifying what is already there.

index.html
```diff
@@ -13002,7 +13002,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p>If all items in a set are present in the document structure, it is not necessary to set this property, as the <a>user agent</a> can automatically calculate the set size and position for each item. However, if only a portion of the set is present in the document structure at a given moment (in order to reduce document size), this property is needed to provide an explicit indication of set size.</p>
-				<p>Authors MUST set the value of <code>aria-setsize</code> to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.</p>
+				<p>Authors MUST set the value of <code>aria-setsize</code> to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.
```
carmacleod, line 13005: This statement in setsize needs a conditional phrase similar to @mcking65's suggestion for posinset, above:
> Because it is not always required, the normative requirement needs a conditional phrase at the beginning. Otherwise, it can be read as requiring the attribute.

```suggestion
				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.</p>
```

index.html
```diff
@@ -13002,7 +13002,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p>When exposing <code>aria-setsize</code> on a <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, or <rref>menuitemradio</rref>, authors SHOULD set the value of <code>aria-setsize</code> based on the total number of items in the <rref>menu</rref>, excluding any separators.</p>
```
carmacleod, line 13006: We should probably change this "exposing" to "specifying", for the same reason as above: usually the word "exposing" is used to describe User Agent requirements (not authors).
```suggestion
				<p>When specifying <code>aria-setsize</code> on a <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, or <rref>menuitemradio</rref>, authors SHOULD set the value of <code>aria-setsize</code> based on the total number of items in the <rref>menu</rref>, excluding any separators.</p>
```

carmacleod (approved):
Thanks, @MelSumner !

Giving +1 assuming the 2 suggestions about adding a conditional phrase will be committed, plus the little one-word suggestion about changing 'exposing' to 'specifying' in setsize (was already changed in posinset).

carmacleod (approved):
+1

Thanks @MelSumner - I think this adds clarity.

So, the only normative change is:
> If authors specify `aria-posinset`,  authors MUST also specify a value for `aria-setsize`.

And (correct me if I'm wrong) we didn't require the opposite (i.e. if setsize then must posinset) because we're allowing the UA's to do the posinset calculation if they know the setsize.

I won't hold up the merge for the following, but just wondering if this paragraph needs to be reworded? (Showing the posinset paragraph, but it is similar in setsize except ending in "an explicit indication of set size.")
> If all items in a set are present in the document structure, it is not necessary to set this attribute , as the user agent can automatically calculate the set size and position for each item. However, if only a portion of the set is present in the document structure at a given moment, this property is needed to provide an explicit indication of an element's position.

jnurthen (changes requested):
Agree with @carmacleod
I think a small change is needed in the aria-posinset definition. Not totally happy with the wording so feel free to modify if something better comes to mind!

from
				<p>If all items in a set are present in the document structure, it is not necessary to set this <a>attribute</a>, as the <a>user agent</a> can automatically calculate the set size and position for each item. However, if only a portion of the set is present in the document structure at a given moment, this <a>property</a> is needed to provide an explicit indication of an element's position.</p>

to

carmacleod:
@MelSumner I think if we just put James' suggested change into the aria-posinset definition, then this PR is ready to merge! :)

index.html
```diff
@@ -13002,8 +13002,8 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				<p>Defines the number of items in the current set of listitems or treeitems. Not required if all elements in the set are present in the <abbr title="Document Object Model">DOM</abbr>. See related <pref>aria-posinset</pref>.</p>
 				<p>This <a>property</a> is marked on the members of a set, not the container element that collects the members of the set. To orient the user by saying an element is "item X out of Y," the <a>assistive technologies</a> would use X equal to the <pref>aria-posinset</pref> <a>attribute</a> and Y equal to the <code>aria-setsize</code> attribute.</p>
 				<p>If all items in a set are present in the document structure, it is not necessary to set this property, as the <a>user agent</a> can automatically calculate the set size and position for each item. However, if only a portion of the set is present in the document structure at a given moment (in order to reduce document size), this property is needed to provide an explicit indication of set size.</p>
```
MelSumner, line 13004:
```suggestion
				<p>If all items up to the current item in a set are present in the document structure, it is not necessary to set this <a>attribute</a>, as the <a>user agent</a> can automatically calculate the position for these items. However, if all previous items in the set are not present in the document structure at a given moment, the author MUST set this <a>attribute</a>to provide an explicit indication of an element's position.</p>
```

scottaohara (approved):

jnurthen (approved):

PR #1335, opened 2020-10-07
Add a conditional on when aria-controls is needed

### LINKED ISSUE #1334: Combobox shouldn't need aria-controls until the popup is rendered
Combobox states
“Authors MUST set aria-controls on a combobox element to a value that refers to the combobox popup element.”

This only needs to be true once the popup has rendered. Until that time the idref is unknown.

jnurthen:
resolves #1334

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1335.html" title="Last updated on Nov 20, 2020, 12:30 AM UTC (eef2e15)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1335/19e5b99...eef2e15.html" title="Last updated on Nov 20, 2020, 12:30 AM UTC (eef2e15)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2113,7 +2113,7 @@
 				</p>
 				<p>
 				Authors MUST ensure the popup element associated with a <code>combobox</code> has a role of <rref>listbox</rref>, <rref>tree</rref>, <rref>grid</rref>, or <rref>dialog</rref>.
-				Authors MUST set <pref>aria-controls</pref> on a <code>combobox</code> element to a value that refers to the <code>combobox</code> popup element.
+				When the popup is displayed, authors MUST set <pref>aria-controls</pref> on a <code>combobox</code> element to a value that refers to the <code>combobox</code> popup element.
 				</p>
 				<p>
 				Elements with the role <code>combobox</code> have an implicit <pref>aria-haspopup</pref> value of <code>listbox</code>.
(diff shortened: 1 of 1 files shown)


carmacleod:
Does it make sense to add an Implied

carmacleod (approved):
Approving as-is, but would like an answer on whether or not adding Implicit value for role makes sense (or would just add more confusion).
Thanks @jnurthen !

jnurthen:
Yeah - we nee

> Approving as-is, but would like an answer on whether or not adding Implicit value for role makes sense (or would just add more confusion).
> Thanks @jnurthen !

We have to do something - either add an implicit value or move to supported... Lets add to today's agenda and see what others think.

mcking65 (changes requested):
This PR makes it more clear that aria-controls needs to be supported not required. If it is not present when the combobox is collapsed, the combobox would not validate.

carmacleod:
+1 for the new state of this PR, which does 2 things:
- Adds "When the popup is displayed," so that the requirement for authors says the following:
  > When the popup is displayed, authors MUST set aria-controls on a combobox element to a value that refers to the combobox popup element.
- Moves aria-controls from required to supported in the characteristics table.

MelSumner (approved):
👍

mcking65 (approved):
Looks great!

PR #1340, opened 2020-10-11
Listbox and tree: clarify requirements for selected and checked

### LINKED ISSUE #700: Should aria-selected be listed as required for option role?

### LINKED ISSUE #1052: aria-checked on option and treeitem

mcking65:
This PR will resolve #700 and resolve #1052. It supercedes #799.

Changes include the following.

Removes "selectable" from the definition of option to support cases where either an item is checkable instead of selectable or an item is neither selectable nor checkable (i.e., similar to disabled). The definition of option is then:

>An item in a listbox.

Similarly simplifies the definition of treeitem.

Replaces the statement:

>Elements with the role option have an implicit aria-selected value of false.

with a list of conditions that must be met before a user agent can provide an implicit value for aria-selected.

Adds the same restrictions on user agent provision of implicit aria-selected values to treeitem and tab.

These user agent restrictions enable authors to:
1. Use aria-checked instead of aria-selected in a listbox or tree.
2. Indicate that an option or treeitem is not selectable or checkable by setting checked or selected to undefined.

Adds authoring guidance that:
1. Weakly suggests use of checked instead of selected in multi-select listboxes and trees.
2. Strongly discourages use of both checked and selected in the same listbox or tree.

In the characteristics table for option:
1. Removes aria-selected from required properties and adds to supported properties.
2. Removes the implicit value for aria-selected.

***

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Faria%2Fc51f1486be17adf29d164be548147d860232aa5f%2Findex.html%3FisPreview%3Dtrue)

```
Navigation timeout of 26205 ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231340.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5992,7 +5992,7 @@
 			<rdef>option</rdef>
 			<div class="role-description">
 				<p>An item in a <rref>listbox</rref>.</p>
-				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or <a>owned</a> by, an element with the <a>role</a> <rref>listbox</rref> or <rref>group</rref> within a <code>listbox</code>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
+				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or <a>owned</a> by, an element with <a>role</a> <rref>listbox</rref> or an element with <a>role</a> <rref>group</rref> that is contained in, or <a>owned</a> by, an element with <a>role</a> <code>listbox</code>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
         <p>User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>option</rref> in a <rref>listbox</rref> if the following conditions are met, but user agents MUST NOT provide an implicit value for <sref>aria-selected</sref> if any of the following conditions are not met:</p>
         <ul>
           <li>The value of <pref>aria-multiselectable</pref> on the <rref>listbox</rref> is <code>false</code> or <code>undefined</code>.</li>
--- a/index.html
+++ b/index.html
@@ -9924,8 +9924,7 @@
         <p>
           Authors MAY indicate selection for <rref>treeitem</rref> elements  using either <sref>aria-selected</sref> or <sref>aria-checked</sref>.
           Some user interfaces indicate selection with <sref>aria-selected</sref> in single-select trees and with <sref>aria-checked</sref> in multi-select trees.
-          Authors SHOULD NOT specify both <sref>aria-selected</sref> and <sref>aria-checked</sref> on <rref>treeitem</rref> elements contained by the same <rref>tree</rref>.
-          However, authors MAY use both in the same <rref>tree</rref> in the extremely rare circumstances where all the following conditions are met:
+          Authors SHOULD NOT specify both <sref>aria-selected</sref> and <sref>aria-checked</sref> on <rref>treeitem</rref> elements contained by the same <rref>tree</rref> except in the extremely rare circumstances where all the following conditions are met:
         </p>
         <ul>
           <li>The meaning and purpose of <sref>aria-selected</sref> is different from the meaning and purpose of <sref>aria-checked</sref> in the user interface.</li>
--- a/index.html
+++ b/index.html
@@ -9916,7 +9916,10 @@
 				<p>An item in a <rref>tree</rref>.</p>
         				<p>A <rref>treeitem</rref> <a>element</a> may contain a sub-level group of elements that can be expanded or collapsed. An expandable collection of <code>treeitem</code> elements are enclosed in an element with the <rref>group</rref> <a>role</a>.</p>
 				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>treeitem</code> are contained in, or <a>owned</a> by, an element with role <rref>tree</rref> or an element with role <rref>group</rref> that is contained in, or owned by, an element with role <rref>treeitem</rref>.</p>
-        <p>User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>treeitem</rref> in a <rref>tree</rref> if the following conditions are met, but user agents MUST NOT provide an implicit value for <sref>aria-selected</sref> if any of the following conditions are not met:</p>
+        <p>
+          User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>treeitem</rref> in a <rref>tree</rref> only if the following conditions are met.
+          If a user agent provides an implicit value for <sref>aria-selected</sref> on a <rref>treeitem</rref>, the user agent MUST ensure the following conditions are met before providing the implicit value:
+        </p>
         <ul>
           <li>The value of <pref>aria-multiselectable</pref> on the <rref>tree</rref> is <code>false</code> or <code>undefined</code>.</li>
           <li>None of the <rref>treeitem</rref> elements in the <rref>tree</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>  
@@ -9926,8 +9929,10 @@
           Otherwise, if a user agent provides an implicit <sref>aria-selected</sref> value for a <rref>treeitem</rref>, the value SHOULD be <code>false</code>.
         </p>
         <p>
-          The <rref>treeitem</rref> role supports both <sref>aria-selected</sref> and <sref>aria-checked</sref> because it is recommended that authors indicate selection with <sref>aria-selected</sref> in single-select trees and with <sref>aria-checked</sref> in multi-select trees.
-          Authors SHOULD NOT specify both <sref>aria-selected</sref> and <sref>aria-checked</sref> on <rref>treeitem</rref> elements contained by the same <rref>tree</rref> except in the extremely rare circumstances where all the following conditions are met:
+          Authors MAY indicate selection for <rref>treeitem</rref> elements  using either <sref>aria-selected</sref> or <sref>aria-checked</sref>.
+          Some user interfaces indicate selection with <sref>aria-selected</sref> in single-select trees and with <sref>aria-checked</sref> in multi-select trees.
+          Authors SHOULD NOT specify both <sref>aria-selected</sref> and <sref>aria-checked</sref> on <rref>treeitem</rref> elements contained by the same <rref>tree</rref>.
+          However, authors MAY use both in the same <rref>tree</rref> in the extremely rare circumstances where all the following conditions are met:
         </p>
         <ul>
           <li>The meaning and purpose of <sref>aria-selected</sref> is different from the meaning and purpose of <sref>aria-checked</sref> in the user interface.</li>
--- a/index.html
+++ b/index.html
@@ -12971,12 +12971,8 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
 			<sdef>aria-selected</sdef>
 			<div class="state-description">
 				<p>Indicates the current "selected" <a>state</a> of various <a>widgets</a>. See related <sref>aria-checked</sref> and <sref>aria-pressed</sref>.</p>
-				<p>This <a>attribute</a> is used with single-selection and multiple-selection widgets:</p>
-				<ol>
-					<li>Single-selection containers where the currently focused item is not selected. The selection normally follows the focus, and is managed by the <a>user agent</a>.</li>
-					<li>Multiple-selection containers. Authors SHOULD ensure that any selectable descendant of a container in which the <pref>aria-multiselectable</pref> attribute is <code>true</code> specifies a value of either <code>true</code> or <code>false</code> for the <sref>aria-selected</sref> attribute.</li>
-				</ol>
-				<p>Any explicit assignment of <sref>aria-selected</sref> takes precedence over the implicit selection based on focus. If no <abbr title="Document Object Model">DOM</abbr> element in the widget is explicitly marked as selected, assistive technologies MAY convey implicit selection which follows the keyboard focus of the <a href="#managingfocus">managed focus</a> widget. If any <abbr title="Document Object Model">DOM</abbr> element in the widget is explicitly marked as selected, the user agent MUST NOT convey implicit selection for the widget.</p>
+				<p>This <a>attribute</a> is used to indicate which elements within single-selection and multiple-selection <rref>composite</rref> widgets are selected.</p>
+        <p>The <rref>option</rref>, <rref>tab</rref>, and <rref>treeitem</rref> roles permit user agents to provide an implicit value for <sref>aria-selected</sref> when specified conditions are met. User agents MUST NOT provide an implicit value for aria-selected in any other circumstance.</p>
 			</div>
 			<table class="state-features">
 				<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -5993,7 +5993,7 @@
 			<div class="role-description">
 				<p>An item in a <rref>listbox</rref>.</p>
 				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or <a>owned</a> by, an element with the <a>role</a> <rref>listbox</rref> or <rref>group</rref> within a <code>listbox</code>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
-        <p>User agents MAY provide an implicit value of <sref>aria-selected</sref> for each <rref>option</rref> in a <rref>listbox</rref> if the following conditions are met, but user agents MUST NOT provide an implicit value of <sref>aria-selected</sref> if any of the following conditions are not met:</p>
+        <p>User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>option</rref> in a <rref>listbox</rref> if the following conditions are met, but user agents MUST NOT provide an implicit value for <sref>aria-selected</sref> if any of the following conditions are not met:</p>
         <ul>
           <li>The value of <pref>aria-multiselectable</pref> on the <rref>listbox</rref> is <code>false</code> or <code>undefined</code>.</li>
           <li>None of the <rref>option</rref> elements in the <rref>listbox</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>  
--- a/index.html
+++ b/index.html
@@ -5993,9 +5993,7 @@
 			<div class="role-description">
 				<p>An item in a <rref>listbox</rref>.</p>
 				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or <a>owned</a> by, an element with <a>role</a> <rref>listbox</rref> or an element with <a>role</a> <rref>group</rref> that is contained in, or <a>owned</a> by, an element with <a>role</a> <code>listbox</code>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
-        <p>
-          In certain conditions, a user agent MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>option</rref> in a <rref>listbox</rref>, and if it does, the user agent MUST ensure the following conditions are met before providing an implicit value:
-        </p>
+        <p>In certain conditions, a user agent MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>option</rref> in a <rref>listbox</rref>, and if it does, the user agent MUST ensure the following conditions are met before providing an implicit value:</p>
         <ul>
           <li>The value of <pref>aria-multiselectable</pref> on the <rref>listbox</rref> is <code>false</code> or <code>undefined</code>.</li>
           <li>None of the <rref>option</rref> elements in the <rref>listbox</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>  
--- a/index.html
+++ b/index.html
@@ -9927,7 +9927,7 @@
           <li>None of the <rref>treeitem</rref> elements in the <rref>tree</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>  
         </ul>
         <p>
-          If a user agent provides an implicit <sref>aria-selected</sref> value for a <rref>treeitem</rref>, the value SHOULD be <code>true</code> if the <rref>treeitem</rref> has DOM focus or the <ref>tree</ref> has DOM focus and the <rref>treeitem</rref> is referenced by <pref>aria-activedescendant</pref>.
+          If a user agent provides an implicit <sref>aria-selected</sref> value for a <rref>treeitem</rref>, the value SHOULD be <code>true</code> if the <rref>treeitem</rref> has DOM focus or the <rref>tree</rref> has DOM focus and the <rref>treeitem</rref> is referenced by <pref>aria-activedescendant</pref>.
           Otherwise, if a user agent provides an implicit <sref>aria-selected</sref> value for a <rref>treeitem</rref>, the value SHOULD be <code>false</code>.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -6003,8 +6003,8 @@
           Otherwise, if a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>false</code>.
         </p>
         <p>
-          The <rref>option</rref> role supports both <sref>aria-selected</sref> and <sref>aria-checked</sref> because it is common for authors to use selection in single-select list boxes and check marks in multi-select list boxes.
-          However, authors SHOULD NOT specify both <sref>aria-selected</sref> and <sref>aria-checked</sref> on <rref>option</rref> elements contained by the same <rref>listbox</rref> except in the extremely rare circumstances where all the following conditions are met:
+          The <rref>option</rref> role supports both <sref>aria-selected</sref> and <sref>aria-checked</sref> because it is recommended that authors indicate selection with <sref>aria-selected</sref> in single-select list boxes and with <sref>aria-checked</sref> in multi-select list boxes.
+          Authors SHOULD NOT specify both <sref>aria-selected</sref> and <sref>aria-checked</sref> on <rref>option</rref> elements contained by the same <rref>listbox</rref> except in the extremely rare circumstances where all the following conditions are met:
         </p>
         <ul>
           <li>The meaning and purpose of <sref>aria-selected</sref> is different from the meaning and purpose of <sref>aria-checked</sref> in the user interface.</li>
--- a/index.html
+++ b/index.html
@@ -5994,7 +5994,7 @@
 				<p>An item in a <rref>listbox</rref>.</p>
 				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or <a>owned</a> by, an element with <a>role</a> <rref>listbox</rref> or an element with <a>role</a> <rref>group</rref> that is contained in, or <a>owned</a> by, an element with <a>role</a> <code>listbox</code>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
         <p>
-          User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>option</rref> in a <rref>listbox</rref>, however the user agent MUST ensure the following conditions are met before providing the implicit value:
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5991,9 +5991,26 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
...
+          Otherwise, if a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>false</code>.
+        </p>
+        <p>
+          The <rref>option</rref> role supports both <sref>aria-selected</sref> and <sref>aria-checked</sref> because it is common for authors to use selection in single-select list boxes and check marks in multi-select list boxes.
+          However, authors SHOULD NOT specify both <sref>aria-selected</sref> and <sref>aria-checked</sref> on <rref>option</rref> elements contained by the same <rref>listbox</rref> except in the extremely rare circumstances where all the following conditions are met:
+        </p>
+        <ul>
+          <li>The meaning and purpose of <sref>aria-selected</sref> is different from the meaning and purpose of <sref>aria-checked</sref> in the user interface.</li>
+          <li>The user interface makes the mening and purpose of each state apparent.</li>
+          <li>The user interface provides a separate method for controling each state.</li>
```
carmacleod, line 6012: Couple of typos:
```suggestion
          <li>The user interface makes the meaning and purpose of each state apparent.</li>
          <li>The user interface provides a separate method for controlling each state.</li>
```

carmacleod:
I like this! 🎉 
Thanks @mcking65, @aleventhal and @smhigley !

carmacleod:
Still need to fix up the wording in [aria-selected](https://pr-preview.s3.amazonaws.com/w3c/aria/1340/02fee23...2929945.html#aria-selected) to match these new words.

This sentence is particularly problematic:
> The selection normally follows the focus, and is managed by the user agent.

It should either say "may be managed by the user agent", or be completely deleted.

carmacleod:
(Accidentally clicked the wrong button and closed this - have reopened.)

mcking65:
This is now ready for working group review. Please see the top comment for a complete summary of changes.

jnurthen:
@jcsteh can you please review this. Thanks.

cookiecrook:
Matt King requests reviewers re-read conversation in #1052 before providing PR review.

carmacleod:
> The option role supports both aria-selected and aria-checked because it is recommended that authors indicate selection with aria-selected in single-select list boxes and with aria-checked in multi-select list boxes.

This is no longer "Weakly suggesting use of checked instead of selected in multi-select listboxes and trees."

I preferred the previous wording:
> ... because it is common for authors to use selection in single-select list boxes and check marks in multi-select list boxes.

Are we certain that we want to make this a strong recommendation?

index.html
```diff
@@ -5991,9 +5991,26 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
...
+        </ul>
+        <p>
+          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or the <ref>listbox</ref> has DOM focus and the option is referenced by <pref>aria-activedescendant</pref>.
+          Otherwise, if a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>false</code>.
+        </p>
+        <p>
+          The <rref>option</rref> role supports both <sref>aria-selected</sref> and <sref>aria-checked</sref> because it is recommended that authors indicate selection with <sref>aria-selected</sref> in single-select list boxes and with <sref>aria-checked</sref> in multi-select list boxes.
+          Authors SHOULD NOT specify both <sref>aria-selected</sref> and <sref>aria-checked</sref> on <rref>option</rref> elements contained by the same <rref>listbox</rref> except in the extremely rare circumstances where all the following conditions are met:
```
smhigley, line 6007: My one suggestion would be to tweak the "except when" wording, since (colloquially) it currently implies that authors *should* use both selected and checked when they think those three conditions are met.

Could we split up this line into two sentences, something like "Authors SHOULD NOT specify both aria-selected and aria-checked on option elements contained by the same listbox. However, authors MAY use both in the extremely rare circumstances where all the following conditions are met: [...]" (though probably with some more wording tweaks)

smhigley (approved):
I just left one comment, but I think overall this will be a really valuable update 🎉

jnurthen:
@aleventhal @jcsteh @cookiecrook please review as would love to merge this

carmacleod (changes requested):
> The treeitem role supports both aria-selected and aria-checked because it is recommended that authors indicate selection with aria-selected in single-select trees and with aria-checked in multi-select trees.

Note that the new wording for `treeitem` (and for `option`, as mentioned in https://github.com/w3c/aria/pull/1340#issuecomment-722538496) uses the word "recommended", which is an [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.txt) word equivalent to **SHOULD**.

I think we need to be careful about saying that authors SHOULD use aria-checked instead of aria-selected in multi-select trees and listboxes. Many (most?) of the multi-select trees and listboxes I've seen do not use checkboxes on the items. For example:
- File Explorer's listview (default does not have checkboxes),
- Finder's "list" view is a tree,
- VSCode file tree, and other IDE file trees (desktop and online).

The only time I can think of where checkboxes make more sense than selection in a multi-select listbox or tree is when the control represents a bunch of user settings, for example:
- macOS Settings - Keyboard (Services is a checkbox tree, other keyboard categories are checkbox lists),
- Windows Control Panel "Windows Features" dialog has a checkbox tree.

index.html
```diff
@@ -9891,9 +9908,27 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
 			<rdef>treeitem</rdef>
 			<div class="role-description">
-				<p>An option item of a <rref>tree</rref>. This is an <a>element</a> within a tree that may be expanded or collapsed if it contains a sub-level group of tree item elements.</p>
-				<p>A collection of <code>treeitem</code> elements to be expanded and collapsed are enclosed in an element with the <rref>group</rref> <a>role</a>.</p>
-				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>treeitem</code> are contained in, or <a>owned</a> by, an element with the role <rref>group</rref> or <rref>tree</rref>.</p>
+				<p>An item in a <rref>tree</rref>.</p>
+        				<p>A <rref>treeitem</rref> <a>element</a> may contain a sub-level group of elements that can be expanded or collapsed. An expandable collection of <code>treeitem</code> elements are enclosed in an element with the <rref>group</rref> <a>role</a>.</p>
+				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>treeitem</code> are contained in, or <a>owned</a> by, an element with role <rref>tree</rref> or an element with role <rref>group</rref> that is contained in, or owned by, and element with role <rref>treeitem</rref>.</p>
```
jcsteh, line 9913: typo
```suggestion
				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>treeitem</code> are contained in, or <a>owned</a> by, an element with role <rref>tree</rref> or an element with role <rref>group</rref> that is contained in, or owned by, an element with role <rref>treeitem</rref>.</p>
```

index.html
```diff
@@ -8694,11 +8708,14 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
+				<p>Authors SHOULD ensure that a selected tab has its <sref>aria-selected</sref> attribute set to <code>true</code>, that inactive tab elements have their <sref>aria-selected</sref> attribute set to <code>false</code>, and that the currently selected tab provides a visual indication that it is selected.</p>
+        <p>User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>tab</rref> in a <rref>tablist</rref> if the following conditions are met, but user agents MUST NOT provide an implicit value for <sref>aria-selected</sref> if any of the following conditions are not met:</p>
+        <ul>
+          <li>The value of <pref>aria-multiselectable</pref> on the <rref>tablist</rref> is <code>false</code> or <code>undefined</code>.</li>
+          <li>None of the <rref>tab</rref> elements in the <rref>tablist</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-expanded</sref>.</li>
```
jcsteh, line 8717: It's probably silly to use aria-expanded in a single-select tablist. However, I think this sentence means that if aria-expanded is set in a single-select tablist, there is no implicit aria-selected. Why does aria-expanded have this effect in this case?

jcsteh (commented):
I agree with @carmacleod's [concerns about saying authors should use aria-checked instead of aria-selected in multi-select listboxes/trees](https://github.com/w3c/aria/pull/1340#pullrequestreview-533918960). This might be what we want to happen in future (and there are even some conversations about changing this in the HTML select widget), but as far as I've seen, this isn't what's happening today.

index.html
```diff
@@ -5991,9 +5991,26 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
...
-				<p>Elements with the role <code>option</code> have an implicit <sref>aria-selected</sref> value of <code>false</code>.</p>
+				<p>An item in a <rref>listbox</rref>.</p>
+				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or <a>owned</a> by, an element with the <a>role</a> <rref>listbox</rref> or <rref>group</rref> within a <code>listbox</code>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
+        <p>User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>option</rref> in a <rref>listbox</rref> if the following conditions are met, but user agents MUST NOT provide an implicit value for <sref>aria-selected</sref> if any of the following conditions are not met:</p>
```
cookiecrook, line 5996: Editorial nit. Is is possible to rephrase some of the negative inversions: "…but user agents MUST **NOT** provide an implicit value [if conditions] are **not** met"… Also, some of the referenced list items start with "None of" so I'm having to triple invert when reading this. (A "None of" condition is "not met" so therefore the UA "must not" do what?)

Perhaps "MUST NOT provide an implicit value [unless one/some/all of the following conditions are true]" and then include related inversions of the bullets? (e.g. "One or more of" instead of "None of") Okay to keep the extra nots if inverting the list items becomes too problematic. I think it's accurate, just a little hard to read.

index.html
```diff
@@ -8694,11 +8708,14 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
+				<p>Authors SHOULD ensure that a selected tab has its <sref>aria-selected</sref> attribute set to <code>true</code>, that inactive tab elements have their <sref>aria-selected</sref> attribute set to <code>false</code>, and that the currently selected tab provides a visual indication that it is selected.</p>
+        <p>User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>tab</rref> in a <rref>tablist</rref> if the following conditions are met, but user agents MUST NOT provide an implicit value for <sref>aria-selected</sref> if any of the following conditions are not met:</p>
```
cookiecrook, line 8714: ditto re: "MUST NOT" if "none of" are "not met"

index.html
```diff
@@ -9891,9 +9908,27 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
-				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>treeitem</code> are contained in, or <a>owned</a> by, an element with the role <rref>group</rref> or <rref>tree</rref>.</p>
+				<p>An item in a <rref>tree</rref>.</p>
+        				<p>A <rref>treeitem</rref> <a>element</a> may contain a sub-level group of elements that can be expanded or collapsed. An expandable collection of <code>treeitem</code> elements are enclosed in an element with the <rref>group</rref> <a>role</a>.</p>
+				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>treeitem</code> are contained in, or <a>owned</a> by, an element with role <rref>tree</rref> or an element with role <rref>group</rref> that is contained in, or owned by, an element with role <rref>treeitem</rref>.</p>
+        <p>User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>treeitem</rref> in a <rref>tree</rref> if the following conditions are met, but user agents MUST NOT provide an implicit value for <sref>aria-selected</sref> if any of the following conditions are not met:</p>
```
cookiecrook, line 9914: ditto

cookiecrook (approved):

index.html
aleventhal: How about "if and only if" the following conditions are met -- that seems clearer than what's there now.

index.html
```diff
@@ -5991,9 +5991,26 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
...
-				<p>Elements with the role <code>option</code> have an implicit <sref>aria-selected</sref> value of <code>false</code>.</p>
+				<p>An item in a <rref>listbox</rref>.</p>
+				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or <a>owned</a> by, an element with the <a>role</a> <rref>listbox</rref> or <rref>group</rref> within a <code>listbox</code>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
+        <p>User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>option</rref> in a <rref>listbox</rref> if the following conditions are met, but user agents MUST NOT provide an implicit value for <sref>aria-selected</sref> if any of the following conditions are not met:</p>
+        <ul>
+          <li>The value of <pref>aria-multiselectable</pref> on the <rref>listbox</rref> is <code>false</code> or <code>undefined</code>.</li>
```
aleventhal, line 5998: So this includes when the author literally has the attribute aria-multiselectable="undefined"

index.html
```diff
@@ -5991,9 +5991,26 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
...
+          <li>None of the <rref>option</rref> elements in the <rref>listbox</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>
```
aleventhal, line 5999: If there are 300 options, do you want us to go through each one and check? Or can the UA optimize a bit here?

index.html
```diff
@@ -9891,9 +9908,27 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
+        				<p>A <rref>treeitem</rref> <a>element</a> may contain a sub-level group of elements that can be expanded or collapsed. An expandable collection of <code>treeitem</code> elements are enclosed in an element with the <rref>group</rref> <a>role</a>.</p>
+				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>treeitem</code> are contained in, or <a>owned</a> by, an element with role <rref>tree</rref> or an element with role <rref>group</rref> that is contained in, or owned by, an element with role <rref>treeitem</rref>.</p>
+        <p>User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>treeitem</rref> in a <rref>tree</rref> if the following conditions are met, but user agents MUST NOT provide an implicit value for <sref>aria-selected</sref> if any of the following conditions are not met:</p>
+        <ul>
+          <li>The value of <pref>aria-multiselectable</pref> on the <rref>tree</rref> is <code>false</code> or <code>undefined</code>.</li>
+          <li>None of the <rref>treeitem</rref> elements in the <rref>tree</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>
```
aleventhal, line 9917: Since is is basically the same text as for option, does it make sense to create a separate section for Multiselectable Container Widgets Supporting Checked or Selected, and then redirect the reader to there? That way we don't have to wonder if it's exactly the same.

aleventhal: Also, that will make it easier to provide the same functionality in the future, e.g. if we ever add a listgrid role.

Finally, why are we only doing this for tree and listbox? What about treegrid? Can't rows be checked or selected there? Imagine a threaded email interface.

index.html
```diff
@@ -5991,9 +5991,28 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
 		<div class="role" id="option">
 			<rdef>option</rdef>
 			<div class="role-description">
-				<p>A selectable item in a <rref>listbox</rref>.</p>
-				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or <a>owned</a> by, an element with the <a>role</a>  <rref>listbox</rref> or <rref>group</rref> within a <code>listbox</code>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
-				<p>Elements with the role <code>option</code> have an implicit <sref>aria-selected</sref> value of <code>false</code>.</p>
+				<p>An item in a <rref>listbox</rref>.</p>
+				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or <a>owned</a> by, an element with the <a>role</a> <rref>listbox</rref> or <rref>group</rref> within a <code>listbox</code>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
```
carmacleod, line 5995: For consistency with the nicely-updated clarifying words in treeitem, please change this sentence about groups within listboxes:

```suggestion
				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or <a>owned</a> by, an element with <a>role</a> <rref>listbox</rref> or an element with <a>role</a> <rref>group</rref> that is contained in, or <a>owned</a> by, an element with <a>role</a> <code>listbox</code>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
```

index.html
mcking65: Fixed with 03732dc

index.html
mcking65: Fixed with 2df964f.
@aleventhal  I kept both the may only if and the then changed must not to must. I think it's important to have a MUST so the requirement is tested. Hopefully the new wording is much easier to read.

index.html
mcking65: Is that a problem? Should we word it so it does not? That would be some real hair splitting.

index.html
mcking65: Good question.

I think we've figured out that if UA never provides implicit values, nothing will break in a way that harms users. So, one approach could be to limit conditions where implicit values are provided if you want a UA to provide them at all.

When building the AX tree, isn't it necessary to process every option and at the same time set its states and props? So, you might be doing this anyway just to create the tree.

index.html
mcking65: I don't think it would simplify the wording to consolidate; it would still have to explicitly mention the option, listbox, tree, and treeitem roles.

In grid and treegrid, aria-checked is not supported on rows. In those widgets, if you were crazy enough to use both checked and selected states, you can just put  a checkbox in a cell. That is not possible in listbox and tree; you can't put a checkbox inside an option or treeitem.

index.html
mcking65: Resolved by d97ef54

index.html
mcking65: Resolved by d97ef54

index.html
mcking65: Same as the argument for aria-checked similarly effecting a listbox.

If the author is using aria-expanded, we assume it is in lieu of aria-selected. If it is not, then it is up to the author to explicitly declare both. We don't want the user agent to assume the author wants both.

mcking65:
@carmacleod @jcsteh 

I hope the following language addresses your concerns about the language about the suitability of selected for single-selects and checked for multi-selects.

>Authors MAY indicate selection for option elements using either aria-selected or aria-checked. Some user interfaces indicate selection with aria-selected in single-select list boxes and with aria-checked in multi-select list boxes. Authors SHOULD NOT specify both aria-selected and aria-checked on option elements contained by the same listbox. However, authors MAY use both in the same listbox in the extremely rare circumstances where all the following conditions are met:
>
>* The meaning and purpose of aria-selected  is different from the meaning and purpose of aria-checked  in the user interface. 
>* The user interface makes the meaning and purpose of each state apparent. 
>* The user interface provides a separate method for controlling each state. 

Similar language is also in the definition of treeitem.

mcking65:
@jnurthen @smhigley @carmacleod @aleventhal @jcsteh @cookiecrook 

I'm pretty sure I've addressed all feedback now. Please let me know if I've overlooked something or the changes are not satisfactory.

sinabahram:
@mcking65  thanks for the work on this.

We have this sentence:
Authors SHOULD NOT specify both aria-selected and aria-checked on option elements contained by the same listbox.

I do not agree with that sentence on its own in the case of multiselect. I thought our group consensus centered around limiting conveyance of selection, but not restricting both being used, since that's why there are criteria listed after that restriction for when both can be used.

Also, as written, it is a touch vague about whether we mean on the same option element or across option elements, plural, in the listbox.

Can we qualify that sentence to make it far less absolute and to only apply to the conveyance of selection, but not to restrict both being used on the same option element , as that was the point of the restriction carveouts?

For example:
Authors SHOULD NOT use both aria-selected and aria-checked to convey the same state on option elements contained by the same listbox.

That way, it is ok to use checked and selected for conveying different states e.g. the to-do list example with them being marked as done VS selected and marked as done as a group.

mcking65:
@sinabahram commented:

> We have this sentence:
> Authors SHOULD NOT specify both aria-selected and aria-checked on option elements contained by the same listbox.
> 
> I do not agree with that sentence on its own in the case of multiselect. I thought our group consensus centered around limiting conveyance of selection, but not restricting both being used, since that's why there are criteria listed after that restriction for when both can be used.

That sentence does not stand on its own. It is a should not, not a must not. The following sentence contains a normative may statement and provides for the exception you want. The two sentences together read:

>Authors SHOULD NOT specify both aria-selected and aria-checked on option elements contained by the same listbox. However, authors MAY use both in the same listbox in the extremely rare circumstances where all the following conditions are met:

@sinabahram commented:

> Also, as written, it is a touch vague about whether we mean on the same option element or across option elements, plural, in the listbox.

You can use either or both properties on some or all of the options. I don't think it is necessary to explicitly spell that out. In other words, if the language doesn't explicitly prohibit something, it allows it.

index.html
```diff
@@ -5991,9 +5991,31 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
...
-				<p>Elements with the role <code>option</code> have an implicit <sref>aria-selected</sref> value of <code>false</code>.</p>
+				<p>An item in a <rref>listbox</rref>.</p>
+				<p>Authors MUST ensure <a>elements</a> with <a>role</a> <code>option</code> are contained in, or <a>owned</a> by, an element with <a>role</a> <rref>listbox</rref> or an element with <a>role</a> <rref>group</rref> that is contained in, or <a>owned</a> by, an element with <a>role</a> <code>listbox</code>. Options not associated with a <rref>listbox</rref> might not be correctly mapped to an <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a>.</p>
+        <p>
+          User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>option</rref> in a <rref>listbox</rref> only if the following conditions are met.
+          If a user agent provides an implicit value for <sref>aria-selected</sref> on an <rref>option</rref>, the user agent MUST ensure the following conditions are met before providing the implicit value:
```
carmacleod, line 5998: It feels like there are too many words and sentences here, and some of them feel redundant.

Maybe reducing into one sentence and using "if and only if" (as suggested by @aleventhal earlier) can still have the same meaning?
```suggestion
          User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>option</rref> in a <rref>listbox</rref> if and only if the following conditions are met:
```

Alternatively, if you feel strongly that a MUST needs to follow the MAY, then perhaps the following would work?
```suggestion
          User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>option</rref> in a <rref>listbox</rref>, however the user agent MUST ensure the following conditions are met before providing the implicit value:
```

If you are ok with either of these suggestions, then the same fix would need to be made for treeitem and tab.

index.html
```diff
@@ -5991,9 +5991,31 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
...
+        </p>
+        <ul>
+          <li>The value of <pref>aria-multiselectable</pref> on the <rref>listbox</rref> is <code>false</code> or <code>undefined</code>.</li>
+          <li>None of the <rref>option</rref> elements in the <rref>listbox</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>  
+        </ul>
+        <p>
+          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or the <ref>listbox</ref> has DOM focus and the option is referenced by <pref>aria-activedescendant</pref>.
```
carmacleod, line 6005: Need rref for listbox (not just ref). Also add rref for last instance of option:
```suggestion
          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or the <rref>listbox</rref> has DOM focus and the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
```

index.html
```diff
@@ -9891,9 +9916,32 @@ <h5>Presentational Roles Conflict Resolution</h5>
...
+        <p>
+          User agents MAY provide an implicit value for <sref>aria-selected</sref> for each <rref>treeitem</rref> in a <rref>tree</rref> only if the following conditions are met.
+          If a user agent provides an implicit value for <sref>aria-selected</sref> on a <rref>treeitem</rref>, the user agent MUST ensure the following conditions are met before providing the implicit value:
+        </p>
+        <ul>
+          <li>The value of <pref>aria-multiselectable</pref> on the <rref>tree</rref> is <code>false</code> or <code>undefined</code>.</li>
+          <li>None of the <rref>treeitem</rref> elements in the <rref>tree</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>  
+        </ul>
+        <p>
+          If a user agent provides an implicit <sref>aria-selected</sref> value for a <rref>treeitem</rref>, the value SHOULD be <code>true</code> if the <rref>treeitem</rref> has DOM focus or the <ref>tree</ref> has DOM focus and the <rref>treeitem</rref> is referenced by <pref>aria-activedescendant</pref>.
```
carmacleod, line 9931: Need rref for tree (not just ref):
```suggestion
          If a user agent provides an implicit <sref>aria-selected</sref> value for a <rref>treeitem</rref>, the value SHOULD be <code>true</code> if the <rref>treeitem</rref> has DOM focus or the <rref>tree</rref> has DOM focus and the <rref>treeitem</rref> is referenced by <pref>aria-activedescendant</pref>.
```

carmacleod (approved):
This is really close!
I only have a couple of suggestions left. :)
Giving an approving review anyhow, although I do feel that that 2-sentence MAY-MUST structure (in option, treeitem, and tab) could use some reduction for clarity.

mcking65:
@sinabahram @carmacleod @smhigley 

Editorial changes based on your feedback are all complete now. Thank you for all the great feedback. All the wording seems much easier to understand now.

@joanmarie Note for testing: This PR introduces a new user agent must in option, treeitem, and tab. However, it is only a must for a UA that chooses to provide implicit values for aria-selected.

jnurthen:
@sinabahram @carmacleod @smhigley @cookiecrook @jcsteh @aleventhal this is ready to merge. Any objections please speak up before tomorrow's ARIA meeting.

index.html
smhigley: @mcking65 I like the change in that commit a lot! I think that commit might've gotten lost along the way, though -- I don't see it in this PR right now

index.html
aleventhal: The main problem occurs when there are updates. We try to only process changed nodes, because as a page changes due to interactions, users expect things to be responsive. However, if we optimize so that once an implicit value is used, that it can be used going forward, then we should be ok on the browser performance front. I don't think the optimization needs any verbiage here, but we should clarify that it's a good approach.

aleventhal (approved):
I'm ok with this, but @mcking65 and @jcsteh, I'd like your feedback about by optimization comment. Namely that I want to avoid constantly checking whether there are any descendants that have aria-selected or aria-checked. I'd much prefer that if there ever are, we cache that info and never recompute it, and therefore the implicit selected decision would be based on a prior state.

PR #1387, opened 2021-01-28
focusable->focused to resolve #1381

### LINKED ISSUE #1381: Follow-up: tree inclusion of focusable elements from #1100
Regarding Issues #841 and #1234, as well as PR #1100.

Currently the accessibility tree inclusion section includes this line:
> [include in the accessibility tree…] "Elements that are focusable even if the element or one of its ancestor elements has its aria-hidden attribute set to true."

I had some prior concerns about including all "focusable" elements (even if hidden) versus including hidden elements at the time they became "focused." In #841, I originally suggested:
> …include something like both "visible elements that are focusable" and ~"hidden elements if/when they become focused."

@carmacleod waited on @aleventhal and me to [resolve this comment](https://github.com/w3c/aria/pull/1100#discussion_r407750951) in PR #1100:
> @aleventhal Are you good with this? Does it need to be more specific about whether or not it is exposed to the tree at the time that the element is focused?

And then I [deferred and approved the PR](https://github.com/w3c/aria/pull/1100#issuecomment-614254555):
> We can address that in a follow-up issue. I don't think it needs to delay this one... Marking PR as Reviewed.

But I can't see that I filed the "follow-up issue" and don't see it linked... So filing this new one for more discussion.

This is coming up now because [the implementation of this change in WebKit](https://webkit.org/b/220534) will cause a number of layout tests to fail... I think these are probably expected and can be updated, but I wanted to verify with @aleventhal, @joanmarie, and @carmacleod before it ships. IIRC the discussion from April 2020 correctly, the WG came to the consensus that the language "focusable" was intended, not the "only when focused" language originally proposed. I think that's probably correct, but wanted to point out that this makes it more difficult for authors to use aria-hidden to hide larger containers. 

```html
<div aria-hidden="true">
   <p>This paragraph text is hidden, but the <a href="#">link</a> is exposed to AT out of context.</p>
</div>
```

In order to hide that link (or any other focusable) from AT, the web author would need to do one of the following:
- explicitly add `tabindex="-1" on every focusable element in the "hidden" section (maybe this will work?)
- make the container unrendered (e.g. `hidden=""` a.k.a. `display:none`) instead of `aria-hidden`
- use `inert=""` on the container instead of `aria-hidden`

Another way to look at this, is that the ARIA change ~encourages~ _incentivizes_ authors to use `inert=""` instead of `aria-hidden="true"`... This is probably the right call in the long term, but `inert` is not yet in WebKit.

Long story short... This may not be an issue. If everyone responds, "We're good", this can probably be closed.

cookiecrook:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1387.html" title="Last updated on Feb 4, 2021, 6:13 PM UTC (783dd3d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1387/02fee23...783dd3d.html" title="Last updated on Feb 4, 2021, 6:13 PM UTC (783dd3d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -13297,9 +13297,9 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
 	  <h2>Including Elements in the Accessibility Tree</h2>
           <p>If not excluded from the accessibility tree per the rules above in <a href="#tree_exclusion">Excluding Elements in the Accessibility Tree</a>, user agents MUST provide an <a>accessible object</a> in the <a>accessibility tree</a> for <abbr title="Document Object Model">DOM</abbr> <a>elements</a> that meet any of the following criteria:</p>
           <ul>
-            <li>Elements that are not <a>hidden</a> and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>:
+            <li>Elements that are not <a>hidden</a> and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>, including:
               <ul>
-                <li>Elements that are <a>focusable</a> even if the element or one of its ancestor elements has its <sref>aria-hidden</sref> attribute set to <code>true</code>.</li>
+                <li>Elements that are currently <a>focused</a>, even if the element or one of its ancestor elements has its <sref>aria-hidden</sref> attribute set to <code>true</code>.</li>
                 <li>Elements that are a valid target of an <pref>aria-activedescendant</pref> attribute.</li>
               </ul>
 	        </li>
--- a/index.html
+++ b/index.html
@@ -13295,7 +13295,7 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
 	</section>
 	<section id="tree_inclusion">
 	  <h2>Including Elements in the Accessibility Tree</h2>
-          <p>If not excluded from the accessibility tree per the rules above in <a href="#tree_exclusion">Excluding Elements in the Accessibility Tree</a>, user agents MUST provide an <a>accessible object</a> in the <a>accessibility tree</a> for <abbr title="Document Object Model">DOM</abbr> <a>elements</a> that meet any of the following criteria:</p>
+          <p>If not excluded from or marked as hidden in the accessibility tree per the rules above in <a href="#tree_exclusion">Excluding Elements in the Accessibility Tree</a>, user agents MUST provide an <a>accessible object</a> in the <a>accessibility tree</a> for <abbr title="Document Object Model">DOM</abbr> <a>elements</a> that meet any of the following criteria:</p>
           <ul>
             <li>Elements that are not <a>hidden</a> and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>, including:
               <ul>
--- a/index.html
+++ b/index.html
@@ -13299,7 +13299,7 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
           <ul>
             <li>Elements that are not <a>hidden</a> and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>, including:
               <ul>
-                <li>Elements that are currently <a>focused</a>, even if the element or one of its ancestor elements has its <sref>aria-hidden</sref> attribute set to <code>true</code>.</li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -13297,9 +13297,9 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
 	  <h2>Including Elements in the Accessibility Tree</h2>
           <p>If not excluded from the accessibility tree per the rules above in <a href="#tree_exclusion">Excluding Elements in the Accessibility Tree</a>, user agents MUST provide an <a>accessible object</a> in the <a>accessibility tree</a> for <abbr title="Document Object Model">DOM</abbr> <a>elements</a> that meet any of the following criteria:</p>
           <ul>
-            <li>Elements that are not <a>hidden</a> and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>:
+            <li>Elements that are not <a>hidden</a> and may fire an <a>accessibility <abbr title="Application Program Interface">API</abbr></a> <a>event</a>, including:
               <ul>
-                <li>Elements that are <a>focusable</a> even if the element or one of its ancestor elements has its <sref>aria-hidden</sref> attribute set to <code>true</code>.</li>
+                <li>Elements that are currently <a>focused</a>, even if the element or one of its ancestor elements has its <sref>aria-hidden</sref> attribute set to <code>true</code>.</li>
```
carmacleod, line 13302: We should take out the definition link for now.

The definition of "focusable" (and/or "focused") may be added some time in future in https://github.com/w3c/aria-common/pull/39, and at that time, we can decide which of the 61 instances of "focusable" and 17 instances of "focused" in the ARIA spec should be linked to the definition. 

Since we haven't needed that definition before, we probably shouldn't add it right before 1.2.

```suggestion
                <li>Elements that are currently focused, even if the element or one of its ancestor elements has its <sref>aria-hidden</sref> attribute set to <code>true</code>.</li>
```

carmacleod (approved):
+1
This is (almost) perfect. :)

Approving on the assumption that the broken link to an as-yet-unwritten definition of "focusable/focused" is removed - see suggestion, below.

Thanks @cookiecrook !

scottaohara (approved):

aleventhal:
This is better than what there was but I don't think it goes far enough?
- IIRC we had a plan for the UA to ignore the entire aria-hidden=true subtree when something inside of it gets focused. The benefit of this is that the user may need context around the focused object. This is in one of the many issues filed on aria-hidden somewhere :)
- What about @sinabahram's concern regarding virtual buffers in JAWS and NVDA. They won't be informed of anything in the aria-hidden subtree, and therefore focus won't occur. Unless the user knows to try to turn off the virtual buffer and tab around, this new rule won't help.

aleventhal:
Alright, after talking to @jnurthen, he convinced me that this only changes the requirement to add focused nodes to the tree. It doesn't say you must not add focusable nodes to the tree (e.g. if there's a way to mark them hidden on that platform).

I'm ok +1'ing it if it's understood that we will have slightly different behavior depending on the platform:
- On Mac, where there is no way to mark something hidden, we will add focused aria-hidden nodes to the tree, but keep other focusable nodes pruned out
- On other platforms (or at least on Windows where there's a virtual buffer), we will add all focusable nodes to the tree and mark them as hidden (e.g. STATE_SYTEM_INVISIBLE). This will allows JAWS and NVDA to show the user something is there even when they're in a virtual buffer.

joanmarie (approved):

index.html
```diff
@@ -13297,9 +13297,9 @@ <h2>Excluding Elements from the Accessibility Tree</h2>
 	  <h2>Including Elements in the Accessibility Tree</h2>
           <p>If not excluded from the accessibility tree per the rules above in <a href="#tree_exclusion">Excluding Elements in the Accessibility Tree</a>, user agents MUST provide an <a>accessible object</a> in the <a>accessibility tree</a> for <abbr title="Document Object Model">DOM</abbr> <a>elements</a> that meet any of the following criteria:</p>
```
cookiecrook, line 13298:
```suggestion
          <p>If not excluded from or marked as hidden in the accessibility tree per the rules above in <a href="#tree_exclusion">Excluding Elements in the Accessibility Tree</a>, user agents MUST provide an <a>accessible object</a> in the <a>accessibility tree</a> for <abbr title="Document Object Model">DOM</abbr> <a>elements</a> that meet any of the following criteria:</p>
```

cookiecrook: @aleventhal wrote:

> I'm ok +1'ing it if it's understood that we will have slightly different behavior depending on the platform:

Does this change help clarify that difference?

jnurthen (approved):

PR #1410, opened 2021-02-24
resolves #1394, remove misleading note about AT modifying attrs directly

### LINKED ISSUE #1394: Misleading note about AT changing aria attrs.
From https://w3c.github.io/aria/#ua_domchanges

```html
<p class="note">Many state and properties can be changed by assistive technologies through existing accessibility APIs by responding to a default action event. For example, the <sref>aria-selected</sref> state of a <rref>tab</rref> in a <rref>tabpanel</rref> can be changed by triggering the default action on the element.</p>
```

I don't think this is actually true. The AT just triggers the default action... That may convey user intent, but the AT isn't actually changing anything. The web app makes the change in response to the "intent."

We could reword or remove the note. Leaning toward remove, b/c I'm not sure it adds anything useful.

cookiecrook:
resolves #1394

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1410.html" title="Last updated on Feb 25, 2021, 7:25 PM UTC (21a1215)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1410/8c9d7b6...21a1215.html" title="Last updated on Feb 25, 2021, 7:25 PM UTC (21a1215)">Diff</a>


jnurthen (approved):

carmacleod (approved):
+1

Thanks @cookiecrook !

Issue #574, opened 2017-04-26
role switch should have additional attribute aria-valuetext


stes-acc:
The values on/off are not sufficient for role switch since often on switches are two text items that have the character of "labels" ("red" and "green" for instance) that stand for two discrete "values".

We need an additional aria-valuetext attribute (like aria-valuetext for slider) that reflects this in string format. Declaration of this attribute can add to the default on/off announcement or replaces it entirely, if declared.
Of course the text value must be updated on state changes.

Additional labelling using aria-label and/or aria-labelledby should still be possible, also aria-describedby.

You cannot use aria-labelledby for this additional value text since screen readers typically poll only for toggle change events of object type boolean, accessing changed labels is additional workload typically not done. For this they should additionally listen to control value changes if it comes to switch role. 
This requires that switch role also supports a value pattern. State changes must be mapped as changed control values with content=valuetext in respective valuechange events of the user agent.

spectranaut:
closing as duplicate of more recent issue: https://github.com/w3c/aria/issues/2085

spectranaut:
closed during issue triage in this meeting: https://www.w3.org/2024/05/30-aria-minutes.html#t08

PR #1413, opened 2021-03-02
draft addition of aria-valuetext on more things

### LINKED ISSUE #711: aria-valuetext would be useful on many more roles

jnurthen:
Very draft PR to close #711

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1413.html" title="Last updated on Mar 2, 2021, 1:30 AM UTC (808afa9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1413/8c9d7b6...808afa9.html" title="Last updated on Mar 2, 2021, 1:30 AM UTC (808afa9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2207,7 +2207,7 @@
 						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
 						<td class="role-properties">
 							<ul>
-                <li><pref>aria-activedescendant</pref></li>
+                				<li><pref>aria-activedescendant</pref></li>
 								<li><pref>aria-autocomplete</pref></li>
 								<li><pref>aria-controls</pref></li>
 								<li><pref>aria-errormessage</pref></li>
@@ -2215,6 +2215,7 @@
 								<li><sref>aria-invalid</sref></li>
 								<li><pref>aria-readonly</pref></li>
 								<li><pref>aria-required</pref></li>
+								<li><pref>aria-valuetext</pref></li>
 							</ul>
 						</td>
 					</tr>
@@ -4655,6 +4656,7 @@
 								<li><pref>aria-multiselectable</pref></li>
 								<li><pref>aria-readonly</pref></li>
 								<li><pref>aria-required</pref></li>
+								<li><pref>aria-valuetext</pref></li>
 							</ul>
 						</td>
 					</tr>
@@ -8639,6 +8641,7 @@
 			<div class="role-description">
 				<p>A type of checkbox that represents on/off values, as opposed to checked/unchecked values. See related <rref>checkbox</rref>.</p>
 				<p>The <sref>aria-checked</sref> <a>attribute</a> of a <code>switch</code> indicates whether the input is on (<code>true</code>) or off (<code>false</code>). The <code>mixed</code> value is invalid, and user agents MUST treat a <code>mixed</code> value as equivalent to <code>false</code> for this role.</p>
+				<p>The <pref>aria-valuetext</pref> <a>attribute</a> is used to specify the value to be used when the switch's true and false values are not on and off. If specified, authors MUST change the <pref>aria-valuetext</pref> <a>attribute</a> when the switch's aria-checked is modified.</li>
 				<p class="note">A <code>switch</code> provides approximately the same functionality as a <code>checkbox</code> and toggle <code>button</code>, but makes it possible for assistive technologies to present the widget in a fashion consistent with its on-screen appearance.</p>
 			</div>
 			<table class="role-features">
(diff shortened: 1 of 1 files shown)


stes-acc:
related: [role switch should have additional attribute aria-valuetext · Issue #574](https://github.com/w3c/aria/issues/574)

index.html
```diff
@@ -4655,6 +4656,7 @@ <h2>Definition of Roles</h2>
 								<li><pref>aria-multiselectable</pref></li>
 								<li><pref>aria-readonly</pref></li>
 								<li><pref>aria-required</pref></li>
+								<li><pref>aria-valuetext</pref></li>
```
mcking65, line 4659: This idea makes sense for the collabsible listbox where there may be an element representing a listbox that does not have any options. However, I think we should deprecate that pattern in APG. Select-only combobox completely replaces it; the combobox is simpler to code and supports aria-required. In any other listbox, the selected option represents the value.

index.html
```diff
@@ -9317,6 +9321,7 @@ <h5>Presentational Roles Conflict Resolution</h5>
 			<rdef>textbox</rdef>
 			<div class="role-description">
 				<p>A type of input that allows free-form text as its value.</p>
+				<p>If the <pref>aria-valuetext</pref> <a>attribute</a> is set it indicates the <a>value</a> of the textbox. If it is not set then the value is the child content of the textbox.</p>
```
mcking65, line 9324: I don't think this should be permitted if the textbox role is on an html input element. It needs language to the effect:

>If the textbox element is a host language element that provides a value, such as an HTML input  element, authors must not specify aria-valuetext.
>If the textbox element is a host language element that provides a value, such as an HTML input  element, user agents MuST ignore aria-valuetext if it is specified.

index.html
```diff
@@ -13286,11 +13292,22 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+					<li>Authors SHOULD only set the <pref>aria-valuetext</pref> attribute when the rendered value cannot be meaningfully represented as a number. For example, a slider may have rendered values of <code>small</code>, <code>medium</code>, and <code>large</code>. In this case, the values of <pref>aria-valuenow</pref> could range from 1 through 3, which indicate the position of each value in the value space, but the <pref>aria-valuetext</pref> would be one of the strings: <code>small</code>, <code>medium</code>, or <code>large</code>. If the <pref>aria-valuetext</pref> attribute is absent, the <a>assistive technologies</a> will rely solely on the <pref>aria-valuenow</pref> attribute for the current value.</li>
+					<li>If <pref>aria-valuetext</pref> is specified, assistive technologies SHOULD render that value instead of the value of <pref>aria-valuenow</pref>.</li>
+				</ul>
+				<p>When this property is used on <rref>textbox</rref>, <rref>combobox</rref> or <rref>listbox</rref>:</p>
```
mcking65, line 13302: Per above, I don't think we should include listbox. Also, not sure if it should be supported on textbox unless it is readonly and only if not backed by a native element that does not have its own value.

mcking65 (commented):
I think this is a great start. I think the application to switch is most interesting. It is good for combobox as long as it is not editable, i.e., only on select-only comboboxes. I am fairly certain that we should not add support for listbox. And, I think textbox support needs additional restrictions. Details below.

carmacleod (commented):
I'm good with switch.

I wonder if prose similar to the UA instructions we added for combobox (pasted below) would go a long way towards helping authors understand where the value for textbox and searchbox usually comes from. UAs already do this for `div role="textbox"` or `div role="searchbox"` anyhow, so we're not saying anything new.

> User agents MUST expose the value of elements with role combobox to assistive technologies. The value of a combobox is represented by one of the following:
>
> - If the combobox element is a host language element that provides a value, such as an HTML input element, the value of the combobox is the value of that element.
> - Otherwise, the value of the combobox is represented by its descendant elements and can be determined using the same method used to compute the name of a button from its descendant content.

We could maybe add another bullet point to the effect of aria-valuetext only being useful if it's (a) not a host language element that already provides a value, and (b) if the value doesn't match the descendant content.

Regarding listbox, I agree with Matt that it probably shouldn't support aria-valuetext. The selected option(s) represent the listbox value, and ATs look for that. We could maybe make this explicit in the prose. Note that UAs don't set the actual value of the listbox. It's implied by the selected option(s).

Regarding CSS content, UAs should always be including CSS text content as specified in the name calculation in https://w3c.github.io/accname/#step2F. So, "the value doesn't match the CSS content" shouldn't be a justification for adding aria-valuetext support.

pkra:
Is 1.3 still realistic for this one?

PR #1414, opened 2021-03-02
Implicit values for required properties section needs revising

### LINKED ISSUE #966: Implicit values for required properties section needs revising
Section 8.6 states
http://w3c.github.io/aria/#state_property_processing

> However, required states and properties that are absent are an author error. Missing required states and properties are treated as if they were present and have an implicit neutral value that is not necessarily their default value. For example, the default value of aria-expanded is undefined, meaning neither expandable nor collapsible. But that does not apply to the case of a combobox. In this case, aria-expanded is needed to convey the expandable/collapsible nature of the combobox. Thus, the implicit value of aria-expanded for the combobox role is false, meaning expandable (and currently collapsed). The characteristics table associated with each WAI-ARIA role has an "Implicit Value for Role" entry that specifies the value of a state or property to use in the context of that role when the state or property is missing. 

This needs revising due to the fact required states and properties no longer have implicit values

jnurthen:
Fixes #966

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1414.html" title="Last updated on Mar 2, 2021, 11:51 PM UTC (142a541)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1414/8c9d7b6...142a541.html" title="Last updated on Mar 2, 2021, 11:51 PM UTC (142a541)">Diff</a>


carmacleod (approved):
Nice! +1
Thanks, @jnurthen !

pkra (approved):

scottaohara (approved):

PR #1426, opened 2021-03-15
Move abstract roles out of main spec

### LINKED ISSUE #725: Move abstract roles out of main spec
Migrated from [ACTION-1624](https://www.w3.org/WAI/ARIA/track/actions/1624). We need the abstract roles in the taxonomy and scripted inheritence, but they just confuse authors.

jnurthen:
Fixes #725
This would need to be split and mostly moved to aria-common. Here as a Proof of concept at the moment

[Preview Link](https://raw.githack.com/w3c/aria/removeAbstract/index.html)

### DIFF
--- a/common/terms.html
+++ b/common/terms.html
@@ -63,6 +63,12 @@
     <dd>
       <p>A set of instance <a class="termref" href="#dfn-object">objects</a> that share similar characteristics.</p>
     </dd>
+    <dt><dfn>Composite</dfn></dt>
+    <dd>
+      <p>A <a>widget</a> that may contain navigable descendants or <a>owned</a> children.</p>
+      <p>Authors SHOULD ensure that a composite widget exists as a single navigation stop within the larger navigation system of the web page. Once the composite widget has focus, authors SHOULD provide a separate navigation mechanism for users to navigate to <a>elements</a> that are descendants or owned children of the composite element.</p>
+      <p><code>composite</code> is an <a>abstract role</a> used for the ontology.</p>
+    </dd>    
     <dt><dfn data-lt="deprecated|deprecate|deprecation">Deprecated</dfn></dt>
     <dd>
       <p>A deprecated <a class="termref" href="#dfn-role">role</a>, <a class="termref" href="#dfn-state">state</a>, or <a class="termref" href="#dfn-property">property</a> is one which has been outdated by newer constructs or changed circumstances, and which may be removed in future versions of the <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> specification. <a class="termref" data-lt="user agent">User agents</a> are encouraged to continue to support items identified as deprecated for backward compatibility. For more information, see <a href="#deprecated" class="specref">Deprecated Requirements</a> in the Conformance section.</p>
@@ -96,6 +102,11 @@
     <dt><dfn data-lt="informative|non-normative">Informative</dfn></dt>
     <dd>
       <p>Content provided for information purposes and not required for conformance. Content required for conformance is referred to as <a>normative</a>.</p>
+    </dd>
+    <dt><dfn>Input</dfn></dt>
+    <dd>
+      <p>A generic type of <a>widget</a> that allows user input.</p>
+      <p class="note"><code>input</code> is an <a>abstract role</a> used for the ontology.</p>
     </dd>
     <dt><dfn>Keyboard Accessible</dfn></dt>
     <dd>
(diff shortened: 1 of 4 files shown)


carmacleod:
Interestingly, this isn't as confusing as I initially thought it would be. :)
(Although it does feel a bit weird not to be able to walk the hierarchy by default).

A couple of notes:
1. The "Show/Hide Abstract Roles" button is cool, but I'll never find it again - can we incorporate a URL query string as well as the button, like maybe `https://raw.githack.com/w3c/aria/removeAbstract/index.html?showAbstract=true` (or false)?
2. What do you want to do about the `<rref>abstractrole</rref>` and `<a>abstractrole</a>` that are scattered throughout the document? (Particularly as there's a show/hide button that fixes/breaks the links).
3. The [definition of inherited attributes](https://raw.githack.com/w3c/aria/removeAbstract/index.html#inheritedattributes) mentions superclass roles, but of course, there are no superclasses listed in the characteristics tables by default any more, so maybe this definition needs a Note explaining that.
4. A Note in the [Required Owned Elements definition](https://raw.githack.com/w3c/aria/removeAbstract/index.html#mustContain) talks about superclass and subclass roles - might need an explanation.
5. The [Required States and Properties definition](https://raw.githack.com/w3c/aria/removeAbstract/index.html#requiredState) talks about subclass roles.
6. The [Supported States and Properties definition](https://raw.githack.com/w3c/aria/removeAbstract/index.html#supportedState) contains the following strange sentence... was "child" intended to be "subclass"?:
   > States and properties specifically applicable to the role and **_child_** roles.

jnurthen:
> The "Show/Hide Abstract Roles" button is cool, but I'll never find it again - can we incorporate a URL query string as well as the button, like maybe https://raw.githack.com/w3c/aria/removeAbstract/index.html?showAbstract=true (or false)?

Probably

> What do you want to do about the <rref>abstractrole</rref> and <a>abstractrole</a> that are scattered throughout the document? (Particularly as there's a show/hide button that fixes/breaks the links).

Can you point to an example. I changed all the `<a href="#someID">abstract role</a>` fragments to `<a>abstract role</a>` and then created the new abstract role defintion so these things could all still point somewhere.

>The definition of inherited attributes mentions superclass roles, but of course, there are no superclasses listed in the characteristics tables by default any more, so maybe this definition needs a Note explaining that.

We should probably just combine inherited and supported - but not sure how to do that easily

> A Note in the Required Owned Elements definition talks about superclass and subclass roles - might need an explanation.

Yes - I noticed this once I submitted the Draft PR - but didn't get round to suggesting something

>The Required States and Properties definition talks about subclass roles.

as previous

> The Supported States and Properties definition contains the following strange sentence... was "child" intended to be "subclass"?: States and properties specifically applicable to the role and child roles.

I think so :)

carmacleod:
> Can you point to an example.

`<rref>composite</rref>`
`<rref>input</rref>`
`<rref>landmark</rref>`
`<rref>range</rref>`
`<rref>section</rref>`
`<rref>select</rref>`
`<rref>structure</rref>`
`<rref>widget</rref>`

usually in role-description or state-description or property-description.

Was wrong about `<a>abstractrole</a>` - I think you got all of those.

common/terms.html
```diff
@@ -63,6 +63,12 @@
     <dd>
       <p>A set of instance <a class="termref" href="#dfn-object">objects</a> that share similar characteristics.</p>
     </dd>
+    <dt><dfn>Composite</dfn></dt>
+    <dd>
+      <p>A <a>widget</a> that may contain navigable descendants or <a>owned</a> children.</p>
+      <p>Authors SHOULD ensure that a composite widget exists as a single navigation stop within the larger navigation system of the web page. Once the composite widget has focus, authors SHOULD provide a separate navigation mechanism for users to navigate to <a>elements</a> that are descendants or owned children of the composite element.</p>
```
carmacleod, line 69: (For some reason, github won't let me create a "suggestion". Draft, maybe?).

I realize that these are the words that were currently in the spec, but they're weirdly confusing. Is there really some other "larger navigation system of the web page" out there (not including AT navigation, because that's not really relevant to the description of a composite)?
Shouldn't we just say "tab order" so as not to confuse authors?

`<p>Authors SHOULD ensure that a composite widget exists as a single tab stop within the tab order of the web page. Once the composite widget has focus, authors SHOULD provide a separate navigation mechanism, such as arrow keys, for users to navigate to <a>elements</a> that are descendants or owned children of the composite element.</p>`

carmacleod:
Missing the following in the definitions:
- command
- roletype
- sectionhead
- window

Also, `landmark` and `widget` need to have their definition massaged so that they define both the concept and the abstract role.
For example, they're both missing the note about "... is an abstract role used for the ontology."

`Widget` may need to mention the bit about AT browse/application mode.

`Landmark` even has some normative text in it, which doesn't belong in a definition as-is, but could say something like "ATs provide a mechanism to to quickly navigate to landmark regions. UAs could provide a mechanism to to quickly navigate to landmark regions."
Fortunately the normative text is fully reflected in the landmark subclasses in PR https://github.com/w3c/aria/pull/1315. Please merge that PR - it's harmless! I guarantee it. :)

jnurthen:
> Missing the following in the definitions:
> 
> * command
> * roletype
> * sectionhead
> * window
> 
I didn't see those referenced anywhere so didn't include definitions for them

jnurthen:
> Also, `landmark` and `widget` need to have their definition massaged so that they define both the concept and the abstract role.
> For example, they're both missing the note about "... is an abstract role used for the ontology."
> 
> `Widget` may need to mention the bit about AT browse/application mode.
> 
> `Landmark` even has some normative text in it, which doesn't belong in a definition as-is, but could say something like "ATs provide a mechanism to to quickly navigate to landmark regions. UAs could provide a mechanism to to quickly navigate to landmark regions."
> Fortunately the normative text is fully reflected in the landmark subclasses in PR #1315. Please merge that PR - it's harmless! I guarantee it. :)

All of them need changing. Where a definition already existed I used that. Where one was missing I copied it verbatim from the role. IMO all of them need to be more like the existing ones but I was lazy and wanted to test the concept of moving these to definitons rather than changing too much and spending too much time on something that might not be practical.

PR #1445, opened 2021-04-01
add local abstract role usage warnings. fixes #1428

### LINKED ISSUE #1428: make it clearer that authors should not use abstract roles
There are a dozen abstract roles in ARIA: https://w3c.github.io/aria/#abstract_roles

At the above overview link, it clearly states that:
> Abstract roles are used for the ontology. Authors MUST NOT use abstract roles in content.

However, in each abstract role definition, it's not quite as clear that authors aren't allowed to use these roles, and some authors do try to use some of the roles, like input, range, section, select, because they "seem like real roles". :)

There is a green note box in each abstract role definition that states, for example in [composite](https://w3c.github.io/aria/#composite):
> `composite` is an abstract role used for the ontology.

This needs to also say (e.g. for composite):
> Authors MUST NOT use `composite` role in content.

However, since normative statements like "MUST NOT" are not allowed in Note boxes, we also need to remove the "Note box" styling (`class="note"`).

Also, consider moving this text nearer to the beginning of the definition, perhaps right after the first sentence. (In some abstract roles there's only one sentence, so the text wouldn't need to move).

Thanks for taking this on, @jorycunningham!

jorycunningham:
Fixes #1428 

- Adding in "Authors MUST NOT use <code>{role name}</code> role in content." language to each role description.
- Remove class="note" from the paragraphs which contain the above statement
- Moving this statement closer to the beginning of the description for longer descriptions.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 8, 2021, 7:55 PM UTC)_.

<details>
<summary>More</summary>

:rotating_light: [HTML Diff Service](http://services.w3.org/htmldiff) - The HTML Diff Service is used to create HTML diffs of the spec changes suggested in a pull request.

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fw3c%2Faria%2Fpull%2F1445%2F16dfd46.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fw3c%2Faria%2Fpull%2F1445.html)

```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>500 Internal Server Error</title>
</head><body>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error or
misconfiguration and was unable to complete
your request.</p>
<p>Please contact the server administrator at 
 sysreq@w3.org to inform them of the time this error occurred,
 and the actions you performed just before this error.</p>
<p>More information about this error may be available
in the server error log.</p>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231445.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2449,7 +2449,7 @@
 			<rdef>command</rdef>
 			<div class="role-description">
 				<p>A form of widget that performs an action but does not receive input data.</p>
-				<p class="note"><code>command</code> is an <a href="#isAbstract">abstract role</a> used for the ontology.</p>
+				<p><code>command</code> is an <a href="#isAbstract">abstract role</a> used for the ontology. Authors MUST NOT use <code>commmand</code> role in content.</p>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
@@ -2710,7 +2710,7 @@
 			<div class="role-description">
 				<p>A <a>widget</a> that may contain navigable descendants or <a>owned</a> children.</p>
 				<p>Authors SHOULD ensure that a composite widget exists as a single navigation stop within the larger navigation system of the web page. Once the composite widget has focus, authors SHOULD provide a separate navigation mechanism for users to navigate to <a>elements</a> that are descendants or owned children of the composite element.</p>
-				<p class="note"><code>composite</code> is an <a href="#isAbstract">abstract role</a> used for the ontology.</p>
+				<p><code>composite</code> is an <a href="#isAbstract">abstract role</a> used for the ontology. Authors MUST NOT use <code>composite</code> role in content.</p>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4430,11 +4430,11 @@ <h2>Definition of Roles</h2>
 			<rdef>landmark</rdef>
 			<div class="role-description">
 				<p>A perceivable <rref>section</rref> containing content that is relevant to a specific, author-specified  purpose and sufficiently important that users will likely want to be able to navigate to the section easily and to have it listed in a summary of the page. Such a page summary could be generated dynamically by a user agent or assistive technology.</p>
+				<p><code>landmark</code> is an abstract role used for the ontology. Authors should not use this role in content. Authors MUST NOT use <code>landmark</code> role in content.</p>
```
carmacleod, line 4433:
```suggestion
				<p><code>landmark</code> is an abstract role used for the ontology. Authors MUST NOT use <code>landmark</code> role in content.</p>
```

carmacleod (commented):
Very nice! Just needs the old sentence in landmark role removed, and it's good to go!
Thanks, @jorycunningham !

carmacleod (approved):
Thanks, @jorycunningham !
Looks good !

carmacleod:
(By the way, don't worry about PR Preview failing... it fails a lot).

scottaohara (approved):

jnurthen (approved):

cookiecrook (approved):

PR #1446, opened 2021-04-01
reorder meter in role listings

### LINKED ISSUE #1441: Correct alphabetical ordering of roles
Presently `meter` is listed before `menu` in the listing of roles / its location in the spec.  

`meter` should be moved so that it comes before `navigation`.

jorycunningham:
Fixes #1441 

- Update the order of roles in aria.js so that meter falls before navigation
- Move the meter role entry in index.html

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1446.html" title="Last updated on Apr 1, 2021, 7:26 PM UTC (5d3c112)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1446/16dfd46...5d3c112.html" title="Last updated on Apr 1, 2021, 7:26 PM UTC (5d3c112)">Diff</a>

### DIFF
--- a/common/script/roleInfo.js
+++ b/common/script/roleInfo.js
@@ -4241,22 +4241,6 @@ var roleInfo = {
     ],
     "localprops": []
   },
-  "meter": {
-    "name": "meter",
-    "fragID": "meter",
-    "parentRoles": [
-      "range"
-    ],
-    "localprops": [
-      {
-        "is": "property",
-        "name": "aria-valuenow",
-        "required": true,
-        "disallowed": false,
-        "deprecated": false
-      }
-    ]
-  },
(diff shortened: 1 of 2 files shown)


scottaohara (approved):

carmacleod (approved):
Nice!
+1

Thanks, @jorycunningham !

jnurthen:
I missed that it is not necessary to change aria.js. This file gets generated periodically and will get generated from the spec in the correct place. In general we don't edit these in the aria repository - but I merged this before noticing. 
There should be no impact though as it will just get overwritten when aria-common gets updated next.

Issue #1300, opened 2020-07-16
aria-busy language update in Required Owned Element section


jnurthen:
Based on https://github.com/w3c/aria/pull/1224#discussion_r426795713

jnurthen:
@smhigley we think this might be taken care of by your other work on Required owned elements

PR #1454, opened 2021-04-13
Update required owned elements and required context role

### LINKED ISSUE #1033: Clarify "required owned element"

### LINKED ISSUE #1440: Secondary actions on items in composite widget roles
I've been repeatedly running into a difficulty between a relatively common app UI pattern and the required owned elements part of the ARIA spec. The problem is specifically with secondary actions with composite interactive widget roles. I'm using "secondary action" to refer to cases where the child of the composite widget role (e.g. tab, treeitem, option, etc.) performs a primary action that makes sense for its role, but then also has secondary actions like edit/delete/etc., exposed through an associated button.

Here are some examples of UI that have this pattern:

1. Close buttons in tabs (probably the most familiar example):
  the VS Code tab list of open files, the selected tab has a visible close button
2. A menu of recently opened files, where each file has share, favorite, and more actions buttons:
  A single item from within a menu or other interactive list; the item has an icon, a formatted name (Document125) and file location, then at the right there are three icon buttons for share, star, and a kebab menu
3. A tree of participants in a meeting, where each participant has mute and more actions buttons:
  A part of the Teams meeting UI, showing only the first meeting participant under the tree section titled "In this meeting". The participant tree item shows a small avatar, a name, and to the right are a mute button and kebab menu
4. A tree of emails, where each email has flag, complete, and more actions buttons:
  A screenshot of a series of notifications in two categories, today and older, similar to Outlook emails. Each notification has an avatar, a formatted title like "Theo assigned you a task", document name, and time, and three buttons in the top right corner for flag, a check mark, and a kebab menu.

## Context menu approach

A solution we've used in the past was to make the visual buttons presentation-only, and duplicate all secondary actions in a context menu. The problem is the context menu wasn't sufficiently discoverable in practice, and sighted users who relied on keyboard navigation (either with or without a screen reader) expected to be able to directly reach the visible buttons.

If a context menu alone were a good approach, this would be covered by #762, but the discoverability problem is enough to make me think there needs to be a pattern for surfacing the secondary action buttons directly.

## Nested vs. sibling options

Exposing the buttons directly runs into two spec + support problems:
1. Putting the secondary action buttons inside the tab/option/treeitem/etc. runs afoul of "Children: presentational" in the spec, and has [known support problems](https://github.com/dequelabs/axe-core/issues/601).
2. Putting the secondary actions adjacent to the tab/option/treeitem/etc may or may not violate the "Required Owned Elements" part of the spec, depending on how #1033 falls out.

I did some testing of practical support for nested vs. sibling buttons within composite widgets, with both the working test cases and results tables on [this test page](https://a11y-screenreader-demos.netlify.app/studies/secondary-actions/). It seems to pretty clearly come down on the side of sibling buttons having better support.

Aside from the test results, I don't personally have much of an opinion on the specific approach used, but I think we have a strong need for explicitly defining some sort of pattern defined for these cases.

### LINKED ISSUE #1486: Allowed children of role=tablist includes any role including tabpanel?
The frequently observed pattern for `tablist`, `tab` and `tabpanel` is for a `tablist` to only contain children with `role=tab`:

```
<div role=tablist><span role=tab>poot 1</span> <span role=tab>poot 2</span></div>
```
but there appears to be nothing (I can find) to prohibit a `tablist` being a container for both `tab`s and `tabpanel`s

```
<div role=tablist>
<span role=tab>poot 1</span>
<div role=tabpanel>...</div> 
<span role=tab>poot 2</span>
<div role=tabpanel>...</div> 
</div>

```

Is it the case that `tablist` with both `tab`s and `tabpanel`s as children is OK? 
Am asking as I found an example of a widget using this latter pattern.

smhigley:
Resolves #1033

Updates "Required Owned Elements" to "Allowed Child Elements", and "Required Context Role" to "Required Parent Role", with added definitions for parent, child, controlled element, and controlling element.

I did some support tests for having intermediate elements in composite widgets, and secondary actions in composite widgets, here: https://cdpn.io/smhigley/debug/vYgLZgK

This is still a POC, so I didn't update all the references to required owned elements/context role. Hopefully that makes it a bit easier to see the relevant diffs for discussion.

Should also resolve #1486 

# PR merge check list

* [x] Related APG Issue/PR: N/A (expandable listbox example already removed)
* [X] validator tests: https://github.com/w3c/aria/issues/1917
* [X] WPT tests: [Aria issue to discuss](https://github.com/w3c/aria/issues/1870)
* [X] Browser implementations (link to issue or when done, link to commit):
   * WebKit: n/a
   * Gecko: n/a
   * Blink: n/a
* [X] Does this need AT implementations? no

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1454.html" title="Last updated on Nov 21, 2021, 5:06 AM UTC (c2e5165)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1454/627b3a6...c2e5165.html" title="Last updated on Nov 21, 2021, 5:06 AM UTC (c2e5165)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1454.html" title="Last updated on May 1, 2023, 9:38 PM UTC (1be7640)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1454/6e45edd...1be7640.html" title="Last updated on May 1, 2023, 9:38 PM UTC (1be7640)">Diff</a>


index.html
```diff
@@ -654,18 +654,17 @@ <h3>Prohibited States and Properties</h3>
...
-			<p class="note">An element with a <a href="#subclassroles">subclass role</a> of the 'required owned element' does not fulfill this requirement. For example, the <rref>listbox</rref> role requires ownership of an element using the <rref>option</rref> or <rref>group</rref> role. Although the <rref>group</rref> role is the superclass of <rref>row</rref>, adding an <a>owned</a> element with a role of <rref>row</rref> will not fulfill the requirement that <rref>listbox</rref> owns an <rref>option</rref> or a <rref>group</rref>.</p>
+			<h3>Allowed Child Elements</h3>
+			<p>Any <a>element</a> that is a direct <a>child</a> of the element with this <a>role</a>. For example, an element with the role <rref>list</rref> may own elements with the role <rref>listitem</rref>, but not elements with the role <rref>option</rref>.</p>
+			<p>To determine whether an element is the direct <a>child</a> of a role, any elements with the role <rref>generic</rref> or <rref>presentation</rref> are ignored.</p>
```
scottaohara, line 659: do we need to explicitly tell user agents what to do here, re: to ignore generic elements?  i ask since that's presently something that isn't consistently happening

index.html
```diff
@@ -654,18 +654,17 @@ <h3>Prohibited States and Properties</h3>
...
+			<p>Non-child descendents of the element are not included in <em>allowed child elements</em>. For example, an <code>image</code> is not an allowed child of a <code>list</code>, but it may exist within a <code>listitem</code> in a <code>list</code>.</p>
```
scottaohara, line 660: here's a bit of a rewording idea which also takes care of the lowercase 'may'
```suggestion
			<p>Descendants which are not direct children of an element ancestor are not considered <em>allowed child elements</em>. For example, an <code>image</code> is not an allowed child of a <code>list</code>, but it is a valid descendant so long as it is also a descendant of the <code>list</code> element's allowed child <code>listitem</code>.</p>
```

index.html
```diff
@@ -654,18 +654,17 @@ <h3>Prohibited States and Properties</h3>
...
+			<p class="note">A role that has 'allowed child elements' does not imply the reverse relationship. Elements with roles in this list do not always have to be found within elements of the given role. See <a href="#scope">allowed parent roles</a> for requirements about the context where elements of a given role will be contained.</p>
+			<p class="note">An element with a <a href="#subclassroles">subclass role</a> of the 'allowed child element' does not fulfill this requirement. For example, the <rref>listbox</rref> role allows a child element using the <rref>option</rref> or <rref>group</rref> role. Although the <rref>group</rref> role is the superclass of <rref>row</rref>, adding a child element with a role of <rref>row</rref> will not fulfill the requirement that <rref>listbox</rref> children must be an <rref>option</rref> or a <rref>group</rref>.</p>
```
scottaohara, line 662: lowercase must => "children must be"

scottaohara (commented):
since this is still a draft, just a few comments/suggestions.  hope they're helpful

common/terms.html
```diff
@@ -53,10 +53,22 @@
     <dd>
       <p>In this specification, attribute is used as it is in markup languages. Attributes are structural features added to <a class="termref" href="#dfn-element">elements</a> to provide information about the <a class="termref" href="#dfn-state">states</a> and <a class="termref" href="#dfn-property">properties</a> of the <a class="termref" href="#dfn-object">object</a> represented by the element.</p>
     </dd>
+    <dt><dfn data-lt="child element|child|children|child elements">Child Element</dfn></dt>
+    <dd>
+      <p>An 'child element' is a direct <abbr title="Document Object Model">DOM</abbr> child of the <a>element</a>, an descendant of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, or a child specified via <pref>aria-owns</pref> on the element.</p>
```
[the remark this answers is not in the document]
aleventhal: I find it a little confusing to use "element" here, since a child might be text, which is not an element.
And I find "child element" to be confusing because it contradicts what a child element means in the DOM, given that aria-owns children are not child elements.

In Chrome, we would avoid saying "element" unless something is truly an element, and would avoid saying child element if something is not actually a DOM child element. We might say child or child object. Maybe we want to say owned child here?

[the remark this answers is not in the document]
smhigley: I'll updated with "owned child", and I'll also change "Allowed Child Elements" to "Allowed Child Roles" both to match the "Required Context Role" and to avoid element 👍

index.html
smhigley: might as well, since it's included in the parent section 👍

smhigley: @scottaohara how do you feel about the updated wording?

index.html
smhigley: reworded to avoid "must"

smhigley:
I updated the PR based on the comments, updated all references to "required owned elements"/"required context role", and moved the `aria-controls` allowed child stuff to a separate branch. It should be ready for a full review now.

common/terms.html
```diff
@@ -144,6 +148,10 @@
     <dd>
       <p>An 'owning element' is any <abbr title="Document Object Model">DOM</abbr> ancestor of the <a>element</a>, or any element with an <pref>aria-owns</pref> attribute which references the ID of the   element. </p>
     </dd>
+    <dt><dfn data-lt="parent element|parent">Parent Element</dfn></dt>
+    <dd>
+      <p>An 'parent element' is a direct <abbr title="Document Object Model">DOM</abbr> parent of the <a>element</a>, an ancestor of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, or any element specified as a child via <pref>aria-owns</pref>.</p>
```
jnurthen, line 153:
```suggestion
      <p>A 'parent element' is a direct <abbr title="Document Object Model">DOM</abbr> parent of the <a>element</a>, an ancestor of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, or any element specified as a child via <pref>aria-owns</pref>.</p>
```

jnurthen (approved):

index.html
```diff
@@ -10909,7 +10909,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p>If changes to a rendered <rref>widget</rref> would create a state where the <rref>widget</rref> is missing <a href="#mustContain">required owned elements</a> during script execution, authors MUST set <code>aria-busy</code> to <code>true</code> on the <rref>widget</rref> during the update process. For example, if a rendered tree grid required a set of simultaneous updates to multiple discontiguous branches, an alternative to replacing the complete tree element with a single update would be to mark the tree busy while each of the branches are modified.</p>
+				<p>If changes to a rendered <rref>widget</rref> would create a state where the <rref>widget</rref> is modifying <a href="#mustContain">allowed child roles</a> during script execution, authors MAY set <code>aria-busy</code> to <code>true</code> on the <rref>widget</rref> during the update process. For example, if a rendered tree grid required a set of simultaneous updates to multiple discontiguous branches, an alternative to replacing the complete tree element with a single update would be to mark the tree busy while each of the branches are modified.</p>
```
jnurthen, line 10912: Can we just remove this. Does setting aria-busy in this state actually do anything?

jnurthen:
Please also move the new terms to index.html. The content of terms.html has been moved there. You'll need to pull the changes from main though to pick this up.

index.html
smhigley: I'm happy to remove it, I just didn't want to muddy this PR with other changes if they'd be controversial 😅

smhigley:
All the previous comments should be addressed. @scottaohara / @aleventhal, how does it look now?

smhigley:
also pinging @jcsteh because for some reason github isn't letting me add him to the reviewers ¯\_(ツ)_/¯

index.html
```diff
@@ -663,18 +671,18 @@ <h3>Prohibited States and Properties</h3>
...
-			<p class="note">An element with a <a href="#subclassroles">subclass role</a> of the 'required owned element' does not fulfill this requirement. For example, the <rref>listbox</rref> role requires ownership of an element using the <rref>option</rref> or <rref>group</rref> role. Although the <rref>group</rref> role is the superclass of <rref>row</rref>, adding an <a>owned</a> element with a role of <rref>row</rref> will not fulfill the requirement that <rref>listbox</rref> owns an <rref>option</rref> or a <rref>group</rref>.</p>
+			<h3>Allowed Child Roles</h3>
+			<p>Any <a>element</a> that is a direct <a>child</a> of the element with this <a>role</a>. For example, an element with the role <rref>list</rref> may own elements with the role <rref>listitem</rref>, but not elements with the role <rref>option</rref>.</p>
+			<p>To determine whether an element is the direct <a>child</a> of a role, <a>user agents</a> MUST ignore any elements with the role <rref>generic</rref> or <rref>presentation</rref>.</p>
```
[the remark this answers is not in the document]
smhigley: I updated to "intervening elements" -- thanks for the suggestion, I like that better :).

And yup! This is guidance on allowed role structure, and isn't prescriptive of API mappings.

index.html
```diff
@@ -396,6 +400,10 @@ <h1>Important Terms</h1>
 			<dd>
 			  <p>An 'owning element' is any <abbr title="Document Object Model">DOM</abbr> ancestor of the <a>element</a>, or any element with an <pref>aria-owns</pref> attribute which references the ID of the   element. </p>
 			</dd>
+			<dt><dfn data-export="" data-lt="parent element|parent">Parent Element</dfn></dt>
+			<dd>
+				<p>An 'parent element' is a direct <abbr title="Document Object Model">DOM</abbr> parent of the <a>element</a>, an ancestor of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, or any element with a child specified via <pref>aria-owns</pref>.</p>
```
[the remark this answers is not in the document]
smhigley: updated! (both the typo, and insertion of "owning")

smhigley:
Just want to ping @jcsteh, @scottaohara, and @aleventhal on this to see if the updates have addressed past comments as desired :)

scottaohara (commented):
wrote these review comments some time ago... sorry for the delay.

index.html
```diff
@@ -663,18 +671,18 @@ <h3>Prohibited States and Properties</h3>
...
-			<p class="note">An element with a <a href="#subclassroles">subclass role</a> of the 'required owned element' does not fulfill this requirement. For example, the <rref>listbox</rref> role requires ownership of an element using the <rref>option</rref> or <rref>group</rref> role. Although the <rref>group</rref> role is the superclass of <rref>row</rref>, adding an <a>owned</a> element with a role of <rref>row</rref> will not fulfill the requirement that <rref>listbox</rref> owns an <rref>option</rref> or a <rref>group</rref>.</p>
+			<h3>Allowed Child Roles</h3>
+			<p>Any <a>element</a> that is a direct <a>child</a> of the element with this <a>role</a>. For example, an element with the role <rref>list</rref> may own elements with the role <rref>listitem</rref>, but not elements with the role <rref>option</rref>.</p>
```
jnurthen, line 675: This reads really strangely (it did before too).
Could we start this sentence with something like "A list of roles which may be a direct child..."?

index.html
```diff
@@ -663,18 +671,18 @@ <h3>Prohibited States and Properties</h3>
...
-			<p class="note">A role that has 'required context role' does not imply the reverse relationship. While an element with the given role needs to appear within an element of the listed role(s) in order to be meaningful, elements of the listed roles do not always need descendant elements of the given role in order to be meaningful. See <a href="#mustContain">required owned elements</a> for requirements about elements that require presence of a given descendant to be processed properly.</p>
+			<h3>Required Parent Role</h3>
+			<p>The required <a>parent</a> role defines the parent container where this <a>role</a> is allowed. If a role has a required parent, authors MUST ensure that an element with the role is a direct <a>child</a> of an element with the required parent role. For example, an element with role <code>listitem</code> is only meaningful when it is a child of an element with role <code>list</code>.</p>
+			<p>To determine whether an element has a parent with the required role, <a>user agents</a> MUST ignore any elements with the role <rref>generic</rref> or <rref>presentation</rref>.</p>
```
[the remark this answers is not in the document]
smhigley: I thought we discussed this at some point and decided not to be prescriptive about it? I don't remember exactly, though.

index.html
```diff
@@ -396,6 +400,10 @@ <h1>Important Terms</h1>
 			<dd>
 			  <p>An 'owning element' is any <abbr title="Document Object Model">DOM</abbr> ancestor of the <a>element</a>, or any element with an <pref>aria-owns</pref> attribute which references the ID of the   element. </p>
 			</dd>
+			<dt><dfn data-export="" data-lt="parent element|parent|owning parent">Parent Element</dfn></dt>
+			<dd>
+				<p>An owning parent element is a direct <abbr title="Document Object Model">DOM</abbr> parent of the <a>element</a>, an ancestor of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, or any element with a child specified via <pref>aria-owns</pref>.</p>
```
scottaohara, line 405: the `dt` is for "parent element", but this line references "owning parent element"

index.html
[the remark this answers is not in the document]
scottaohara: not sure?  i don't recall that discussion.   but to quickly illustrate my point:

```
<div role=list>
  <div tabindex=-1>
    <div role=listitem>1</div>
    <div role=listitem>2</div></div>
</div>
```

because there is a `tabindex` on the `div`, chromium is presently calculating a name for the `div` from its children.  With VoiceOver, this list is now exposed as a "list of 1 item" and its direct child is a "group" with the name "1 2".

common/terms.html
```diff
@@ -136,6 +136,10 @@
     <dd>
       <p>Usable by users in ways they can control. References in this document relate to <cite><a href="https://www.w3.org/TR/WCAG21/#operable"><abbr title="Web Content Accessibility Guidelines">WCAG</abbr> 2.1 Principle 2: Content must be operable</a></cite> [[WCAG21]]. See <a>Keyboard Accessible</a>.</p>
     </dd>
+    <dt><dfn data-lt="owned child|child element|child|children|child elements">Owned Child</dfn></dt>
+    <dd>
+      <p>An 'owned child' is a direct <abbr title="Document Object Model">DOM</abbr> child of the <a>element</a>, a descendant of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, or a child specified via an <pref>aria-owns</pref> relationship to the element.</p>
```
pkra, line 141: If an intervening generic aria-owns something, is it an owned child?

common/terms.html
```diff
@@ -144,6 +148,10 @@
     <dd>
       <p>An 'owning element' is any <abbr title="Document Object Model">DOM</abbr> ancestor of the <a>element</a>, or any element with an <pref>aria-owns</pref> attribute which references the ID of the   element. </p>
     </dd>
+    <dt><dfn data-lt="parent element|parent">Parent Element</dfn></dt>
+    <dd>
+      <p>A 'parent element' is a direct <abbr title="Document Object Model">DOM</abbr> parent of the <a>element</a>, an ancestor of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, or any element with a child specified via <pref>aria-owns</pref>.</p>
```
pkra, line 153: How about  "An 'owning parent'"? (To be more consistent with owned child, owned element, owning element)

As with 'owned child', do intermediate generics with aria-owns factor in?

(Also, one of these two could also refer to the other.)

index.html
```diff
@@ -663,18 +671,18 @@ <h3>Prohibited States and Properties</h3>
...
-			<p class="note">An element with a <a href="#subclassroles">subclass role</a> of the 'required owned element' does not fulfill this requirement. For example, the <rref>listbox</rref> role requires ownership of an element using the <rref>option</rref> or <rref>group</rref> role. Although the <rref>group</rref> role is the superclass of <rref>row</rref>, adding an <a>owned</a> element with a role of <rref>row</rref> will not fulfill the requirement that <rref>listbox</rref> owns an <rref>option</rref> or a <rref>group</rref>.</p>
+			<h3>Allowed Child Roles</h3>
+			<p>A list of roles which may be a direct <a>child</a> of the element with this <a>role</a>. For example, an element with the role <rref>list</rref> may own elements with the role <rref>listitem</rref>, but not elements with the role <rref>option</rref>.</p>
```
pkra, line 675: From 

> A list of roles which may be a direct child of the element with this role.

to

> A list of roles which may be allowed on an owned child of the element with this role.

pkra: Sorry, posted my earlier comment too early. Since roles are not the children, this seemed better to me.

pkra: (And also "owned child" vs "direct child")

common/terms.html
smhigley: No, it would be an owned descendant, but not an owned child.

jnurthen:
should also resolve #1486

jnurthen:
@smhigley would this PR take care of #1300 ?

common/terms.html
smhigley: Updated, with a reference back to "owned child"

index.html
smhigley: Updated to "owning parent" to match terms.

smhigley:
@pkra / @scottaohara I updated based on both of your feedback -- let me know if you have any more suggestions!

scottaohara (approved):
having reviewed this a few times now, I think this is looking good.

pkra (approved):

pkra:
Thanks for the changes @smhigley - looks good to me!

index.html
```diff
@@ -6664,11 +6672,11 @@ <h5>Presentational Role Inheritance</h5>
...
 					</tr>
 					<tr>
-						<th class="role-scope-head" scope="row">Required Context Role:</th>
+						<th class="role-scope-head" scope="row">Required Parent Role:</th>
 						<td class="role-scope"> </td>
 					</tr>
 					<tr>
-						<th class="role-mustcontain-head" scope="row">Required Owned Elements:</th>
+						<th class="role-mustcontain-head" scope="row">Allowed Child Roles:</th>
 						<td class="role-mustcontain"><rref>radio</rref></td>
```
WilcoFiers, line 6680: This one seems wrong now. Radiogroups do not require radio nodes as children, any descendant is fine. This is the only one of the required owned elements where I think it made sense to 1. require at least one, and 2. allow descendants.

If you wanted to keep it, you could add a MUST or SHOULD to L6642. In either case, this should come out.

```suggestion
						<td class="role-mustcontain"> </td>
```

index.html
```diff
@@ -596,18 +604,18 @@ <h3>Prohibited States and Properties</h3>
...
-			<p class="note">A role that has 'required owned elements' does not imply the reverse relationship. While processing of a role may be incomplete without elements of given roles present as descendants, elements with roles in this list do not always have to be found within elements of the given role. See <a href="#scope">required context role</a> for requirements about the context where elements of a given role will be contained.</p>
-			<p class="note">An element with a <a href="#subclassroles">subclass role</a> of the 'required owned element' does not fulfill this requirement. For example, the <rref>listbox</rref> role requires ownership of an element using the <rref>option</rref> or <rref>group</rref> role. Although the <rref>group</rref> role is the superclass of <rref>row</rref>, adding an [=ARIA/owned=] element with a role of <rref>row</rref> will not fulfill the requirement that <rref>listbox</rref> owns an <rref>option</rref> or a <rref>group</rref>.</p>
+			<h3>Allowed Child Roles</h3>
```
WilcoFiers, line 607: ❤

index.html
```diff
@@ -6347,7 +6355,7 @@ <h5>Presentational Role Inheritance</h5>
   &lt;foo&gt; Sample Content &lt;/foo&gt;
   &lt;foo&gt; More Sample Content &lt;/foo&gt;
 &lt;/foo&gt;</pre>
-				<p class="note">There are other <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> roles with required children for which this situation is applicable (e.g., radiogroups and listboxes), but tables and lists are the most common real-world cases in which the presentation inheritance is likely to apply.</p>
+				<p class="note">There are other <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> roles with specific allowed children for which this situation is applicable (e.g., radiogroups and listboxes), but tables and lists are the most common real-world cases in which the presentation inheritance is likely to apply.</p>
```
WilcoFiers, line 6358: See my comment on radiogroups. Don't think this should have allowed child roles.

```suggestion
				<p class="note">There are other <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> roles with specific allowed children for which this situation is applicable (e.g., feeds and listboxes), but tables and lists are the most common real-world cases in which the presentation inheritance is likely to apply.</p>
```

WilcoFiers (changes requested):

WilcoFiers:
Had a couple comments @smhigley. Thank you for putting this together. It looks great.

index.html
smhigley: Good catch, I'll remove the allowed child role from `radiogroup`.

index.html
```diff
@@ -353,14 +353,22 @@ <h1>Important Terms</h1>
 			<dd>
 			  <p>Usable by users in ways they can control. References in this document relate to <cite><a href="https://www.w3.org/TR/WCAG21/#operable"><abbr title="Web Content Accessibility Guidelines">WCAG</abbr> 2.1 Principle 2: Content must be operable</a></cite> [[WCAG21]]. See <a>Keyboard Accessible</a>.</p>
 			</dd>
-			<dt><dfn data-export="" data-dfn-for="ARIA" data-local-lt="owned|owned element's">Owned Element</dfn></dt>
+			<dt><dfn data-export="" data-lt="owned child|child element|child|children|child elements">Owned Child</dfn></dt>
+			<dd>
+				<p>An 'owned child' is a direct <abbr title="Document Object Model">DOM</abbr> child of the <a>element</a>, a descendant of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, or a child specified via an <pref>aria-owns</pref> relationship to the element.</p>
```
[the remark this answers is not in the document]
smhigley: This is a good question, but I don't think this PR meaningfully changes the question or the answer. Would you be up for opening a separate issue against the spec for clarifying DOM parent/child relationship mappings when `aria-owns` is also present?

My personal assumption would be a replacement of the parent relationship rather than an addition, but I'm curious what other people think. A quick test shows Chrome and Firefox on Windows move the owned element in the accessibility tree, but do not duplicate it.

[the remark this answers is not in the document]
smhigley: The generic element may still be included in the accessibility tree, but it should not disrupt the association between the parent/child relationship between a listbox and option in something like this example:

```html
<div role="listbox">
  <div tabindex-"-1">
    <button role="option">op</button>
  </div>
</div>
```

A `<div>` should never have a role of `presentation` unless it's explicitly applied. I think I remember some discussion around giving it a `role="group"` instead of `generic` based on some combination of naming & focused state, but I don't believe that's currently the case. If that were to change and `<div tabindex="-1">` were to be mapped as `group`, then my understanding is the above code would be disallowed.

smhigley:
@WilcoFiers thanks for the review! With the `radiogroup` changes and my above comments, does this look good to you?

index.html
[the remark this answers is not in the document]
WilcoFiers: https://github.com/w3c/aria/issues/1817

WilcoFiers (approved):
@smhigley This looks good. Though please do take note that the way `generic` is described in this PR is different from how AT operate today. It's a good change, but it'll require a change to browsers/AT.

index.html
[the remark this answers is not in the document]
smhigley: It's a good callout, and I think there are two potential ways forward:
1. The mapped role of a focusable `<div>` is no longer generic
2. Webkit allows this code to functionally work (note: this does not require them to remove the `<div>` from the a11y tree, just to not have it to disrupt the listbox/option relationship)

This came up some time ago in the ARIA WG discussions, and I think I remember @cookiecrook wanting to verify that this does not require UAs to remove generics from the a11y tree. I'll ask him to confirm though :)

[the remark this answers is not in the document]
cookiecrook: Some comments... 

1. I don't think "owned child" is a good term, especially when it can contain non-child descendants.
2. The proposed def does not count for shadow roots, which may also need accounting for? If so, in a separate issue.
3. Re: "The mapped role of a focusable `<div>` is no longer generic", @scottaohara may wish to address the mappings of `<div tabindex>` and `<div aria-globalattrs>` in separate rows for HTML-AAM?
4. Re: "Webkit allows this code to functionally work" IIRC this was a VoiceOver change. WebKit exposes the tree as it did, but VO (years ago) wouldn't step deeply into it if there were many layers of irrelevant container elements in between the listbox and its options.
5. I think all the code samples above are purposefully author ambiguous, so AT tries to do its best. If there's a specific "bug" or something the spec should address, let's call that out in an engine problem tracker or an ARIA issue. E.g. I think the focusable "text" role description is reasonable, as is labeling by the first descendant element ("op"), but we could discuss diffs of opinion in a new issue?

I'm hopeful I have responded to all points. If not, please call out again.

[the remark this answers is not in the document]
cookiecrook: IMO, "owned child" is the wrong term for the reasons outlined in the previous comment. 

I concur with Sarah that the owned element should be be moved in the AX tree, not copied. Note that this is a more recent change than the original understanding of aria-owns in ARIA 1.0. Several of those earlier meetings referenced author based aria-owns changes affecting the tree as the Ghostbusters meme: "Don't cross the streams… It would be bad."

However, most of the implementation challenges have been resolved since then, and I think all implementors are now in agreement that "move in the AX tree" should replace the "keep in place but add a cross-ref pointer" which was the original understanding. Those changes are are only partially implemented though.

smhigley:
Based on @cookiecrook's comments, I updated the text from "Allowed child roles"/"Required parent role" to "Allowed direct descendant roles"/"Required direct ancestor role".

Pinging @jnurthen for an editorial review of the update

cookiecrook:
Sorry to nitpick but I don't think "direct" is right either...

What about "Required container" (Accessibility Ancestor) and "Required contained roles" (Accessibility Descendant)

Or perhaps more direct: "Required Accessibility Ancestor Context Role" and "Required Accessibility Descendant Roles"

cookiecrook:
"Required Accessibility Parent"/"Required Accessibility Child"

dd8:
How are optional child roles handled? For example `menu` can have optional `seperator` elements between the menu items, and `table` \ `grid` \ `treegrid` can have an optional `caption` child. The text 'Allowed Child Elements' suggests they need to be included somehow.

There seem to be two types of allowed a11y tree children:

- required children like `radio` inside `radiogroup` or `row` \ `rowgroup` inside `table`
- optional children like `caption` inside `table` and `seperator` inside `menu`

Seems like you need to list both if there's normative text forbidding any roles except the listed child roles for a given parent role. At the moment the optional roles appear in the description of `caption` and `menuitem` but this makes it hard to make a testable statement about what's allowed as an a11y child.

This came up when reviewing one of the ARIA Act Rules - it currently only allows ARIA 1.2 required owned roles, which causes problems with optional roles like `caption`:
https://www.w3.org/WAI/standards-guidelines/act/rules/bc4a75/proposed/

cookiecrook:
> How are optional child roles handled? 

I think the wording "allowed" may address that need. Technically, even the "required" children aren't required when a container is marked as busy, so perhaps.

dd8:
> > The text 'Allowed Child Elements' suggests they need to be included somehow.
> 
> I don't understand how "allowed" implies they need to be included.

I mean included in the list of Allowed Direct Descendant Roles in the spec. If a different term like Required Direct Descendant Roles is used then it's less of an issue (but makes it harder to determine which Direct Descendant Roles are allowed).

The term Allowed Direct Descendant Roles implies anything not on the list is disallowed (and the optional elements don't currently appear in the Allowed Direct Descendant Role lists in the PR). For example the list for role=menu shows:

```
<th class="role-mustcontain-head" scope="row">Allowed Direct Descendant Roles:</th>
<td class="role-mustcontain">
    <ul>
        <!-- keep in sync with #menubar -->
        <li><rref>group</rref> with child <rref>menuitem</rref></li>
        <li><rref>group</rref> with child <rref>menuitemradio</rref></li>
        <li><rref>group</rref> with child <rref>menuitemcheckbox</rref></li>
        <li><rref>menuitem</rref></li>
        <li><rref>menuitemcheckbox</rref></li>
        <li><rref>menuitemradio</rref></li>
    </ul>
</td>
```

but the separator role is also allowed:

> Authors MAY separate menu items into sets by use of a <rref>separator</rref> or an element with an equivalent role from the native markup language.

No matter what term is finally used it would be useful listing optional child elements on the parent roles since it takes a lot of hunting through the individual role descriptions to find which parent roles allow them in addition to required roles.

smhigley:
One of the goals of this change was to fully remove the implication that any roles are "required", since even for treeitem within tree, or menuitem within menu, it is valid to have a menu with no items, or a tree with no content. The "required owned element" wording incorrectly implies that the container role cannot be empty, which has caused issues in the past (e.g. https://github.com/dequelabs/axe-core/issues/3543 and other linked issues). My preference for resolving required vs. optional is that all children are allowed/optional, and none are required.

@dd8 thanks for pointing out the missing separator and things like paragraph missing allowed roles! I'll do another pass on catching all the roles that should be permitted, but I do imagine that we'll find more and add them iteratively in the future :)

dd8:
@smhigley That makes perfect sense. I think role=caption should probably appear in the allowed list for table, grid, and treegrid. 

I was going to mention role=radiogroup, which required role=radio children in ARIA 1.2 but looks like that's been removed in the current draft.

smhigley:
I added separator and caption to their respective parent roles, but didn't add anything for paragraph and other general content containers, since they already do not have an entry for "Allowed Child Roles", so that should be enough to imply that there are no restrictions IMO (which is the same as it was before).

index.html
```diff
@@ -353,14 +353,22 @@ <h1>Important Terms</h1>
 			<dd>
 			  <p>Usable by users in ways they can control. References in this document relate to <cite><a href="https://www.w3.org/TR/WCAG21/#operable"><abbr title="Web Content Accessibility Guidelines">WCAG</abbr> 2.1 Principle 2: Content must be operable</a></cite> [[WCAG21]]. See <a>Keyboard Accessible</a>.</p>
 			</dd>
-			<dt><dfn data-export="" data-dfn-for="ARIA" data-local-lt="owned|owned element's">Owned Element</dfn></dt>
+			<dt><dfn data-export="" data-lt="accessibility child|owned child|child element|child|children|child elements">Accessibility child</dfn></dt>
+			<dd>
+				<p>An 'accessibility child' is a <abbr title="Document Object Model">DOM</abbr> child of the <a>element</a>, a descendant of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, a child specified via an <pref>aria-owns</pref> relationship to the element, or a descendant of a <rref>generic</rref> or <rref>presentation</rref> child specified via <pref>aria-owns</pref> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening.</p>
```
cookiecrook, line 358: Interesting. I didn't realize (or didn't remember) the last scenario was expected to work. Seems okay though.

index.html
```diff
@@ -596,18 +604,18 @@ <h3>Prohibited States and Properties</h3>
...
-			<p class="note">An element with a <a href="#subclassroles">subclass role</a> of the 'required owned element' does not fulfill this requirement. For example, the <rref>listbox</rref> role requires ownership of an element using the <rref>option</rref> or <rref>group</rref> role. Although the <rref>group</rref> role is the superclass of <rref>row</rref>, adding an [=ARIA/owned=] element with a role of <rref>row</rref> will not fulfill the requirement that <rref>listbox</rref> owns an <rref>option</rref> or a <rref>group</rref>.</p>
+			<h3>Allowed Accessibility Child Roles</h3>
+			<p>A list of roles which may be allowed on an <a>accessibility child</a> (simplified as "child") of the element with this <a>role</a>. For example, an element with the role <rref>list</rref> may own child elements with the role <rref>listitem</rref>, but not elements with the role <rref>option</rref>.</p>
+			<p>To determine whether an element is the <a>accessibility child</a> of a role, <a>user agents</a> MUST ignore any intervening elements with the role <rref>generic</rref> or <rref>presentation</rref>.</p>
```
cookiecrook, line 609:
```suggestion
			<p>To determine whether an element is the <a>accessibility child</a> of a role, <a>user agents</a> MUST ignore any intervening elements with the role <rref>generic</rref> or <rref>none</rref>.</p>
```

index.html
```diff
@@ -353,14 +353,22 @@ <h1>Important Terms</h1>
...
 			<dd>
 			  <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child.</p>
 			</dd>
 			<dt><dfn data-local-lt="owning">Owning Element</dfn></dt><!--Not used-->
 			<dd>
 			  <p>An 'owning element' is any <abbr title="Document Object Model">DOM</abbr> ancestor of the <a>element</a>, or any element with an <pref>aria-owns</pref> attribute which references the ID of the   element. </p>
 			</dd>
+			<dt><dfn data-export="" data-lt="accessibility parent|parent element|parent">Accessibility parent</dfn></dt>
+			<dd>
+				<p>An 'accessibility parent' is a <abbr title="Document Object Model">DOM</abbr> parent of the <a>element</a>, an ancestor of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, any element with an <a>owned child</a> specified via <pref>aria-owns</pref>, or an ancestor of an <a>owned element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening.</p>
```
cookiecrook, line 370:
```suggestion
				<p>An 'accessibility parent' is a <abbr title="Document Object Model">DOM</abbr> parent of the <a>element</a>, an ancestor of the <a>element</a> with only elements of role <rref>generic</rref> or <rref>none</rref> intervening, any element with an <a>owned child</a> specified via <pref>aria-owns</pref>, or an ancestor of an <a>owned element</a> with only elements of role <rref>generic</rref> or <rref>none</rref> intervening.</p>
```

index.html
cookiecrook:
```suggestion
				<p>An 'accessibility child' is a <abbr title="Document Object Model">DOM</abbr> child of the <a>element</a>, a descendant of the <a>element</a> with only elements of role <rref>generic</rref> or <rref>none</rref> intervening, a child specified via an <pref>aria-owns</pref> relationship to the element, or a descendant of an element with role <rref>generic</rref> or <rref>none</rref> specified via <pref>aria-owns</pref> with only elements of role <rref>generic</rref> or <rref>none</rref> intervening.</p>
```

index.html
```diff
@@ -596,18 +604,18 @@ <h3>Prohibited States and Properties</h3>
...
-			<p class="note">A role that has 'required context role' does not imply the reverse relationship. While an element with the given role needs to appear within an element of the listed role(s) in order to be meaningful, elements of the listed roles do not always need descendant elements of the given role in order to be meaningful. See <a href="#mustContain">required owned elements</a> for requirements about elements that require presence of a given descendant to be processed properly.</p>
+			<h3>Required Accessibility Parent Role</h3>
+			<p>The required <a>accessibility parent</a> (simplified as "parent") role defines the container where this <a>role</a> is allowed. If a role has a required accessibility parent, authors MUST ensure that an element with the role is an <a>accessibility child</a> of an element with the required accessibility parent role. For example, an element with role <code>listitem</code> is only meaningful when it is a child of an element with role <code>list</code>.</p>
+			<p>To determine whether an element has a parent with the required role, <a>user agents</a> MUST ignore any elements with the role <rref>generic</rref> or <rref>presentation</rref>.</p>
```
cookiecrook, line 618:
```suggestion
			<p>To determine whether an element has a parent with the required role, <a>user agents</a> MUST ignore any elements with the role <rref>generic</rref> or <rref>none</rref>.</p>
```

cookiecrook (approved):
Looks great, Sarah! Thanks for working so hard to get this finished.

My editorial changes here are just changing "generic or presentation elements" to "elements of role generic or none"... b/c I think `none` as a role is more understandable than `presentation`... We added the `none` synonym after 1.0 shipped b/c authors had so much trouble confusing `role="presentation"` with `aria-hidden="true"`... IMO, any time we can avoid using the `presentation` synonym in prose increases the understandability.

index.html
```diff
@@ -353,14 +353,22 @@ <h1>Important Terms</h1>
 			<dd>
 			  <p>Usable by users in ways they can control. References in this document relate to <cite><a href="https://www.w3.org/TR/WCAG21/#operable"><abbr title="Web Content Accessibility Guidelines">WCAG</abbr> 2.1 Principle 2: Content must be operable</a></cite> [[WCAG21]]. See <a>Keyboard Accessible</a>.</p>
 			</dd>
-			<dt><dfn data-export="" data-dfn-for="ARIA" data-local-lt="owned|owned element's">Owned Element</dfn></dt>
+			<dt><dfn data-export="" data-lt="accessibility child|owned child|child element|child|children|child elements">Accessibility child</dfn></dt>
+			<dd>
+				<p>An 'accessibility child' is a <abbr title="Document Object Model">DOM</abbr> child of the <a>element</a>, a descendant of the <a>element</a> with only elements of role <rref>generic</rref> or <rref>none</rref> intervening, a child specified via an <pref>aria-owns</pref> relationship to the element, or a descendant of an element with role <rref>generic</rref> or <rref>none</rref> specified via <pref>aria-owns</pref> with only elements of role <rref>generic</rref> or <rref>none</rref> intervening.</p>
```
scottaohara, line 358: i wonder if this could be written as a list of conditions?  i'm finding this incredibly difficult to read.

maybe if examples were added?  i had to write out an example for each to make sure i was actually following along.

DOM child
```
<div role=listbox>
  <div role=option ...>...</div>
</div>
```

Descendent of an element
```
<div role=listbox>
  <div>
    <div role=option ...>...</div>
  </div>
</div>
```

aria-owns relationship.
```
<div role=listbox aria-owns=f></div>
<div role=option id=f ...>...</div>
```

aria-owns relationship with intervening elements.
this one definitely seems covered.
```
<div role=listbox aria-owns=f></div>
<div id=f>
  <div>
    <div role=option ...>...</div>
  </div>
</div>
```

this one seems like it has to be covered, but the wording keeps making me think maybe it could not be?
```
<div role=listbox aria-owns=f></div>
<div id=f>
  <div role=option ...>...</div>
</div>
```

i'm a bit stuck on the 'with only...' where that to me says what follows is part of the condition.  But 'with or without only' to me makes both of my above examples true.

sorry, i wish i hadn't spent the last 35 minutes just on this one line.

index.html
```diff
@@ -596,18 +604,18 @@ <h3>Prohibited States and Properties</h3>
...
-			<p class="note">An element with a <a href="#subclassroles">subclass role</a> of the 'required owned element' does not fulfill this requirement. For example, the <rref>listbox</rref> role requires ownership of an element using the <rref>option</rref> or <rref>group</rref> role. Although the <rref>group</rref> role is the superclass of <rref>row</rref>, adding an [=ARIA/owned=] element with a role of <rref>row</rref> will not fulfill the requirement that <rref>listbox</rref> owns an <rref>option</rref> or a <rref>group</rref>.</p>
+			<h3>Allowed Accessibility Child Roles</h3>
+			<p>A list of roles which may be allowed on an <a>accessibility child</a> (simplified as "child") of the element with this <a>role</a>. For example, an element with the role <rref>list</rref> may own child elements with the role <rref>listitem</rref>, but not elements with the role <rref>option</rref>.</p>
```
scottaohara, line 608: does this need to be a "but MUST not own elements with..."

the use of 'may' (but not MAY) in the first part of the sentence has me interpreting the second part as being a bit squishy with this expectation.

index.html
```diff
@@ -596,18 +604,18 @@ <h3>Prohibited States and Properties</h3>
...
+			<p>To determine whether an element is the <a>accessibility child</a> of a role, <a>user agents</a> MUST ignore any intervening elements with the role <rref>generic</rref> or <rref>none</rref>.</p>
```
scottaohara, line 609: child of a role?    

this seems to be the only time its referred to this way, where other times its referred to as a child of the or an element, or a specific element/role type.

not 100% opposed to this wording if others are fine with it.  just seemed odd, as there's no other wording like this in the spec at present.

scottaohara (commented):
i feel bad even posting these comments, this is looking so good and so close to done.  i hope my comments are helpful.

index.html
```diff
@@ -353,14 +353,22 @@ <h1>Important Terms</h1>
...
+			</dd>
+			<dt><dfn data-export="" data-dfn-for="ARIA" data-local-lt="owned element|owned|owned element's|owned elements">Owned Element</dfn></dt>
 			<dd>
 			  <p>An 'owned element' is any <abbr title="Document Object Model">DOM</abbr> descendant of the <a>element</a>, any element specified as a child via <pref>aria-owns</pref>, or any <abbr title="Document Object Model">DOM</abbr> descendant of the owned child.</p>
 			</dd>
 			<dt><dfn data-local-lt="owning">Owning Element</dfn></dt><!--Not used-->
 			<dd>
 			  <p>An 'owning element' is any <abbr title="Document Object Model">DOM</abbr> ancestor of the <a>element</a>, or any element with an <pref>aria-owns</pref> attribute which references the ID of the   element. </p>
```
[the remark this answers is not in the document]
Jym77: > * In Scenario #⁠1, the “Child node” `li` should be _hidden_ (i.e., excluded from the accessibility tree) because it has an “accessibility parent” — the intervening `div` — which is not rendered in host language terms.

> In this way, “owning elements” would become a subset of “accessibility parents”.

I think that work, but I'm not sure the distinction is needed.

## Scenario 3
```html
<ul style="display: none" aria-owns="li"></ul>
<div>
  <li id="li">Child node</li>
</div>
```
Here, the `li` is exposed because `display: none` prevents the `aria-owns` to take effect, therefore the `ul` is not an "accessibility parent" and it not being rendered doesn't matter.
Where this goes (I feel), is that we actually seem to only need "**DOM ancestor** not rendered => hidden" That is, I cannot really see a case where an "accessibility parent" would be not rendered, thus hiding the element, without that accessibility parent being a DOM ancestor. Because the "accessibility parent" is either a DOM ancestor or an `aria-own`ing element, but in the second case if it's not rendered, then it doesn't `aria-own` anything 🤔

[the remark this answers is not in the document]
spectranaut: Two things:

1. For your PR,  @adampage, given what is said above, can you just use the term "accessibility parent" and DOM ancestor, and drop the "owning element"/"owned element" language?

2. For this PR, @smhigley, I'm realizing its a bit odd that "own child" and "accessibility child" have the same definition, but "owned parent" and "accessibility parent" are different. Is this intentional?

index.html
spectranaut:
```suggestion
			<p>A list of roles which may be allowed on an <a>accessibility child</a> (simplified as "child") of the element with this <a>role</a>. For example, an element with the role <rref>list</rref> MAY own child elements with the role <rref>listitem</rref>, but MUST NOT own elements with the role <rref>option</rref>.</p>
```

I think this is the fix?

index.html
spectranaut:
```suggestion
			<p>To determine whether an element is the <a>accessibility child</a> of an element, <a>user agents</a> MUST ignore any intervening elements with the role <rref>generic</rref> or <rref>none</rref>.</p>
```

I think scott's right, so I'm just suggesting it

index.html
spectranaut:
```suggestion
				<p>An 'accessibility child' refers to the child in the <a>accessibility tree<a> ignoring any intervening elements of role <rref>generic</rref> or <rref>none</rref>.</p>
				<p>Related: <a href="#mustContain">Allowed Accessibility Child Roles</a>, <a>Accessibility Parent</a>, <pref>aria-owns</pref><p>
```

Ok, I decided to try my hand at this, what do you guys think with a similar definition for  accessibility parent? I think it covers all of @scottaohara's and @smhigley cases. It was @pkra review of my definition of hidden which makes me propose this (see hidden: https://w3c.github.io/aria/#dfn-hidden)

spectranaut (commented):
I just tried to make suggestions based on Scott's comments to try to help this CL along, take them or leave them, @smhigley ! :)

index.html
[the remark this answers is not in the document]
adampage: re: #1 — I think so, depending on how the final language for “accessibility parent” shakes out.  One of the puzzles I need to solve in #1839 is to explain how the `<div aria-hidden="true">` in this scenario:

```html
<ul aria-owns="li"></ul>
<div aria-hidden="true">
  <li id="li">Child node</li>
</div>
```

...will _not_ cause the `<li>` to be excluded from the accessibility tree.  The current definition of “accessibility parent” will still apply to the `<div>` since it will continue to be a DOM ancestor after the `<ul>`’s `aria-owns` has taken effect.

So that’s where I figured a more tightly scoped definition of “owning element” would give me a way to distinguish: the `<div>` might still be both a DOM ancestor and an “accessibility parent” after `aria-owns` has been resolved, but it would no longer be an “owning element” and therefore its `aria-hidden="true"` would have no effect.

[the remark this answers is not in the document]
smhigley: My understanding was that `aria-owns` moved an element in the accessibility tree from one branch to another, so the owned `<li>` would no longer have an accessibility parent/child relationship with the `<div>`.

For the scenario with `display: none`, I'm not sure whether that works in practice. I think that would be a somewhat separate concern, related to #1839, in determining whether `aria-owns` works with e.g. `hidden` an `display: none`. My understanding is that in any case in which `aria-owns` is processed, the `<div>` is no longer the parent of the `<li>`.

@spectranaut I think "owning element" is a superset of "accessibility parent", since it already includes the entire tree, not just parent-chlid relationships. The generic clarification isn't necessary there, since it's an ancestor thing anyway and not a direct parent one. Same for "owned element", not sure why the wording is slightly different between the two.

index.html
smhigley: Hm, I actually think this goes too far in the direction of simplification. I think I'd prefer to explicitly list the ways the child can be associated with the parent, i.e.
- DOM child
- DOM descendent with only generics intervening
- direct aria-owns relationship
- descendant of aria-owns relationship with only generics intervening

I think the issue with "child in the accessibility tree" is it becomes a sort of circular definition -- an "accessibility child" is a "child in the accessibility tree", but without specifying how to construct the accessibility tree.

smhigley: I'll add the list of examples though, with one for each :)

smhigley:
@spectranaut, @scottaohara, and @adampage -- I updated with a lot of the suggestions you mentioned, and some comments. Let me know what you think!

spectranaut (approved):
Looks great!

I don't think there are any changes to the validator tests, right, Sarah? Seems to me like this is all clarification/rewording of existing "author MUST" requirements.

smhigley:
@spectranaut I think the validator change would be checking something like this:
```
<div role="listbox">
  <div>
    <div role="option">hi I'm valid now</div>
  </div>
</div>
```

spectranaut:
Oh right what was I thinking, we even talked about this. I just got so wrapped up in the "rewrite" and new terms I forgot about this key change XD

Do you want to do that here, or in a follow up issue? If a follow issue, please make one, otherwise, you can add a validator test here.

index.html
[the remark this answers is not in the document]
adampage: Got it, thank you.  I had misunderstood the scope of the accessibility parent/child relationship.  @spectranaut, I’ll do what you suggested and update #1839 to just differentiate between “accessibility parent” and “DOM ancestor”. 👍🏻

index.html
```diff
@@ -13501,7 +13547,7 @@ <h3>Roles</h3>
...
-		<p>Certain landmark roles require names from authors. In situations where an author has not specified names for these landmarks, it is considered an authoring error. The user agent MUST treat such elements as if no <a>role</a> had been provided. If a valid fallback role had been specified, or if the element had an implicit ARIA role, then user agents would continue to expose that role, instead.  Instances of such roles are as follows:</p>
+		<p>Certain landmark roles require names from authors. In situations where an author has not specified names for these landmarks, it is considered an authoring error. The user agent MUST treat such elements as if no <a>role</a> had been provided. Igitf a valid fallback role had been specified, or if the element had an implicit ARIA role, then user agents would continue to expose that role, instead.  Instances of such roles are as follows:</p>
```
adampage, line 13550:
```suggestion
		<p>Certain landmark roles require names from authors. In situations where an author has not specified names for these landmarks, it is considered an authoring error. The user agent MUST treat such elements as if no <a>role</a> had been provided. If a valid fallback role had been specified, or if the element had an implicit ARIA role, then user agents would continue to expose that role, instead.  Instances of such roles are as follows:</p>
```

A wild git tried to sneak in. 😄

adampage (approved):

index.html
smhigley: oh my god, thanks for catching this 🤣

smhigley:
@spectranaut issue is made! https://github.com/w3c/aria/issues/1917

spectranaut:
@pkra or @jnurthen do you think we can merge this? :)

index.html
```diff
@@ -601,18 +609,51 @@ <h3>Prohibited States and Properties</h3>
...
-			<p class="note">An element with a <a href="#subclassroles">subclass role</a> of the 'required owned element' does not fulfill this requirement. For example, the <rref>listbox</rref> role requires ownership of an element using the <rref>option</rref> or <rref>group</rref> role. Although the <rref>group</rref> role is the superclass of <rref>row</rref>, adding an [=ARIA/owned=] element with a role of <rref>row</rref> will not fulfill the requirement that <rref>listbox</rref> owns an <rref>option</rref> or a <rref>group</rref>.</p>
+			<h3>Allowed Accessibility Child Roles</h3>
+			<p>A list of roles which may be allowed on an <a>accessibility child</a> (simplified as "child") of the element with this <a>role</a>. For example, an element with the role <rref>list</rref> MAY own child elements with the role <rref>listitem</rref>, but MUST NOT own elements with the role <rref>option</rref>.</p>
```
spectranaut, line 613: Is this first may an normative MAY? Seems like yes actually
```suggestion
			<p>A list of roles which MAY be allowed on an <a>accessibility child</a> (simplified as "child") of the element with this <a>role</a>. For example, an element with the role <rref>list</rref> MAY own child elements with the role <rref>listitem</rref>, but MUST NOT own elements with the role <rref>option</rref>.</p>
```

or replace with "A list of roles that are allowed"

spectranaut (dismissed):
I spoke too soon once again! The new CI check caught an errant "may", we need to fix that THEN  I think it is ready for merge

index.html
cookiecrook: Actually, I think we'd need an actor (Author or User Agent) in each of the RFC-2119 normative statements, otherwise they are ambiguous and unenforceable with author warnings or automated browser tests.

cookiecrook:
```suggestion
			<p>A list of roles which are allowed on an <a>accessibility child</a> (simplified as "child") of the element with this <a>role</a>. For example, an element with the role <rref>list</rref> can own child elements with the role <rref>listitem</rref>, but cannot own elements with the role <rref>option</rref>.</p>
```

cookiecrook (changes requested):

index.html
spectranaut: ok I see, @smhigley I thnk you can just commit @cookiecrook suggestiong above.

spectranaut, line 613:
```suggestion
			<p>A list of roles which are allowed on an <a>accessibility child</a> (simplified as "child") of the element with this <a>role</a>. Authors MUST only add child element with allowed roles. For example, an element with the role <rref>list</rref> MAY own child elements with the role <rref>listitem</rref>, but MUST NOT own elements with the role <rref>option</rref>.</p>
```

jnurthen: I like it

index.html
```diff
@@ -6371,10 +6415,10 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
...
 				<p>For any element with an explicit or inherited role of presentation and which is not focusable, user agents MUST ignore role-specific <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> states and properties for that element. For example, in HTML, a <code>ul</code> or <code>ol</code> element with a role of <code>presentation</code> will have the implicit native semantics of its <code>li</code> elements removed because the <rref>list</rref> role to which the <code>ul</code> or <code>ol</code> corresponds has a <a href="#mustContain">required owned element</a> of <rref>listitem</rref>. Likewise, the implicit native semantics of an HTML <code>table</code> element's <code>thead</code>/<code>tbody</code>/<code>tfoot</code>/<code>tr</code>/<code>th</code>/<code>td</code> descendants will also be removed, because the HTML specification indicates that these are required structural descendants of the <code>table</code> element.</p>
```
jnurthen, line 6420:
```suggestion
				<p>For any element with an explicit or inherited role of presentation and which is not focusable, user agents MUST ignore role-specific <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> states and properties for that element. For example, in HTML, a <code>ul</code> or <code>ol</code> element with a role of <code>presentation</code> will have the implicit native semantics of its <code>li</code> elements removed because the <rref>list</rref> role to which the <code>ul</code> or <code>ol</code> corresponds has an<a href="#mustContain">Allowed Accessibility Child Role</a> of <rref>listitem</rref>. Likewise, the implicit native semantics of an HTML <code>table</code> element's <code>thead</code>/<code>tbody</code>/<code>tfoot</code>/<code>tr</code>/<code>th</code>/<code>td</code> descendants will also be removed, because the HTML specification indicates that these are required structural descendants of the <code>table</code> element.</p>
```

jnurthen (commented):
I think a #mustContain was missed

index.html
spectranaut: looks right!

PR #1457, opened 2021-04-15
removal of redundant content

### LINKED ISSUE #1236: Processing of the role attribute not clearly defined
For the `role` attribute, https://w3c.github.io/aria/#introroles says

> User agents MUST use the first token in the sequence of tokens in the role attribute value that matches the name of any non-abstract WAI-ARIA role. The following steps will correctly identify the applicable WAI-ARIA role:

The above only has one MUST requirement; the second sentence and the list below are statements of fact. I think the first sentence should be removed, and the second sentence should say "MUST".

>
> 1. Use the rules of the host language to detect that an element has a role attribute and to identify the attribute value string for it.

What does this mean? Is it applying the requirements from html-aam for HTML?

> 2. Separate the attribute value string for that attribute into a sequence of whitespace-free substrings by separating on whitespace.

What does "whitespace" mean? Should this use https://infra.spec.whatwg.org/#split-on-ascii-whitespace ?

> 3. Compare the substrings to all the names of the non-abstract WAI-ARIA roles. Case-sensitivity of the comparison inherits from the case-sensitivity of the host language.

What does this mean? HTML has some attribute values case-sensitive, some ASCII case-insensitive.

> 4. Use the first such substring in textual order that matches the name of a non-abstract WAI-ARIA role.

(This step seems clear enough.)

scottaohara:
closes #1236

remove second sentence / steps to correctly identify WAI-ARIA roles from intro to roles section.

add new sentence linking to 8.1 Role Attribute steps for `role` implementation in the host language

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1457.html" title="Last updated on Apr 15, 2021, 7:10 PM UTC (cbe9c63)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1457/16dfd46...cbe9c63.html" title="Last updated on Apr 15, 2021, 7:10 PM UTC (cbe9c63)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -499,13 +499,7 @@
 			<li>supported <a>states</a> and <a>properties</a> for each role (e.g., a <rref>checkbox</rref> supports being checked via <sref>aria-checked</sref>).</li>
 		</ul>
 		<p>Attaching a role gives <a>assistive technologies</a> information about how to handle each element. When <abbr title="Accessible Rich Internet Application">WAI-ARIA</abbr> roles override host language semantics, there are no changes in the <abbr title="Document Object Model">DOM</abbr>, only in the <a class="termref">accessibility tree</a>.</p>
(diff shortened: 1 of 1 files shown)


carmacleod (approved):
+1
Looks good.
Thanks @scottaohara !

cookiecrook (approved):

jnurthen (approved):

PR #1467, opened 2021-04-21
aria-disabled and descendant elements

### LINKED ISSUE #1130: aria-disabled non-focusable descendents
Currently, the ARIA spec says the following about using aria-disabled to disable descendants: 

> https://www.w3.org/TR/wai-aria-1.1/#aria-disabled
> The state of being disabled applies to the current element and all focusable descendant elements of the element on which the aria-disabled attribute is applied.

That would suggest that in the following code snippet the textfield would not be disabled:

```html
<div role="group" aria-disabled="true" aria-labelledby="legend">
  <div id="legend">Personal information</div>
  <div id="givenname">Given name</div>
  <div role="textbox" tabindex="-1" aria-labelledby="givenname"></div>
</div>
```

I checked this against Chrome and Firefox. Chrome considers the textbox disabled. Firefox seems to ignore aria-disabled on ancestors regardless of if it is focusable or not. So neither browser seem to conform to spec.

I think Chrome's behaviour is the sensible one here, as it is consistent with what the HTML specification says should happen when using native elements. In the following example the disabled attribute on the fieldset disables he input element. The input element is not focusable because of it:

```html
<fieldset>
  <legend>Personal information</legend>
  <label for="givenname">Given name</label>
  <input type="text" id="givenname">
</fieldset>
```

I do not know if this "all focusable descendant elements" is intentional, but it is inconsistent with the HTML specification, and the implementation for it seems to be lacking. Would it be possible to align this with the behaviour of the disabled attribute in HTML?

scottaohara:
This is some draft wording I have in place on how to resolve #1130.  

Essentially, I *do not* want to break expectations with how `<fieldset disabled>` works compared to `<div role=group aria-disabled=true>` in regards to nested links.  In the former links are _not_ disabled because HTML does **not** allow `<a href>` to be set to the disabled state.  The latter will expose an `<a href>` or a `<div role=link>` as being in the _disabled_ state. This should not happen and user agents that presently do this are likely not matching the reality of what the HTML author expected would happen.

So, yeh.  Happy to get feedback on what I've done so far and make adjustments as need be.  But, am quite against the idea of the disabled state on an ancestor element impacting descendant links. So whatever we do here, I want to make sure _that_ is clearly _not_ what is expected here, unless it can be very clear this would only apply to `role=link` and NOT `<a href>` HTML links.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1467.html" title="Last updated on Jul 1, 2021, 7:26 PM UTC (b00c79a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1467/627b3a6...b00c79a.html" title="Last updated on Jul 1, 2021, 7:26 PM UTC (b00c79a)">Diff</a>

***

### :boom: Error: 504 Gateway Timeout :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 5, 2022, 12:47 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Faria%2F19ad712cf2881a14851e64709afa8af48cce0f1a%2Findex.html%3FisPreview%3Dtrue)

```
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231467.)._
</details>


index.html
```diff
@@ -11552,8 +11552,8 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p>The <a>state</a> of being disabled applies to the current element and all focusable descendant elements of the element on which the <sref>aria-disabled</sref> <a>attribute</a> is applied.</p>
-                <p class="note">While <sref>aria-disabled</sref> and proper scripting can successfully disable an element with role <rref>link</rref>, fully disabling a host language equivalent can be problematic. Authors are advised not to use <sref>aria-disabled</sref> on elements that cannot be disabled through features of the host language alone.</p>
+				<p>The <a>state</a> of being disabled applies to the element which the <sref>aria-disabled</sref> <a>attribute</a> is specified, and any  descendants of that element which inherit from the <rref>input</rref> abstract role, to which the <sref>aria-disabled</sref> <a>attribute</a> can be applied. For instance, User Agents MUST NOT set a <rref>link</rref>, or its host language equivalent, to the disabled state if it were to be a descendant of an ancestor element with the <sref>aria-disabled</sref> attribute specified.</p>
```
scottaohara, line 11555:
```suggestion
				<p>The <a>state</a> of being disabled applies to the element which the <sref>aria-disabled</sref> <a>attribute</a> is specified, and any descendants of that element which inherit from the <rref>input</rref> abstract role, to which the <sref>aria-disabled</sref> <a>attribute</a> can be applied. For instance, it would be expected for <rref>textbox</rref>, <rref>combobox</rref>, or their host language equivalents to be exposed in the disabled state if they are descendants of an ancestor element with a specified <sref>aria-disabled</sref> attribute set to the "true" value. However, User Agents MUST NOT set a <rref>link</rref>, or its host language equivalent, to the disabled state if it is a descendant of an ancestor element with a specified <sref>aria-disabled</sref> attribute set to the "true" value</p>
```

index.html
```diff
@@ -11329,8 +11329,8 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>The <a>state</a> of being disabled applies to the element which the <sref>aria-disabled</sref> <a>attribute</a> is specified, and any descendants of that element which inherit from the <rref>input</rref> abstract role, to which the <sref>aria-disabled</sref> <a>attribute</a> can be applied. For instance, it would be expected for <rref>textbox</rref>, <rref>combobox</rref>, or their host language equivalents to be exposed in the disabled state if they are descendants of an ancestor element with a specified <sref>aria-disabled</sref> attribute set to the "true" value. However, User Agents MUST NOT set a <rref>link</rref>, or its host language equivalent, to the disabled state if it is a descendant of an ancestor element with a specified <sref>aria-disabled</sref> attribute set to the "true" value</p>
+        <p class="note">While <sref>aria-disabled</sref> and proper scripting can successfully disable an element with role <rref>link</rref>, fully disabling a host language equivalent can be problematic. Authors are advised not to use <sref>aria-disabled</sref> on elements that cannot be disabled through features of the host language alone.</p>
```
cookiecrook, line 11333: looks like there are some whitespace only diffs (tabs versus spaces?). I think it's reasonable to correct the whitespace when you edit a line, but this doesn't match the previous line or the lines around it. I request that you either correct that portion of the diff or remove it.

screen shot showing inconsistent whitespace

```suggestion
				<p class="note">While <sref>aria-disabled</sref> and proper scripting can successfully disable an element with role <rref>link</rref>, fully disabling a host language equivalent can be problematic. Authors are advised not to use <sref>aria-disabled</sref> on elements that cannot be disabled through features of the host language alone.</p>
```

index.html
```diff
@@ -11329,8 +11329,8 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-                <p class="note">While <sref>aria-disabled</sref> and proper scripting can successfully disable an element with role <rref>link</rref>, fully disabling a host language equivalent can be problematic. Authors are advised not to use <sref>aria-disabled</sref> on elements that cannot be disabled through features of the host language alone.</p>
+				<p>The <a>state</a> of being disabled applies to the element which the <sref>aria-disabled</sref> <a>attribute</a> is specified, and any descendants of that element which inherit from the <rref>input</rref> abstract role, to which the <sref>aria-disabled</sref> <a>attribute</a> can be applied. For instance, it would be expected for <rref>textbox</rref>, <rref>combobox</rref>, or their host language equivalents to be exposed in the disabled state if they are descendants of an ancestor element with a specified <sref>aria-disabled</sref> attribute set to the "true" value. However, User Agents MUST NOT set a <rref>link</rref>, or its host language equivalent, to the disabled state if it is a descendant of an ancestor element with a specified <sref>aria-disabled</sref> attribute set to the "true" value</p>
```
cookiecrook, line 11332: Mostly editorial but IMO, normative statements should not follow inline behind a non-normative "for instance" example. I've added a line break, among another minor punctuation change (closing period) and a grammatical change ('on which').

```suggestion
				<p>The <a>state</a> of being disabled applies to the element on which the <sref>aria-disabled</sref> <a>attribute</a> is specified, and any descendants of that element which inherit from the <rref>input</rref> abstract role, to which the <sref>aria-disabled</sref> <a>attribute</a> can be applied. For instance, it would be expected for <rref>textbox</rref>, <rref>combobox</rref>, or their host language equivalents to be exposed in the disabled state if they are descendants of an ancestor element with a specified <sref>aria-disabled</sref> attribute set to the "true" value.</p>
				<p>User Agents MUST NOT set a <rref>link</rref>, or its host language equivalent, to the disabled state if it is a descendant of an ancestor element with a specified <sref>aria-disabled</sref> attribute set to the "true" value.</p>
```

cookiecrook (dismissed):

index.html
scottaohara: thank you

index.html
scottaohara: ah that might have come from me resolving the conflict in the github ui.  i'll see if i can't fix that.

scottaohara:
friendly ping @jnurthen and @a11ydoer (or @smhigley since you thumbs upped) to get 3 reviews so we can merge this

index.html
```diff
@@ -11329,8 +11329,9 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-                <p class="note">While <sref>aria-disabled</sref> and proper scripting can successfully disable an element with role <rref>link</rref>, fully disabling a host language equivalent can be problematic. Authors are advised not to use <sref>aria-disabled</sref> on elements that cannot be disabled through features of the host language alone.</p>
+				<p>The <a>state</a> of being disabled applies to the element on which the <sref>aria-disabled</sref> <a>attribute</a> is specified, and any descendants of that element which inherit from the <rref>input</rref> abstract role, to which the <sref>aria-disabled</sref> <a>attribute</a> can be applied. For instance, it would be expected for <rref>textbox</rref>, <rref>combobox</rref>, or their host language equivalents to be exposed in the disabled state if they are descendants of an ancestor element with a specified <sref>aria-disabled</sref> attribute set to the "true" value.</p>
+				<p>User Agents MUST NOT set a <rref>link</rref>, or its host language equivalent, to the disabled state if it is a descendant of an ancestor element with a specified <sref>aria-disabled</sref> attribute set to the "true" value.</p>
```
smhigley, line 11333: Small question: is `link` meant to be an example here, based on the previous line specifying only roles that inherit from `input`? It reads a bit confusingly to me right now, since `link` isn't the only non-input role that can be disabled. Maybe this can be combined with the earlier "For instance...", or get its own "Another example..." wording?

scottaohara: yup that makes sense.  that current wording isn't sufficient to catch everything that would need to inherit being disabled.

As there are far more roles which can be disabled, and not all of them inherit from `input`, maybe this is the wrong wording to go about this.  Maybe what's really needed here is just a clear indication that a `link` must not inherit the disabled state from an ancestor, and rather authors must explicitly specify a `role=link` as disabled so as to not accidently indicate nested links within a group are disabled, when really the author is only trying to bulk disable the form controls.  This would then match functional parity with HTML `<fieldset disabled>` - where only nested form controls are disabled, but not hyperlinks.

jnurthen:
Lets agenda+ this as I'm totally confused as to why we would want to prevent `role=link` from inheriting aria-disabled.

scottaohara:
@jnurthen just making a note that when this is discussed, another proposal is https://github.com/w3c/aria/issues/1632

scottaohara:
Resolution from today's meeting.

close this PR.  instead resolve to make a new PR that deprecates `aria-disabled` from `role=group`.

spectranaut:
Notes from meeting: https://www.w3.org/2022/08/11-aria-minutes#t07

PR #1468, opened 2021-04-22
Editorial: remove text that links nowhere

marcoscaceres:
Currently,  the `<a class="specref" href="#statevsprop">clarification of states versus properties</a>` doesn't link to anything in terms.html, so the link checker (and ReSpec) is complaining.

### DIFF
--- a/common/terms.html
+++ b/common/terms.html
@@ -150,7 +150,7 @@
     </dd>
     <dt><dfn data-lt="property|properties">Property</dfn></dt>
     <dd>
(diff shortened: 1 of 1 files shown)


jnurthen:
The terms are in the process of being moved to the relevant specs as requested. These links will make sense once this is done (the specific ones here are valid in the aria spec where they are now)
This can be closed but leaving here for now. 

If you cannot wait then a similar PR can be made against aria-common if you like.

marcoscaceres:
Thanks @jnurthen. The affected spec is [html-aam](https://github.com/w3c/html-aam/). Not sure if there is a pull request there for this.

jnurthen:
I'm working on PRs for the AAMs tomorrow.

jnurthen:
This can be closed as the sentence is now in the spec where the reference is valid.

PR #1465, opened 2021-04-21
Editorial: capitalize "true"

scottaohara:
all other instances are capital “True”, so updates this one to follow suit.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1465.html" title="Last updated on Apr 21, 2021, 2:44 PM UTC (4448d10)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1465/16dfd46...4448d10.html" title="Last updated on Apr 21, 2021, 2:44 PM UTC (4448d10)">Diff</a>

PR #1477, opened 2021-04-29
Revisions to accessible name required

### LINKED ISSUE #1466: Roles with required accessible names
When looking into issue #1421, it got me reviewing all the roles that can be named / require naming, and a few things stuck out:  I have broken down the topics of conversation by heading 2, with heading 3 being proposed actions.  My last topic, 'questions' has no actions at present.  All actions are of course my opinion and open to further discussion if people disagree.

## name true, false, n/a
In the characteristics table, "Accessible Name Required" is marked as "true", "false" and also not at all for some roles which can be named.  I would expect that it be "true" for all instances of roles which require a name.  False for all roles which can be named, but do not require a name.  Roles which prohibit naming would not need this table row

### action item
All roles which require names need to be marked as "Accessible name required: true".  All roles which can be named but do not require a name either need to be consistently marked as "Accessible name required: false" or not have that row displayed.

## landmarks SHOULD
For landmark roles only `form` and `region` require names, which makes sense because without them these roles won't generally be exposed. However, I think other landmark roles would benefit from a "authors SHOULD name..." for when there are multiple instances of landmarks in a document.

### action item
add an "authors SHOULD name..." sentence to landmark roles where more than one could exist in a document at a time.  

## roles which are missing name required
* `group` - won't be exposed otherwise, so needs it similarly to `role=region` and `role=form`
* `tab` - like other interactive controls, needs a name
* `form` - is marked as required, but it isn't marked as such under the listing in 5.2.8.4 Roles Supporting Name from Author - apparently my PR #1465 fixes this (thanks for the heads up @carmacleod)
* `separator` - if focusable, it would make sense to make sure it has a name

### action item
Mark these roles as naming required - or fix instances where they aren't referred to as such.

## roles which should not require a name
* `tooltip` - a tooltip should be represented by its contents / almost never interacted with directly. so... not convinced it should have a required name.
* `label` and `legend` - a `caption` is prohibited from being named.  these should be prohibited from being named too as they would perform similar functions.

### action item
Do not require names on these roles.  `label` and `legend` should be changed to name prohibited to match `caption`.

## questions
* `tree` and `treegrid` require a name, but other similar roles (of varying similarities mind you) `menu`, `menubar`, `toolbar`, `tablist`, and `navigation` do not. should `tree`/`treegrid` continue to require a name, or should some of these other roles require a name too?
* do `alertdialog`, `dialog`, `table`, `grid`, `tabpanel` need a quantifier to their required name status?  e.g., if these roles contained a heading, or were preceded by one, would that not provide the necessary context?  I think they all would benefit from having a name, but I'm a bit torn on that being a requirement, especially when host language equivalents do not have such requirements, which then leads to the author confusion as outlined in #1421.

scottaohara:
closes #1466

* Removes instances of "accessible name required: false" - only when accName is required will this row have a "True" value.
* Requires `tab` have an accessible name.
* Removes requirement that a `tooltip` needs an accessible name.

Note: #1465 is related to this issue and accepting that PR will mark `form` as being marked as requiring an accessible name in sections such as 5.2.8.4 Roles Supporting Name from Author

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1477.html" title="Last updated on May 18, 2021, 11:42 PM UTC (1e0c5ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1477/f6998c7...1e0c5ee.html" title="Last updated on May 18, 2021, 11:42 PM UTC (1e0c5ee)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3254,7 +3254,7 @@
 					</tr>
 					<tr>
 						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
-						<td class="role-namerequired">False</td>
+						<td class="role-namerequired"> </td>
 					</tr>
 					<tr>
 						<th class="role-namerequired-inherited-head" scope="row">Inherits Name Required:</th>
@@ -3429,7 +3429,7 @@
 					</tr>
 					<tr>
 						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
-						<td class="role-namerequired">False</td>
+						<td class="role-namerequired"> </td>
 					</tr>
 					<tr>
 						<th class="role-namerequired-inherited-head" scope="row">Inherits Name Required:</th>
@@ -3510,7 +3510,7 @@
 					</tr>
 					<tr>
 						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
-						<td class="role-namerequired">False</td>
+						<td class="role-namerequired"> </td>
 					</tr>
 					<tr>
 						<th class="role-namerequired-inherited-head" scope="row">Inherits Name Required:</th>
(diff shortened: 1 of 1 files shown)


charmarkk (approved):

carmacleod (approved):
+1
Totally agree with these changes.

shirsha (approved):

Issue #1288, opened 2020-06-16
Prohibit authors from causing conflict resolutions


WilcoFiers:
I was on an [ACT](https://act-rules.github.io/) related call yesterday, where there was pretty unanimous agreement that accessibility tools should flag occurrences of [presentational roles conflict resolution](https://w3c.github.io/aria/#conflict_resolution_presentation_none) as a best practice issue. Something has clearly gone wrong if you put role="none" and aria-label on the same `img` element. Right now it seems like ARIA condones this, and just relies on UA to address it. it seems to me that in addition to having the UAs address the problem, authors should be prohibited from doing this in the first place.

As a comparison, in HTML, browsers always use the first title element as the page title, but despite that it is still invalid HTML to have more than one title element. I think that idea, of both prohibiting an obviously bad practice, while still ensuring UA handle things consistently is a good one to follow here too.

pkra:
> Something has clearly gone wrong if you put role="none" and aria-label on the same img element. Right now it seems like ARIA condones this,

In the current draft, role=presentation lists aria-label under "Prohibited States and Properties": 

https://github.com/w3c/aria/blob/646f093848a1943c8343f34ad80f72dbd78f54e6/index.html#L6409-L6416

jnurthen:
closing as suggested

WilcoFiers:
I don't think this should have been closed. Yes, `aria-label` is now explicitly prohibited on role=presentation. But the problem is just as valid on other global ARIA attributes. There is nothing in ARIA that prohibits content authors from doing stuff like this, which causes `role=none` to be ignored:

```html
<h1 role="none" aria-hidden="false">
```

pkra:
@WilcoFiers your https://github.com/w3c/aria/issues/1489 was added to the v1.4 milestone on the last WG call (which is where most issues related to aria-hidden are right now).

joanmarie:
There's this example in our spec:
```
<!-- 1. [role="presentation"] is ignored due to the global aria-describedby property. -->
<h1 role="presentation" aria-describedby="comment-1"> Sample Content </h1>
```

I agree with @WilcoFiers that we don't want to encourage this. Setting `none`/`presentation` to having 0 properties in its characteristics table seems like it might be a way to discourage authors. Perhaps also we can add a note like, "If you make this focusable, you are a bad person." :grinning:

WilcoFiers:
I opened a PR that sort of fits with what I discussed with @jnurthen on Slack.

[...]

PR #1494, opened 2021-06-01
Prohibit presentation / none on where it causes conflict resolution

WilcoFiers:
Related to #1288

***
<a href="https://pr-preview.s3.amazonaws.com/WilcoFiers/aria/pull/1494.html" title="Last updated on Jun 1, 2021, 5:04 PM UTC (d85f7ad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1494/d494e74...WilcoFiers:d85f7ad.html" title="Last updated on Jun 1, 2021, 5:04 PM UTC (d85f7ad)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/WilcoFiers/aria/pull/1494.html" title="Last updated on Feb 22, 2024, 7:54 PM UTC (36dc9b8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1494/d494e74...WilcoFiers:36dc9b8.html" title="Last updated on Feb 22, 2024, 7:54 PM UTC (36dc9b8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -13843,7 +13843,7 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
 <span class="comment">&lt;!-- 2. [role="presentation"] negates both the implicit 'heading' and the non-global aria-level. --&gt;</span>
 &lt;h1 role="presentation" aria-level="2"&gt; Sample Content &lt;/h1&gt;
 		</pre>
-		<p>Authors MUST NOT use <rref>presentation</rref> and <rref>none</rref> on elements where users agents will ignore that role because it conflicts with one of the above items.</p>
+		<p>Authors MUST NOT use <rref>presentation</rref> and <rref>none</rref> on elements where user agents will ignore that role because it conflicts with one of the above items.</p>
 	</section>
 </section>
 <section id="idl-interface" class="normative">
--- a/index.html
+++ b/index.html
@@ -13843,6 +13843,7 @@ button.ariaPressed; // "undefined" (Note: button is no longer a toggle button.)<
 <span class="comment">&lt;!-- 2. [role="presentation"] negates both the implicit 'heading' and the non-global aria-level. --&gt;</span>
 &lt;h1 role="presentation" aria-level="2"&gt; Sample Content &lt;/h1&gt;
 		</pre>
+		<p>Authors MUST NOT use <rref>presentation</rref> and <rref>none</rref> on elements where users agents will ignore that role because it conflicts with one of the above items.</p>
 	</section>
 </section>
 <section id="idl-interface" class="normative">


joanmarie:
Thanks for doing this!

Some general thoughts:
1. I think that anything we want authors to be aware of shouldn't be down in the conflict resolution section; it should be with the associated state/property/role.
2. Anything we want authors to do should not require reading the conflict resolution section. I had to read it yesterday for a different issue. It's rather cumbersome content in parts.
3. I'm curious as to why you decided to prohibit the use of the role rather than prohibit the use of properties on that role. If, for instance `aria-label` is prohibited on `presentation`/`none`, why not also prohibit `aria-description`?

WilcoFiers:
@joanmarie I'll look into moving the requirement to a different place. Would you prefer putting it in the description of the `presentation` role? I wanted to keep it close to the list of scenarios where `presentation` needs to get ignored. There is no name for this list of items that cause conflict, making it difficult to reference. Do you have any suggestions on how best to do this?

As for why I'm suggesting to prohibit the role. The way I figure it, the ARIA spec can only prohibit using things that are in ARIA. It seems strange for ARIA to have a requirement that says authors must ensure elements with a presentation role are not focusable or otherwise interactive. I don't think we can even guarantee that's always possible. Is there some way to put role=none on a `summary` element and have it not cause conflict?

joanmarie:
Note that my replies are as me as an individual; I think the WG should discuss/review your PR as others might disagree with me and like your PR as-is. In the meantime, my answers:

> @joanmarie I'll look into moving the requirement to a different place. Would you prefer putting it in the description of the `presentation` role?

Yes.

> I wanted to keep it close to the list of scenarios where `presentation` needs to get ignored. There is no name for this list of items that cause conflict, making it difficult to reference. Do you have any suggestions on how best to do this?

I would argue that we don't even have a (concrete) list, let alone a named one. :slightly_smiling_face:  We have three bullets which describe stuff. For instance, in the third bullet it mentions global states and properties trumping `presentation`. But it doesn't list those global states and properties. So authors need to then go find a list. I think we should have a list of specific properties we don't want authors to use on something where they have also used `role="presentation"`. And I think we can do that via the characteristics table. See below.

That said, could you just tweak your existing language a tad, like:

"Authors MUST NOT use `presentation` and `none` on elements where users  agents will ignore that role as described in [Presentational Roles Conflict Resolution](https://w3c.github.io/aria/#conflict_resolution_presentation_none)."

> As for why I'm suggesting to prohibit the role. The way I figure it, the ARIA spec can only prohibit using things that are in ARIA. 

Ohhhh. Yeah, that's a good point. :smiley: Ok, what about a mixed approach:

1. In the characteristics table under `presentation`, move anything which would trump `presentation` from "Inherited States and Properties" to "Prohibited States and Properties". This would make it easier to find by authors and also not require them to figure out what is a global property.
2. In the description of `presentation`, maybe right above "For any element with a role of `presentation` and which is not focusable, the user agent MUST NOT expose the implicit native semantics of the element....", insert a new paragraph which states "Authors MUST NOT use `presentation` and `none` on an element which is focusable or otherwise interactive." (Wordsmith as you see fit.)

I'm not yet sure if we should consider doing something about the middle conflict-resolution bullet about required owned elements. But I'm leaning towards no. For instance, it's already an author error for an explicit `listitem` to not be inside a `list`.

> It seems strange for ARIA to have a requirement that says authors must ensure elements with a presentation role are not focusable or otherwise interactive.

Does the mixed approach above address your concern?

> I don't think we can even guarantee that's always possible. Is there some way to put role=none on a `summary` element and have it not cause conflict?

Is there a use case for that? Or is it merely a challenge? :grin:

WilcoFiers:
@joanmarie I've kind of been putting this off. It is rather daunting to make changes in this file. Adding in a line here or there I can manage, but I don't think I can make larger changes like what you're suggesting without it taking me half a day of work. If there's a better way to do this, please let me know. If not, I'd prefer to leave this up to someone who's comfortable working in this file.

index.html
```diff
@@ -13843,6 +13843,7 @@ <h3>Presentational Roles Conflict Resolution</h3>
 <span class="comment">&lt;!-- 2. [role="presentation"] negates both the implicit 'heading' and the non-global aria-level. --&gt;</span>
 &lt;h1 role="presentation" aria-level="2"&gt; Sample Content &lt;/h1&gt;
 		</pre>
+		<p>Authors MUST NOT use <rref>presentation</rref> and <rref>none</rref> on elements where users agents will ignore that role because it conflicts with one of the above items.</p>
```
spectranaut, line 13846:
```suggestion
		<p>Authors MUST NOT use <rref>presentation</rref> and <rref>none</rref> on elements where user agents will ignore that role because it conflicts with one of the above items.</p>
```

spectranaut:
Follow up issue: https://github.com/w3c/aria/issues/2126

spectranaut (approved):

spectranaut:
@Jaunita-George @MelSumner or @pkra, can one of you take a look so we can get this merged?

pkra:
@spectranaut I had meant to talk about this at the (canceled) editors call this week since Joanie had suggested a lot of additional work.

index.html
```diff
@@ -13843,6 +13843,7 @@ <h3>Presentational Roles Conflict Resolution</h3>
 <span class="comment">&lt;!-- 2. [role="presentation"] negates both the implicit 'heading' and the non-global aria-level. --&gt;</span>
 &lt;h1 role="presentation" aria-level="2"&gt; Sample Content &lt;/h1&gt;
 		</pre>
+		<p>Authors MUST NOT use <rref>presentation</rref> and <rref>none</rref> on elements where user agents will ignore that role because it conflicts with one of the above items.</p>
```
MelSumner, line 13846:
```suggestion
		<p>Authors MUST NOT use <rref>presentation</rref> and <rref>none</rref> on elements where user agents will ignore that role as described in [Presentational Roles Conflict Resolution](https://w3c.github.io/aria/#conflict_resolution_presentation_none).</p>
```

MelSumner (approved):
I'm suggesting the language change included in one of Joanie's comments, but I don't think it's a blocker.

index.html
jnurthen: it is already in this section - not sure we need a link to itself.

mookey86 (approved):

PR #1496, opened 2021-06-02
Do not restrict img naming to aria-label/labelledby

### LINKED ISSUE #1398: role=img should allow accName other than aria-label or aria-labelledby
Requested change:

I'm requesting `role=img` be edited to use @mcking65's phrase from pull request #785 comments: _authors MUST provide a label, which can be done with the aria-label or aria-labelledby attribute_

This was in response to @carmacleod's draft in issue #783 comments: _In order for elements with a role of img to be perceivable, authors MUST provide a label determined by the accessible name calculation._

Currently:

The ARIA spec for `role=img` currently says, "authors MUST provide a label using the aria-label or aria-labelledby attribute."

However, in some cases `role=img` can get its accessible name from a name from author mechanism, without actually using either `aria-label` or `aria-labelledby`. The ARIA spec should not prohibit these cases.

1. [ARIA in HTML](https://www.w3.org/TR/html-aria/#allowed-aria-roles-states-and-properties) allows `role=img` on an `<iframe>` element, which can get its accessible name from a `title` attribute.
2. `role=img` is "NOT RECOMMENDED" for an `` element, but a valid reason may exist to use it, such as @scottaohara's [workaround for a Safari bug](https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html#image-elements-with-svg-source-exception). In this case the `alt` attribute of an `` can still provide its accessible name.

scottaohara:
rewording of how a _name_ (not necessarily a “label” - so changed that word) can be specified for a `role=img`, so as to not imply that a host language naming mechinism would not be allowed. However, I’m wary of explicitly stating a host language naming mechanism could be used here, as I don’t think it’s necessary to do so in this spec.

closes #1398

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1496.html" title="Last updated on Jun 2, 2021, 2:49 PM UTC (9e7051d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1496/d494e74...9e7051d.html" title="Last updated on Jun 2, 2021, 2:49 PM UTC (9e7051d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4115,7 +4115,7 @@
 			<rdef>img</rdef>
 			<div class="role-description">
 				<p>A container for a collection of <a>elements</a> that form an image. See synonym <rref>image</rref>.</p>
-				<p>An <code>img</code> can contain captions and descriptive text, as well as multiple image files that when viewed together give the impression of a single image. An <code>img</code> represents a single graphic within a document, whether or not it is formed by a collection of drawing <a>objects</a>. In order for elements with a <a>role</a> of <code>img</code> to be <a>perceivable</a>, authors MUST provide a label using the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attribute.</p>
+				<p>An <code>img</code> can contain captions and descriptive text, as well as multiple image files that when viewed together give the impression of a single image. An <code>img</code> represents a single graphic within a document, whether or not it is formed by a collection of drawing <a>objects</a>. In order for an element with a <a>role</a> of <code>img</code> to be <a>perceivable</a>, authors MUST provide the element with an <a>accessible name</a>. This can be done using the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attribute.</p>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>


index.html
```diff
@@ -4115,7 +4115,7 @@ <h5>Note regarding the ARIA 1.3 <code>image</code> role.</h5>
...
 				<p>A container for a collection of <a>elements</a> that form an image. See synonym <rref>image</rref>.</p>
-				<p>An <code>img</code> can contain captions and descriptive text, as well as multiple image files that when viewed together give the impression of a single image. An <code>img</code> represents a single graphic within a document, whether or not it is formed by a collection of drawing <a>objects</a>. In order for elements with a <a>role</a> of <code>img</code> to be <a>perceivable</a>, authors MUST provide a label using the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attribute.</p>
+				<p>An <code>img</code> can contain captions and descriptive text, as well as multiple image files that when viewed together give the impression of a single image. An <code>img</code> represents a single graphic within a document, whether or not it is formed by a collection of drawing <a>objects</a>. In order for an element with a <a>role</a> of <code>img</code> to be <a>perceivable</a>, authors MUST provide the element with an <a>accessible name</a>. This can be done using the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attribute.</p>
```
joanmarie, line 4118: Do we want to say anything about native host language attributes? (e.g. `title`)

scottaohara: I thought about that, but shied away from the idea per my original PR comment.  Specifically didn't want to make someone think that the example we referenced was always the best way to add a name with the host language, and also didn't want to make someone think "oh, i could use the `alt` attribute here."

but if i'm being too wary and others don't think that's an issue, then yes. let's add it :)

MelSumner (approved):
Looks good! 👍

index.html
cookiecrook: I agree with @scottaohara on this point.

cookiecrook (approved):

PR #1497, opened 2021-06-02
Editorial: remove unnecessary role from example

scottaohara:
In section 4.2 there is a markup example which contains an `` with both an `alt=“”` and `role=presentation`. The role is unnecessary here so this PR removes it.  I do not want authors to think they need the role when they’ve already set the image to have no alternative text.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1497.html" title="Last updated on Jun 2, 2021, 3:02 PM UTC (f70420e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1497/d494e74...f70420e.html" title="Last updated on Jun 2, 2021, 3:02 PM UTC (f70420e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -527,7 +527,7 @@
 [aria-checked=&quot;true&quot;]::before { background-image: url(checked.gif); }</pre>
 		<p>If CSS is not used to toggle the visual representation of the check mark, the author could include additional markup and scripts to manage an image that represents whether or not the <rref>menuitemcheckbox</rref> is checked.</p>
 		<pre class="example highlight">&lt;li role=&quot;menuitemcheckbox&quot; aria-checked=&quot;true&quot;&gt;
-  &lt;img src=&quot;checked.gif&quot; role=&quot;presentation&quot; alt=&quot;&quot;&gt;
+  &lt;img src=&quot;checked.gif&quot; alt=&quot;&quot;&gt;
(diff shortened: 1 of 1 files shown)


joanmarie:
On the one hand, makes sense. Not just because of the explicitly empty `alt` but also because `menuitemcheckbox` is Children Presentational: True.

On the other hand, there seems to be a growing trend in which user agents are are ignoring -- or actively considering ignoring -- children-presentational and leaving it up to the ATs to handle it.

I just tried the test case in Firefox nightly. The image doesn't show up in the accessibility tree when role is set to presentation; when there is no role, the explictly-empty `alt` doesn't seem sufficient to prune it.

Maybe this is two issues: 

1. It's officially unnecessary so remove it from the example.
2. Our spec and reality are drifting further apart regarding children presentational and when images should be exposed.

MelSumner:
FWIW, it makes it easier to check this with a linter if there is `alt=""` AND `role="presentation"` or `role="none"`. It clarifies intent.

scottaohara:
@joanmarie would agree
1. it is officially unnecessary.
2. James Teh mentioned in the HTML AAM issue about the name calc for images that Firefox was actively doing something different here.  But chromium/webkit (at least when i checked this a few weeks ago) are in alignment with the spec and appear to be pruning `alt=""` images.  That is the expectation.

jnurthen (approved):

PR #1498, opened 2021-06-03
Draft: Add secondary action buttons as allowed child roles

### LINKED ISSUE #1440: Secondary actions on items in composite widget roles

smhigley:
Resolves #1440 

This is a proof of concept for allowing secondary actions in composite widget roles, and relies on the changes in #1454.

The relevant changes in this PR are these additional terms:
> **Controlled Element**
A 'controlled element' is any element referenced by `id` via `aria-controls`.
> 
> **Controlling Element**
A 'controlling element' is any element with an `aria-controls` attribute which references the ID of the element.

And this addition to the allowed child roles of `listbox`, as an example:
`<li><rref>button</rref> if <a>controlling</a> a child <rref>option</rref></li>`

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1498.html" title="Last updated on Jun 3, 2021, 10:56 AM UTC (8160130)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1498/d494e74...8160130.html" title="Last updated on Jun 3, 2021, 10:56 AM UTC (8160130)">Diff</a>

### DIFF
--- a/common/terms.html
+++ b/common/terms.html
@@ -53,9 +53,9 @@
     <dd>
       <p>In this specification, attribute is used as it is in markup languages. Attributes are structural features added to <a class="termref" href="#dfn-element">elements</a> to provide information about the <a class="termref" href="#dfn-state">states</a> and <a class="termref" href="#dfn-property">properties</a> of the <a class="termref" href="#dfn-object">object</a> represented by the element.</p>
     </dd>
-    <dt><dfn data-lt="child element|child|children|child elements">Child Element</dfn></dt>
+    <dt><dfn data-lt="owned child|child element|child|children|child elements">Owned Child</dfn></dt>
     <dd>
-      <p>An 'child element' is a direct <abbr title="Document Object Model">DOM</abbr> child of the <a>element</a>, an descendant of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, or a child specified via <pref>aria-owns</pref> on the element.</p>
+      <p>An 'owned child' is a direct <abbr title="Document Object Model">DOM</abbr> child of the <a>element</a>, a descendant of the <a>element</a> with only <rref>generic</rref> or <rref>presentation</rref> elements intervening, or a child specified via an <pref>aria-owns</pref> relationship to the element.</p>
     </dd>
     <dt><dfn data-lt="class|classes">Class</dfn></dt>
     <dd>
--- a/common/terms.html
+++ b/common/terms.html
@@ -53,9 +53,21 @@
     <dd>
       <p>In this specification, attribute is used as it is in markup languages. Attributes are structural features added to <a class="termref" href="#dfn-element">elements</a> to provide information about the <a class="termref" href="#dfn-state">states</a> and <a class="termref" href="#dfn-property">properties</a> of the <a class="termref" href="#dfn-object">object</a> represented by the element.</p>
     </dd>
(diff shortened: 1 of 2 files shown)


JAWS-test:
I don't understand the concept yet: the button should be on the same level in the DOM as option (as a child of listbox)? I assumed so far that the buttons should be child elements of option and the listbox contains only options. 
For me the following questions would be open:
* How is a listbox with option and buttons operated with the keyboard?
* Can the Accessbility APIs handle buttons inside listbox at all?

smhigley:
Test page for support of nested vs. sibling action buttons is here: https://github.com/w3c/aria/issues/1440
(cc/ @joanmarie)

smhigley:
Closing in favor of aria-actions (https://github.com/w3c/aria/pull/1805/)

PR #1503, opened 2021-06-24
Initial creation of validator tests repo

joanmarie:
Adds:
* Documentation
* A test for name prohibited
* @axe-core/cli results for the name-prohibited test


joanmarie:
@jnurthen I'm not sure if we'll need to do anything else to get these to show up on w3c.github.io/aria/validator-tests (e.g. cause them to be included in stuff that gets copied to ARIA's gh-pages branch).

PR #1508, opened 2021-07-01
Validator Tests

design1online:
* Abstract Roles
* Must Have Owned Elements

### DIFF
--- a/validator-tests/must-have-owned-elements.html
+++ b/validator-tests/must-have-owned-elements.html
@@ -10,9 +10,12 @@
     * example, until a page is fully initialized and complete, an author could
     * mark the document element as busy."
     -->
-    <div id="while-loading" aria-label="while-loading" role="list">
-      <div class="loader" aria-busy="true">Loading List...</div>
+    <div id="while-loading" aria-label="while-loading" role="list" aria-busy="true"></div>
+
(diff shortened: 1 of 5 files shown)


validator-tests/must-have-owned-elements.html
```diff
@@ -0,0 +1,21 @@
...
+    <!--
+    URL: https://www.w3.org/TR/wai-aria-1.2/#mustContain
+    RULE: "There may be times that required owned elements are missing,
+    * for example, while editing or while loading a data set. When a widget is
+    * missing required owned elements due to script execution or loading,
+    * authors MUST mark a containing element with aria-busy equal to true. For
+    * example, until a page is fully initialized and complete, an author could
+    * mark the document element as busy."
+    -->
+    <div id="while-loading" aria-label="while-loading" role="list">
```
jnurthen, line 13: aria-busy should be on the parent (while-loading) in this case.

design1online: Oh interesting. I assumed that the busy should be on the required child element since it's in the required child owned elements. Does that mean in the second case it should have no child elements?

design1online: Updated to add aria-busy to the parent element

jnurthen:
closes #1543 
closes #1519

PR #1583, opened 2021-07-27
added validation text for heading role to have a valid aria-lavel at…

jongund:
added validation text for `heading` role to have a valid `aria-lavel` attribute

### DIFF
--- a/validator-tests/heading-role-must-have-level.html
+++ b/validator-tests/heading-role-must-have-level.html
@@ -10,8 +10,6 @@ RULE: "Authors MUST use the aria-level attribute to indicate the proper nesting 
 <!-- Heading element that should fail -->
 <div id="heading-1" role="heading" class="fail"></div>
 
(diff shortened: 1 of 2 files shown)


validator-tests/heading-role-must-have-level.html
```diff
@@ -0,0 +1,27 @@
...
+
+<!--
+URL: https://www.w3.org/TR/wai-aria-1.2/#heading
+RULE: "Authors MUST use the aria-level attribute to indicate the proper nesting level."
+-->
+
+<!-- Heading element that should fail -->
+<div id="heading-1" role="heading" class="fail"></div>
+
+<div id="heading-2" role="heading" class="fail" aria-level="0"></div>
```
jnurthen, line 13: While I agree that this is not valid it doesn't fail this particular rule. Can you remove this one?

jnurthen (approved):

Issue #1566, opened 2021-07-22
Author test case: group role (ARIA 1.2)


cyns:
group role
However, when a group is used in the context of a listbox, authors MUST limit its children to option elements.

spectranaut:
isn't this "must" just ONE of the cases that falls under this "must"? https://github.com/w3c/aria/issues/1522

spectranaut:
Oh no! I wrote up a test case for this AND THEN notice @jongund had a PR opened for it :'( @jongund can you assign yourself to issues that you are working on? I see you somehow linked the PR to this issue but I don't see the PR from this issue...

spectranaut:
Oh wait! Now I see that PR #1578 is slightly different! its about "option" being in a "group" or "listbox", as opposed to this issue, which is about "group" necessarily only having "option" child elements. I'll see if there is actually any test cases to write that are separate from what @jongund wrote.

Sorry for the confusion!

jongund:
@spectranaut 
I am just making the test case html files for some of the ARIA 1.2 authoring requirement.  I couldn't get axe installed so I was not able to generate the `json` results files.  I was waiting for some feedback from James or Joanie before making anymore.
So I not sure what the status of my pull requests are, so feel free to to work on them, especially generating the `json` files.
What is your role?

spectranaut:
hi @jongund I actually do have axe running so I'll try adding the json to this PR! I just started working at igalia on accessibility related work :)

jongund:
@spectranaut 
Good news, you will be a great addition to the work on ARIA specs!
I assume you are working with Joanie, and therefore probably have more information on the work plan for the authoring requirement tests, so let me know how I can help.

spectranaut:
Thanks @jongund I'm glad to be helping here and glad to see your name again! And I am working with Joanie :)

PR #1586, opened 2021-08-02
Add validation for children of groups in listbox context

spectranaut:
Add author test case for [group role](https://www.w3.org/TR/wai-aria-1.2/#group): `However, when a group is used in the context of a listbox, authors MUST limit its children to option elements.`

See Issue  #1566

### DIFF
--- a/validator-tests/README.md
+++ b/validator-tests/README.md
@@ -12,6 +12,8 @@ corresponding changes in validators.
 * [Dialog Must Have Name](dialog-must-have-name.html)
   * @axe-core/cli: [Results](dialog-must-have-name-axe.json)
 * [Img Role Must Have Name](img-role-must-have-name.html)
+* [Listbox Group Children Must Be Option](listbox-group-children-must-be-option.html)
+  * @axe-core/cli: [Results](listbox-group-children-must-be-option-axe.json), [bug](https://github.com/dequelabs/axe-core-npm/issues/313)
 * [Menuitemcheckbox Owned By Menu](menuitemcheckbox-owned-by-menu.html)
 * [Menuitem Owned By Menu](menuitem-owned-by-menu.html)
 * [Menuitemradio Owned By Menu](menuitemradio-owned-by-menu.html)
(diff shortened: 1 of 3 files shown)


spectranaut:
Below are the AXE results, which I believe indicate a bug in AXE?

First bug: All 8 examples (both the "pass" and "fail") have `Violation of "aria-required-children"`, but these violations are all on the listbox element. I believe the violation is because an element with role `group` is the child of an element with role `listbox`, which I also believe should *not* be a violation, as you can see in [listbox with group Aria Practices example](https://w3c.github.io/aria-practices/examples/listbox/listbox-grouped.html). If you run the browser extension AXE on this page, it finds the same error with the listbox element.

Second bug: There are no violations on the group elements which do not have the correct children.

Third bug: Additionally, there are violations for every use of `role="option"`, when it should be valid for `role="option"` to have the parent `role="group"` in the context of a listbox. If I understand the [spec on option](https://www.w3.org/TR/wai-aria-1.2/#option) correctly. Maybe this bug shouldn't be considered here though, because this requirement is about children, not parents?

```
~/bin/local-axe$ npx axe http://localhost:8000/listbox-group-children-must-be-option.html --tags cat.aria
Running axe-core 4.3.2 in chrome-headless

Testing http://localhost:8000/listbox-group-children-must-be-option.html ... please wait, this may take a minute.

  Violation of "aria-required-children" with 8 occurrences!
    Ensures elements with an ARIA role that require child roles contain them. Correct invalid elements at:
     - div[role="listbox"][aria-label="foo"]:nth-child(1)
     - div[role="listbox"][aria-label="foo"]:nth-child(2)
     - div[role="listbox"][aria-label="foo"]:nth-child(3)
     - div[role="listbox"][aria-label="foo"]:nth-child(4)
     - div[role="listbox"][aria-label="foo"]:nth-child(5)
     - div[role="listbox"][aria-label="foo"]:nth-child(6)
     - div[role="listbox"][aria-label="foo"]:nth-child(7)
     - body > ul
    For details, see: https://dequeuniversity.com/rules/axe/4.3/aria-required-children

  Violation of "aria-required-parent" with 5 occurrences!
    Ensures elements with an ARIA role that require parent roles are contained by them. Correct invalid elements at:
     - #listbox-group-2 > div[role="option"][aria-label="bar"]
     - #listbox-group-4 > ul > li
     - #listbox-group-6 > div[role="option"][aria-label="bar"]
     - #listbox-group-7 > li
     - #listbox-group-8 > li
    For details, see: https://dequeuniversity.com/rules/axe/4.3/aria-required-parent
```

spectranaut:
Ok I added [the bug on AXE](https://github.com/dequelabs/axe-core-npm/issues/313)

And this is ready for review!

jongund (approved):
LGTM

Issue #1576, opened 2021-07-22
Author test case: aria-errormessage


cyns:
aria-errormessage
Similarly, when aria-errormessage is not pertinent, authors MUST either ensure the content is hidden or remove the aria-errormessage attribute or its value.

PR #1588, opened 2021-08-03
aria-errormessage is hidden or removed when not pertinent 

jongund:


spectranaut:
I just fixed the conflict in the README and added a json file for the AXE results. It seems to me there is a bug in AXE, but I'm having trouble reporting it because I'm not sure I understand it. 

Here are the results:
```
```

spectranaut (approved):
I approve this PR but I also added a AXE file, so @joanmarie or @schne324  should take a look too.

validator-tests/errormessage-hidden-removed.html
```diff
@@ -0,0 +1,40 @@
...
+<input type="text" id="error-message-hidden-1" aria-invalid="true" aria-errormessage="id-message-1" class="fail"></div>
+<div id="id-message-1" hidden>Error message 1</div>
+
+<input type="text" id="error-message-hidden-2" aria-invalid="true" aria-errormessage="id-message-2" class="fail"></div>
+<div id="id-message-2" style="display: none">Error message 2</div>
+
+<input type="text" id="error-message-hidden-3" aria-invalid="true" aria-errormessage="id-message-3" class="fail"></div>
+<div id="id-message-3" style="visibility: hidden">Error message 3</div>
+
+<input type="text" id="error-message-empty-1" aria-invalid="true" aria-errormessage="" class="fail"></div>
```
jnurthen, line 21: How does this fail this specific statement? IMO this is equivalent to aria-errormessage not being present. I don't think this should fail this statement

jnurthen: From the spec "Sometimes states and properties are present in the DOM but have a zero-length string ("") as their value. Authors MAY specify a zero-length string ("") for any supported (but not required) state or property. User agents SHOULD treat state and property attributes with a value of "" the same as they treat an absent attribute."

pkra:
ping @joanmarie @jnurthen 

This testcases resolves #1576 (which is technically in the 1.2 milestone).

validator-tests/errormessage-hidden-removed-axe.json
```diff
@@ -0,0 +1,100 @@
+[
+  {
+    "description": "Ensures all ARIA attributes have valid values",
+    "help": "ARIA attributes must conform to valid values",
+    "helpUrl": "https://dequeuniversity.com/rules/axe/4.3/aria-valid-attr-value?application=webdriverjs",
```
MelSumner, line 5: Question, non-blocker: Is the query param needed here? It seems like the URL is valid without it. Just curious.

validator-tests/errormessage-hidden-removed.html
```diff
@@ -0,0 +1,37 @@
+<html lang="en-US">
+<head><title>When aria-errormessage is pertinent, authors MUST ensure the content is not hidden so users can navigate to and examine the error message</title></head>
+<body>
+
+<!--
+URL: https://www.w3.org/TR/wai-aria-1.2/#aria-errormessage
+RULE: " Authors MUST use aria-invalid in conjunction with aria-errormessage and when aria-errormessage is pertinent, authors MUST ensure the content is not hidden so users can navigate to and examine the error message"
```
MelSumner, line 7: Question, potential blocker: Should an example be included that has no `aria-invalid` attribute? 

From the spec, emphasis mine:

> Initially, the object is in a valid state and either has aria-invalid set to false **or no aria-invalid attribute**, and the element referenced by aria-errormessage is not applicable.

validator-tests/errormessage-hidden-removed-axe.json
jnurthen: This is the file retrieved from aXe. I don't think we should be changing it.

validator-tests/errormessage-hidden-removed.html
jnurthen: I think we would if this were a test file for all of aria-errormessage. 

However this is only a testfile for the added authors MUST statements in aria 1.2. Without aria-invalid, aria-errormessage is not pertinent so the entire sentence would not apply.

pkra:
@jnurthen can we merge this? It's actually closing the last open issue for the 1.2 milestone.

schne324 (approved):
looks good to me.

@spectranaut, you probably have already seen this but it looks like the underlying axe-core issue you raised has been resolved: https://github.com/dequelabs/axe-core/issues/3149

PR #1589, opened 2021-08-03
added validation test for `form` role having accessible name

jongund:


spectranaut:
hey @jongund , can you add a line to the README file [here](https://github.com/w3c/aria/blob/main/validator-tests/README.md#aria-12) when you add a new validator test? I checked with @joanmarie that's what we should do :)

spectranaut:
Sorry @jongund -- you are going to have to rebase because added all of those other links to the README.md file so there is a conflict. Notice I put all the files in alphabetical order, so that there will be less conflicts when multiple people are adding tests at the same time, and so that it will be easier to read!

spectranaut (approved):
Looks good! But, again, I added AXE files and an AXE bug, if @joanmarie or @schne324 could review :)

joanmarie:
It also doesn't seem consistent with the guidelines in the readme. If the group doesn't like those guidelines, so be it. But then let's change the readme rather than have inconsistent tests.

What was done by Jon in slider-role-aria-valuenow.html follows the current guidelines. This test not as much:

* The headings don't need to be in there, do they?
* The thing being tested along with the url pointing to the relevant spec text is not in a comment
* Nit: Personally I'd prefer to keep the content being tested simple as well.
* Each item (even if there is only one) should have a unique id so that we can immediately (visually, and potentially programmatically in a to-be-created parser) identify it in the json file (e.g. the "target" for axe) without having to read the markup.

joanmarie (approved):

PR #1595, opened 2021-08-12
added validation test for aria-roledescription on generic

jongund:
Added validation test for roles with prohibited `role-description` property for the `generic` role.


joanmarie:
Isn't this redundant to https://github.com/w3c/aria/blob/main/validator-tests/name-prohibited.html

jongund:
@joanmarie 
I guess it is, but `aria-label` and `aria-labelledby` are the only prohibited properties in the spec.  Is a better definition for this issue "deprecated properties and states"?  If so that would generate a different set of test cases.

joanmarie:
aria-roledescription is prohibited on the generic role. There may be other prohibited things besides name.

spectranaut (approved):
I approve but again I added AXE results, so @joanmarie or @schne324 should take a look too.

I'm not sure, but I think one of the AXE errors could be wrong and worth a bug report -- and think the AXE results indicate "generic" is an incorrect role value?

joanmarie:
It also doesn't seem consistent with the guidelines in the readme. If the group doesn't like those guidelines, so be it. But then let's change the readme rather than have inconsistent tests.

What was done by Jon in slider-role-aria-valuenow.html follows the current guidelines. This test not as much:

* The headings don't need to be in there, do they?
* The thing being tested along with the url pointing to the relevant spec text is not in a comment
* Nit: Personally I'd prefer to keep the content being tested simple as well.
* Each item (even if there is only one) should have a unique id so that we can immediately (visually, and potentially programmatically in a to-be-created parser) identify it in the json file (e.g. the "target" for axe) without having to read the markup.

joanmarie (changes requested):
See my comment about consistency with other tests and the guidelines in the readme.

jongund:
@joanmarie 
updated file names and the file format to remove the H1 element.

PR #1594, opened 2021-08-05
Add AXE output files for all validation tests, fix bugs and file issues

spectranaut:
Add the AXE tests for the following files:

* [Img Role Must Have Name](https://github.com/w3c/aria/blob/main/validator-tests/img-role-must-have-name.html)
* [Menuitemcheckbox Owned By Menu](https://github.com/w3c/aria/blob/main/validator-tests/menuitemcheckbox-owned-by-menu.html)
* [Menuitem Owned By Menu](https://github.com/w3c/aria/blob/main/validator-tests/menuitem-owned-by-menu.html)
* [Menuitemradio Owned By Menu](https://github.com/w3c/aria/blob/main/validator-tests/menuitemradio-owned-by-menu.html)
* [Option Owned By Listbox](https://github.com/w3c/aria/blob/main/validator-tests/option-owned-by-listbox.html)
* [Row Must Not in Table Grid](https://github.com/w3c/aria/blob/main/validator-tests/row-must-not-in-table-grid.html)
* [Scrollbar Role Aria Controls](https://github.com/w3c/aria/blob/main/validator-tests/scrollbar-role-aria-controls.html)
* [Scrollbar Role Aria Valuenow](https://github.com/w3c/aria/blob/main/validator-tests/scrollbar-role-aria-valuenow.html)
* [Slide Role Aria Valuenow](https://github.com/w3c/aria/blob/main/validator-tests/slider-role-aria-valuenow.html)

This included fixing some files to limit ARIA errors, updating the README.md and filing the following bug reports on AXE:
* [Bug: axe-core/cli validator does not report missing "aria-controls" attribute from element with role "scrollbar" (ARIA 1.2)](https://github.com/dequelabs/axe-core-npm/issues/318)
* [Bug: axe-core/cli validator does not report error incorrect use of aria attributes on element with role="row" (ARIA 1.2)](https://github.com/dequelabs/axe-core-npm/issues/317)
* [Bug: axe-core/cli validator does not give warning about lack of aria-label or aria-labelledby on role="img" element(ARIA 1.2 )](https://github.com/dequelabs/axe-core-npm/issues/316)

Issue #1593, opened 2021-08-03
Upload remining AXE json result files for validator tests


spectranaut:
@jongund wrote a lot of validator tests but did not upload AXE json result files for them, we should add them! :) The following files need results:

Val is working on:
* [Img Role Must Have Name](https://github.com/w3c/aria/blob/main/validator-tests/img-role-must-have-name.html)
* [Menuitemcheckbox Owned By Menu](https://github.com/w3c/aria/blob/main/validator-tests/menuitemcheckbox-owned-by-menu.html)
* [Menuitem Owned By Menu](https://github.com/w3c/aria/blob/main/validator-tests/menuitem-owned-by-menu.html)
* [Menuitemradio Owned By Menu](https://github.com/w3c/aria/blob/main/validator-tests/menuitemradio-owned-by-menu.html)

Unclamed:
* [Option Owned By Listbox](https://github.com/w3c/aria/blob/main/validator-tests/option-owned-by-listbox.html)
* [Row Must Not in Table Grid](https://github.com/w3c/aria/blob/main/validator-tests/row-must-not-in-table-grid.html)
* [Scrollbar Role Aria Controls](https://github.com/w3c/aria/blob/main/validator-tests/scrollbar-role-aria-controls.html)
* [Scrollbar Role Aria Valuenow](https://github.com/w3c/aria/blob/main/validator-tests/scrollbar-role-aria-valuenow.html)
* [Slide Role Aria Valuenow](https://github.com/w3c/aria/blob/main/validator-tests/slider-role-aria-valuenow.html)

spectranaut:
@schne324 I think in the ARIA working group meeting you volunteered to upload some json files? Are you working on that already? I'm happy to also work on this if you'd like to split the list with me!

It can take a little work because:
1. You might have to fix the test file if you get an error about something unrelated (like a missing accessible name)
2. You want to figure out if there is an appropriate AXE error and if there is, produce the json with `--rules <appropriate rule>`. If it seems like there isn't an appropriate rule or AXE hasn't implemented the validation, I just produce the json with `--tags cat.aria`.
2. If AXE doesn't catch the validation error, then you need to write a bug report.

spectranaut:
I'll start working on the first four!

spectranaut:
Actually it's pretty easy so I'm going to do them all!

PR #1597, opened 2021-08-12
Adds validator-tests‘ axe-core results

schne324:


spectranaut:
Hi @schne324! I'm sorry you dropped the meeting right before I could catch you, I did produce this files in PR #1594 last week so I'm going to close the PR as a duplicate for #1594. I tried making sure we didn't do duplicate work by opening issue #1593, but unfortunately I think you missed it!

For the future, when adding the AXE files, please update the [README](https://github.com/w3c/aria/blob/main/validator-tests/README.md) with links to the new results, as you can see demonstrated in PR #1594.

Additionally, it looks like your files are different than my files, and you are the AXE expert between us. Can you review #1594, and the [issues I opened on axe core](https://github.com/dequelabs/axe-core-npm/issues/created_by/spectranaut) ?

PR #1606, opened 2021-08-25
Add validator test results for validator.nu

spectranaut:
We had some discussion in the working group about whether to do this. There is some ARIA validation that happens in the validator.nu parser, so it seems like we should produce validator.nu json and file bugs against the project.

I want to hear from others first, before I begin to file bugs. Should they be filed against https://github.com/validator/validator ?


spectranaut:
@jnurthen thanks for merging all the other PRs! This one has been rebased and doctype added everywhere.

PR #1624, opened 2021-10-04
Remove bad license

marcoscaceres:
"document" is not a valid license. ReSpec automatically defaults to "w3c-software-doc", which is what we want here.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1624.html" title="Last updated on Oct 5, 2021, 12:00 AM UTC (3a36195)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1624/0d06a71...3a36195.html" title="Last updated on Oct 5, 2021, 12:00 AM UTC (3a36195)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -23,7 +23,6 @@
 		// if you wish the publication date to be other than today, set this
 		//publishDate:  "2014-12-11",
 		copyrightStart:  "2013",
-		license: "document",
 
 		// if there is a previously published draft, uncomment this and set its YYYY-MM-DD date
 		// and its maturity status


jnurthen:
@marcoscaceres we are chartered to publish under https://www.w3.org/Consortium/Legal/2015/doc-license
How do we specify this?

jnurthen (changes requested):
We are chartered under 
https://www.w3.org/Consortium/Legal/2015/doc-license

This is not the same as the software and document license. Please advise

marcoscaceres:
The “document” license was deprecated with this new one a while ago. This one does what the old document did, plus more.

marcoscaceres:
Oh, I see, that’s weird? Why is the working group using that license? That’s out of step with the rest of the w3c?

jnurthen:
> Oh, I see, that’s weird? Why is the working group using that license? That’s out of step with the rest of the w3c?

Its in our charter. it is also in the AG charter and they will need this too.
Why was this changed? respec should support all the W3C licenses which are still valid.

marcoscaceres:
> Why was this changed?

I’ll track down the history, but my recollection is that we really don’t want people using that old license for the reasons the new software-doc license was created to supersede it.

jnurthen:
> > Why was this changed?
> 
> I’ll track down the history, but my recollection is that we really don’t want people using that old license for the reasons the new software-doc license was created to supersede it.

I have no problems changing at our charter renewal (happening now btw and hadn't received this as a comment from anyone) but folks can't change something like this without rechartering so can't really change it out from under them.

marcoscaceres:
Ok, cool. I added comment to https://github.com/w3c/strategy/issues/282#issuecomment-935408044 

I'll update ReSpec to silently add "document", but maybe show a warning to consider switching to the more permissive license.

PR #1633, opened 2021-11-02
updates per discussion in #1598

### LINKED ISSUE #1598: Updating ARIA 1.2 due to IDL implementations (exit and re-enter CR?)
As described in https://github.com/w3c/aria/issues/1585#issuecomment-893892641 as well as in ARIA WG meetings [1](https://www.w3.org/2021/07/29-aria-minutes.html#t08) and [2](https://www.w3.org/2021/08/12-aria-minutes.html#t06), what implementors have done for ARIA reflection does not appear to match what is in the spec.

In particular, default/implicit values for missing attributes are not being reflected, and no validation is taking place. When an attribute is not present in the DOM, the reflected value is null. Thus as it currently stands, I believe we have 0 implementations for the IDL interface as spec'ed in ARIA 1.2.

What is currently implemented seems consistent with the properties being a nullable `DOMString`. Therefore, unless there is a strong belief that what is in the spec now (attributes are a non-nullable `DOMString`) is correct and that the implementations are wrong, the thing to do might be to update ARIA 1.2 to reflect (pun intended) reality.

jnurthen:
resolves #1598

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1633.html" title="Last updated on Nov 11, 2021, 9:53 PM UTC (a7a5e2a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1633/0d06a71...a7a5e2a.html" title="Last updated on Nov 11, 2021, 9:53 PM UTC (a7a5e2a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -10246,25 +10246,18 @@
 			[[[#supportedState]]], and in [[[CORE-AAM]]].</p>
 		</section>
 		<section id="enumerated-attribute-values-html">
-			<h3>ARIA Enumerated Attributes in HTML</h3>
+			<h3>ARIA nullable DOMString Attributes in HTML</h3>
 			<p>As noted in [[[#typemapping]]], attributes are included in host languages, and the syntax for representation of enumerated value types is governed by the host language.</p>
-			<p>When the host language is HTML, ARIA enumerated attributes MUST follow the <a data-cite="html/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes">HTML rules</a> for reflecting [=nullable type|nullable=] {{DOMString}} IDL attributes current as of September 3, 2021.</p>
-			<div class="note informative" data-cite="HTML">
-				<p>Below is a copy of the HTML specification text for [=nullable type|nullable=] {{DOMString}} IDL attributes as of September 3, 2021. 
-				The HTML spec is a living document, and this section may have changed since. Refer to the link above for the most current version.</p>
-				
-				<blockquote>
-					a reflecting IDL attribute is a nullable {{DOMString}}
-					attribute whose content attribute is an <a data-cite="html/common-microsyntaxes.html#enumerated-attribute">enumerated attribute</a>, then, on getting, if
-					the corresponding content attribute is in its <i>missing value default</i> state then the IDL
-					attribute must return null, otherwise, the IDL attribute must return the keyword value associated
-					with the state the attribute is in. If there are multiple keyword values for the state, then
-					return the conforming one. If there are multiple conforming keyword values, then one will be
-					designated the <a data-cite="html/common-dom-interfaces.html#canonical-keyword">canonical keyword</a>; choose that one. On setting, if the new value is
-					null, the content attribute must be removed, and otherwise, the content attribute must be set to
-					the specified new value.
-				</blockquote>
-			</div>
+			<p>The following algorithm should be used for nullable {{DOMString}} attributes in HTML:</p>
+			<p>If a reflecting IDL attribute is a nullable {{DOMString}} attribute whose content attribute is not an <a data-cite="html/common-microsyntaxes.html#enumerated-attribute">enumerated attribute</a>,
+			then, on getting, if the corresponding content attribute is not present, then the IDL attribute must return null, 
+			otherwise, the IDL attribute must get the value in a transparent, case-preserving manner. 
+			On setting, if the new value is null, the content attribute must be removed, and otherwise, 
+			the content attribute must be set to the specified new value in a transparent, case-preserving manner.</p>
+			<p class="note">
+			Note: As of ARIA 1.2, all ARIA attributes exposed via IDL are defined as nullable {{DOMStrings}}. This matches the current implementation of all major rendering engines, and has for some time. This specification change should result in no implementation changes; it will merely represent the current reality of web engines.
+			However, in a future draft, the ARIA Working Group intends to change several ARIA enumerated attributes to non-nullable DOMStrings, and seek implementations. The proposed change will bring ARIA into alignment with the HTML’s usage of enumerated attributes.
+			</p>
 			<section class="informative" id="enumeration-example">
 				<h4>Example Enumerated Attribute Usage</h4>
 				<aside class="example">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -10246,25 +10246,18 @@ <h3>Operating System Accessibility API mapping of ARIA enumerated attributes</h3
 			[[[#supportedState]]], and in [[[CORE-AAM]]].</p>
 		</section>
 		<section id="enumerated-attribute-values-html">
-			<h3>ARIA Enumerated Attributes in HTML</h3>
+			<h3>ARIA nullable DOMString Attributes in HTML</h3>
```
cookiecrook, line 10249: Does this need to be limited to "in HTML"? HTML includes all the common use cases I can think of offhand (SVG, MathML, EPUB) but I don't know of a reason to limit this to just HTML and other technologies incorporated into HTML.

index.html
```diff
@@ -10246,25 +10246,18 @@ <h3>Operating System Accessibility API mapping of ARIA enumerated attributes</h3
...
+			then, on getting, if the corresponding content attribute is not present, then the IDL attribute must return null, 
+			otherwise, the IDL attribute must get the value in a transparent, case-preserving manner. 
+			On setting, if the new value is null, the content attribute must be removed, and otherwise, 
+			the content attribute must be set to the specified new value in a transparent, case-preserving manner.</p>
+			<p class="note">
+			Note: As of ARIA 1.2, all ARIA attributes exposed via IDL are defined as nullable {{DOMStrings}}. This matches the current implementation of all major rendering engines, and has for some time. This specification change should result in no implementation changes; it will merely represent the current reality of web engines.
+			However, in a future draft, the ARIA Working Group intends to change several ARIA enumerated attributes to non-nullable DOMStrings, and seek implementations. The proposed change will bring ARIA into alignment with the HTML’s usage of enumerated attributes.
```
cookiecrook, line 10259:
```suggestion
			However, in a future draft, the ARIA Working Group intends to change several ARIA enumerated attributes to non-nullable DOMStrings, and seek implementations. The proposed change will bring ARIA into alignment with the HTML’s usage of <a data-cite="html/common-microsyntaxes.html#enumerated-attribute">enumerated attributes</a>.
```

index.html
```diff
@@ -10246,25 +10246,18 @@ <h3>Operating System Accessibility API mapping of ARIA enumerated attributes</h3
...
-				</blockquote>
-			</div>
+			<p>The following algorithm should be used for nullable {{DOMString}} attributes in HTML:</p>
+			<p>If a reflecting IDL attribute is a nullable {{DOMString}} attribute whose content attribute is not an <a data-cite="html/common-microsyntaxes.html#enumerated-attribute">enumerated attribute</a>,
+			then, on getting, if the corresponding content attribute is not present, then the IDL attribute must return null, 
+			otherwise, the IDL attribute must get the value in a transparent, case-preserving manner. 
+			On setting, if the new value is null, the content attribute must be removed, and otherwise, 
+			the content attribute must be set to the specified new value in a transparent, case-preserving manner.</p>
+			<p class="note">
+			Note: As of ARIA 1.2, all ARIA attributes exposed via IDL are defined as nullable {{DOMStrings}}. This matches the current implementation of all major rendering engines, and has for some time. This specification change should result in no implementation changes; it will merely represent the current reality of web engines.
```
cookiecrook, line 10258:
```suggestion
			Note: As of ARIA 1.2, all ARIA attributes exposed via IDL are defined as nullable {{DOMStrings}}. This matches the current implementation of all major rendering engines. This specification change should result in no implementation changes; it will merely represent the current reality of web engines.
```

index.html
cookiecrook, line 10249:
```suggestion
			<h3>ARIA nullable DOMString Attributes</h3>
```

cookiecrook (approved):

jnurthen:
@cookiecrook @aleventhal @cyns @domenic @annevk 
Can folks please review this and let us know if this will work.

domenic:
So I think the reflection text looks good. However it exposes a confusion that makes the spec kind of hard to understand, which is that you talk a lot about "ARIA enumerated attributes". But these do not match HTML's notion of [enumerated attribute](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute).

This is most stark in how the new reflection text says

> If a reflecting IDL attribute is a nullable DOMString attribute whose content attribute **is not an enumerated attribute**

but then apparently the goal is to apply this reflection algorithm to the various ARIA attributes, per

> When the ARIA attribute definition includes a table enumerating the attribute's allowed values, that attribute is a nullable enumerated attribute

Another confusing issue is that "nullable enumerated attribute" is confusing content attributes and IDL attributes. IDL attributes can be nullable, and content attributes can be enumerated, but which does "nullable enumerated" apply to?

It would be best to remove all mention of "enumerated attribute" and "nullable enumerated attribute" from this document, I think. Maybe replace it with a new term like "multi-value ARIA attribute"?

domenic:
Stated from another perspective: HTML already defines reflection algorithms for nullable IDL attributes that reflect enumerated content attributes. They perform validation/canonicalization.

You want to use a different algorithm that does not perform validation/canonicalization. And you want to use a nullable IDL attribute. Thus, your content attributes must not be enumerated attributes; they must be some new category of thing (which I proposed calling "multi-value ARIA attribute").

cyns (approved):

domenic (approved):
Love it!

index.html
```diff
@@ -10217,56 +10217,50 @@ <h3>Value</h3>
...
 			</dl>
 			<p>These are generic types for states and properties, but do not define specific representation. See <a href="#state_property_processing">State and Property Attribute Processing</a> for details on how these values are expressed and handled in host languages.</p>
 		</section>
 	</section>
-	<section id="enumerated-attribute-values">
-		<h2>Enumerated Attribute Values</h2>
-		<p>When the ARIA attribute definition includes a table enumerating the attribute's allowed <span>values</span>, 
-		that attribute is a nullable <a data-cite="html/common-microsyntaxes.html#enumerated-attribute">enumerated attribute</a>. 
-		Each value in the table is a keyword for the attribute, mapping to a state of the same name. </p>
+	<section>
```
cookiecrook, line 10227: I see why you moved the "Multi-Value" section down to an H3 sub-section, but should this higher-level section get a new IDREF? (No strong preference. Approving as-is.)

cookiecrook (approved):

Issue #1623, opened 2021-10-03
Deploy w3c/spec-prod


saschanaz:
It seems the current Travis action somehow takes the previous state of the repository instead of the latest one.

For example, 3c8c9e364f0932181a7c3cc52f3228b7d19a7358 is triggered by 0d06a71dbedcc0aaca9a49bc1a5d83802da0d74e but it instead applied 7c5efbe270eaab1b9b255b9f2642aca2cf80effd (the previous commit).

I guess maybe time to migrate to w3c/spec-prod.

jnurthen:
We have an open PR on this https://github.com/w3c/aria/pull/1504 but it seems @plehegar is stuck. If you know of any way to help unstick the PR that would be great.

saschanaz:
Hmm, seems there is a ReSpec issue. Filed https://github.com/w3c/respec/issues/3805

saschanaz:
The duplication issue was because of aria/common script, will open a PR for that.

saschanaz:
w3c/aria-common#76 and #1631 should resolve other issues, and the remaining one is `role=time`. Still not sure who should be pinged?

saschanaz:
Okay, `role=time` is being newly added by ARIA 1.2 and it's weird that the definition already uses itself. I think it can use `<time>` for now, what do you think @jnurthen?

plehegar:
I believe gh-pages is [no longer outdated](https://github.com/w3c/aria/commits/gh-pages). Can we close this issue?

saschanaz:
Can we use w3c/spec-prod to simplify the settings? Each repository having their own script leads to maintenance problem.

saschanaz:
#1648 fixed this, thanks!

PR #1638, opened 2021-11-04
Escape examples for `role="time"`

saschanaz:
Fixes another issue from #1623 

Not sure whether using `role="time"` in the definition of `role="time"` makes sense, but please tell me if it actually does.

***
<a href="https://pr-preview.s3.amazonaws.com/saschanaz/aria/pull/1638.html" title="Last updated on Nov 5, 2021, 9:14 PM UTC (fa4235d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1638/43b08b9...saschanaz:fa4235d.html" title="Last updated on Nov 5, 2021, 9:14 PM UTC (fa4235d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -9376,19 +9376,21 @@
 				<p>An element that represents a specific point in time.</p>
 				<p class="note">At the present time, there are no <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> properties corresponding to the <code>datetime</code> attribute supported on <code>&lt;time&gt;</code> in [[HTML]]. The addition of this property will be considered for ARIA version 1.3.</p>
 				<p>Authors SHOULD limit text contents to a valid date- or time-related string, or apply this future <code>datetime</code>-equivalent property to the element which has role <code>time</code>.</p>
-				<p>Examples of valid date- or time-related strings as text contents of an element with the <code>time</code> role:</p>
-				<ul>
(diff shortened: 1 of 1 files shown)


pkra:
The line above the list reads:

> Examples of valid date- or time-related strings as text contents of an element with the time role:

I think the change would make this less clear. Readers would need to know the `time` element's implicit role.

On the other hand, these example lack an example wrapper (class="example") and the section could probably use some work in general.

saschanaz:
`<time>` will not be visible as-is to readers, unless you mean screen readers, which are expected to understand `<time>` by now if I'm not wrong?

pkra:
Right, sorry for having been unclear. I mixed my first and second point. 

I find this whole "example" confusing. Usually, examples in the spec are marked up separately and present code samples. 

If this is supposed to be such a code example, then I'm not sure what it's covering. If it's just about strings, there's no need for any markup. If it's about role=time markup, the current markup seems better -- but needs to be escaped.

Not sure I'm making more sense now.

saschanaz:
Escaping sounds good so I just did that, hope that's more clear now.

scottaohara (approved):
thanks @saschanaz, this helps!

i have some other editorial nits with this definition, but that is beyond scope for what you were doing here, so i'll open another issue.

pkra (approved):

pkra:
Thanks @saschanaz !

pkra:
ping @jnurthen

Issue #1629, opened 2021-10-29
Flesh out role=note


medeaoblongata:
I'm finding the spec for the "note" role quite vague and thin on detail. I think I need to use this role quite a bit, but I am simply not certain.

The preamble currently consists of a single line of text, which could mean many or few things, depending on interpretation.

> "A section whose content is parenthetic or ancillary to the main content of the resource."

I'm particularly bothered by the use of the word "main" here. Does this mean role=main? Does "resource" mean the note itself? The page?

I have put in the hours with Funk and Lewis, so "parenthetic or ancillary" don't have me reaching for a dictionary, however it strikes me that "parenthetic or ancillary" is not exactly "plain language". Can we find replacements for these that doesn't sound like jargon, or perhaps offer some definitions?

That's the copy mentioned. Now for the actual spec.

The following semantic cases may or may not call for a "note", but the spec leaves plenty of room for doubt.

- An note in (part of) the page that is not the main content (e.g. contentinfo or navigation)
- An annotation of a specific element or part
- A comment by an editor about a particular chunk of copy
- A notification
- A note with a close button

If "note" is the wrong thing for annotations. Do we need another role, or is there already a better candidate today?

Also, for authoring practices: 
- How do notes best fit into read order? 
- If annotations are notes, should the note come before/after the element they annotate? 
- Is there a class of notes that might be live regions by default? (e.g. notifications).
- Would it be useful to include an example of an annotation using (say) aria-describedby to link the annotation to the annotated. 
- What other attribute might reliably connect an annotation to something?

jnurthen:
any interest in creating a PR for this @brennanyoung ?

brennanyoung:
Yes, I can do that. I assume that any/all of the semantic cases I mentioned could be 'notes'.

I'm thinking that some examples in Aria Practices might be the best way to handle this, but a brief list of example cases would be useful for guiding authors and UA/AT vendors. 

I'd also like to get @aleventhal 's take on this, since he has put so much work into the aria-annotations spec, which seems to cover similar ground. A synthesis of the two would be ideal, if possible.

brennanyoung:
Unless there's a compellingly more appropriate aria attribute, I suggest that we explicitly recommend `aria-details` as the mechanism for associating an element with a `note`, since this matches the proposal in the aria-annotations spec.

AT/UA vendors might use such associations to provide useful kinds of interaction (e.g. to move focus from element to note).

aleventhal:
[...]

PR #1639, opened 2021-11-05
role=note fleshed out a little.

brennanyoung:
Language a little less terse. Explicit mention of aria-details to associate a note with another element.

***
<a href="https://pr-preview.s3.amazonaws.com/brennanyoung/aria/pull/1639.html" title="Last updated on Nov 9, 2021, 2:21 PM UTC (d9fea75)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1639/43b08b9...brennanyoung:d9fea75.html" title="Last updated on Nov 9, 2021, 2:21 PM UTC (d9fea75)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/brennanyoung/aria/pull/1639.html" title="Last updated on Jan 20, 2023, 6:46 PM UTC (1591365)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1639/43b08b9...brennanyoung:1591365.html" title="Last updated on Jan 20, 2023, 6:46 PM UTC (1591365)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5931,7 +5931,8 @@
 		<div class="role" id="note">
 			<rdef>note</rdef>
 			<div class="role-description">
-				<p>A <rref>section</rref> whose content represents additional information or parenthetical context to the primary content it supplements. A <code>note</code> is content provided by the author of the page or document, it is not to be used for providing reactions or suggestions. For these purposes, please review <rref>comment</rref> and <rref>suggestion</rref>.</p>
+				<p>A <rref>section</rref> whose content represents additional information or parenthetical context to the primary content it supplements.</p>
+				<p>A <code>note</code> is content provided by the author of the page or document, it is not to be used for providing reactions or suggestions. For these purposes, please review <rref>comment</rref> and <rref>suggestion</rref>.</p>
 				<p>When used within the normal flow of a page's content, a <code>note</code> has an implicit association with the content that it supplements. The following example demonstrates using a <code>note</code> to call out additional information in the natural reading order of a page:</p>
 				<pre class="example highlight">
 					&lt;p>... the following results outline support for the tested features.&lt;/p>
--- a/index.html
+++ b/index.html
@@ -5942,7 +5942,7 @@
 					&lt;/div>
 					&lt;p>...&lt;/p>
 				</pre>		
-				<p>In cases where an element with role <rref>note</rref> has been determiend to need a programmatic association with the content it supplements, authors can use one of the following mechanisms to associate the elements:</p>
+				<p>In cases where an element with role <rref>note</rref> has been determined to need a programmatic association with the content it supplements, authors can use one of the following mechanisms to associate the elements:</p>
 				<ul>
 					<li>If the <code>note</code> contains structured or interactive content (for example, a link, button, list, table, etc.) use <pref>aria-details</pref>.</li>
 					<li>If the <code>note</code> is brief and consists of static text, use <pref>aria-describedby</pref>.</li>
--- a/index.html
+++ b/index.html
@@ -5931,8 +5931,12 @@
 		<div class="role" id="note">
 			<rdef>note</rdef>
 			<div class="role-description">
-				<p>A <rref>section</rref> whose content qualifies, explains, annotates or is subordinate to some other piece of content</p>
-				<p>In cases where an element with role <rref>note</rref> has an explicit association with another element, authors SHOULD point to the note from that other element using <pref>aria-details</pref>.</p>
+				<p>A <rref>section</rref> whose content is supplementary to some other piece of content</p>
+				<p>In cases where an element with role <rref>note</rref> has an explicit association with another element, authors can use one of the following mechanisms to point to the note from the associated element:</p>
+				<ul>
+					<li>If the note is structured (for example with rich text, or operable features such as a close button), use <pref>aria-details</pref>.</li>
+					<li>If the note is brief or simple enough to be represented as plain text, use <pref>aria-describedby</pref>.</li>
+				</ul>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -5935,7 +5935,7 @@
 				<p>When used within the normal flow of a page's content, a <code>note</code> has an implicit association with the content that it supplements. The following example demonstrates using a <code>note</code> to call out additional information in the natural reading order of a page:</p>
 				<pre class="example highlight">
 					&lt;p>... the following results outline support for the tested features.&lt;/p>
-					&lt;div>
+					&lt;div role="note">
 					  &lt;p>Please keep in mind that at the time of publishing this page all results were accurate.&lt;/p>
 					  &lt;p>If you find any variations in results, please let us know!&lt;/p>
 					&lt;/div>
--- a/index.html
+++ b/index.html
@@ -5931,7 +5931,8 @@
 		<div class="role" id="note">
 			<rdef>note</rdef>
 			<div class="role-description">
-				<p>A section whose content is parenthetic or ancillary to the main content of the resource.</p>
+				<p>A <rref>section</rref> whose content qualifies, explains, annotates or is subordinate to some other piece of content</p>
+				<p>In cases where an element with role <rref>note</rref> has an explicit association with another element, authors SHOULD point to the note from that other element using <pref>aria-details</pref>.</p>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -5931,12 +5931,31 @@
 		<div class="role" id="note">
 			<rdef>note</rdef>
 			<div class="role-description">
-				<p>A <rref>section</rref> whose content is supplementary to some other piece of content.</p>
-				<p>In cases where an element with role <rref>note</rref> has an explicit association with another element, authors can use one of the following mechanisms to point to the note from the associated element:</p>
+				<p>A <rref>section</rref> whose content represents additional information or parenthetical context to primary content it supplements. A <code>note</code> is content provided by the author of the page or document, it is not to be used for providing reactions or suggestions. For these purposes, please review <rref>comment</rref> and <rref>suggestion</rref>.</p>
+				<p>When used within the normal flow of a page's content, a <code>note</code> has an implicit association with the content that it supplements. The following example demonstrates using a <code>note</code> to call out additional information in the natural reading order of a page:</p>
+				<pre class="example highlight">
+				  &lt;p>... the following results outline support for the tested features.&lt;/p>
+					&lt;div>
+					  &lt;p>Please keep in mind that at the time of publishing this page all results were accurate.&lt;/p>
+						&lt;p>If you find any variations in results, please let us know!&lt;/p>
+					&lt;/div>
+					&lt;p>...&lt;/p>
+				</pre>		
+				<p>In cases where an element with role <rref>note</rref> has been determiend to need an explicit association with another element, authors can use one of the following mechanisms to programmatically associate the elements:</p>
 				<ul>
-					<li>If the note is structured (for example with rich text, or operable features such as a close button), use <pref>aria-details</pref>.</li>
-					<li>If the note is brief or simple enough to be represented as plain text, use <pref>aria-describedby</pref>.</li>
+					<li>If the <code>note</code> contains structured or interactive content (for example, a link, button, list, table, etc.) use <pref>aria-details</pref>.</li>
+					<li>If the <code>note</code> is brief and consists of static text, use <pref>aria-describedby</pref>.</li>
 				</ul>
+				<pre class="example highlight">
+				  &lt;!-- using aria-detials to reference a note containing a link -->
+				  ...
+					&lt;button aria-details="info-note">Get Started&lt;/button>
+					...
+					&lt;div role="note" id="info-note">
+					  &lt;p>Need more information before you get started?&lt;/p> 
+						&lt;p>Visit our &lt;a href="...">product description page&lt;/a> to get all the information you need.&lt;/p>
+					&lt;/div>
+				</pre>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5931,7 +5931,8 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
 		<div class="role" id="note">
 			<rdef>note</rdef>
 			<div class="role-description">
-				<p>A section whose content is parenthetic or ancillary to the main content of the resource.</p>
+				<p>A <rref>section</rref> whose content qualifies, explains, annotates or is subordinate to some other piece of content</p>
```
scottaohara, line 5934: I was going to say we don't need to reference `section` here, having just looked through other definitions that don't reference `section`... but there is quite a bit of inconsistency.  E.g,  there are roles that do explicitly refer to themselves as `section` (`listitem`), those that define themselves as "sections" of content (`blockquote`, `article`, `code`, etc.)  but do not reference the abstract role, and then even those like `math` which instead just state "content that represents...", bypassing the term section all together.

I don't have a strong preference one way or the other, but calling this out because I do have a preference we make the spec consistent (in another PR, but the decision could impact the linking/wording for this one).

scottaohara, line 5934:
```suggestion
				<p>A <rref>section</rref> whose content qualifies, explains, annotates or is subordinate to some other piece of content.</p>
```

index.html
```diff
@@ -5931,7 +5931,8 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
...
+				<p>In cases where an element with role <rref>note</rref> has an explicit association with another element, authors SHOULD point to the note from that other element using <pref>aria-details</pref>.</p>
```
scottaohara, line 5935: This introduces a normative change that is arguably subjective.  I would personally prefer that this be changed to an "authors can use..." 

For instance, 
```
<p>Some words about something.</p>
<div role=note><p>supplementary content that supports the previous paragraph...</p></div>
```

authors should not be expected to create an `aria-details` association there.

There could be cases where an `aria-details` association might be helpful... so again, happy to introduce the idea.  just do not want a SHOULD here at all.

index.html
scottaohara, line 5934: >qualifies, explains, annotates or is subordinate

do not think "annotates" or "subordinate" should be added here.  i understand why they would be... but their inclusion could muddle `note` with `comment` or `suggestion`, if not also `doc-footnote` or `doc-endnote`.

suggest "supplementary" is more appropriate, if we even need additional words beyond "qualifies" and "explains"

scottaohara (changes requested):

index.html
brennanyoung: That's fair. I wondered about it too, and just copied the first pattern I found. 
FWIW the inheritance relationship is mentioned in the table of characteristics. Maybe that's enough.

For clarity, I think that if the word 'section' refers to the abstract ARIA role (or the HTML element), it should get the appropriate markup sugar. In every other case, we should try to avoid it, and use an appropriate synonym.

brennanyoung: Good. I replaced that mini-thesaurus wiith "supplementary".

cookiecrook:
Should this PR be linked to a particular issue?

brennanyoung:
@cookiecrook it arose from https://github.com/w3c/aria/issues/1629

index.html
```diff
@@ -5931,7 +5931,12 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
 		<div class="role" id="note">
 			<rdef>note</rdef>
 			<div class="role-description">
-				<p>A section whose content is parenthetic or ancillary to the main content of the resource.</p>
+				<p>A <rref>section</rref> whose content is supplementary to some other piece of content.</p>
+				<p>In cases where an element with role <rref>note</rref> has an explicit association with another element, authors can use one of the following mechanisms to point to the note from the associated element:</p>
+				<ul>
+					<li>If the note is structured (for example with rich text, or operable features such as a close button), use <pref>aria-details</pref>.</li>
```
scottaohara, line 5937: this should probably say "If the note contains..."  rather than "is structured".  

Are people commonly using notes with close buttons?  Is this going to confuse people to think that a "note" is a "notification" - a notification generally being more appropriate to inject into an `alert` or `status` element, arguably a `region` landmark to promote its importance?

index.html
```diff
@@ -5931,7 +5931,12 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
 		<div class="role" id="note">
 			<rdef>note</rdef>
 			<div class="role-description">
-				<p>A section whose content is parenthetic or ancillary to the main content of the resource.</p>
+				<p>A <rref>section</rref> whose content is supplementary to some other piece of content.</p>
+				<p>In cases where an element with role <rref>note</rref> has an explicit association with another element, authors can use one of the following mechanisms to point to the note from the associated element:</p>
```
scottaohara, line 5935: I still worry this addition would make people think they _need_ to do this.   I most commonly come across notes in specs, and I'll use them (when i remember) in my own writing.  Notes in these cases are in line with the content they are adding supplementary info to.  e.g., `<p>...</p> <p role=note>...</p> <p>...</p>`.

If adding in this additional information, we should be very clear it is NOT for this basic usage.   Having re-read this again today after some time away, I'm having a hard time coming up with a scenario where i would advise someone to follow these additional steps.  Not saying they might not be appropriate, but each time I think of something, I talk myself out of it thinking "well that's more like a footnote pattern, using links to ensure all users can navigate to and from the supporting content."  or  "now we're overlapping with the annotation roles".

index.html
brennanyoung: Note and notification are not adequately distinguished from one another right now. And that's a problem.

I work on range of RIAs which all include notes with close buttons. **They are not websites, they are RIAs**

Whether this is **common** seems irrelevant to the fact that it is a genuine, real world case that's not forbidden by the spec. Such a thing is not unusual when a new feature is added to an existing UI - there will be a kind of 'sticky' tooltip with a close button drawn next to the new UI element with a label along the lines of "You can try the new XYZ feature here". That's basically what we have.

scottaohara: Thanks for the context.  So then again are these "notes" or "notifications"?   To be clear, I'm not saying that a close button inside of a note is wrong or bad.  I just want to make sure we don't confuse people to think notes = notifications, where it is more common for a notification to have a close button.

index.html
brennanyoung: Doesn't "In cases where..." copy suggest that the use of relationship attributes is required?

Also: Do we have a use case of a note that it is **not** associated with some other particular element? I can't think of one, although they may exist. Dissociated notes seem to be the exception, not the rule. AFAICT a note **always** has a context. A note is (arguably) a kind of annotation, although the annotation roles do not include "note" for reasons that I have not yet discovered.

scottaohara: >Do we have a use case of a note that it is not associated with some other particular element?

Yes. I already mentioned it, and you referenced it in your response.  If a note is already presented in line with the content it further describes, there's no need to provide a programmatic association. It's implicit with the natural reading order of the content.  

Sure, I suppose one could consider a note a type of annotation, but I'd submit that a note is something added by the author for clarity, and is a part of the document itself.  Annotations, e.g., `comment` and `suggestion` are reactions _other_ people, but also including an author, to existing content.  They can be incorporated into a document, but are not part of that document in its initial state.

In line with my previous comment, I want to reiterate that I'm not saying there isn't a use case for associating a note with another element... but I want the spec to be clear on what that use case is.  Again, I would reference the ARIA spec as an example of where programmatically associating notes (the content blocks that start with the word "NOTE" are marked up as `role=note`) with other content isn't necessary.

index.html
brennanyoung: I wouldn't say they're notifications.

I conceive of a notification as something which typically appears during use, triggered by some background process, rather than a mechanism for orienting a user about a feature of a user interface.

Our notes are drawn when the user encounters the UI for the first time - as a result of _their_ choice to begin the activity. They're "affordances", really. 

This kind of thing is quite common in games and simulations, especially to provide a little extra guidance about the UI in a "first run" situation, "training level" or "onboarding" phase.

index.html
brennanyoung: Well, the association is still there, it's just implicit and contextual, rather than explicitly and programmatically specified.

Anyway I get your point. I can see that we need to distinguish between 

- notes within rich text, which are usually contextually associated, and may have no additional need of aria relationship attributes
- notes which might appear as signposts, hints or supplementary guidance within a user interface, where the note's relationship to a particular part might need to be programmatically indicated

, where the non-operable/static parts (including any notes) would typically be 'skipped' as a user tabs around.

index.html
scottaohara: so "teaching ui" then.  thank you.  this is kinda helpful for another discussion i've been having.

index.html
scottaohara: resolving this thread as I have made updates to the content to address my concerns here.

index.html
scottaohara: resolving this thread as I have made updates to the content to address my concerns here.

scottaohara:
While my commit message mentioned I was creating a separate PR to review and to then merge into this branch, unfortunately something was not working with that process, and my separate branch kept errorring out.  Fortunately, through the magic of version control, we can revert / make changes to anything that need be.  But as my commit message mentions, @brennanyoung please review these updates when you can.  Thanks

scottaohara:
@jnurthen hmm, tried to remove myself from the reviewers since i've now added a bunch of content to this PR - clearly that didn't work.  Worth getting someone else besides me to review now.

index.html
```diff
@@ -5931,7 +5931,31 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
 		<div class="role" id="note">
 			<rdef>note</rdef>
 			<div class="role-description">
-				<p>A section whose content is parenthetic or ancillary to the main content of the resource.</p>
+				<p>A <rref>section</rref> whose content represents additional information or parenthetical context to the primary content it supplements. A <code>note</code> is content provided by the author of the page or document, it is not to be used for providing reactions or suggestions. For these purposes, please review <rref>comment</rref> and <rref>suggestion</rref>.</p>
+				<p>When used within the normal flow of a page's content, a <code>note</code> has an implicit association with the content that it supplements. The following example demonstrates using a <code>note</code> to call out additional information in the natural reading order of a page:</p>
+				<pre class="example highlight">
+					&lt;p>... the following results outline support for the tested features.&lt;/p>
+					&lt;div>
```
jnurthen, line 5938: should this have role="note"? If not I'm not sure what this example is doing here.

jnurthen (changes requested):

index.html
scottaohara: see?  told you i needed a reviewer :)

index.html
```diff
@@ -5931,7 +5931,31 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
...
+					&lt;/div>
+					&lt;p>...&lt;/p>
+				</pre>		
+				<p>In cases where an element with role <rref>note</rref> has been determiend to need a programmatic association with the content it supplements, authors can use one of the following mechanisms to associate the elements:</p>
+				<ul>
+					<li>If the <code>note</code> contains structured or interactive content (for example, a link, button, list, table, etc.) use <pref>aria-details</pref>.</li>
+					<li>If the <code>note</code> is brief and consists of static text, use <pref>aria-describedby</pref>.</li>
+				</ul>
+				<pre class="example highlight">
+				  &lt;!-- using aria-detials to reference a note containing a link -->
```
jnurthen, line 5950:
```suggestion
				  &lt;!-- using aria-details to reference a note containing a link -->
```

brennanyoung (commented):
Good!

jnurthen (approved):

index.html
```diff
@@ -5931,7 +5931,31 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
 		<div class="role" id="note">
 			<rdef>note</rdef>
 			<div class="role-description">
-				<p>A section whose content is parenthetic or ancillary to the main content of the resource.</p>
+				<p>A <rref>section</rref> whose content represents additional information or parenthetical context to the primary content it supplements. A <code>note</code> is content provided by the author of the page or document, it is not to be used for providing reactions or suggestions. For these purposes, please review <rref>comment</rref> and <rref>suggestion</rref>.</p>
```
cookiecrook, line 5934:
```suggestion
				<p>A <rref>section</rref> whose content represents additional information or parenthetical context to the primary content it supplements.</p>
				<p>A <code>note</code> is content provided by the author of the page or document, it is not to be used for providing reactions or suggestions. For these purposes, please review <rref>comment</rref> and <rref>suggestion</rref>.</p>
```

cookiecrook: Keeping the first paragraph short, since it's used in the short role description list here: https://w3c.github.io/aria/#role_definitions

cookiecrook (approved):
Approve with suggested editorial change.

index.html
```diff
@@ -5931,7 +5931,31 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
...
+				<p>When used within the normal flow of a page's content, a <code>note</code> has an implicit association with the content that it supplements. The following example demonstrates using a <code>note</code> to call out additional information in the natural reading order of a page:</p>
+				<pre class="example highlight">
+					&lt;p>... the following results outline support for the tested features.&lt;/p>
+					&lt;div role="note">
+					  &lt;p>Please keep in mind that at the time of publishing this page all results were accurate.&lt;/p>
+					  &lt;p>If you find any variations in results, please let us know!&lt;/p>
+					&lt;/div>
+					&lt;p>...&lt;/p>
+				</pre>		
+				<p>In cases where an element with role <rref>note</rref> has been determiend to need a programmatic association with the content it supplements, authors can use one of the following mechanisms to associate the elements:</p>
```
adampage, line 5944:
```suggestion
				<p>In cases where an element with role <rref>note</rref> has been determined to need a programmatic association with the content it supplements, authors can use one of the following mechanisms to associate the elements:</p>
```

Minor typo.

adampage (approved):
I added a code suggestion for one small typo, otherwise this looks terrific to me. 🚀

scottaohara:
thank you @adampage and @cookiecrook 

@jnurthen seems we can merge this now :)

scottaohara:
@pkra @jnurthen reminder about merging

PR #1631, opened 2021-10-30
Use rref instead of ref, which is probably a typo

saschanaz:
Resolves one of the issues found in #1623.

***
<a href="https://pr-preview.s3.amazonaws.com/saschanaz/aria/pull/1631.html" title="Last updated on Nov 4, 2021, 6:18 PM UTC (439b495)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1631/0d06a71...saschanaz:439b495.html" title="Last updated on Nov 4, 2021, 6:18 PM UTC (439b495)">Diff</a>

PR #1648, opened 2021-11-20
Use w3c/spec-prod

saschanaz:
This should:

1. Simplify the script
2. Reduce the maintenance cost, since spec-prod is being actively maintained by its own maintainer
3. Allow better validation for HTML and Web IDL

All the previous errors are now solved by https://github.com/w3c/respec/pull/3806, https://github.com/w3c/aria-common/pull/76, #1631, and #1638.


saschanaz:
cc @plehegar, @jnurthen

saschanaz:
@marcoscaceres Does https://github.com/w3c/input-events/issues/125#issuecomment-972807387 also apply to this repo?

marcoscaceres:
It does, but this looks good 👍 it does exactly the same thing as the script.

marcoscaceres (approved):

PR #1650, opened 2021-11-24
Editorial: clarify expectations for name/desc verbosity

### LINKED ISSUE #1649: aria-describedby, verbose vs. concise and arbitrary 'state' information
There is a semantic aspect of `aria-describedby` which I think may get lost with the current spec copy, or perhaps I misunderstand something. Here are a couple of rather common use cases by way of illustration:

- A notifications button on a social media site which includes an additional 'badge' displaying the number of unread posts.
- A shopping cart on an e-commerce site which includes an additional 'badge' displaying the number of items in the cart.
- A process queue of items in a dashboard whose completion state is indicated with a small icon: Pending, processing, done.

These cases seem semantically identical. 

I assume that the **most correct** way to associate the item count 'badge' or the process state icon with the element is to point to the id of the badge/icon with `aria-describedby` - is that right? 

If so, I am concerned that the spec for `aria-describedby` states

> The aria-labelledby attribute is similar to the aria-describedby in that both reference other elements to calculate a text alternative, but **a label should be concise, where a description is intended to provide more verbose information**.

(Emphasis mine). What bothers me here is that the item count in both the cases mentioned above is **just a single value**, and is certainly concise, rather than verbose. Strictly speaking, it's **state**, pure and simple. The current wording might lead authors to look beyond these attributes for expressing state values, or to 'hack' the accessible name to include this additional (and semantically distinct) text. 

I strongly believe that concatenating an accessible name with **semantically distinct** strings that express **state** is an anti-pattern which indicates that a role or relationship may be missing. Similarly, it's preferred to rely on `aria-pressed` to express the state of a toggle button, rather than manipulating the accessible name.

I suppose that what is _intended_ by the words "concise" and "verbose" is that `aria-describedby` calls for 'flat' strings (i.e. without further semantic variation). 

The point appears to be to distinguish this attribute from `aria-details` which _can_ point to rich text ("verbose"). Yet `aria-details` is not mentioned as a possible alternative in the spec for `aria-describedby`, in the useful way that certain other parts of the spec guide authors to make the other appropriate choices where there may be similarities in the semantics.

The distinction **is** made in [the ARIA annotations spec](https://github.com/aleventhal/aria-annotations). I approve of this clarity, and hope the distinction will be included in the 1.3 spec for `aria-describedby`.

So there are two problems here, one is to guide readers of the ARIA spec to make the correct choice between `aria-labelledby`, `aria-describedby` and `aria-details` based **not** on the razor of verbose/concise but on whether the indicated element is

- The accessible name of the element (`aria-labelledby`)
- A flat string describing the element (`aria-describedby`)
- A rich text structure describing the element (`aria-details`)

... and secondly, to make it fully legitimate that **concise** 'state' annotations (such as the item counts or process state in the examples above) are valid use cases for `aria-describedby` - and I don't think that APG examples will be quite enough to make this unambiguous.

Or maybe I've got this all wrong, and someone could clear up my doubt about this. Is `aria-describedby` the wrong thing for arbitrary 'state' information associated with an element? Do we need something else?

scottaohara:
closes #1649 

These editorial changes attempt to clarify that a description _doesn't have to be verbose_.

It also adjusts some wording so when talking about an accessible name, the spec is saying that, and not using the term "label"

This update touches `aria-labelledby` and `aria-describedby` - because they share the same paragraph between the two.  It also touches `aria-description` which also was semi-reusing the paragraph that was repeated between `aria-labelledby` and `aria-describedby` - but it also needed a few wording changes.

Happy for others to add onto this or make further wording adjustments as need be.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1650.html" title="Last updated on Nov 26, 2021, 6:12 PM UTC (0e28ff1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1650/fcea743...0e28ff1.html" title="Last updated on Nov 26, 2021, 6:12 PM UTC (0e28ff1)">Diff</a>


jnurthen (approved):

index.html
```diff
@@ -11217,7 +11217,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 			<pdef>aria-describedby</pdef>
 			<div class="property-description">
 				<p><a>Identifies</a> the <a>element</a> (or elements) that describes the <a>object</a>. See related <pref>aria-labelledby</pref> and <pref>aria-description</pref>.</p>
-				<p>The <pref>aria-labelledby</pref> attribute is similar to <pref>aria-describedby</pref> in that both reference other elements to calculate a text alternative, but a label should be concise, where a description is intended to provide more verbose information.</p>
+				<p>The <pref>aria-labelledby</pref> attribute is similar to <pref>aria-describedby</pref> in that both reference other elements to calculate a text alternative. While a concise accessible name is preferrable, a description can either be concise, or provide more verbose information.</p>
```
scottaohara, line 11220:
```suggestion
				<p>The <pref>aria-labelledby</pref> attribute is similar to <pref>aria-describedby</pref> in that both reference other elements to calculate a text alternative (an accessible name, and description, respectively). While a concise accessible name is preferable, a description can either be concise, or provide more verbose information.</p>
```

index.html
```diff
@@ -12000,7 +12000,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p>If the interface is such that it is not possible to have a visible label on the screen, authors SHOULD use <pref>aria-label</pref> and SHOULD NOT use <pref>aria-labelledby</pref>. As required by the <a href="#textalternativecomputation">accessible name and description computation</a>, user agents give precedence to <pref>aria-labelledby</pref> over <pref>aria-label</pref> when computing the accessible name property.</p>
-				<p>The <pref>aria-labelledby</pref> attribute is similar to <pref>aria-describedby</pref> in that both reference other elements to calculate a text alternative, but a label should be concise, where a description is intended to provide more verbose information.</p>
+				<p>The <pref>aria-labelledby</pref> attribute is similar to <pref>aria-describedby</pref> in that both reference other elements to calculate a text alternative (an accessible name, and description, respectively). While a concise accessible name is preferrable, a description can either be concise, or provide more verbose information.</p>
```
scottaohara, line 12003:
```suggestion
				<p>The <pref>aria-labelledby</pref> attribute is similar to <pref>aria-describedby</pref> in that both reference other elements to calculate a text alternative (an accessible name, and description, respectively). While a concise accessible name is preferable, a description can either be concise, or provide more verbose information.</p>
```

scottaohara (commented):
edit to keep paragraphs in sync and fix a typo

pkra (approved):

pkra:
@jnurthen @michael-n-cooper  What is the process to update the stable branch? Do I simply cherry pick the merged commit and push it to stable?

PR #1660, opened 2021-12-10
remove overly prescriptive distinctions from aria-current values.

### LINKED ISSUE #1656: Suggest change to aria-current step (not only a link)
The Specification for aria-current=Step
"_A step token used to indicate a link within a step indicator for a step-based process, where the link is visually styled to represent the current step._"

Both aria-current page and step are described in the normative text of the specification as associate with links. It is more lightly that aria-current page will be applied to a link, however this is not necessarily the case, for example, the current page in a breadcrumb is usually not linked. Arguably aria-current location should be used for a breadcrumb, but, you will find that 'page' is used more than 'location' on breadcrumbs.

The problem I see is that aria-current step may not be linked, and this would be an acceptable pattern, for example, a progress bar showing the current step in a purchase journey, in my experience, is not usually linked.

I recommend that the Spec for 1.2 be changed to "_A step token used to indicate a step within a progress indicator for a step-based process, where the element is visually styled to represent the current step._"

Would like to see a discussion on extending this to other values on inactive elements.

Reference
Accessible Rich Internet Applications (WAI-ARIA) 1.2 > [aria-current](https://www.w3.org/TR/wai-aria-1.2/#aria-current) (state)

cookiecrook:
Resolves #1656.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1660.html" title="Last updated on Dec 10, 2021, 10:02 PM UTC (e523750)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1660/9653692...e523750.html" title="Last updated on Dec 10, 2021, 10:02 PM UTC (e523750)">Diff</a>

***

### :boom: Error: 504 Gateway Timeout :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 24, 2022, 8:22 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Faria%2F624fef245892f351643bc932041789f06f53b499%2Findex.html%3FisPreview%3Dtrue)

```
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231660.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -11139,7 +11139,7 @@ button.ariaPressed; // null</pre>
 				<p>The <sref>aria-current</sref> <a>attribute</a> is a token type. Any value not included in the list of allowed values SHOULD be treated by <a>assistive technologies</a> as if the value <code>true</code> had been provided. If the attribute is not present or its value is an empty string or <code>undefined</code>, the default value of <code>false</code> applies and the <sref>aria-current</sref> <a>state</a> MUST NOT be exposed by user agents or assistive technologies.</p>
 				<p>The <sref>aria-current</sref> attribute is used when an element within a set of related elements is visually styled to indicate it is the current item in the set. For example:</p>
 				<ul>
-					<li>A <code>page</code> token used to indicate a page within a page-based process, where the element is visually styled to represent the current page.</li>
+					<li>A <code>page</code> token used to indicate a page within a set of pages, where the element is visually styled to represent the current page.</li>
 					<li>A <code>step</code> token used to indicate a step within a step-based process, where the element is visually styled to represent the current step.</li>
 					<li>A <code>location</code> token used to indicate the element that is visually styled as the current component, such as within a flow chart.</li>
 					<li>A <code>date</code> token used to indicate the current date within a calendar or other date collection.</li>
--- a/index.html
+++ b/index.html
@@ -11139,9 +11139,9 @@ button.ariaPressed; // null</pre>
 				<p>The <sref>aria-current</sref> <a>attribute</a> is a token type. Any value not included in the list of allowed values SHOULD be treated by <a>assistive technologies</a> as if the value <code>true</code> had been provided. If the attribute is not present or its value is an empty string or <code>undefined</code>, the default value of <code>false</code> applies and the <sref>aria-current</sref> <a>state</a> MUST NOT be exposed by user agents or assistive technologies.</p>
 				<p>The <sref>aria-current</sref> attribute is used when an element within a set of related elements is visually styled to indicate it is the current item in the set. For example:</p>
 				<ul>
-					<li>A <code>page</code> token used to indicate a page within a a progress indicator for a page-based process, where the element is visually styled to represent the current page.</li>
-					<li>A <code>step</code> token used to indicate a step within a progress indicator for a step-based process, where the element is visually styled to represent the current step.</li>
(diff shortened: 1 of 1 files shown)


scottaohara (approved):

cookiecrook:
Note: I also removed overly prescriptive context limitations on the other values (e.g. location was limited to an "image" in a "flow chart"), so this is a little more than editorial.

scottaohara:
I'm glad you made that change... i personally never interpreted that bit about `location` to mean that was its only use case. Especially since that listing is preceded by the words "for example".

Was it really a normative limitation being set there?   That might be a gap in my understanding on how normative declared examples are if they are not contained within notes or non-normative sections

smhigley (approved):

index.html
```diff
@@ -11139,11 +11139,11 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p>The <sref>aria-current</sref> attribute is used when an element within a set of related elements is visually styled to indicate it is the current item in the set. For example:</p>
 				<ul>
-					<li>A <code>page</code> token used to indicate a link within a set of pagination links, where the link is visually styled to represent the currently-displayed page.</li>
-					<li>A <code>step</code> token used to indicate a link within a step indicator for a step-based process, where the link is visually styled to represent the current step.</li>
-					<li>A <code>location</code> token used to indicate the image that is visually highlighted as the current component of a flow chart.</li>
-					<li>A <code>date</code> token used to indicate the current date within a calendar.</li>
-					<li>A <code>time</code> token used to indicate the current time within a timetable.</li>
+					<li>A <code>page</code> token used to indicate a page within a page-based process, where the element is visually styled to represent the current page.</li>
```
carmacleod, line 11142: Not sure I like "page-based process". It doesn't have to be a process.
I think I prefer "set of pages". It can just be in a nav in a single-page app (for example), where the author wants to show which page is currently being displayed.
```suggestion
					<li>A <code>page</code> token used to indicate a page within a set of pages, where the element is visually styled to represent the current page.</li>
```

index.html
```diff
@@ -11139,11 +11139,11 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+					<li>A <code>step</code> token used to indicate a step within a step-based process, where the element is visually styled to represent the current step.</li>
+					<li>A <code>location</code> token used to indicate the element that is visually styled as the current component, such as within a flow chart.</li>
```
carmacleod, line 11144: Just wondering whether to say "current location". Not sure. I could go either way with this.
```suggestion
					<li>A <code>location</code> token used to indicate the element that is visually styled as the current location, such as within a flow chart.</li>
```

carmacleod (commented):
2 comments (in suggestions below):
1. I really do prefer "set of pages" in the first suggestion below, because "page-based process" feels overly prescriptive. ;)
2. Just something to think about with "location" instead of "component" in the second suggestion below, but I think it's fine either way.

index.html
jnurthen: I prefer component over location here.

cookiecrook: I do too. Not incorporating the last edit.

PR #1663, opened 2021-12-16
native attribute -> native feature

### LINKED ISSUE #1610: Ambiguity in #host_general_conflict when native host language feature is not a Content Attribute
From [ARIA §8.5 Conflicts with Host Language Semantics](https://w3c.github.io/aria/#host_general_conflict):

> Therefore, to prevent providing conflicting states and properties to assistive technologies, host languages MUST explicitly declare where the use of WAI-ARIA attributes on each host language element conflicts with native **attributes** for that element. When a host language declares a WAI-ARIA attribute to be in direct semantic conflict with a native **attribute** for a given element, user agents MUST ignore the WAI-ARIA attribute and instead use the host language **attribute** with the same implicit semantic.

The topic of host language conflict came up during [implementation of the dialog element in WebKit](https://bugs.webkit.org/show_bug.cgi?id=229815). 

In this case, the ARIA feature is a Content Attribute, but the native Dialog API feature (`is modal`) is not a Content Attribute. 

Question: Since there is a direct semantic conflict, I assume the ARIA WG members would expect `aria-modal` to have have no effect on the HTML `dialog` element. Is there anyone who disagrees, and if so, why?

If there is no disagreement, I believe there's a easy fix in the spec: Change these three instances of the word "attribute" to the more general term "feature" as is used in the rest of that section. So "native attribute(s)" and "host language attribute(s)" become "native feature(s)" and "host language feature(s)" respectively.

cookiecrook:
Resolves #1610

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1663.html" title="Last updated on Dec 16, 2021, 12:19 AM UTC (bfd1a67)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1663/9653692...bfd1a67.html" title="Last updated on Dec 16, 2021, 12:19 AM UTC (bfd1a67)">Diff</a>


cookiecrook:
Diff not apparent, but this changes 3 instances of "native attribute" or "host language attribute" to "native feature" or "host language feature" respectively. 

screen shot of specific non-highlighted diff

jnurthen (approved):

scottaohara (approved):

pkra (approved):

PR #1666, opened 2021-12-21
add note to aria-roledescription

### LINKED ISSUE #1651: Question should aria-roledescription be prohibited on additional roles?
Should we be more prescriptive of what roles should be marked as being prohibited from using aria-roledescription?  `generic` is the only role that presently prohibits this (note: would have also expected this on none/presentation - but actually i kinda expect all global attributes to be prohibited on those roles since they undo the role... but i digress).

For instance, `role=paragraph` or the `<p>` element.  While it has an implicit / explicit ARIA role, what would the expectation here be?  That screen readers, which generally don't announce a role like this, instead start announcing this modified role 'description'?

Upon a quick test, `<p aria-roledescription="oh no">hi there</p>` NVDA actually does announce "oh no" when i navigate to the paragraph.  But is that a good idea?  JAWS and Narrator don't do this when navigating by paragraph... people don't generally expect a "paragraph" role to be announced.  

Thoughts?

scottaohara:
closes #1651

adds a note indicating that `aria-roledescription` may not be conveyed depending on the element/AT defaults or user settings.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1666.html" title="Last updated on Dec 23, 2021, 3:04 AM UTC (edeea7d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1666/9653692...edeea7d.html" title="Last updated on Dec 23, 2021, 3:04 AM UTC (edeea7d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -12736,11 +12736,12 @@ button.ariaPressed; // null</pre>
 				<p><a>Defines</a> a human-readable, author-localized description for the <a>role</a> of an <a>element</a>.</p>
 				<p>Some <a>assistive technologies</a>, such as screen readers, present the role of an element as part of the user experience. Such assistive technologies typically localize the name of the role, and they may customize it as well. Users of these assistive technologies depend on the presentation of the role name, such as "region," "button," or "slider," for an understanding of the purpose of the element and, if it is a widget, how to interact with it.</p>
 				<p>The <code>aria-roledescription</code> property gives authors the ability to override how assistive technologies localize and express the name of a role. Thus inappropriately using <code>aria-roledescription</code> may inhibit users' ability to understand or interact with an element. Authors SHOULD limit use of <code>aria-roledescription</code> to clarifying the purpose of non-interactive container roles like <rref>group</rref> or <rref>region</rref>, or to providing a <em>more specific</em> description of a <rref>widget</rref>.</p>
-				<p> When using <code>aria-roledescription</code>, authors SHOULD also ensure that:</p>
+				<p>When using <code>aria-roledescription</code>, authors SHOULD also ensure that:</p>
 				<ol>
 					<li>The element to which <code>aria-roledescription</code> is applied has a valid <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> role or has an implicit WAI-ARIA role semantic.</li>
 					<li>The value of <code>aria-roledescription</code> is not empty or does not contain only whitespace characters.</li>
 				</ol>
+				<p class="note">Depending on the assistive technology, user verbosity settings, or other factors, certain elements and their implicit or explicit ARIA role descriptions might not be conveyed. Authors need to be aware that if specifying <code>aria-roledescription</code> on these elements, then their custom role descriptions may not be conveyed by these assistive technologies, as well.</p>
 				<p>User agents MUST NOT expose the <code>aria-roledescription</code> property if any of the following conditions exist:</p>
 				<ol>
 					<li>The element to which <code>aria-roledescription</code> is applied has an explicit or implicit WAI-ARIA role where <code>aria-roledescription</code> is <a href="#prohibitedattributes">prohibited</a>.</li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -12736,11 +12736,12 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p> When using <code>aria-roledescription</code>, authors SHOULD also ensure that:</p>
+				<p>When using <code>aria-roledescription</code>, authors SHOULD also ensure that:</p>
 				<ol>
 					<li>The element to which <code>aria-roledescription</code> is applied has a valid <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> role or has an implicit WAI-ARIA role semantic.</li>
 					<li>The value of <code>aria-roledescription</code> is not empty or does not contain only whitespace characters.</li>
 				</ol>
+				<p class="note">Depending on the assistive technology, user verbosity settings, or other factors, certain elements and their implicit or explicit ARIA role descriptions might not be conveyed. Authors need to be aware that if specifying <code>aria-roledescription</code> on these elements, then their custom role descriptions may not be conveyed by these assistive technologies, as well.</p>
```
jnurthen, line 12744: reads awkwardly - would it be better without the last "as well"?

scottaohara, line 12744:
```suggestion
				<p class="note">Depending on the assistive technology, user verbosity settings, or other factors, certain elements' role descriptions might not be conveyed. If specifying <code>aria-roledescription</code> on such elements, then the custom role descriptions may also not be conveyed by these assistive technologies.</p>
```

scottaohara: agreed.  definitely needed to step away from that for a bit.  have tried to make it less bad.

index.html
```diff
@@ -12736,11 +12736,12 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p> When using <code>aria-roledescription</code>, authors SHOULD also ensure that:</p>
+				<p>When using <code>aria-roledescription</code>, authors SHOULD also ensure that:</p>
 				<ol>
 					<li>The element to which <code>aria-roledescription</code> is applied has a valid <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> role or has an implicit WAI-ARIA role semantic.</li>
 					<li>The value of <code>aria-roledescription</code> is not empty or does not contain only whitespace characters.</li>
 				</ol>
+				<p class="note">Depending on the assistive technology, user verbosity settings, or other factors, certain elements' role descriptions might not be conveyed. If specifying <code>aria-roledescription</code> on such elements, then the custom role descriptions may also not be conveyed by these assistive technologies.</p>
```
spectranaut, line 12744: I'm not sure, but should the first sentence be: "certain elements' role might not be conveyed" instead of "certain elements' role _description_ might not be conveyed"?

When a AT does communicate the role like "link", we say the "role" is being communicated not the "role description" is being communicated, right?

spectranaut (changes requested):
Just one question! :)

index.html
scottaohara: I was trying to use similar language per James's comment in the original issue](https://github.com/w3c/aria/issues/1651#issuecomment-982369197).  specifically "At low verbosity, most role descriptions (custom or otherwise) should not be spoken at all"

That said, i'm honestly open to either wording.  whatever people think is most appropriate

spectranaut: I see! then I'm agnostic too -- so long as @jnurthen thinks it the correct term.

spectranaut (approved):

cookiecrook (approved):

scottaohara:
@jnurthen anything else you'd like to see here, or this ok to merge?

index.html
pkra: "role description" is used in https://w3c.github.io/aria/#terms

> When the button receives focus, assistive technologies may concatenate the platform's role description with the accessible name. For example, a screen reader may speak "push-button OK" or "OK button". The order of concatenation and specifics of the role description (e.g., "button", "push-button", "clickable button") are determined by platform [accessibility API](https://w3c.github.io/aria/#dfn-accessibility-api)s or [assistive technologies](https://w3c.github.io/aria/#dfn-assistive-technology).

So I think this is good to go.

PR #1671, opened 2022-01-11
progressbar: clarify described-by

### LINKED ISSUE #1654: Clarify statement in progressbar role definition
In _[5.4 > progressbar](https://www.w3.org/TR/wai-aria-1.2/#progressbar)_

> If the progressbar is describing the loading progress of a particular region of a page, the author SHOULD use aria-describedby to point to the status, and set the aria-busy attribute to true on the region until it is finished loading.

This statement leaves it somewhat unclear whether the author should put the `aria-describedby` attribute on the region of the page that is loading and point it to the progressbar or vice-versa. Contrast this with the instructions for `aria-busy`, where it is made clear that attribute should be placed on the region and guidance is given on what to do regarding the attribute when the loading has finished.

TLDR: clarify which element `aria-describedby` should go on and which element it should point to in this case, and what should be done when the loading finishes.

Related: https://github.com/mdn/content/pull/10707

pkra:
Resolves #1654

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1671.html" title="Last updated on Jan 11, 2022, 2:24 PM UTC (bc7be54)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1671/1e0383a...bc7be54.html" title="Last updated on Jan 11, 2022, 2:24 PM UTC (bc7be54)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6462,7 +6462,7 @@
 					</ul>
 				</p>
 				<p>The author SHOULD supply a <span>value</span> for <pref>aria-valuenow</pref> unless the value is indeterminate, in which case the author SHOULD omit the <pref>aria-valuenow</pref> attribute.
-				Authors SHOULD update this value when the visual progress indicator is updated. If the <code>progressbar</code> is describing the loading progress of a particular region of a page, the author SHOULD use <pref>aria-describedby</pref> on the region to point to the status, and set the <sref>aria-busy</sref> attribute to <code>true</code> on the region until it is finished loading. It is not possible for the user to alter the value of a <code>progressbar</code> because it is always read-only.</p>
+				Authors SHOULD update this value when the visual progress indicator is updated. If the <code>progressbar</code> is describing the loading progress of a particular region of a page, authors SHOULD both use <pref>aria-describedby</pref> to reference the progressbar status, and set the <sref>aria-busy</sref> attribute to <code>true</code> on the region until it is finished loading. It is not possible for the user to alter the value of a <code>progressbar</code> because it is always read-only.</p>
 				<p class="note">Assistive technologies generally will render the value of <pref>aria-valuenow</pref> as a percent of a range between the value of <pref>aria-valuemin</pref> and <pref>aria-valuemax</pref>, unless <pref>aria-valuetext</pref> is specified.</p>
 			</div>
 			<table class="role-features">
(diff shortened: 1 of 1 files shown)


pkra:
I wasn't sure if we should replace the next "status" with "progressbar" - since status is vague (i.e., I guess could point within the progressbar) I think that would go beyond an editorial change.

index.html
```diff
@@ -6462,7 +6462,7 @@ <h5>Note regarding the ARIA 1.1 <rref>none</rref> role.</h5>
...
 				<p>The author SHOULD supply a <span>value</span> for <pref>aria-valuenow</pref> unless the value is indeterminate, in which case the author SHOULD omit the <pref>aria-valuenow</pref> attribute.
-				Authors SHOULD update this value when the visual progress indicator is updated. If the <code>progressbar</code> is describing the loading progress of a particular region of a page, the author SHOULD use <pref>aria-describedby</pref> to point to the status, and set the <sref>aria-busy</sref> attribute to <code>true</code> on the region until it is finished loading. It is not possible for the user to alter the value of a <code>progressbar</code> because it is always read-only.</p>
+				Authors SHOULD update this value when the visual progress indicator is updated. If the <code>progressbar</code> is describing the loading progress of a particular region of a page, the author SHOULD use <pref>aria-describedby</pref> on the region to point to the status, and set the <sref>aria-busy</sref> attribute to <code>true</code> on the region until it is finished loading. It is not possible for the user to alter the value of a <code>progressbar</code> because it is always read-only.</p>
```
scottaohara, line 6465: per your comment in the issue, how about "progressbar status"?

Made some other wording changes to get away from saying "on the region" twice in the same sentence.  

```suggestion
				Authors SHOULD update this value when the visual progress indicator is updated. If the <code>progressbar</code> is describing the loading progress of a particular region of a page, authors SHOULD both use <pref>aria-describedby</pref> to reference the progressbar status, and set the <sref>aria-busy</sref> attribute to <code>true</code> on the region until it is finished loading. It is not possible for the user to alter the value of a <code>progressbar</code> because it is always read-only.</p>
```

pkra: Much better than me -- thanks, Scott!

scottaohara (approved):

MelSumner (approved):
👍

carmacleod (approved):
+1
Looks good.
Thanks @pkra !

PR #1678, opened 2022-01-14
Update related concepts for landmarks

### LINKED ISSUE #1677: Inconsistency with related concepts for landmarks
Presently `region`, `form`, and `navigation` all point to their HTML counterparts in their characteristics table.

In comparison, `banner`, `complementary`, `main`, and `contentinfo` do not.

`search` presently has no related concept / element - but that will hopefully change soon.

Seems to me that one of two things happen here:
1. they all just point to their related HTML elements.  Seems easy enough. 
2. only the landmarks that have no special conditions point to their related HTML elements.  e.g., `main` should point to `<main>`, but maybe `region` and `form` should not point to `<section>` and `<form>` because those elements are not always exposed as landmarks.

I think either makes sense, with option 1 being my preference. But, maybe the inconsistency is because of the reasoning for option 2?  

Either way, I'll make a PR to at least make a reference to  `main`, and then I can finish it up based on whatever option the WG decides.

scottaohara:
closes  #1677

The initial PR adds the related concept row to the `main` landmark, referencing HTML's `<main>` element. 

Follow-on updates to this PR will be contingent on the direction decided in the linked issue.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1678.html" title="Last updated on Jan 17, 2022, 1:40 PM UTC (28ca1e2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1678/b2503c8...28ca1e2.html" title="Last updated on Jan 17, 2022, 1:40 PM UTC (28ca1e2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1572,7 +1572,7 @@
 					</tr>
 					<tr>
 						<th class="role-related-head" scope="row">Related Concepts:</th>
-						<td class="role-related"> </td>
+						<td class="role-related"><code>&lt;header&gt;</code> in [[HTML]]</td>
 					</tr>
 					<tr>
 						<th class="role-scope-head" scope="row">Required Context Role:</th>
@@ -2663,7 +2663,7 @@
 					</tr>
 					<tr>
 						<th class="role-related-head" scope="row">Related Concepts:</th>
-						<td class="role-related"> </td>
+						<td class="role-related"><code>&lt;aside&gt;</code> in [[HTML]]</td>
 					</tr>
 					<tr>
 						<th class="role-scope-head" scope="row">Required Context Role:</th>
@@ -2823,7 +2823,7 @@
 					</tr>
 					<tr>
 						<th class="role-related-head" scope="row">Related Concepts:</th>
-						<td class="role-related"> </td>
+						<td class="role-related"><code>&lt;footer&gt;</code> in [[HTML]]</td>
 					</tr>
 					<tr>
 						<th class="role-scope-head" scope="row">Required Context Role:</th>
--- a/index.html
+++ b/index.html
@@ -4910,7 +4910,7 @@
 					</tr>
 					<tr>
 						<th class="role-related-head" scope="row">Related Concepts:</th>
-						<td class="role-related"> </td>
+						<td class="role-related"><code>&lt;main&gt;</code> in [[HTML]]</td>
 					</tr>
 					<tr>
 						<th class="role-scope-head" scope="row">Required Context Role:</th>


carmacleod (commented):
I prefer your option 1, which would add to banner, complementary and contentinfo as well. :)

scottaohara:
added in the remaining per opt1

pkra (approved):

carmacleod (approved):
+1
Thanks, @scottaohara !

carmacleod:
Might make sense to change "Base Concept" to "Related Concepts" for `form` in this PR? (because it's not really identical to `<form>`)
https://w3c.github.io/aria/#form
https://w3c.github.io/aria/#baseConcept
https://w3c.github.io/aria/#relatedConcept

Also kind of wondering if `<main>` and `<nav>` should use Base Concept instead of Related Concepts because they're pretty much identical to the aria role.

Also kind of wondering if having 2 categories (Base and Related) is really all that useful? Thoughts? The only roles that have both are [button](https://w3c.github.io/aria/#button) and [option](https://w3c.github.io/aria/#option).

scottaohara:
hmmm.  something to think about there.  thanks @carmacleod

carmacleod:
@scottaohara 
> hmmm. something to think about there.

Ok, I've opened https://github.com/w3c/aria/issues/1679 because I don't think this PR needs to be held up by that discussion. I gave it an Agenda label so that the group can discuss it.

I think the only thing I would change in this "landmarks related concepts" PR is to move `<form>` from Base Concept to Related Concepts in https://w3c.github.io/aria/#form. Do you agree?

scottaohara:
hmm.  i guess i'm not so sure.  reading through the various roles I could see an argument for either.

carmacleod:
I guess the key differentiator is:
> A base concept is like a related concept except that the base concept is almost identical to the role definition.

So my thinking is that `<div role="form">` doesn't have any of the `<form>` element attributes (autocomplete, action, name, target, ...), and the form magic doesn't work, e.g. a submit button with a formaction does nothing inside a div.

So I would say they're not really "almost identical".

That said, I'm happy to put this PR in as-is, and address whether `<form>` is a base or related concept for role form in #1679.

scottaohara:
yeah. I would tend to agree that we use 1679 to cleanup any existing content (`form`) as well as determine what to do about the base/related, even if only just being more accurate in their usage.

carmacleod:
@pkra This is ready to be merged. :)

pkra:
@carmacleod thanks. I was leaving it open until we have triaged this on a WG call.

PR #1682, opened 2022-01-21
Update paragraph in option to remove DOM focus requirement

### LINKED ISSUE #1661: Inconsistency between native and ARIA listboxes when implicit aria-selected is provided
According to the [spec](https://w3c.github.io/aria/#option):

> If a user agent provides an implicit `aria-selected` value for an `option`, the value SHOULD be true if the `option` has DOM focus or the `listbox` has DOM focus and the `option` is referenced by `aria-activedescendant`. Otherwise, if a user agent provides an implicit `aria-selected` value for an `option`, the value SHOULD be false.

This can lead to an inconsistency between native and ARIA listboxes. Consider the following:

```html
<select size="3">
<option>blue</option>
<option>red</option>
<option>green</option>
</select>
<button>ok</button>
```

If the user arrows down to the "red" `option` and then presses Tab, "red" is still selected even though the "ok" `button` is focused. Furthermore, and more importantly, a screen reader's review mode could be used to examine the `select` list and obtain the currently-selected value without ever changing focus. In the case of Linux, the AtkSelection interface implemented on that `select` element will return the accessible associated with the "red" `option`.

In contrast, we have this example:

```html
<div role="listbox" aria-activedescendant="red" tabindex="0">
<div role="option">blue</div>
<div role="option" id="red">red</div>
<div role="option">green</div>
</div>
<button>ok</button>
```

Assume the author provided keyboard navigation, the user arrowed to "red", the author updated the value of `aria-activedescendant` accordingly. Then the user pressed Tab to give focus to the "ok" `button`.

If the user agent provides an implicit `aria-selected` (which Chrome/Chromium does), the spec states that it should only report the selected item to ATs when the `listbox` or `option` is focused. Chrome/Chromium follows this. As a result,  a screen reader's review mode cannot be used to examine the `listbox` and obtain the currently-selected value without ever changing focus.

What is the rationale behind the language in the spec? Is there a benefit to not exposing the selected item when the `listbox` is no longer focused and introducing this inconsistency between native elements and ARIA?

carmacleod:
Fixes #1661

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1682.html" title="Last updated on Jan 27, 2022, 5:58 PM UTC (fabb8b6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1682/157d711...fabb8b6.html" title="Last updated on Jan 27, 2022, 5:58 PM UTC (fabb8b6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6025,7 +6025,7 @@
           <li>None of the <rref>option</rref> elements in the <rref>listbox</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>
         </ul>
         <p>
-          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has <code>tabindex="0"</code> (or tabindex with a value greater then 0), or the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
+          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or has <code>tabindex="0"</code> (or tabindex with a value greater then 0), or the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
           Otherwise, if a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>false</code>.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -6025,7 +6025,7 @@
           <li>None of the <rref>option</rref> elements in the <rref>listbox</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>
         </ul>
         <p>
-          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or has <code>tabindex="0"</code> (or tabindex with a value greater then 0), or the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
+          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or has <code>tabindex="0"</code> (or <code>tabindex</code> with a value greater than 0), or the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
           Otherwise, if a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>false</code>.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -6025,7 +6025,7 @@
           <li>None of the <rref>option</rref> elements in the <rref>listbox</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>
         </ul>
         <p>
-          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or the <rref>listbox</rref> has DOM focus and the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
+          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has <code>tabindex="0"</code> (or tabindex with a value greater then 0), or the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
           Otherwise, if a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>false</code>.
         </p>
         <p>


JAWS-test:
@carmacleod If I understand correctly, you have removed the variant that the list entry (role=option) has DOM focus. I think that is wrong. Our discussion was about what happens if the option does not have DOM focus. But with DOM focus, the option should also be considered selected (aria-selected=true), regardless of the tabindex. In practice, DOM focus and tabindex will usually coincide, but this does not necessarily have to be the case. My suggestion would be to reinsert "has DOM focus or".

index.html
```diff
@@ -6025,7 +6025,7 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
           <li>None of the <rref>option</rref> elements in the <rref>listbox</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>
         </ul>
         <p>
-          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or the <rref>listbox</rref> has DOM focus and the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
+          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has <code>tabindex="0"</code> (or tabindex with a value greater then 0), or the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
```
carmacleod, line 6028:
```suggestion
          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or has <code>tabindex="0"</code> (or tabindex with a value greater then 0), or the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
```

index.html
```diff
@@ -6025,7 +6025,7 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
           <li>None of the <rref>option</rref> elements in the <rref>listbox</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>
         </ul>
         <p>
-          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or the <rref>listbox</rref> has DOM focus and the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
+          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or has <code>tabindex="0"</code> (or tabindex with a value greater then 0), or the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
```
carmacleod, line 6028:
```suggestion
          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or has <code>tabindex="0"</code> (or <code>tabindex</code> with a value greater than 0), or the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
```

spectranaut (commented):
Darn I have to say I'm not a huge fan of leaving "has dom focus" in this list, @JAWS-test -- what scenario satisfies "has dom focus" while not satisfying something else from the list? Happy to be convinced otherwise but it seems unnecessary to add.

Otherwise I'm good with the change just waiting to leave a +1 until I understand :)

JAWS-test:
@spectranaut 

Two reasons for leaving "has dom focus":
1. this was in it so far, removing this now would make it no longer compatible with old versions.
2. if a list entry has dom focus, the other conditions don't necessarily have to be met (e.g. focus can be set via JS without tabindex=0 at the list entry). The new conditions were only added so that the selected item can be determined even if the item does not have the focus.

But I don't mind if dom focus is removed either. In the end it's just about providing a repair mechanism when no aria-selected is used

smhigley (approved):

index.html
```diff
@@ -6025,7 +6025,7 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
           <li>None of the <rref>option</rref> elements in the <rref>listbox</rref> have an explicitly declared value for <sref>aria-selected</sref> or <sref>aria-checked</sref>.</li>
         </ul>
         <p>
-          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or the <rref>listbox</rref> has DOM focus and the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
+          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <rref>option</rref> has DOM focus or has <code>tabindex="0"</code> (or <code>tabindex</code> with a value greater than 0), or the <rref>option</rref> is referenced by <pref>aria-activedescendant</pref>.
```
scottaohara, line 6028: trying to simplify this (DOM focus, tabindex=0, tabindex=1+ all seem to just imply the option has keyboard focus... am i missing something to think otherwise?) and bringing in James's requested change.

```suggestion
          If a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>true</code> if the <code>option</code> has keyboard focus, or the <code>option</code> is referenced by <pref>aria-activedescendant</pref> so long as the <code>option</code> is not hidden or inert.
```

JAWS-test: @scottaohara But this would not solve the problem of #1661, that a listbox has a selected option only if it has keyboard focus (unless it uses aria-selected or aria-activedescandant). #1661 can only be solved with additional conditions for the non-focused state or by removing implicit aria-selected.

JAWS-test: If we define hidden and inert as exclusion reasons, then disabled should be one as well

smhigley:
> I think more work is need to avoid hidden or inert elements that still match this criteria... It may be as simple as adding, "…and are not hidden or inert…" but I'm not in a position to consider prose suggestions deeply at the moment.

@cookiecrook question b/c I have no idea how this is implemented in browsers -- how important is it to specify that hidden/inert options are not implicitly selected if they aren't exposed anyway? Just from an authoring standpoint, I don't think I would care about that detail unless future AOM stuff makes it possible to somehow query the implicit selection of an ARIA listbox.

I do agree with @JAWS-test that disabled options should not get implicit selection no matter what.

index.html
scottaohara: i'm resolving this as talkign with Sarah about this, we understand why this won't work.

scottaohara:
what happens if all options are set to tabindex 0+?   everything is selected?  that seems wrong.... but also, trying to guess at which is the correct selection is more wrong.

smhigley:
Closing this PR, since #1719 was merged

PR #1683, opened 2022-01-23
Handling Author Errors: form & region roles

### LINKED ISSUE #1083: Add handling author errors update for unnamed region and form landmarks?
`region` requires a name from authors, and [`form` will require author naming](https://github.com/w3c/aria/pull/1071).

If names are not provided, should we note in the Handling Author Errors section that these landmarks should not be exposed if authors fail to provide them?

HTML AAM already notes that if names are not provided, then `section` and `form` have no corresponding role (will be updated to `role=generic`).

scottaohara:
closes #1083

adds initial draft content to the handling author error section to call out expected behavior for `form` and `region` landmark roles which have not been provided accessible names.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1683.html" title="Last updated on Mar 7, 2022, 8:00 PM UTC (ea92b0c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1683/157d711...ea92b0c.html" title="Last updated on Mar 7, 2022, 8:00 PM UTC (ea92b0c)">Diff</a>


smhigley (commented):
For `form`, this looks like a substantive change from what happens currently. It looks like `<div role="form">` is still exposed as a form landmark even without a name (checked Firefox and Chromium on this quick example: https://jsfiddle.net/py96hL7b/, and Safari appears to map to `Group` whether named or not).

Is that something we actually want to change? Out of all landmarks with name required, I can most see `form` being valuable to expose anyway.

scottaohara:
Fair point.  The intent here was to align how HTML AAM treats unnamed `<form>`.  

If we do not want to make this change then this is really just to better specify what happens with `region`, and arguably that could just be part of the definition of that role, since it'd be unique to that one alone.

smhigley:
To be fair, aligning to an unnamed `<form>` also makes sense. I'm not really against either approach, maybe we should just make sure browser folks are on board with updating the mappings?

scottaohara:
Adding agenda label to hopefully get this quickly resolved.

Solutions - keep PR as is if people agree that an unnamed `form` role should behave like HTML's `<form>` and not be exposed if unnamed.   Webkit mapping to group seems odd regardless.

Alternative, we decide that `form` if not named should still show up as a landmark (in which case, is it really required to name it???).  But if so, then maybe this PR changes to note `region`'s behavior solely within the `region` definition.  Not sure if we need to call this out in handling authors errors if it's behavior just for this one landmark.

scottaohara:
@smhigley @MelSumner I have finished my TODO, so this is ready to review again.

@cookiecrook I added you as a reviewer as it was your comment (if i recall correctly) that prompted me to add the new bits to this PR.  If you do not wish to review, please ignore.

MelSumner (approved):
As an aside, I wonder if browsers could add some sort of warning to the console log. 🤔 

Anyway, thank you for doing this work @scottaohara 👍  :shipit:

smhigley (approved):

scottaohara:
question, but should I also update the `form` and `region` roles following text:
>[User agents](https://w3c.github.io/aria/#dfn-user-agent) SHOULD treat elements with role form as navigational [landmarks](https://w3c.github.io/aria/#dfn-landmark).

to 

>[User agents](https://w3c.github.io/aria/#dfn-user-agent) SHOULD treat elements with **named** role form as navigational [landmarks](https://w3c.github.io/aria/#dfn-landmark).

bold "named" for emphasis, but not because i think it should be bold

smhigley:
@scottaohara that sounds like a good idea to me 👍

scottaohara:
thanks @smhigley.  added some words to those lines for each role.

cookiecrook (approved):

PR #1698, opened 2022-03-04
proposed rewording for the complementary landmark

### LINKED ISSUE #1396: Discussion: aside vs complementary
Note: This issue is a proxy for https://github.com/w3c/html-aam/issues/86, but I want to bring the issue to the wider group and discuss how we want to proceed. If we decide to make this change, I can provide PRs for HTML-AAM and HTML-ARIA.

---

The [ARIA definition of complementary](https://w3c.github.io/aria/#complementary) says:
> A landmark that is designed to be **complementary to the main content at a similar level in the DOM hierarchy**…

The [HTML definition of aside](https://html.spec.whatwg.org/multipage/sections.html#the-aside-element) says:
> The aside element represents a section of a page that consists of content that is **tangentially related to the content around the aside element**… The element can be used for typographical effects like pull quotes or sidebars, for advertising, for groups of nav elements...

Here's the thing: I don't think pull quotes should be landmarks. I don't think sidebars within an article should be landmarks, either.

I think the following points in the [APG's guidance for complementary](https://w3c.github.io/aria-practices/#aria_lh_complementary) say it well:

- complementary landmarks should be top level landmarks (e.g. not contained within any other landmarks).
- If the complementary content is not related to the main content, a more general role should be assigned (e.g. region).

Since HTML `aside` is more permissive than ARIA `complementary`, and since [`aside` maps directly to `complementary`](https://w3c.github.io/html-aam/#el-aside) without restriction, authors are defining landmarks that they didn't intend to be landmarks. Possibly lots of them.

Interestingly, [axe-core flags a violation](https://github.com/dequelabs/axe-core/issues/795) for an [aside or complementary that is not at the top level](https://dequeuniversity.com/rules/axe/3.3/landmark-complementary-is-top-level). Of course, this means that authors who don't want any axe-core violations can't use asides in the "put 'em anywhere" way defined in the HTML spec... unless they specify some other role on the aside, like perhaps `role="presentation"` or `role="note"`.

But that puts the onus on authors to know what's going on, understand ARIA, and come up with a work around. More likely, they would just delete any flagged asides and wonder why they misunderstood the HTML spec.

I think the axe-core rule should stay. I think it's a good rule. And I think we can help authors by refining the mapping for the aside element, which I believe would make this better for users of AT.

I propose that we restrict the mapping of aside so that it only maps to complementary if it's a top level landmark (i.e. scoped to the body element), similar to the language for header and footer in HTML-AAM and HTML-ARIA, except that instead of mapping to "No corresponding role" when it's **not** scoped to the body element, I think it probably makes more sense to map it to `role="note"`. This gives AT the opportunity to announce when entering/leaving the note (some already do), or AT could provide a means for users to skip over the note if desired.

I think this would greatly clarify what asides are and what they are for, and it would reduce the number of frivolous complementary landmarks out there that had no business being landmarks in the first place.

scottaohara:
closes #1396

might need some more wordsmithing, but something to go on / talk about at least

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1698.html" title="Last updated on Aug 30, 2022, 10:00 PM UTC (2c3bae3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1698/d925596...2c3bae3.html" title="Last updated on Aug 30, 2022, 10:00 PM UTC (2c3bae3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2630,7 +2630,7 @@
 		<div class="role" id="complementary">
 			<rdef>complementary</rdef>
 			<div class="role-description">
-				<p>A <rref>landmark</rref> that is designed to be complementary to the main content it is a direct descendant of, or is at a similar level in the DOM hierarchy. The contents of a complementary landmark would be expected to remain meaningful if it were to be separated from the main content it is relevant to.</p>
+				<p>A <rref>landmark</rref> that is designed to be complementary to the main content it is a sibling to, or a direct descendant of. The contents of a complementary landmark would be expected to remain meaningful if it were to be separated from the main content it is relevant to.</p>
 				<p>There are various types of content that would appropriately have this <a>role</a>. For example, in the case of a portal, this may include but not be limited to show times, current weather, related articles, or stocks to watch. If the complementary content is completely separable from the main content, it may be appropriate to use a more general role.</p>
 				<p><a>Assistive technologies</a> SHOULD enable users to quickly navigate to elements with role <code>complementary</code>.
   				<a>User agents</a> SHOULD treat elements with role <code>complementary</code> as navigational <a>landmarks</a>.
--- a/index.html
+++ b/index.html
@@ -2630,8 +2630,8 @@
 		<div class="role" id="complementary">
 			<rdef>complementary</rdef>
 			<div class="role-description">
-				<p>A <rref>landmark</rref> that is designed to be complementary to the main content at a similar level in the DOM hierarchy, but remaining meaningful when separated from the main content.</p>
-				<p>There are various types of content that would appropriately have this <a>role</a>. For example, in the case of a portal, this may include but not be limited to show times, current weather, related articles, or stocks to watch. The complementary role indicates that contained content is relevant to the main content. If the complementary content is completely separable from the main content, it may be appropriate to use a more general role.</p>
+				<p>A <rref>landmark</rref> that is designed to be complementary to the main content it is a direct descendant of, or is at a similar level in the DOM hierarchy. The contents of a complementary landmark would be expected to remain meaningful if it were to be separated from the main content it is relevant to.</p>
+				<p>There are various types of content that would appropriately have this <a>role</a>. For example, in the case of a portal, this may include but not be limited to show times, current weather, related articles, or stocks to watch. If the complementary content is completely separable from the main content, it may be appropriate to use a more general role.</p>
 				<p><a>Assistive technologies</a> SHOULD enable users to quickly navigate to elements with role <code>complementary</code>.
   				<a>User agents</a> SHOULD treat elements with role <code>complementary</code> as navigational <a>landmarks</a>.
   				<a>User agents</a> MAY enable users to quickly navigate to elements with role <code>complementary</code>.</p>
--- a/index.html
+++ b/index.html
@@ -2630,7 +2630,7 @@
 		<div class="role" id="complementary">
 			<rdef>complementary</rdef>
 			<div class="role-description">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2630,8 +2630,8 @@ <h2>Definition of Roles</h2>
 		<div class="role" id="complementary">
 			<rdef>complementary</rdef>
 			<div class="role-description">
-				<p>A <rref>landmark</rref> that is designed to be complementary to the main content at a similar level in the DOM hierarchy, but remaining meaningful when separated from the main content.</p>
-				<p>There are various types of content that would appropriately have this <a>role</a>. For example, in the case of a portal, this may include but not be limited to show times, current weather, related articles, or stocks to watch. The complementary role indicates that contained content is relevant to the main content. If the complementary content is completely separable from the main content, it may be appropriate to use a more general role.</p>
+				<p>A <rref>landmark</rref> that is designed to be complementary to the main content it is a direct descendant of, or is at a similar level in the DOM hierarchy. The contents of a complementary landmark would be expected to remain meaningful if it were to be separated from the main content it is relevant to.</p>
```
jnurthen, line 2633: Not sure about this. I think what we mean is that it should be either a sibling landmark or a child landmark in the landmark hierarchy. I'm not sure why it matters where it is in the DOM hierarchy. (ie <main><div><div><div><complementary></div></div></div>...</main> should be absolutely fine). I don't think we've ever defined a concept of landmark hierarchy though so don't think we can use this terminology.

scottaohara: simplifying the terminology sounds good to me.  I think "DOM hierarchy" is likely a cause of confusion for people here anyway, so I probably should have gotten rid of that term anyway.

scottaohara, line 2633:
```suggestion
				<p>A <rref>landmark</rref> that is designed to be complementary to the main content it is a sibling to, or a direct descendant of. The contents of a complementary landmark would be expected to remain meaningful if it were to be separated from the main content it is relevant to.</p>
```

thoughts?

spectranaut: Add a second "that" for readability: `A landmark that is designed to be complementary to the main content that it is a sibling to, or a direct descendant of.`
other than that looks good!

spectranaut (approved):

index.html
```diff
@@ -2630,8 +2630,8 @@ <h2>Definition of Roles</h2>
 		<div class="role" id="complementary">
 			<rdef>complementary</rdef>
 			<div class="role-description">
-				<p>A <rref>landmark</rref> that is designed to be complementary to the main content at a similar level in the DOM hierarchy, but remaining meaningful when separated from the main content.</p>
-				<p>There are various types of content that would appropriately have this <a>role</a>. For example, in the case of a portal, this may include but not be limited to show times, current weather, related articles, or stocks to watch. The complementary role indicates that contained content is relevant to the main content. If the complementary content is completely separable from the main content, it may be appropriate to use a more general role.</p>
+				<p>A <rref>landmark</rref> that is designed to be complementary to the main content it is a sibling to, or a direct descendant of. The contents of a complementary landmark would be expected to remain meaningful if it were to be separated from the main content it is relevant to.</p>
```
scottaohara, line 2633:
```suggestion
				<p>A <rref>landmark</rref> that is designed to be complementary to the main content that it is a sibling to, or a direct descendant of. The contents of a complementary landmark would be expected to remain meaningful if it were to be separated from the main content it is relevant to.</p>
```

MelSumner (approved):
This seems like a bit of an improvement. 👍  :shipit:

pkra:
ping @jnurthen

jnurthen (approved):

scottaohara:
@jnurthen i guess this is what happens when PRs linger for a long time, as I forgot I even made this and in turn made https://github.com/w3c/aria/pull/1778 the other day.  Do we like one wording over the other?  :(

PR #1703, opened 2022-03-12
revise caption definition

### LINKED ISSUE #1696: Labelling listbox option groups
Right now if I were to tell someone how to provide a visible label for their listbox option groups, i'd have to recommend a markup pattern like:

```html
<div role=listbox ...>
  <div role=group aria-labelledby=f>
    <div id=f aria-hidden=true>group label</div>
    <div role=option ...>...</div>
    ...
  </div>
</div>
```

And fine.  we can get by with that.  But is there any reason we could not say that `caption` was an allowed child of a `group` in this context?  

```html
<div role=listbox ...>
  <div role=group aria-labelledby=f>
    <div id=f role=caption>group label</div>
    <div role=option ...>...</div>
    ...
  </div>
</div>
```
In this context, the `caption` would only ever be accessed directly by AT if someone were to switch out of forms mode when interacting with the listbox.  Arguably, we could also tell AT that in this context the `caption` should be ignored... though I could see good reason to not do that as well.

Also, if we re-pickup the idea of naming an ancestor from a descendant role, this could eventually turn into:

```html
<div role=listbox ...>
  <div role=group>
    <div role=caption>group label</div>
    <div role=option ...>...</div>
    ...
  </div>
</div>
```

scottaohara:
This change in the definition is related to the changes in `figure` and `figcaption` in HTML AAM: https://github.com/w3c/html-aam/pull/359 and introduces the idea that a `caption` may contain structured content - and in this PR - `aria-details` is referenced as a way authors should reference such content if within a `caption`

Additionally, this PR extends the definition to allow `caption` to be used for purposes of naming/describing a `group` or `radiogroup`, which fills some gaps from the dropped `legend` role.

**NOTE:** `label` and `legend` are still listed on [the wiki page for role parity](https://github.com/w3c/aria/wiki/Plans-regarding-role-parity#originally-planned-but-moved-out-of-1-2-potentially-to-1-3).  Are these still relevant or not?

If these updates are accepted, this would help pave the way to resolve #1696 as well.

***

### :boom: Error: 504 Gateway Time-out :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 28, 2022, 11:53 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Faria%2F7a58493dbc32f74d6eb8ecb06bb5e02318915111%2Findex.html%3FisPreview%3Dtrue)

```
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%231703.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1830,7 +1830,7 @@
 				      &lt;div id="name"&gt;Sales information for 20XX&lt;/div&gt;
 				      &lt;div id="details"&gt;
 				        This barchart represents the total amount of sales over the course
-				        of five years. <a href="...">Sales information for last year</a> can
+				        of five years. &lt;a href="...">Sales information for last year&lt;/a> can
 				        be reviewed, or you can overlay <button aria-pressed="false">previous year</button>
 				        information in this graphic.
 				      &lt;/div&gt;
--- a/index.html
+++ b/index.html
@@ -1800,8 +1800,8 @@
 				<p>If the <code>caption</code> represents an accessible name for its containing element, authors SHOULD specify <pref>aria-labelledby</pref> on the parent element to reference the element with role <code>caption</code>.</p>
 				
 				<pre class="example highlight">
-					&lt;div role="radiogroup" aria-labelledby="name"&gt;
-				    &lt;div role="caption"&gt;
+					&lt;div role="radiogroup" aria-labelledby="cap"&gt;
+				    &lt;div role="caption" id="cap"&gt;
 				      Choose your favorite fruit
 				    &lt;/div&gt;
 				    &lt;!-- ... --&gt;
--- a/index.html
+++ b/index.html
@@ -1790,15 +1790,25 @@
 		<div class="role" id="caption">
 			<rdef>caption</rdef>
 			<div class="role-description">
-				<p>Visible content that names, and may also describe, a <rref>figure</rref>, <rref>table</rref>, <rref>grid</rref>, or <rref>treegrid</rref>.</p>
+				<p>Visible content that names, or describes a <rref>group</rref>, <rref>figure</rref>, <rref>table</rref>, <rref>grid</rref>, <rref>radiogroup</rref>, or <rref>treegrid</rref>.</p>
 				<p>When using <code>caption</code> authors SHOULD ensure:</p>
 				<ul>
-					<li>The <code>caption</code> is a direct child of a <rref>figure</rref>, <rref>table</rref>, <rref>grid</rref>, or <rref>treegrid</rref>.</li>
-					<li>The <code>caption</code> is the first child of a <rref>table</rref>, <rref>grid</rref>, or <rref>treegrid</rref>.</li>
-					<li>The <code>caption</code> is the first or last child of a <rref>figure</rref>.</li>
+					<li>The <code>caption</code> is a direct child of a <rref>group</rref>, <rref>figure</rref>, <rref>grid</rref>, <rref>radiogroup</rref>, <rref>table</rref>, or <rref>treegrid</rref>.</li>
+					<li>The <code>caption</code> is the first non-<code>generic</code> child of a <rref>group</rref>, <rref>radiogroup</rref>, <rref>grid</rref>, <rref>table</rref>, or <rref>treegrid</rref>.</li>
+					<li>The <code>caption</code> is the first or last non-<code>generic</code> child of a <rref>figure</rref>.</li>
 				</ul>
-				<p>Authors SHOULD set <pref>aria-labelledby</pref> on the parent <code>figure</code>, <code>table</code>, <code>grid</code>, or <rref>treegrid</rref> to reference the element with role <code>caption</code>. However, if a <code>caption</code> contains content that serves as both a name and description for its parent, authors MAY instead set <pref>aria-labelledby</pref> to reference an element within the <code>caption</code> that contains a concise name, and set <pref>aria-describedby</pref> to reference an element within the <code>caption</code> that contains the descriptive content.</p>
-
+				<p>If the <code>caption</code> represents an accessible name for its containing element, authors SHOULD specify <pref>aria-labelledby</pref> on the parent element to reference the element with role <code>caption</code>.</p>
+				
+				<pre class="example highlight">
+					&lt;div role="radiogroup" aria-labelledby="name"&gt;
+				    &lt;div role="caption"&gt;
+				      Choose your favorite fruit
+				    &lt;/div&gt;
+				    &lt;!-- ... --&gt;
+				</pre>
+				
+				<p>If a <code>caption</code> contains content that serves as both a name and description for its parent, authors MAY instead specify <pref>aria-labelledby</pref> to reference an element within the <code>caption</code> that represents the "name" of the parent element, and specify <pref>aria-describedby</pref> to reference an element within the <code>caption</code> that contains the descriptive content.</p>
+				
 				<pre class="example highlight">
 					&lt;div role="table" aria-labelledby="name" aria-describedby="desc"&gt;
 				    &lt;div role="caption"&gt;
@@ -1810,6 +1820,24 @@
 				    &lt;/div&gt;
 				    &lt;!-- ... --&gt;
 				</pre>
+				
+				<p>If the <code>caption</code> represents a long-form description, or if the description contains semantic elements which are important in understanding the description, authors MAY instead specify <pref>aria-labelledby</pref> to reference an element within the <code>caption</code> that represents the "name" of the parent element, and specify <pref>aria-details</pref> to reference an element within the <code>caption</code> that contains the descriptive content.</p>
+
+				<pre class="example highlight">
+					&lt;div role="figure" aria-labelledby="name" aria-details="details"&gt;
+					  &lt;!-- figure content here, such as a complex data viz SVG -->
+				    &lt;div role="caption"&gt;
+				      &lt;div id="name"&gt;Sales information for 20XX&lt;/div&gt;
+				      &lt;div id="details"&gt;
+				        This barchart represents the total amount of sales over the course
+								of five years. <a href="...">Sales information for last year</a> can
+								be reviewed, or you can overlay <button aria-pressed="false">previous year</button>
+								information in this graphic.
+				      &lt;/div&gt;
+				    &lt;/div&gt;
+				    &lt;!-- ... --&gt;
+				</pre>
+				
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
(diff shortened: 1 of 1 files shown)


scottaohara:
A question I have is if we _need_ the requirements about where the `caption` should be placed in these different markup patterns?  We had done this to mimic HTML's requirements for their similar elements.... but... hindsight I guess... do we actually need to be as prescriptive here?  Especially since these are still being associated via IDs, so it's not actually important _where_ the content is actually placed in the DOM.   And if we do re-introduce the idea of named by encapsulation, would being this strict actually be necessary?

index.html
```diff
@@ -1790,15 +1790,25 @@ <h2>Definition of Roles</h2>
 		<div class="role" id="caption">
 			<rdef>caption</rdef>
 			<div class="role-description">
-				<p>Visible content that names, and may also describe, a <rref>figure</rref>, <rref>table</rref>, <rref>grid</rref>, or <rref>treegrid</rref>.</p>
+				<p>Visible content that names, or describes a <rref>group</rref>, <rref>figure</rref>, <rref>table</rref>, <rref>grid</rref>, <rref>radiogroup</rref>, or <rref>treegrid</rref>.</p>
 				<p>When using <code>caption</code> authors SHOULD ensure:</p>
```
smhigley, line 1794: Based on your comment, maybe rather than a "SHOULD", this should be a general recommendation with the justification that is what would match the HTML pattern?

On the other hand, for table/grid, it might actually be a SHOULD. I haven't tested what happens if you throw a random caption in the middle of rows 😆

index.html
```diff
@@ -1810,6 +1820,24 @@ <h2>Definition of Roles</h2>
 				    &lt;/div&gt;
 				    &lt;!-- ... --&gt;
 				</pre>
+				
+				<p>If the <code>caption</code> represents a long-form description, or if the description contains semantic elements which are important in understanding the description, authors MAY instead specify <pref>aria-labelledby</pref> to reference an element within the <code>caption</code> that represents the "name" of the parent element, and specify <pref>aria-details</pref> to reference an element within the <code>caption</code> that contains the descriptive content.</p>
```
smhigley, line 1824: For this (and sort-of for the previous example), why would `aria-labelledby` need to reference an element within the caption?

For example, take this table that is named by a heading preceding it, and described or details-ed by the caption:

```html
<h2 id="table-name">Prices</h2>
<div role="table" aria-labelledby="table-name" aria-describedby/details="table-caption">
  <div role="caption">All prices in this table are blah blah blah something boring and long</div>
  <!-- table rows -->
</div>
```

I wonder if it'd be good to modify one of the details/desc examples to have the caption _only_ provide a description with the name coming from elsewhere, and update the wording to not specify that `aria-labelledby` needs to point within the caption

index.html
scottaohara: Right?  I'm not convinced this needs to be a 'should'.... but if it remains it needs to be clarified that if it is a child of these elements, it needs to reflect the way it's used in HTML...

index.html
scottaohara: Another example that illustrates this makes sense to me... so long as we aren't getting too far into APG example territory?

the intent of the caption being to provide a name, description or both is why the wording is written the way it is.  But to your point, someone could go and use a caption purely for descriptive purposes and get a name from somewhere else or use an aria-label... or leave it nameless if ... REASONS ....  so yeh.  one more example probably makes sense and then future updates could go into APG / some of this could go to APG if/when needed.

index.html
scottaohara: after thinking more about this, i'm leaving this as is for now.  we could always change in another PR if need be.

MelSumner (approved):
Thank you for working on this.

index.html
```diff
@@ -1790,15 +1790,25 @@ <h2>Definition of Roles</h2>
...
 			<rdef>caption</rdef>
 			<div class="role-description">
-				<p>Visible content that names, and may also describe, a <rref>figure</rref>, <rref>table</rref>, <rref>grid</rref>, or <rref>treegrid</rref>.</p>
+				<p>Visible content that names, or describes a <rref>group</rref>, <rref>figure</rref>, <rref>table</rref>, <rref>grid</rref>, <rref>radiogroup</rref>, or <rref>treegrid</rref>.</p>
 				<p>When using <code>caption</code> authors SHOULD ensure:</p>
 				<ul>
-					<li>The <code>caption</code> is a direct child of a <rref>figure</rref>, <rref>table</rref>, <rref>grid</rref>, or <rref>treegrid</rref>.</li>
-					<li>The <code>caption</code> is the first child of a <rref>table</rref>, <rref>grid</rref>, or <rref>treegrid</rref>.</li>
-					<li>The <code>caption</code> is the first or last child of a <rref>figure</rref>.</li>
+					<li>The <code>caption</code> is a direct child of a <rref>group</rref>, <rref>figure</rref>, <rref>grid</rref>, <rref>radiogroup</rref>, <rref>table</rref>, or <rref>treegrid</rref>.</li>
```
pkra, line 1796: Maybe `direct child` be expanded to `owned child`. (I realize it's only a SHOULD statement but it seems like checkers shouldn't complain if it's owned.)

pkra:
> A question I have is if we _need_ the requirements about where the `caption` should be placed in these different markup patterns? We had done this to mimic HTML's requirements for their similar elements.... but... hindsight I guess... do we actually need to be as prescriptive here? Especially since these are still being associated via IDs, so it's not actually important _where_ the content is actually placed in the DOM. And if we do re-introduce the idea of named by encapsulation, would being this strict actually be necessary?

I agree that it seems no unnecessary but it's also just a SHOULD statement; I feel like authors should not mind automated checkers to question their decision a little.

index.html
```diff
@@ -1810,6 +1820,38 @@ <h2>Definition of Roles</h2>
...
+				<p>If the <code>caption</code> represents a long-form description, or if the description contains semantic elements which are important in understanding the description, authors MAY instead specify <pref>aria-labelledby</pref> to reference an element within the <code>caption</code> that represents the "name" of the parent element, and specify <pref>aria-details</pref> to reference an element within the <code>caption</code> that contains the descriptive content.</p>
+
+				<pre class="example highlight">
+					&lt;div role="figure" aria-labelledby="name" aria-details="details"&gt;
+					  &lt;!-- figure content here, such as a complex data viz SVG -->
+				    &lt;div role="caption"&gt;
+				      &lt;div id="name"&gt;Sales information for 20XX&lt;/div&gt;
+				      &lt;div id="details"&gt;
+				        This barchart represents the total amount of sales over the course
+				        of five years. <a href="...">Sales information for last year</a> can
```
scottaohara, line 1833:
```suggestion
				        of five years. &lt;a href="...">Sales information for last year&lt;/a> can
```

index.html
```diff
@@ -1810,6 +1820,38 @@ <h2>Definition of Roles</h2>
...
+				        be reviewed, or you can overlay <button aria-pressed="false">previous year</button>
```
scottaohara, line 1834:
```suggestion
				        be reviewed, or you can overlay &lt;button aria-pressed="false">previous year&lt;/button>
```

index.html
```diff
@@ -1810,6 +1820,38 @@ <h2>Definition of Roles</h2>
...
+				</pre>
+				
+				<p>There may be instances where a <code>caption</code> contains only a description, without a suitable text string to serve as the accessible name for the parent element. In such instances, <pref>aria-label</pref> or <pref>aria-labelledby</pref> MAY be used to provide an accessible name, and the <code>caption</code> MAY be treated solely as descriptive content.
+					
+				<pre class="example highlight">
+					&lt;div role="figure" aria-label="Sales information" aria-details="details"&gt;
+					  &lt;!-- figure content here, such as a complex data viz SVG -->
+				    &lt;div role="caption" id="details"&gt;
+				      This barchart represents the total amount of sales over the course
+				      of five years. <a href="...">Sales information for last year</a> can
```
scottaohara, line 1848:
```suggestion
				      of five years. &lt;a href="...">Sales information for last year&lt;/a> can
```

index.html
```diff
@@ -1810,6 +1820,38 @@ <h2>Definition of Roles</h2>
...
+				      be reviewed, or you can overlay <button aria-pressed="false">previous year</button>
```
scottaohara, line 1849:
```suggestion
				      be reviewed, or you can overlay &lt;button aria-pressed="false">previous year&lt;/button>
```

index.html
scottaohara:
```suggestion
					<li>The <code>caption</code> is an owned child of a <rref>group</rref>, <rref>figure</rref>, <rref>grid</rref>, <rref>radiogroup</rref>, <rref>table</rref>, or <rref>treegrid</rref>.</li>
```

scottaohara: i had made the suggested edit, but deleted it as it is awkward per the following "caption is the first non-generic child of..." requirement.

to explain further, consider if I have the following markup

```
<div role=caption id=c ...>my amazing caption</div>
<div role=radiogroup aria-owns=c>
  <!-- all my radios go here -->
</div>
```
now the caption is 'owned' by the radiogroup.... but it gets slotted at the end of the radiogroup's content resulting as if the following markup were what had been written:

```
<div role=radiogroup ...>
  <!-- all my radios go here -->
  <div role=caption id=c>my amazing caption</div>
</div>
```

so now this fails the second requirement of the caption being the 'first child'.  

Someone would actually have to know to *not* have the radiogroup be the element to use aria-owns on to bring in the caption, but rather insert an empty div as the first child of the radiogroup and then have THAT element own the caption so that it could be slotted in at the right place.

```
<div role=caption id=c ...>my amazing caption</div>
<div role=radiogroup ...>
  <div aria-owns-c></div> <!-- insert an empty div to slot in the caption, rather than just putting the caption here. convoluted markup is convoluted -->
  <!-- all my radios go here -->
</div>
```

This again gets to my other comment that these requirements aren't actually useful in reality without name from encapsulation.

pkra: Yikes - yes, that makes a lot of sense. (It's only a should in the end.)

scottaohara:
>I feel like authors should not mind automated checkers questioning their decision a little.

i wish this were the case (i truly do and personally I agree with you here), but reality has shown me that developers will argue against legitimate errors that conformance checkers flag. warnings like this, where changing the markup wouldn't really result in a more accessible experience, only fuel their arguments of whether or not a11y checks actually matter :(

smhigley (approved):

pkra:
@scottaohara I wanted to merge as is since this had plenty of review. Could you propose the looser positioning requirements in a new PR?

scottaohara:
Thanks @pkra.  have opened an issue to get a quick discussion on the pathway forward there.  Can make a PR based on what people think about that.

PR #1719, opened 2022-04-07
Update authoring requirement for aria-selected on options

smhigley:
A different solution for #1661, discussed on 4/7 and summarized in this comment: https://github.com/w3c/aria/issues/1661#issuecomment-1092024562

I wrangled the existing wording around single vs. multiselect a little because the new `SHOULD` text about `aria-selected` overlapped with the existing single vs. multi-select wording, so I consolidated a bit.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1719.html" title="Last updated on Apr 9, 2022, 6:50 PM UTC (146e602)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1719/5051beb...146e602.html" title="Last updated on Apr 9, 2022, 6:50 PM UTC (146e602)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6030,11 +6030,12 @@
         </p>
         <p>
           Authors SHOULD indicate selection for <rref>option</rref> elements using one of the following:
-				</p>
-				<ul>
-					<li>An <sref>aria-selected</sref> value of <code>true</code> on the selected option within a single-select <rref>listbox</rref>, and optionally <sref>aria-selected</sref> values of <code>false</code> on unselected options.</li>
-					<li>either <sref>aria-selected</sref> or <sref>aria-checked</sref> on all options within a multi-select <rref>listbox</rref></li>
-				<p>
+	</p>
+	<ul>
+		<li>An <sref>aria-selected</sref> value of <code>true</code> on the selected option within a single-select <rref>listbox</rref>, and optionally <sref>aria-selected</sref> values of <code>false</code> on unselected options.</li>
+		<li>Either <sref>aria-selected</sref> or <sref>aria-checked</sref> on all options within a multi-select <rref>listbox</rref></li>
+	</ul>
+	<p>
           Authors SHOULD NOT specify both <sref>aria-selected</sref> and <sref>aria-checked</sref> on <rref>option</rref> elements contained by the same <rref>listbox</rref> except in the extremely rare circumstances where all the following conditions are met:
         </p>
         <ul>
--- a/index.html
+++ b/index.html
@@ -6033,9 +6033,9 @@
         </p>
         <ul>
           <li>An <sref>aria-selected</sref> value of <code>true</code> on the selected option within a single-select <rref>listbox</rref>, and optionally <sref>aria-selected</sref> values of <code>false</code> on unselected options.</li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -6029,8 +6029,12 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
           Otherwise, if a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>false</code>.
         </p>
         <p>
-          Authors MAY indicate selection for <rref>option</rref> elements  using either <sref>aria-selected</sref> or <sref>aria-checked</sref>.
-          Some user interfaces indicate selection with <sref>aria-selected</sref> in single-select list boxes and with <sref>aria-checked</sref> in multi-select list boxes.
+          Authors SHOULD indicate selection for <rref>option</rref> elements using one of the following:
+				</p>
```
scottaohara, line 6033:
```suggestion
	</p>
```

index.html
```diff
@@ -6029,8 +6029,13 @@ <h5>Note regarding the ARIA 1.1 <code>none</code> role.</h5>
           Otherwise, if a user agent provides an implicit <sref>aria-selected</sref> value for an <rref>option</rref>, the value SHOULD be <code>false</code>.
         </p>
         <p>
-          Authors MAY indicate selection for <rref>option</rref> elements  using either <sref>aria-selected</sref> or <sref>aria-checked</sref>.
-          Some user interfaces indicate selection with <sref>aria-selected</sref> in single-select list boxes and with <sref>aria-checked</sref> in multi-select list boxes.
+          Authors SHOULD indicate selection for <rref>option</rref> elements using one of the following:
+        </p>
+        <ul>
+          <li>An <sref>aria-selected</sref> value of <code>true</code> on the selected option within a single-select <rref>listbox</rref>, and optionally <sref>aria-selected</sref> values of <code>false</code> on unselected options.</li>
+          <li>Either <sref>aria-selected</sref> or <sref>aria-checked</sref> on all options within a multi-select <rref>listbox</rref></li>
```
scottaohara, line 6036: Suggest being explicit with the `true` or `false` values to be inline with the previous list item indicating that authors should use a value of true.  Additionally, if not being specific here there's the potential for someone to assume that `mixed` or `undefined` are allowed... 

```suggestion
          <li>Either <sref>aria-selected</sref> or <sref>aria-checked</sref> on all options within a multi-select <rref>listbox</rref>, with the attributes' specified values of `true` or `false`, representing their current selected or unselected state.</li>
```

smhigley: I wordsmithed a bit, let me know if you like the current wording

JAWS-test:
In my opinion, this change should also be made for treeitem and not only for option.

chlane (approved):
The added specificity regarding the value looks good to me.

chlane (approved):

mcking65 (approved):
Thank you @smhigley! This looks great.

I agree with @JAWS-test  that the same language would work in treeitem; that could be a separate PR. The related issue is about option.

jnurthen:
ready to merge

PR #1733, opened 2022-04-29
Tidying up section "Characteristics of Roles"

### LINKED ISSUE #1263: Characteristics of Roles section needs tidying up
The [Characteristics of Roles](https://w3c.github.io/aria/#Properties) section has some inconsistencies that should be fixed up.

#### Problem 1:
A few (3) of the characteristics definitions have a "Value" type listed:
- [Abstract Roles](https://w3c.github.io/aria/#isAbstract)
- [Presentational Children](https://w3c.github.io/aria/#childrenArePresentational)
- [Accessible Name Calculation](https://w3c.github.io/aria/#namecalculation)

In 2 of those cases (Abstract and Presentational Children), this looks odd and out of place, and should probably just be deleted ("boolean" can be inferred from reading the prose). However, the list of values for the "Accessible Name Calculation" definition needs to be kept.

Looking back at the [same section in ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#Properties), it looks like these sections were more formally defined in the past. Informal is fine, but we should make sure they're consistent with each other and with the characteristics tables in the role definitions.

#### Problem 2:
The names of these definition sections do not always exactly match the name used in the role characteristics tables. For example, [Presentational Children](https://w3c.github.io/aria/#childrenArePresentational) is not the same as "Children Presentational".

This is particularly problematic in the case of [Accessible Name Calculation](https://w3c.github.io/aria/#namecalculation) which is "Name From" in the tables.

These definition names should all be identical to the characteristics table entries that they are defining.

#### Problem 3:
Probably because the section for [Accessible Name Calculation](https://w3c.github.io/aria/#namecalculation) is poorly named, it seems to have some extraneous subsections that I believe should be deleted. There's no reason to link to the accessible description or text alternative calculation in this (the "Name From" definition) section, so those 2 subsections (5.8.2 and 5.8.3) should be deleted.

#### Suggestion:
Consider also linking from the characteristics tables to the characteristic definitions. For example, "Supported States and Properties" in each table should link to the [Supported States and Properties](https://w3c.github.io/aria/#supportedState) definition.

pkra:
A first step.

Resolves #1263

***
<a href="https://pr-preview.s3.amazonaws.com/pkra/aria/pull/1733.html" title="Last updated on Sep 7, 2022, 4:14 PM UTC (2d8cdf5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1733/7e37849...pkra:2d8cdf5.html" title="Last updated on Sep 7, 2022, 4:14 PM UTC (2d8cdf5)">Diff</a>


index.html
```diff
@@ -724,14 +705,8 @@ <h4>Roles which cannot be named (Name prohibited)</h4>
...
-			<h3>Presentational Children</h3>
-			<dl class="runin">
-				<dt>Values</dt>
-				<dd>
-					<p>Boolean (<code>true</code> | <code>false</code>)</p>
-				</dd>
-			</dl>
-			<p>The <abbr title="Document Object Model">DOM</abbr> descendants are presentational. <a>User agents</a> SHOULD NOT expose descendants of this <a>element</a> through the platform <a>accessibility <abbr title="Application Programing Interface">API</abbr></a>. If <a>user agents</a> do not hide the descendant nodes, some information may be read twice.</p>
+			<h3>Children Presentational</h3>
+			<p>Indicates whether owned elements are presentational. <a>User agents</a> SHOULD NOT expose descendants of this <a>element</a> through the platform <a>accessibility <abbr title="Application Programing Interface">API</abbr></a>. If <a>user agents</a> do not hide the descendant nodes, some information may be read twice.</p>
```
jnurthen, line 709: Is owned elements the same as DOM descendents here? owned would include aria-owns. Does Children Presentational cascade to that? (asking the question - I'm not sure of the answer). It would seem that aria-owns should probablty be prohibited on anything with children presentational true but we don't currently specify that....

pkra: Sorry for missing this last month.

> Is owned elements the same as DOM descendents here? 

Good question. I must have been overeager.

>  Does Children Presentational cascade to that? (asking the question - I'm not sure of the answer). 

Can we add this question to the agenda?

>  It would seem that aria-owns should probablty be prohibited on anything with children presentational true but we don't currently specify that...

Makes sense to me.

pkra:
Revisiting this, I think this is ready for review.

This does not address on suggestion from #1263

> Consider also linking from the characteristics tables to the characteristic definitions. For example, "[Supported States and Properties](https://w3c.github.io/aria/#supportedState)" in each table should link to the Supported States and Properties definition.

Primarily, because we it would add a lot of noise. Also, because we might consider doing this for all characteristics (not just "supported states and properties").

index.html
jnurthen: we will tackle aria-owns restrictions in https://github.com/w3c/aria/issues/1760

pkra:
Rebased on main.

index.html
```diff
@@ -611,29 +607,14 @@ <h3>Required Context Role</h3>
...
-			</dl>
-			<section id="namecomputation">
-				<h4>Name Computation</h4>
-				<p><a href="#mapping_additional_nd_name" class="accname">Name Computation</a> is defined in the Accessible Name and Description specification.</p>
-			</section>
-			<section id="descriptioncomputation">
-				<h4>Description Computation</h4>
-				<p><a href="#mapping_additional_nd_description" class="accname">Description Computation</a> is defined in the Accessible Name and Description specification.</p>
-			</section>
-			<section id="textalternativecomputation">
```
jnurthen, line 633: #textalternativecomputation, #namecomputation and #descriptioncomputation are referenced from other places (maybe outside this document too - I don't know). Where should these references be updated to?

pkra: > Where should these references be updated to?

ouch, another oversight. They could point to accname directly, no? If they need to be reproduce, we could move this to terms.

chlane:
#1263 mentions, in Problem 1,  that the list values for the "Accessible Name Calculation" need to be kept.  But they have been removed.

pkra:
> https://github.com/w3c/aria/issues/1263 mentions, in Problem 1, that the list values for the "Accessible Name Calculation" need to be kept. But they have been removed.

I'm not sure I follow. They are under the h3 with "Name From" now. Sorry if I misunderstand what you're pointing out.

chlane (dismissed):

index.html
pkra: Fun fact: `#mapping_additional_nd_te` (which I copy&pasted from the old version) also does not seem to match an ID at this point.

pkra: `#namecomputation` is not used in the document anymore.

pkra: `#descriptioncomputation` is not used in the document anymore.

pkra: `#mapping_additional_nd_description` also needs something.

pkra: I think I fixed all relevant ones now. @jnurthen please take another look?

pkra: @jnurthen post TPAC ping.

jnurthen (approved):

jnurthen:
@pkra please fix the conflicts and we can merge

pkra:
@jnurthen I'll need to do a fresh PR. There have been too many changes the relevant parts of the spec (as is probably  expected after 3 years).

pkra:
Replaced by  #2506

PR #1762, opened 2022-06-30
editorial changes for translatable attrs

### LINKED ISSUE #1761: Editorial changes for ‘Translatable Attributes’
@bert-github wrote:

The change to the draft specification (in [section 6.4 ‘Translatable Attributes’](https://w3c.github.io/aria/#translatable-attributes)) mostly defines the language of each attribute that contains text. But a few things remain unclear:

1) The ‘lang’ attribute is not the only thing in HTML that defines the language of an element. The text should probably just say that the language of the attribute is the same as the language of the element, without talking about the ‘lang’ attribute, but referencing the HTML5 specification (in particular the [definition of language](https://html.spec.whatwg.org/#language) in section 3.2.6.2.)

2) The new text implies that the language is just for making translations, but information about the language is also needed for proper pronunciation and layout. (I.e., for ‘rendering’, in the terms of the ARIA specification.)

3) What do you do if the attribute happens to be in a different language than the contents of the element? Or if two attributes have different languages?

   In the case of an aria-label attribute, you could replace aria-label by aria-labelledby and point to a separate element. (Maybe that should be a note or an example?)

   But what do you do for aria-valuetext?

4) For visual rendering, the direction of the text may also be needed. E.g., if the document is in Arabic and the attribute contains a phone number with dashes or spaces between the digits, it is important that the whole number is displayed 
left to right, otherwise the first group of digits will be displayed to the right of the second group. ([Here is an explanation](https://www.w3.org/International/articles/inline-bidi-markup/index.en#usecase7).)

Possibly the direction could be taken from the element's ‘dir’ attribute, but, like for the language, that leaves a problem when the attribute's direction is different from the direction of the element's content. E.g., you wouldn't want to set the ‘dir’ attribute on a slider, because it reverses the slider.

_Originally posted by @bert-github in https://github.com/w3c/aria/issues/1365#issuecomment-1128042375_

cookiecrook:
Resolves #1761

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1762.html" title="Last updated on Aug 25, 2022, 4:50 PM UTC (e16e730)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1762/7e37849...e16e730.html" title="Last updated on Aug 25, 2022, 4:50 PM UTC (e16e730)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -10293,7 +10293,7 @@ button.ariaPressed; // null</pre>
 	</section>
 	<section data-cite="HTML">
 		<h2>Translatable Attributes</h2>
-		<p>The HTML specification states that other specifications can define <a data-cite="html/dom.html#translatable-attributes">translatable attributes</a>. The source translation language of each attribute value is defined by use of the [^html-global/lang^] attribute on the element or its ancestor elements.</p>
+		<p>The HTML specification states that other specifications can define <a data-cite="html/dom.html#translatable-attributes">translatable attributes</a>. The language and directionality of each attribute value is the same as the <a href="html/dom.html#language">language</a> and <a href="html/dom.html#the-directionality">directionality</a> of the element.</p>
 		<p>To be understandable by assistive technology users, the values of the following <a>states</a> and [=ARIA/properties=] are <a data-cite="html/dom.html#translatable-attributes">translatable attributes</a> and should be translated when a page is localized:</p>
 		<ul>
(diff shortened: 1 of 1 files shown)


spectranaut (approved):
Looks good!

jnurthen (approved):

PR #1778, opened 2022-08-18
Resolving user agent must not statement and whitespace defs

### LINKED ISSUE #1476: Problematic User Agents MUST NOT statement needs to be qualified
This [pull request](https://github.com/w3c/aria/pull/1408/files) introduced the following statement into ARIA 1.3:

> User Agents MUST NOT expose an ARIA attribute to the platform accessibilty API when the element's computed role prohibits use of the attribute.

Given something like `<div tabindex="0" aria-label="foo">bar</div>`, in which `aria-label` is prohibited (`generic` role), that means we need implementors to NOT expose "foo". We will not get vendor buy-in for this. Trust me. Yes, it's lousy authoring which is not ARIA compliant. But given the explicit author-provided label, user agents aren't going to ignore it in favor of "bar" or worse, nothing.

I think we need to modify the above statement to allow for User Agents to expose "foo" as part of handling author errors.

scottaohara:
Related to https://github.com/w3c/core-aam/issues/128 
Fixes #1476

- link to definition of whitespace in the aria-braillelabel and aria-brailleroledescription, aria-roledescription sections
- remove reference to whitespace from user agents must not aria-roledescription requirement
- remove author error correction User Agents MUST NOT expose prohibited ARIA attributes statement from correcting for author errors section.
	- adds authors MUST NOT specify aria-roledescription on elements with roles where the attribute is prohibited.
	- removed non-normative "should" from aria-brailleroledescription
	- adds authors MUST NOT use aria-brailleroledescription where prohibited 
	- adds authors MUST NOT use aria-label and aria-labelledby where prohibited

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1778.html" title="Last updated on Aug 25, 2022, 7:47 PM UTC (7f6baae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1778/413dcf9...7f6baae.html" title="Last updated on Aug 25, 2022, 7:47 PM UTC (7f6baae)">Diff</a>


jnurthen:
@jcsteh can you please review this.

index.html
```diff
@@ -12704,13 +12705,14 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				</ol>
 				<p class="note">Depending on the assistive technology, user verbosity settings, or other factors, certain elements' role descriptions might not be conveyed. If specifying <code>aria-roledescription</code> on such elements, then the custom role descriptions may also not be conveyed by these assistive technologies.</p>
+				<p>Additionally, authors MUST NOT specify <code>aria-roledescription</code> on an element which has an explicit or implicit WAI-ARIA role where <code>aria-roledescription</code> is <a href="#prohibitedattributes">prohibited</a>.</p>
 				<p>User agents MUST NOT expose the <code>aria-roledescription</code> property if any of the following conditions exist:</p>
 				<ol>
 					<li>The element to which <code>aria-roledescription</code> is applied has an explicit or implicit WAI-ARIA role where <code>aria-roledescription</code> is <a href="#prohibitedattributes">prohibited</a>.</li>
-					<li>The value of <code>aria-roledescription</code> is empty or contains only whitespace characters.</li>
+					<li>The value of <code>aria-roledescription</code> is the empty string.</li>
```
spectranaut, line 12715: is undefined the same as the empty string? We might want to say do not expose if undefined as well.

spectranaut: but it didn't say that before, and this change is only about removing white space, so maybe that is not necessary.

jnurthen (approved):

index.html
```diff
@@ -10629,19 +10629,20 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				</p>
 				<p>User agents MUST NOT expose the <code>aria-brailleroledescription</code> property if any of the following conditions exist:</p>
 				<ol>
-					<li>The value of <code>aria-brailleroledescription</code> is empty or contains only whitespace characters or the empty Braille pattern: dots-0 (U+2800).</li>
+					<li>The value of <code>aria-brailleroledescription</code> is empty or contains only [=ascii whitespace|whitespace=] characters or the empty Braille pattern: dots-0 (U+2800).</li>
```
cookiecrook, line 10645:
```suggestion
					<li>The value of <code>aria-brailleroledescription</code> is empty or contains only whitespace characters, which includes standard [=ascii whitespace|whitespace=] and the empty Braille pattern: dots-0 (U+2800).</li>
```
For clarify and to rule out an edge case.

index.html
```diff
@@ -11932,7 +11933,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
...label text is available in the DOM (i.e. typically visible text content), authors SHOULD use <pref>aria-labelledby</pref> and SHOULD NOT use <pref>aria-label</pref>. There may be instances where the name of an element cannot be determined programmatically from the DOM, and there are cases where referencing DOM content is not the desired user experience. Most host languages provide an attribute that could be used to name the element (e.g., the <code>title</code> attribute in [[HTML]]), yet this could present a browser tooltip. In the cases where DOM content or a tooltip is undesirable, authors MAY set the accessible name of the element using <pref>aria-label</pref>, if the element does not <a href="#prohibitedattributes">prohibit</a> use of the attribute. Authors MUST NOT specify <code>aria-label</code> on an element which has an explicit or implicit WAI-ARIA role where <code>aria-label</code> is <a href="#prohibitedattributes">prohibited</a>. As required by the <a href="#textalternativecomputation">accessible name and description computation</a>, user agents give precedence to <pref>aria-labelledby</pref> over <pref>aria-label</pref> when computing the accessible name property.</p>
```
cookiecrook, line 11936:
```
In the cases where DOM content or a tooltip is undesirable, authors MAY set the accessible name of the element using <pref>aria-label</pref>, if the element does not <a href="#prohibitedattributes">prohibit</a> use of the attribute. Authors MUST NOT specify <code>aria-label</code> on an element which has an explicit or implicit WAI-ARIA role where <code>aria-label</code> is <a href="#prohibitedattributes">prohibited</a>.
```
This felt redundant while reading... I looked back and realized it was not. 

Reversing the two sentence may resolve it? E.g. list the rule first then exception, rather than exception before the rule.

index.html
cookiecrook:
```suggestion
					<li>The value of <code>aria-roledescription</code> is undefined or the empty string.</li>
```

cookiecrook (approved):
approving after comment threads resolved

index.html
scottaohara: made a commit to try to address this - [7f6baae](https://github.com/w3c/aria/pull/1778/commits/7f6baae8e0f3219e2b7e69ba2939ee8c89fffd46)

spectranaut (approved):

PR #1787, opened 2022-09-01
Removed excess confusing language

### LINKED ISSUE #1455: reword one of examples for role presentation
The example states:
"An image that is in a container with the img role and where the full text alternative is available and is marked up with aria-labelledby and (if needed) aria-describedby;"

This took me multiple times to parse what was meant - needs rewording to avoid confusion.

My reading is that it is trying to describe the following
```
<div role="img" aria-labelledby="a">
  <span id="a">Alternative text for the image</span>
  <img role="presentation">
</div>
```

MelSumner:
If merged, this PR removes excess language in one of the examples for `role presentation`, as it was unnecessarily complex and causing confusion.
 
Resolves #1455

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1787.html" title="Last updated on Aug 29, 2023, 4:45 PM UTC (5fb951e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1787/2617046...5fb951e.html" title="Last updated on Aug 29, 2023, 4:45 PM UTC (5fb951e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6303,7 +6303,6 @@
 				<p>Example use cases:</p>
 				<ul>
 					<li>An element whose content is completely presentational (like a spacer image, decorative graphic, or clearing element);</li>
-					<li>An image that is in a container with the <rref>img</rref> <a>role</a> (all children of an <rref>img</rref> are automatically presentational);</li>
 					<li>An element used as an additional markup "hook" for <abbr title="Cascading Style Sheets">CSS</abbr>; or</li>
 					<li>A layout table and/or any of its associated rows, cells, etc.</li>
 				</ul>
--- a/index.html
+++ b/index.html
@@ -6303,7 +6303,7 @@
 				<p>Example use cases:</p>
 				<ul>
 					<li>An element whose content is completely presentational (like a spacer image, decorative graphic, or clearing element);</li>
-					<li>An image that is in a container with the <rref>img</rref> <a>role</a> and where the full text alternative is available and is marked up with <pref>aria-labelledby</pref> and (if needed) <pref>aria-describedby</pref>;</li>
+					<li>An image that is in a container with the <rref>img</rref> <a>role</a> (all children of an <rref>img</rref> are automatically presentational);</li>
 					<li>An element used as an additional markup "hook" for <abbr title="Cascading Style Sheets">CSS</abbr>; or</li>
 					<li>A layout table and/or any of its associated rows, cells, etc.</li>
 				</ul>


cookiecrook (approved):

chlane (approved):
I agree with this change. It is confusing to address textual alternatives when this is about sematics.

index.html
```diff
@@ -6303,7 +6303,7 @@ <h5>Note regarding the ARIA 1.1 <rref>none</rref> role.</h5>
 				<p>Example use cases:</p>
 				<ul>
 					<li>An element whose content is completely presentational (like a spacer image, decorative graphic, or clearing element);</li>
-					<li>An image that is in a container with the <rref>img</rref> <a>role</a> and where the full text alternative is available and is marked up with <pref>aria-labelledby</pref> and (if needed) <pref>aria-describedby</pref>;</li>
+					<li>An image that is in a container with the <rref>img</rref> <a>role</a> (all children of an <rref>img</rref> are automatically presentational);</li>
```
scottaohara, line 6306: ok... so the wording is definitely an improvement.  but this is still a confusing example.

As I understand this section, this is a list of instances where an author should know to use `role=presentation|none`.  But, if an image is a child of an element with an image role, then it shouldn't be necessary to provide the child image a role of none, since that should happen anyway (really - many implementations treat the children of an image role as if it was completely hidden).  Saying that this is an instance of where one would use role=none|presentation on the child image is just a redundancy.

So, unless i've completely missed the point of this section of content, I think this entire bullet should just be deleted.  Or, maybe if we don't want to lose the bullet, its content could instead be revised to align with this PR https://github.com/w3c/aria/pull/1830

scottaohara (changes requested):
this PR makes the wording less confusing, but the use case doesn't make any sense to me.  comments provided - https://github.com/w3c/aria/pull/1787/files#r1083293683

index.html
MelSumner, line 6306:
```suggestion
```

MelSumner:
After rebasing, it looks like this section was removed entirely in a different PR; can someone else confirm? Maybe my git-foo is off.

scottaohara:
@MelSumner - the latest editor's draft moved all the content that was under `presentation` to be under `none` instead, since that's meant to be the primary role of these two synonyms.

but, per my previous comment, your update is a good change, but i still think this bullet should just be deleted since the use case doesn't make sense for authors

MelSumner:
ok, @scottaohara thanks for clarifying! 

I'm going to open a new PR because that will just be...easier.

Issue #1493, opened 2021-05-31
Consider adding more normative requirements for authors on the textbox role


joanmarie:
The ARIA spec says the following for the [`textbox` role](https://w3c.github.io/aria/#textbox):

> A type of input that allows free-form text as its value.
> 
> If the `aria-multiline` attribute is `true`, the widget accepts line breaks within the input, as in an HTML `textarea`. Otherwise, this is a simple text box. The intended use is for languages that do not have a text input element, or cases in which an element with different semantics is repurposed as a text field.

After that there is a (non-normative) note and the characteristics table. That's it.  That suggests the following is, officially, a valid "simple text box":

```
<div role="textbox" aria-label="good luck!">
  <button>ok</button>
  <a href="#">link</a>
  <label>name: <input></label>
</div>
```

Some observations:
1. I don't think that's a "simple text box"
2. `contenteditable` isn't set but the default/implicit value of `aria-readonly` is `false`. According to the ARIA spec's [State and Property Attribute Processing](https://w3c.github.io/aria/#state_property_processing):

    > When WAI-ARIA roles are used, supported states and properties that are not present in the DOM are treated according to their default value.

    So we have to trust that the author is using some non-contenteditable means to make that `div` editable.

3. `aria-readonly` is `false` as described above. `aria-disabled` is also `false` for the same reason, but this `textbox` is also not focusable, which brings me to the next item:
4. Should this really be exposed as a textbox to ATs?

I'm currently working on fixing a Chromium bug related to authoring like the above which breaks the screen-reader user experience because those focusable children get pruned from the accessibility tree unless `contenteditable` is set on the `div` with the `textbox` role. Why are they getting pruned? Simple text boxes don't have children. (Right??)

Fixing the not-in-the-accessibility-tree bug is not hard. But I think authors need to be stopped from doing the above. And depending on the normative statements we add for authors, we might want to consider handling-author-errors content so that user agents can provide a more consistent implementation and user experience given such authoring.

JAWS-test:
Since an HTML textbox cannot have child elements, I would have expected role=textbox to have "Children Presentational: True". 
If a match is to be created between HTML input/textarea and role=textbox, I think it would make sense to add "Children Presentational: True". (But then contenteditable would be something other than a textbox, for which we might need a new ARIA role. )

[...]

PR #1788, opened 2022-09-02
Adding verbiage to allow only presentational elements in textboxes

chlane:
Adding verbiage to allow only presentational elements in textboxes https://github.com/w3c/aria/issues/1493

***
<a href="https://pr-preview.s3.amazonaws.com/chlane/aria/pull/1788.html" title="Last updated on Sep 6, 2022, 5:08 PM UTC (f4fcd18)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1788/bcf44d1...chlane:f4fcd18.html" title="Last updated on Sep 6, 2022, 5:08 PM UTC (f4fcd18)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -10352,7 +10352,7 @@ button.ariaPressed; // null</pre>
 		</section>
 		<section id="attrs_liveregions">
 			<h3>Live Region Attributes</h3>
-			<p>This section contains [=attributes=] specific to <a>live regions</a> in rich internet applications. These attributes may be applied to any <a>element</a>. The purpose of these attributes is to indicate that content changes may occur without the element having focus, and to provide <a>assistive technologies</a> with information on how to process those content updates. Some <a>roles</a> specify a default value for the <pref>aria-live</pref> attribute specific to that role. An example of a live region is a ticker section that lists updating stock quotes.</p>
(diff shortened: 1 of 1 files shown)


jnurthen:
@adampage please review

index.html
```diff
@@ -9256,6 +9256,7 @@ <h5>Presentational Role Inheritance</h5>
 			<rdef>textbox</rdef>
 			<div class="role-description">
 				<p>A type of input that allows free-form text as its value.</p>
+				<p>Authors MUST limit its children to completely presentational elements (like a spacer image, decorative graphic, or clearing element).</p>
```
scottaohara, line 9259: do we need the parenthetical examples?  

if examples are needed, maybe they should be more specific to what someone might visually display in a textbox - such as a success or error icon that visually communicates the field's `aria-invalid` state?  these could replace the references to a "spacer image" and "decorative graphic".   Those two examples should be enough, if needed, and then the "clearing element" can be dropped all together. I'm at a loss for what this means outside of a "clear button", and that would be the opposite of the guidance this PR is adding.

adampage, line 9259: Another example couple be a decorative icon corresponding to the purpose of the textbox, like the little portrait silhouette that sometimes accompanies a “Name” or “Username” field?

Jaunita-George (changes requested):
Would it be possible to remove "clearing language" and add in the error image example from the first comment and the input purpose example from the second?

PR #1793, opened 2022-09-07
[normative][author] clarify widget in trees can only be group and treeitem

### LINKED ISSUE #1251: May a treeitem contain interactive elements?
According to the specification, a treeitem can contain any interactive elements because it does not have a "Children Presentational: True" (https://w3c.github.io/aria/#treeitem), but maybe this is only so that it can contain other treeitems and groups. If a treeitem may contain interactive elements, the question would be how to access and use them with the keyboard. If it may not contain interactive elements, this should be noted in the specification.

See also: https://github.com/FreedomScientific/VFO-standards-support/issues/50

pkra:
Resolves #1251

***
<a href="https://pr-preview.s3.amazonaws.com/pkra/aria/pull/1793.html" title="Last updated on Sep 7, 2022, 6:42 PM UTC (a0e8f8d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1793/bcf44d1...pkra:a0e8f8d.html" title="Last updated on Sep 7, 2022, 6:42 PM UTC (a0e8f8d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -9700,6 +9700,7 @@
 				<p>A <rref>widget</rref> that allows the user to select one or more items from a hierarchically organized collection.</p>
 				<p>To be <a>keyboard accessible</a>, authors SHOULD manage focus of descendants for all instances of this <a>role</a>, as described in <a href="#managingfocus">Managing Focus</a>.</p>
 				<p>Elements with the role <code>tree</code> have an implicit <pref>aria-orientation</pref> value of <code>vertical</code>.</p>
+				<p>Authors MUST ensure that all [=ARIA/owned=] [=elements=] with a <rref>widget</rref> <a>role</a> have either role <code>treeitem</code> or role <rref>group</rref>.</p>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
--- a/index.html
+++ b/index.html
@@ -9700,7 +9700,7 @@
 				<p>A <rref>widget</rref> that allows the user to select one or more items from a hierarchically organized collection.</p>
 				<p>To be <a>keyboard accessible</a>, authors SHOULD manage focus of descendants for all instances of this <a>role</a>, as described in <a href="#managingfocus">Managing Focus</a>.</p>
 				<p>Elements with the role <code>tree</code> have an implicit <pref>aria-orientation</pref> value of <code>vertical</code>.</p>
-				<p>Authors MUST ensure that all [=ARIA/owned=] [=elements=] with a <rref>widget</rref> <a>role</a> have either role <code>treeitem</code> or role <rref>group</rref>.</p>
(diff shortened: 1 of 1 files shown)


ariellalgilmore (approved):

index.html
```diff
@@ -9700,6 +9700,7 @@ <h5>Presentational Role Inheritance</h5>
 				<p>A <rref>widget</rref> that allows the user to select one or more items from a hierarchically organized collection.</p>
 				<p>To be <a>keyboard accessible</a>, authors SHOULD manage focus of descendants for all instances of this <a>role</a>, as described in <a href="#managingfocus">Managing Focus</a>.</p>
 				<p>Elements with the role <code>tree</code> have an implicit <pref>aria-orientation</pref> value of <code>vertical</code>.</p>
+				<p>Authors MUST ensure that all [=ARIA/owned=] [=elements=] with a <rref>widget</rref> <a>role</a> have either role <rref>treeitem</rref> or role <rref>group</rref>.</p>
```
scottaohara, line 9703: a `group` is not a widget role though?

is this trying to say that a tree may contain elements with a `treeitem` role, or elements with a `group` role, so long as the `group` contains elements with a `treeitem` role, and that only other element must have a generic or presentation role, with no other properties important to accessibility?   e.g., `<div tabindex=lol>`

pkra: > a group is not a widget role though?

Thanks for catching that. I keep thinking they are but you are of course right.

> is this trying to say that a tree may contain elements with a treeitem role, or elements with a group role, so long as the group contains elements with a treeitem role, and that only other element must have a generic or presentation role, with no other properties important to accessibility?

That's not what I was going for. I understand the point but I'm not sure I'd personally make it that strong. treeitem allows name from content so I'd expect more than just "flat" content (at the very least images but I could see more complex content at the leafs of the tree).

Perhaps we should clarify this in the group.

scottaohara: Yeh agreed. And that’s a good idea to define what’s allowed in a group in the group role section. Otherwise we’d need to duplicate whatever’s called out here in other role defs

cookiecrook: Please rephrase to avoid using the abstract role `widget`? If this is referring to any subrole of widget, expand that list.

pkra: Coming back to this I'm wondering: isn't #1440 trying to enable some uses cases that this would prevent (e.g., button inside a treeitem as secondary action)?

spectranaut:
this may be covered by: https://github.com/w3c/aria/pull/1454

smhigley:
This may be addressed by #1454

pkra:
> This may be addressed by #1454

@smhigley that sounds great. I don't see a matching change in the diff though. Does your comment mean you're planning to address #1251 in the PR or am I (more likely) missing the relevant change?

PS I haven't found time to overhaul this PR because I did it all wrong to begin with (as discussed at a WG call a few weeks ago).

pkra:
Closing this. As noted this is the wrong way around and needs a fresh PR.

PR #1802, opened 2022-09-15
Change aria-errormessage to ID reference list

### LINKED ISSUE #1730: Should aria-errormessage be an ID reference list rather than a single reference?
I'm thinking `aria-errormessage`, like `aria-details` might need to be an ID reference list rather than a single reference.

Use case: A password-creation field which may have/display multiple errors:
* "Password must be between 10 and 35 characters"
* "Password must contain at least one punctuation symbol"

mrego:
Fixes #1730.

### DIFF
--- a/index.html
+++ b/index.html
@@ -11467,10 +11467,10 @@ button.ariaPressed; // null</pre>
 		<div class="property" id="aria-errormessage">
 			<pdef>aria-errormessage</pdef>
 			<div class="property-description">
-				<p><a>Identifies</a> the <a>element</a> that provides an error message for an <a>object</a>. See related <sref>aria-invalid</sref> and <pref>aria-describedby</pref>.</p>
-				<p>The <code>aria-errormessage</code> attribute references another element that contains error message text. Authors MUST use <sref>aria-invalid</sref> in conjunction with <code>aria-errormessage</code>.</p>
-				<p>When the value of an object is not valid, <sref>aria-invalid</sref> is set to <code>true</code>, which indicates that the message contained by an element referenced by <code>aria-errormessage</code> is pertinent.</p>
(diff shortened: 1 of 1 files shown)


cyns:
In Core-AAM it already has multiple references for IA2 and ATK/ATSPI. 
Windows UI uses ControllerFor, which takes an array of references, so that can work

AX API (Mac) uses a text property and may not be able to support this. 
@cookiecrook can this get text from multiple elements? Property: AXValidationError: textual content of the referenced element

cyns (approved):
See also https://github.com/w3c/core-aam/issues/138 to make core-aam align with this change

jnurthen:
@jcsteh can you please take a look at this.

cookiecrook (approved):

cookiecrook:
> AX API (Mac) uses a text property and may not be able to support this. @cookiecrook can this get text from multiple elements? Property: AXValidationError: textual content of the referenced element

Yes, WebKit could concatenate the text equivalents into the string for the API, or Apple could consider changing the platform API.

spectranaut:
@jnurthen do you want @jcsteh's review? This hasn't been implemented yet in firefox.

I have a memory of discussing this in a meeting but I just looked through sept and oct meetings and didn't see anything. Should we agenda+ for a meeting in Jan, or do you remember discussing it?

mrego:
BTW, filled issues on [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1403266) and [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1807040).

spectranaut:
Needs WPT test and link to webkit issue before merging.

mrego:
> Needs WPT test and link to webkit issue before merging.

WebKit already supports an ID reference list, so I guess we don't need an issue for that.

adampage (approved):

cookiecrook:
Thanks @rniwa for https://github.com/web-platform-tests/wpt/pull/37856

@spectranaut, I think this can be merged now?

jcsteh:
Sorry I didn't get around to reviewing this before it was merged. FWIW, this looks good to me.

PR #1814, opened 2022-09-22
Consider new abstract role "phrase"

### LINKED ISSUE #908: Consider new abstract role "phrase"
Consider an abstract role for the ontology which would be used for roles equivalent to HTML phrasing elements. Section doesn't seem correct as a parent role for many of these and there doesn't seem to be any other place to put them... Example would be the new subcript/superscript roles

jnurthen:
Fixes #908

Have not spent much time on the actual definition and the name.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1814.html" title="Last updated on Sep 29, 2022, 12:51 AM UTC (4ee6403)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1814/bcf44d1...4ee6403.html" title="Last updated on Sep 29, 2022, 12:51 AM UTC (4ee6403)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1795,7 +1795,7 @@
 					</tr>
 					<tr>
 						<th class="role-parent-head" scope="row">Superclass Role:</th>
-						<td class="role-parent"><rref>section</rref></td>
+						<td class="role-parent"><rref>phrasingcontent</rref></td>
 					</tr>
 					<tr>
 						<th class="role-children-head" scope="row">Subclass Roles:</th>
@@ -2093,7 +2093,7 @@
 					<tr>
 						<th class="role-parent-head" scope="row">Superclass Role:</th>
 						<td class="role-parent">
-							<rref>section</rref>
+							<rref>phrasingcontent</rref>
 						</td>
 					</tr>
 					<tr>
@@ -2863,7 +2863,7 @@
 					</tr>
 					<tr>
 						<th class="role-parent-head" scope="row">Superclass Role:</th>
-						<td class="role-parent"><rref>section</rref></td>
+						<td class="role-parent"><rref>phrasingcontent</rref></td>
 					</tr>
 					<tr>
 						<th class="role-children-head" scope="row">Subclass Roles:</th>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3610,7 +3610,7 @@ <h2>Definition of Roles</h2>
 					</tr>
 					<tr>
 						<th class="role-parent-head" scope="row">Superclass Role:</th>
-						<td class="role-parent"><rref>structure</rref></td>
+						<td class="role-parent"><rref>phrasingcontent</rref></td>
```
pkra, line 3613: I find this one confusing. The generic role specifically mentions `div` which does not match well.

index.html
```diff
@@ -1795,7 +1795,7 @@ <h2>Definition of Roles</h2>
 					</tr>
 					<tr>
 						<th class="role-parent-head" scope="row">Superclass Role:</th>
-						<td class="role-parent"><rref>section</rref></td>
+						<td class="role-parent"><rref>phrasingcontent</rref></td>
```
pkra, line 1798: This feels strange to me; captions tend to allow flow content, e.g., table or figure captions. (Also, all the examples for the caption role use `div`.)

index.html
jnurthen: yet span does ;) - no idea for a solution though!

cookiecrook (approved):
FWIW I found this StackOverflow helpful in sorting out the diffs between HTML flow content and phrasing content. 
https://stackoverflow.com/questions/30233447/what-is-the-difference-between-phrasing-content-and-flow-content

I dislike the role name, but since that's what HTML calls it, I think it's fine as is. Thanks @jnurthen

pkra:
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories is also a good reference.

jnurthen:
Looking at the roles we are discussing - this is very different from HTML Phrasing Content which includes all sorts of interactives. Using the same term would IMO cause confusion.
I've been trying to find a good term but haven't yet come up with one. Until we do I propose that we park this as changing abstract roles from one to another doesn't sound useful unless we can get the naming right.

scottaohara:
guess `textyMcTextFace` just didn't cut it...

PR #1816, opened 2022-09-22
Editorial: change heading level on "Note regarding the ARIA 1.1 none role."

### LINKED ISSUE #1773: heading level "Note regarding the ARIA 1.1 none role." is wrong
E.g., in https://github.com/w3c/aria/blob/6d46a9a0f694a4a0ae9f1fc7cdcf4d9299ac7729/index.html#L6299 the heading in the note is marked up as `h5` but in https://w3c.github.io/aria/#note-regarding-the-aria-1-1-none-role-0 it comes out as `h3`. 

This is problematic since the automatically generated heading for the role's entry itself is an `h4`.

Maybe we shouldn't do headings inside notes. It definitely seems risky to add a fixed heading level below an automatically generated one.

dakahn:
closes #1773 

It's been a minute since this was discussed so let me know if this isn't the solution we want. But I simply adjusted the heading level to fall in line with the programmatically generated heading and make sense with the heading above it 🏄🏽

***
<a href="https://pr-preview.s3.amazonaws.com/dakahn/aria/pull/1816.html" title="Last updated on Sep 29, 2022, 4:22 PM UTC (5b46146)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1816/bcf44d1...dakahn:5b46146.html" title="Last updated on Sep 29, 2022, 4:22 PM UTC (5b46146)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4067,7 +4067,7 @@
 			<div class="role-description">
 				<p>A container for a collection of [=elements=] that form an image. See synonym <rref>img</rref>.</p>
 				<div class="note" id="role-image-note-image">
-					<h5>Note regarding the ARIA 1.3 <code>image</code> role.</h5>
+					Note regarding the ARIA 1.3 <code>image</code> role.
 					<p>The <code>image</code> was added to ARIA in version 1.3 as a
 					synonym of the ARIA 1.0 <rref>img</rref> role. The <code>image</code>
 					role improves syntactic consistency with the names of other roles,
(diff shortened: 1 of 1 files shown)


scottaohara:
hmm.  there is a heading with the text "note" that is injected prior to this heading. making it go from h4 to h3 (this heading text).

I wonder if it would actually make sense to just remove the heading all together, rather than trying to fiddle with this.  this note about none seems to be (from my quick review) the only note that has it's own special heading.

i'm more concerned with the fact that in preview mode, i cannot seem to use a screen reader to read the content of the spec outside of the TOC....  but that's nothing to do with this issue / what you've changed.

jnurthen:
Can we fix the other 2 as well

- https://github.com/w3c/aria/blob/bcf44d193e74f99d63e8f411877b6572c92e13de/index.html#L4070
- https://github.com/w3c/aria/blob/bcf44d193e74f99d63e8f411877b6572c92e13de/index.html#L5910

jnurthen:
Actually - instead of removing the header can we try using the title attribute on the element with class note. I was looking at the respec doc and it seems to do what we want
https://respec.org/docs/#note

jnurthen (approved):

PR #1820, opened 2022-10-03
Permit UA to ignore user-triggered changes inside live regions

### LINKED ISSUE #72: Need: ~"User Agents MAY ignore user-triggered changes to live regions, such as typing in text fields."
We've gotten a bug report against WebKit that VO repeats every character of a text field on Facebook.com. After investigating, the text field is placed inside an ARIA live region.

The spec ~"every change that affects the accessible label" should trigger a live region alert, but I think there should be an exception for user triggered changed like typing. 

This is probably an single-sentence "User Agents MAY" change.

adampage:
Closes #72 by permitting user agents to ignore user-triggered changes inside live regions.

I created [a quick Codepen](https://codepen.io/adampage/pen/jOxYgJG/892e8ddfc9a31a82eec40a993ffe3048) to test the behavior described in #72 and was able to reproduce it in macOS 12.6 + VoiceOver + Safari 16.0.  In that environment, values entered into a text input inside a live region announce themselves _twice_.  In Chrome 106, it’s more severe: the value repeats itself a second time, then the input’s label is read again, then the value repeats a third time.

I didn’t find one obvious location in 1.3 to add a normative statement for this, but my first instinct was to [put it in 6.6.2 Live Region Attributes](https://pr-preview.s3.amazonaws.com/adampage/aria/pull/1820.html#:~:text=User%20agents%20MAY%20ignore%20changes%20triggered%20by%20a%20user%20directly%20on%20an%20object%20inside%20a%20live%20region%20(e.g.%2C%20editing%20the%20value%20of%20a%20text%20field).), where the general concept of live regions is first described.

Alternatively, I wondered if it might belong in [6.7’s `aria-live` description](https://github.com/w3c/aria/blob/3d20a5c3535de735b082f5fa4f14763f7bdd7c2d/index.html#L12045) or even in the [`aria-relevant` description](https://github.com/w3c/aria/blob/3d20a5c3535de735b082f5fa4f14763f7bdd7c2d/index.html#L12576)?

***
<a href="https://pr-preview.s3.amazonaws.com/adampage/aria/pull/1820.html" title="Last updated on Oct 4, 2022, 5:41 PM UTC (1e2f87e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1820/9d1869f...adampage:1e2f87e.html" title="Last updated on Oct 4, 2022, 5:41 PM UTC (1e2f87e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -10352,7 +10352,7 @@ button.ariaPressed; // null</pre>
 		</section>
 		<section id="attrs_liveregions">
 			<h3>Live Region Attributes</h3>
-			<p>This section contains [=attributes=] specific to <a>live regions</a> in rich internet applications. These attributes may be applied to any <a>element</a>. The purpose of these attributes is to indicate that content changes may occur without the element having focus, and to provide <a>assistive technologies</a> with information on how to process those content updates. Some <a>roles</a> specify a default value for the <pref>aria-live</pref> attribute specific to that role. An example of a live region is a ticker section that lists updating stock quotes. User agents MAY ignore changes triggered by a user directly on an <a>object</a> inside a live region (e.g., editing the value of a text field).</p>
+			<p>This section contains [=attributes=] specific to <a>live regions</a> in rich internet applications. These attributes may be applied to any <a>element</a>. The purpose of these attributes is to indicate that content changes may occur without the element having focus, and to provide <a>assistive technologies</a> with information on how to process those content updates. Some <a>roles</a> specify a default value for the <pref>aria-live</pref> attribute specific to that role. An example of a live region is a ticker section that lists updating stock quotes. User agents MAY ignore changes triggered by direct user action on an <a>element</a> inside a live region (e.g., editing the value of a text field).</p>
 			<ul>
 				<li><pref>aria-atomic</pref></li>
 				<li><sref>aria-busy</sref></li>
--- a/index.html
+++ b/index.html
@@ -10352,7 +10352,7 @@ button.ariaPressed; // null</pre>
 		</section>
 		<section id="attrs_liveregions">
 			<h3>Live Region Attributes</h3>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -10352,7 +10352,7 @@ <h3>Widget Attributes</h3>
...
+			<p>This section contains [=attributes=] specific to <a>live regions</a> in rich internet applications. These attributes may be applied to any <a>element</a>. The purpose of these attributes is to indicate that content changes may occur without the element having focus, and to provide <a>assistive technologies</a> with information on how to process those content updates. Some <a>roles</a> specify a default value for the <pref>aria-live</pref> attribute specific to that role. An example of a live region is a ticker section that lists updating stock quotes. User agents MAY ignore changes triggered by a user directly on an <a>object</a> inside a live region (e.g., editing the value of a text field).</p>
```
cookiecrook, line 10355:
```suggestion
			<p>This section contains [=attributes=] specific to <a>live regions</a> in rich internet applications. These attributes may be applied to any <a>element</a>. The purpose of these attributes is to indicate that content changes may occur without the element having focus, and to provide <a>assistive technologies</a> with information on how to process those content updates. Some <a>roles</a> specify a default value for the <pref>aria-live</pref> attribute specific to that role. An example of a live region is a ticker section that lists updating stock quotes. User agents MAY ignore changes triggered by direct user action on an <a>element</a> inside a live region (e.g., editing the value of a text field).</p>
```

cookiecrook (approved):
One editorial ("a user directly" > "direct user action") and one substantive ("object" > "element") change... In case I'm missing something and it really should be referencing "object," the explanation should be more clear. Looks good overall.

index.html
adampage: Thanks very much, I’ve committed both changes.

scottaohara (approved):

spectranaut (approved):

Issue #184, opened 2016-01-12
HTML-AAM: should @headers participate in name calculation?


cyns:

jasonkiss:
This issue was moved to w3c/html-aam#18

PR #1827, opened 2022-10-06
Aria1720

### LINKED ISSUE #1720: Add explicit value calculations for `textbox` and `searchbox` role
A [textbox](https://www.w3.org/TR/wai-aria-1.2/#textbox) has a value according to the spec: `A type of input that allows free-form text as its value.` But ARIA does not specify how the value is calculated and CORE-AAM does not specify how the implicit value is translated into the accessibility APIs. 

[Combobox's](https://w3c.github.io/aria/#combobox) value is specified and we will probably want to follow the same convention.

Whatever we specify for `textbox` probably will apply to `searchbox` as well.

chlane:
Closes #1720

Added explanations for how textboxes and searchboxes obtain thier values

### DIFF
--- a/index.html
+++ b/index.html
@@ -9256,6 +9256,7 @@
 			<rdef>textbox</rdef>
 			<div class="role-description">
 				<p>A type of input that allows free-form text as its value.</p>
+				<p>Authors MUST limit its children to completely presentational elements (like a spacer image, decorative graphic, or clearing element).</p>
 				<p>If the <pref>aria-multiline</pref> <a>attribute</a> is <code>true</code>, the <a>widget</a> accepts line breaks within the input, as in an <abbr title="Hypertext Markup Language">HTML</abbr> <code>textarea</code>. Otherwise, this is a simple text box. The intended use is for languages that do not have a text input <a>element</a>, or cases in which an element with different <a>semantics</a> is repurposed as a text field.</p>
 				<!-- keep the following para synced with its equivalent in #aria-multiline -->
 				<p class="note">In most user agent implementations, the default behavior of the <kbd>ENTER</kbd> or <kbd>RETURN</kbd> key is different between the single-line and multi-line text fields in HTML. When user has focus in a single-line <code>&lt;input type="text"&gt;</code> element, the keystroke usually submits the form. When user has focus in a multi-line <code>&lt;textarea&gt;</code> element, the keystroke inserts a line break. The <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> <code>textbox</code> role differentiates these types of boxes with the <pref>aria-multiline</pref> attribute, so authors are advised to be aware of this distinction when designing the field.</p>
@@ -9332,7 +9333,7 @@
 					</tr>
 					<tr>
 						<th class="role-childpresentational-head" scope="row">Children Presentational:</th>
-						<td class="role-childpresentational"> </td>
+						<td class="role-childpresentational">True</td>
 					</tr>
 					<tr>
 						<th class="role-presentational-inherited-head" scope="row">Inherits Presentational:</th>
--- a/index.html
+++ b/index.html
@@ -7446,6 +7446,14 @@
 			<rdef>searchbox</rdef>
 			<div class="role-description">
 				<p>A type of textbox intended for specifying search criteria. See related <rref>textbox</rref> and <rref>search</rref>.</p>
+				<p>
+					User agents MUST expose the value of elements with role <code>searchbox</code> to <a>assistive technologies</a>. The value of a <code>searchbox</code> is represented by one of the following:
+				</p>
+					<ul>
+						<li>If the <code>searchbox</code> element is a host language element that provides a value, such as an HTML <code>input</code> element, the value of the searchbox is the value of that element.</li>
+						<li>Otherwise, the value of the <code>searchbox</code> is represented by its descendant elements and can be determined using the same method used to compute the name of a <rref>button</rref> from its descendant content.</li>
+					</ul>
+
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>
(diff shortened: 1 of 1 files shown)


spectranaut:
Can you make sure to always make the title of the PR a description of the change?

index.html
```diff
@@ -9256,7 +9264,20 @@ <h5>Presentational Role Inheritance</h5>
 			<rdef>textbox</rdef>
 			<div class="role-description">
 				<p>A type of input that allows free-form text as its value.</p>
+				<p>Authors MUST limit its children to completely presentational elements (like a spacer image, decorative graphic, or clearing element).</p>
```
spectranaut, line 9267: Looks like an unrelated change snuck into this PR! Can you delete this line.

index.html
```diff
@@ -9256,7 +9264,20 @@ <h5>Presentational Role Inheritance</h5>
...
+					User agents MUST expose the value of elements with role <code>textbox</code> to <a>assistive technologies</a>.
+					The value of a <code>textbox</code> is represented by one of the following:
+					</p>
+					<ul>
+						<li>If the <code>textbox</code> element is a host language element that provides a value, such as an HTML <code>input</code> element, the value of the textbox is the value of that element.</li>
+							<li>Otherwise, the value of the <code>textbox</code> is represented by its descendant elements and can be determined using the same method used to compute the name of a <rref>button</rref> from its descendant content.</li>
+					</ul>
+
+
+
```
spectranaut, line 9280: And these empty lines, as they are unnecessary

index.html
```diff
@@ -9332,7 +9353,7 @@ <h5>Presentational Role Inheritance</h5>
 					</tr>
 					<tr>
 						<th class="role-childpresentational-head" scope="row">Children Presentational:</th>
-						<td class="role-childpresentational"> </td>
+						<td class="role-childpresentational">True</td>
```
spectranaut, line 9356: And looks like this snuck in as well -- when making PRs, it's good practice to review the diff, because you probably would have caught these things! :)

spectranaut (changes requested):

index.html
chlane: Thanks @spectranaut I removed the unrelated change in commit [81cd343](https://github.com/chlane/aria/commit/81cd343772f02835c1d27cbf45553726508ee064)

index.html
chlane: @spectranaut I couldn't find these blank lines and they are not showing up in the diff for my latest commit [81cd343](https://github.com/chlane/aria/commit/81cd343772f02835c1d27cbf45553726508ee064)

index.html
chlane: Thanks @spectranaut this has been fixed in [81cd343](https://github.com/chlane/aria/commit/81cd343772f02835c1d27cbf45553726508ee064)

index.html
chlane: @spectranaut found them and removed them thanks. See https://github.com/w3c/aria/pull/1827/commits/55fca79365055aa062e17600503afca9c5b48362

index.html
chlane: Thanks @spectranaut this has been fixed in https://github.com/w3c/aria/pull/1827/commits/0317b3039c121f81949135a8b056015416f52e92

index.html
```diff
@@ -9256,7 +9264,17 @@ <h5>Presentational Role Inheritance</h5>
...
 				<p>A type of input that allows free-form text as its value.</p>
+				<p>Authors MUST limit its children to completely presentational elements (like a spacer image, decorative graphic, or clearing element).</p>
 				<p>If the <pref>aria-multiline</pref> <a>attribute</a> is <code>true</code>, the <a>widget</a> accepts line breaks within the input, as in an <abbr title="Hypertext Markup Language">HTML</abbr> <code>textarea</code>. Otherwise, this is a simple text box. The intended use is for languages that do not have a text input <a>element</a>, or cases in which an element with different <a>semantics</a> is repurposed as a text field.</p>
+				
+				<p>
+					User agents MUST expose the value of elements with role <code>textbox</code> to <a>assistive technologies</a>.
+					The value of a <code>textbox</code> is represented by one of the following:
+					</p>
+					<ul>
+						<li>If the <code>textbox</code> element is a host language element that provides a value, such as an HTML <code>input</code> element, the value of the textbox is the value of that element.</li>
```
scottaohara, line 9275: see previous comment i left for this line in searchbox

index.html
```diff
@@ -7446,6 +7446,14 @@ <h5>Presentational Role Inheritance</h5>
 			<rdef>searchbox</rdef>
 			<div class="role-description">
 				<p>A type of textbox intended for specifying search criteria. See related <rref>textbox</rref> and <rref>search</rref>.</p>
+				<p>
+					User agents MUST expose the value of elements with role <code>searchbox</code> to <a>assistive technologies</a>. The value of a <code>searchbox</code> is represented by one of the following:
+				</p>
+					<ul>
+						<li>If the <code>searchbox</code> element is a host language element that provides a value, such as an HTML <code>input</code> element, the value of the searchbox is the value of that element.</li>
+						<li>Otherwise, the value of the <code>searchbox</code> is represented by its descendant elements and can be determined using the same method used to compute the name of a <rref>button</rref> from its descendant content.</li>
```
scottaohara, line 7454: is there something special about button in particular that it is uniquely called out, instead of alluding to the textbox/searchbox needs to calculate its value based on its subtree, similar to roles who derive their name from content?

index.html
```diff
@@ -7446,6 +7446,14 @@ <h5>Presentational Role Inheritance</h5>
 			<rdef>searchbox</rdef>
 			<div class="role-description">
 				<p>A type of textbox intended for specifying search criteria. See related <rref>textbox</rref> and <rref>search</rref>.</p>
+				<p>
+					User agents MUST expose the value of elements with role <code>searchbox</code> to <a>assistive technologies</a>. The value of a <code>searchbox</code> is represented by one of the following:
+				</p>
+					<ul>
+						<li>If the <code>searchbox</code> element is a host language element that provides a value, such as an HTML <code>input</code> element, the value of the searchbox is the value of that element.</li>
```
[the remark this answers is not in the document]
JAWS-test: Is a copy of the [combobox](https://w3c.github.io/aria/#combobox) text. So that no misunderstandings arise, I find this useful

index.html
JAWS-test: This is also a copy of the [combobox](https://w3c.github.io/aria/#combobox) text. If we change it here, we should change it there. The link to the button role at the combobox text is also slightly misleading, because it suggests that the link points to an explanation of how the button's name is derived from its content, but this is not the case (for that, the AccName would have to be linked)

index.html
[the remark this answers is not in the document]
scottaohara: fari enough.

index.html
scottaohara: yeh.  tbh i should have paid more attention to that and even though this text is coming from combobox, I would then say that my questions also relate to combobox then.

chlane: @scottaohara thanks, I'll close this PR and get these changes to the same content in one PR.  @scottaohara @JAWS-test what do you think of the following verbiage?
						<li>Otherwise, the value of the <code>searchbox</code> is determined by the <a href="https://www.w3.org/TR/accname-1.1/" >Accessible Name and Description Computation</a>.</li>

JAWS-test: @chlane I am not sure if it is useful to refer to the AccName as it is too complicated to understand and in my opinion wrong:
* Imagine that in the textbox there is a div. A div is not allowed to have a label, with that the textbox then has no value?
* Imagine that there is an element in the textbox, which has text content and is labeled by aria-label at the same time. Should the value be the text (like in Edge and Chrome) or the value in the aria-label (like in Firefox)?

Until the AccName resolves such inconsistencies, maybe we should just say the text content is the value, right?

chlane: @JAWS-test thanks, the verbiage "using the same method used to compute the 'name' of a button" has led me to conflate name and value. Do you think "Otherwise, the value is determined by its text content" will work?

JAWS-test: @chlane After all, it's not about the wording, it's about the method used. Simply taking the text content would be different than taking the name of a button as the value. I was just expressing my concern about what happens when we use the button method. But I'm not in a position to decide which method is the right one. That should be decided by the ARIA WG.

chlane: TODO: At the weekly meeting we decided to 1) create a new issue in accname to explain how values are grabbed and 2) remove the current verbiage that describes getting the value for comboboxes from combobox, searchbox and textbox

chlane:
I created this accname issue Explain how values are obtained #184 https://github.com/w3c/accname/issues/184 to move value calculation out of this spec.

spectranaut:
I'm going to close this PR, as we decided to remove the value calculation from ARIA into AccName: https://www.w3.org/2023/01/26-aria-minutes#t05

PR #1828, opened 2022-10-07
Providing more current examples of presentational elements in textboxes

### LINKED ISSUE #1493: Consider adding more normative requirements for authors on the textbox role

chlane:
Closes #1493. Updated examples to be more current.

***
<a href="https://pr-preview.s3.amazonaws.com/chlane/aria/pull/1828.html" title="Last updated on Oct 10, 2022, 8:50 PM UTC (877511f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1828/78af4a5...chlane:877511f.html" title="Last updated on Oct 10, 2022, 8:50 PM UTC (877511f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -9256,7 +9256,6 @@
 			<rdef>textbox</rdef>
 			<div class="role-description">
 				<p>A type of input that allows free-form text as its value.</p>
-				<p>Authors MUST limit the children of a textbox to presentational elements such as icons used to visually convey information. Some examples are error and success icons, portraits or silhouettes of users and status indicators.</p>
 				<p>If the <pref>aria-multiline</pref> <a>attribute</a> is <code>true</code>, the <a>widget</a> accepts line breaks within the input, as in an <abbr title="Hypertext Markup Language">HTML</abbr> <code>textarea</code>. Otherwise, this is a simple text box. The intended use is for languages that do not have a text input <a>element</a>, or cases in which an element with different <a>semantics</a> is repurposed as a text field.</p>
 				<!-- keep the following para synced with its equivalent in #aria-multiline -->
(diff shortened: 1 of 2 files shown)


spectranaut:
https://github.com/w3c/core-aam/pull/144

chlane:
Closing because this PR contains an additional file it should not - index.html.1493

PR #1830, opened 2022-10-10
Providing more current examples of presentational elements in textboxes

### LINKED ISSUE #1493: Consider adding more normative requirements for authors on the textbox role

chlane:
Closes #1493. Provided more current examples of presentational elements in textboxes that were suggested in  Pull Request https://github.com/w3c/aria/pull/1788.

***
<a href="https://pr-preview.s3.amazonaws.com/chlane/aria/pull/1830.html" title="Last updated on Feb 22, 2024, 8:14 PM UTC (75c884d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1830/e88317c...chlane:75c884d.html" title="Last updated on Feb 22, 2024, 8:14 PM UTC (75c884d)">Diff</a>


index.html
```diff
@@ -9257,6 +9257,7 @@ <h5>Presentational Role Inheritance</h5>
 			<div class="role-description">
 				<p>A type of input that allows free-form text as its value.</p>
 				<p>If the <pref>aria-multiline</pref> <a>attribute</a> is <code>true</code>, the <a>widget</a> accepts line breaks within the input, as in an <abbr title="Hypertext Markup Language">HTML</abbr> <code>textarea</code>. Otherwise, this is a simple text box. The intended use is for languages that do not have a text input <a>element</a>, or cases in which an element with different <a>semantics</a> is repurposed as a text field.</p>
+				<p>Authors MUST limit the children of a textbox to presentational elements such as icons used to visually convey information. Some examples are error and success icons, portraits or silhouettes of users and status indicators.</p>
```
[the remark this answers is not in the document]
cookiecrook: I think "non-interactive" is better here than "presentational" but I could be convinced the original term is better.

cookiecrook (changes requested):

index.html
```diff
@@ -9332,7 +9333,7 @@ <h5>Presentational Role Inheritance</h5>
 					</tr>
 					<tr>
 						<th class="role-childpresentational-head" scope="row">Children Presentational:</th>
-						<td class="role-childpresentational"> </td>
+						<td class="role-childpresentational"> </td>
```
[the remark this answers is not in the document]
cookiecrook: rolling back the whitespace diff

cookiecrook:
I think your separate commit marked the rest of my suggestions as outdated. Please look back at the rest of the diff.

For future ref, you can commit reviewers diffs directly if you agree with the changes.

cookiecrook (changes requested):
Most changes from the last review still unaddressed.

index.html
cookiecrook, line 9336: I'm not able to roll back this diff with a change suggestion. Not sure why, but it's not as simple as a different whitespace char.

index.html
```diff
@@ -9257,6 +9257,7 @@ <h5>Presentational Role Inheritance</h5>
 			<div class="role-description">
 				<p>A type of input that allows free-form text as its value.</p>
 				<p>If the <pref>aria-multiline</pref> <a>attribute</a> is <code>true</code>, the <a>widget</a> accepts line breaks within the input, as in an <abbr title="Hypertext Markup Language">HTML</abbr> <code>textarea</code>. Otherwise, this is a simple text box. The intended use is for languages that do not have a text input <a>element</a>, or cases in which an element with different <a>semantics</a> is repurposed as a text field.</p>
+				<p>Authors MUST limit the children of a textbox to non-interactive elements such as icons used to visually convey information. Some examples are error and success icons, portraits or silhouettes of users and status indicators.</p>
```
cookiecrook, line 9260:
```suggestion
				<p>Authors MUST limit the children of a textbox to non-interactive elements such as icons used to visually convey information that is already exposed in an accessible manner. Examples include:
				<ul>
					<li>an error icon, with an associated <sref>aria-invalid</sref> and/or <sref>aria-errormessage</sref>;</li>
					<li>an icon of a user silhouette, where the textbox is labeled as a name or username); and</li>
					<li>status indicators, such as characters remaining, which is exposed in text outside the field.</li>
					</p>
```

cookiecrook (changes requested):
Adding back in the original diff.

scottaohara (commented):
Does this PR also need to call out how it could potentially impact what is being defined in #1827?

For instance, unless explicitly called out as being content that is entirely presentational - could someone not do:
```
<div role=textbox ...>
  actual value 
  <img src=... alt="oh no, now am i'm part of the value too!?"> <!-- error icon that really should probably have been positioned outside of the textbox anyway -->
</div>
```

chlane:
@scottaohara @cookiecrook I think the example Scott provided below favors the usage of "entirely presentational" instead of "non-interactive". Or could we say "entirely presentational and non-interactive"?

```
<div role=textbox ...>
  actual value 
  <img src=... alt="oh no, now am i'm part of the value too!?"> 
  <!-- error icon that really should probably have been positioned outside of the textbox anyway -->
</div>
```

index.html
```diff
@@ -9257,6 +9257,13 @@ <h5>Presentational Role Inheritance</h5>
 			<div class="role-description">
 				<p>A type of input that allows free-form text as its value.</p>
 				<p>If the <pref>aria-multiline</pref> <a>attribute</a> is <code>true</code>, the <a>widget</a> accepts line breaks within the input, as in an <abbr title="Hypertext Markup Language">HTML</abbr> <code>textarea</code>. Otherwise, this is a simple text box. The intended use is for languages that do not have a text input <a>element</a>, or cases in which an element with different <a>semantics</a> is repurposed as a text field.</p>
+				<p>Authors MUST limit the children of a textbox to non-interactive elements such as icons used to visually convey information that is already exposed in an accessible manner. Examples include:
```
cookiecrook, line 9260:
```suggestion
				<p>Authors MUST limit the children of a textbox to non-interactive, entirely presentational elements such as icons used to visually convey information that is already exposed in an accessible manner. Examples include:
```

cookiecrook (approved):

index.html
cookiecrook: Agree with Scott and Chris on the image alt example.

index.html
```diff
@@ -9257,6 +9257,13 @@ <h5>Presentational Role Inheritance</h5>
 			<div class="role-description">
 				<p>A type of input that allows free-form text as its value.</p>
 				<p>If the <pref>aria-multiline</pref> <a>attribute</a> is <code>true</code>, the <a>widget</a> accepts line breaks within the input, as in an <abbr title="Hypertext Markup Language">HTML</abbr> <code>textarea</code>. Otherwise, this is a simple text box. The intended use is for languages that do not have a text input <a>element</a>, or cases in which an element with different <a>semantics</a> is repurposed as a text field.</p>
+				<p>Authors MUST limit the children of a textbox to non-interactive, entirely presentational elements such as icons used to visually convey information that is already exposed in an accessible manner. Examples include:
+					<ul>
+						<li>an error icon, with an associated <sref>aria-invalid</sref> and/or <sref>aria-errormessage</sref>;</li>
+						<li>an icon of a user silhouette, where the textbox is labeled as a name or username; and</li>
```
scottaohara, line 9263:
```suggestion
						<li>an icon of a user silhouette, where the textbox is also visibly labeled or provided an accessible name of "name" or "username"; and</li>
```

index.html
```diff
@@ -9257,6 +9257,13 @@ <h5>Presentational Role Inheritance</h5>
 			<div class="role-description">
 				<p>A type of input that allows free-form text as its value.</p>
 				<p>If the <pref>aria-multiline</pref> <a>attribute</a> is <code>true</code>, the <a>widget</a> accepts line breaks within the input, as in an <abbr title="Hypertext Markup Language">HTML</abbr> <code>textarea</code>. Otherwise, this is a simple text box. The intended use is for languages that do not have a text input <a>element</a>, or cases in which an element with different <a>semantics</a> is repurposed as a text field.</p>
+				<p>Authors MUST limit the children of a textbox to non-interactive, entirely presentational elements such as icons used to visually convey information that is already exposed in an accessible manner. Examples include:
+					<ul>
+						<li>an error icon, with an associated <sref>aria-invalid</sref> and/or <sref>aria-errormessage</sref>;</li>
```
scottaohara, line 9262:
```suggestion
						<li>an error icon, where the containing textbox has been provided an <sref>aria-invalid</sref>,  <sref>aria-errormessage</sref>, or both attributes;</li>
```

index.html
```diff
@@ -9257,6 +9257,13 @@ <h5>Presentational Role Inheritance</h5>
...
+						<li>an icon of a user silhouette, where the textbox is labeled as a name or username; and</li>
+						<li>status indicators, such as characters remaining, which is exposed in text outside the field.</li>
```
scottaohara, line 9264:
```suggestion
						<li>a graphical status indicator, such as a gauge to represent characters remaining, which represents dynamically updating text available outside of the textbox.</li>
```

scottaohara (commented):
just some final wording suggestions, then i think this would be good to go.

chlane:
@scottaohara thanks for providing more detail where needed. I've commit all the suggestions.

index.html
```diff
@@ -9257,6 +9257,13 @@ <h5>Presentational Role Inheritance</h5>
...
 				<p>If the <pref>aria-multiline</pref> <a>attribute</a> is <code>true</code>, the <a>widget</a> accepts line breaks within the input, as in an <abbr title="Hypertext Markup Language">HTML</abbr> <code>textarea</code>. Otherwise, this is a simple text box. The intended use is for languages that do not have a text input <a>element</a>, or cases in which an element with different <a>semantics</a> is repurposed as a text field.</p>
+				<p>Authors MUST limit the children of a textbox to non-interactive, entirely presentational elements such as icons used to visually convey information that is already exposed in an accessible manner. Examples include:
+					<ul>
+						<li>an error icon, where the containing textbox has been provided an <sref>aria-invalid</sref>,  <sref>aria-errormessage</sref>, or both attributes;</li>
+						<li>an icon of a user silhouette, where the textbox is also visibly labeled or provided an accessible name of "name" or "username"; and</li>
+						<li>a graphical status indicator, such as a gauge to represent characters remaining, which represents dynamically updating text available outside of the textbox.</li>
+					</ul>
+				</p>
```
scottaohara, line 9266: remove stray ending paragraph tag.
```suggestion
```

index.html
```diff
@@ -9257,6 +9257,13 @@ <h5>Presentational Role Inheritance</h5>
 			<div class="role-description">
 				<p>A type of input that allows free-form text as its value.</p>
 				<p>If the <pref>aria-multiline</pref> <a>attribute</a> is <code>true</code>, the <a>widget</a> accepts line breaks within the input, as in an <abbr title="Hypertext Markup Language">HTML</abbr> <code>textarea</code>. Otherwise, this is a simple text box. The intended use is for languages that do not have a text input <a>element</a>, or cases in which an element with different <a>semantics</a> is repurposed as a text field.</p>
+				<p>Authors MUST limit the children of a textbox to non-interactive, entirely presentational elements such as icons used to visually convey information that is already exposed in an accessible manner. Examples include:
```
scottaohara, line 9260: close paragraph.  a list cannot be nested within a paragraph.
```suggestion
				<p>Authors MUST limit the children of a textbox to non-interactive, entirely presentational elements such as icons used to visually convey information that is already exposed in an accessible manner. Examples include:</p>
```

scottaohara (approved):
some markup cleanup and approving this with the changes made

index.html
spectranaut: @chlane do you think you can remove this change? other wise this looks good to go!

spectranaut (approved):

index.html
chlane: @spectranaut not sure why this whitespace issue is happening. I couldn't tell what to change on my end but the commit button was enabled after I edited line 9345. I just deleted and replaced the whitespace.

Issue #1150, opened 2020-01-15
Unclear use of the term "descendant"


WilcoFiers:
There are a number of places where the word "descendant" is used. It is unclear what tree this is referring to. There are different trees involved in the user agent, such as the light DOM tree, shadow DOM trees, the flat tree, and the accessibility tree. In some uses of the word "descendant" the intended purpose seems to be flat tree, in others a document fragment, and yet others the accessibility tree.

This problem is new in ARIA 1.2 because it is written for DOM 4, rather than DOM 3 which was used in ARIA 1.1..

jnurthen:
@scottaohara any updates on this one?

cookiecrook:
Also see #1161 which refers to owned elements in the context of descendants.

jongund:
@jnurthen 

You may want to wait until [PR](https://github.com/w3c/aria/pull/1162) is merged, since there will be more references to `descendant` in the document before reviewing this issue.

jnurthen:
<details><summary>Discussed in WG call on 11/12/2020</summary>
**JN:** so the first one of these is 1150, and this is unclear use of the term descendant  
… I think at this stage going through and doing a PR on this isn't necessarily the right way of going about it. I think we need to go through and work out all the different uses of descendant that we have  
… doing an informal search of the spec, we have 4 different uses. There's DOM descendant, descendant element (not sure if the same as DOM descendant), and then we have other uses of descendant, many of which are in activedescendant and keyboard handling sections, and then we have random other things  
… my proposal on this is that someone takes a look and tries to categorize the usages of the word descendant into various things and then we can potentially come up with a different word for some of them and a definition, or a definition for what descendant means in each of these ways  
… e.g. "element descendant" and "DOM descendant", and define each of those  
… so anyone want to voluteer for this?  

**SH:** I wouldn't mind doing some of that  

**MK:** do they have to be fixed before CR?  

**JN:** no, only if they (the ACT) say they have to be  
… so I'm not proposing fixing this in 1.2, I'm proposing as soon in 1.3 as we can, and getting that in the working draft as soon as we can  

> *carmacleod:* Also [https://github.com/w3c/aria/issues/1033](https://github.com/w3c/aria/issues/1033) and [https://github.com/w3c/aria/issues/1161](https://github.com/w3c/aria/issues/1161)

**MK:** so you're not planning of taking the results of this analysis and trying to get something changed in 1.2 after it's gone into CR  

[...]

PR #1839, opened 2022-10-21
Clarify relationship between `aria-hidden` and `aria-owns`

### LINKED ISSUE #1818: Should aria-hidden be effected by aria-owns?
The description of aria-hidden has the following text:

> https://www.w3.org/TR/wai-aria-1.2/#aria-hidden
> An element is considered [hidden](https://www.w3.org/TR/wai-aria-1.2/#dfn-hidden) if it, or any of its ancestors are not rendered or have their aria-hidden attribute value set to true.

This would suggest that if I do something like the following, my list would be empty, because the li is hidden.

```html
  <ul aria-owns="li"></ul>
  <div aria-hidden="true">
    <li id="li">Child node</li>
  </div>
```

This doesn't seem to be the case when testing in Chrome and Firefox. There it seems that `aria-owns` is taking priority over `aria-hidden`. Intuitively, that does make sense to me. So then that makes me wonder whether that phrase in aria-hidden should be changed to maybe something like:

> An element is considered [hidden](https://www.w3.org/TR/wai-aria-1.2/#dfn-hidden) if it, or any of its ancestors are not rendered**, or if any of its [Owner Elements](https://www.w3.org/TR/wai-aria-1.2/#dfn-owning-element)** have their aria-hidden attribute value set to true.

adampage:
Closes #1818.

The [definition of _Owning Element_](https://w3c.github.io/aria/#dfn-owning-element) is already inclusive of both DOM ancestry _and_ ownership via `aria-owns`:

> An 'owning element' is any DOM ancestor of the [element](https://dom.spec.whatwg.org/#concept-element), or any element with an [aria-owns](https://w3c.github.io/aria/#aria-owns) attribute which references the ID of the element.

Because of this, I’m thinking it’ll be sufficient to swap out “_ancestors_” for “_owning elements_” in `aria-hidden`’s description.

I also editorially removed what I believe is an unhelpful comma.


adampage:
Here’s [a Codepen with a few exploratory tests](https://codepen.io/adampage/pen/BaxXaJQ/2543df121f8adfa1ad311a64b5c06a2c).

Jym77:
Two points to mention:

---
I'm encountering a problem similar to #1818 but the other way around, and I'm not sure this PR fully handles it 🤔 

The case in #1818 was:
```html
  <ul aria-owns="li"></ul>
  <div aria-hidden="true">
    <li id="li">Child node</li>
  </div>
```
That is, a not-hidden element `aria-owns` an element with a hidden DOM ancestor. In that case, this PR says that the `li#li` element should also be hidden because it has an "owning element" (here, DOM ancestor) which is hidden.

What happens in the reverse case:
```html
<div aria-hidden="true">
  <div id="owner" aria-owns="li" />
</div>
<ul> <li id="li">Item</li> </ul>
```
In that case, a hidden element `aria-owns` an element with no hidden ancestor. The clarification in this PR also means that `li#li` should be hidden because it has an "owning element" (here, an element with `aria-owns`) which is hidden.
Quick tests, it seems that browsers ignore `aria-owns` in this case.
So, mostly raising the point to make sure this is the intention of the PR.

---
Second point is that the language in [§7.1  Excluding Elements from the Accessibility Tree](https://w3c.github.io/aria/#tree_exclusion) uses a lot of "descendant" and other tree-related wording, notably for the case concerning this PR:
> user agents SHOULD NOT include the following elements in the accessibility tree:
> * Elements, including their descendants, that have [aria-hidden](https://w3c.github.io/aria/#aria-hidden) set to true.

Should there be a precision that "descendants" here mean "either descendant in the DOM tree, or descendant in the accessibility tree where `aria-owns` relations are taken into account"? Most of this Section would benefit from such clarification. Maybe it's a topic for another PR, though 🤔

pkra:
For the second point: https://github.com/w3c/aria/issues/1150

adampage:
> In that case, a hidden element `aria-owns` an element with no hidden ancestor.

Mmm, thanks for poking at this, @Jym77.  I hadn’t considered that reverse case.

[“Owning element” is currently defined](https://pr-preview.s3.amazonaws.com/w3c/aria/1839/be8f3c9...adampage:9870bb2.html#dfn-owning-element:~:text=Owning%20Element,ID%20of%20the%20element.) as “_any DOM ancestor of the element, or any element with an `aria-owns` attribute which references the ID of the element._”

In this reverse scenario, `<div aria-hidden="true">` is neither a DOM ancestor of `li#li` nor does it use `aria-owns` to reference `li#li` directly.  It feels a bit lawyerly 🤔, but I think we could then say that `<div aria-hidden="true">` is **not** an “owning element” of `li#li`.  `div#owner` is hidden because its DOM ancestor is `aria-hidden`, so its use of `aria-owns` is no longer exposed to the accessibility API.  This leaves `li#li` in its original location, owned by its parent `ul`.

Jym77:
@adampage That actually makes sense 😄 
This PR seems to change the behaviour in the related simplified case:
```html
<div aria-hidden="true" aria-owns="li"></div>

<ul> <li id="li">Item</li> </ul>
```
In that case, the current behaviour is to expose `li#li` as it has no _ancestor_ with `aria-hidden` (and Chrome and FF do that), but the new behaviour will be to hide it since it has a _owning element_ with `aria-hidden`.

I guess it's a bit of a question which of `aria-owns` or `aria-hidden` "acts faster". The current behaviour is that `aria-hidden` hides away the owner before it has time to own anything else, but the new behaviour is that `aria-owns` is tweaking the tree before the owner gets hidden 🤔 

---
Can get weirder with a case like:
```html
<div aria-hidden="true" aria-owns="foo"></div>
<div id="foo" aria-owns="bar"></div>
<div id="bar"></div>
```
Now, `#foo` has a owning element with `aria-hidden`, so it is hidden. But since it itself does not have `aria-hidden`, it still owns `#bar`, which is also not hidden in any way. I feel that is at best confusing and might explode in weird ways 🤔

Jym77:
> “_Changes to ownership are resolved in DOM order. If both `aria-hidden` and `aria-owns` are specified on the same element, `aria-hidden` is processed first and `aria-owns` will not take effect._”

> Whew. Having said all that, would you agree that the gist of my additional spec proposal is helpful & accurate? And that it would reasonably lead to the interpretations I described for each of those four scenarios? Are there yet other ambiguous scenarios lurking out there? 😅

Yes, I think that clarification works. Maybe we need to add that change of ownership is only initiated by non-hidden node (that's more or less the second sentence but making it clear that one should stop looking for `aria-owns` upon encountering a `aria-hidden`, i.e. the way case 2 works)¹
From a graph-theory point of view, we'd walk the DOM tree in DOM order, stop the current traversal on `aria-hidden` or already traversed node, and jump to follow `aria-owns` (which does trigger the "already traversed node" in the previous check).

This also makes it clear that `aria-hidden` affects the accessibility tree while `display: none` affects the DOM tree (so replacing `aria-hidden` by `display: none` in case 2 does **not** un-hide the `li`, it is still not rendered, therefore hidden).


Great clarification 👍

scottaohara:
was following along with this until:
>This also makes it clear that aria-hidden affects the accessibility tree while display: none affects the DOM tree (so replacing aria-hidden by display: none in case 2 does not un-hide the li, it is still not rendered, therefore hidden).

this statement has made me wonder if this actually makes sense or not.  since display none does modify the a11y tree... and then i'm not sure what case 2 refers to?  the second of Adam's examples in his last comment, or the second odd case you brought up?

Jym77:
Yep, poor writing on my side 🙈 and messing up things badly…
```html
  <ul aria-owns="li"></ul>
  <div style="display: none">
    <li id="li">Child node</li>
  </div>
```
case **1 not 2** in Adam's list with `aria-hidden` replaced by `display: none`.

This **does not expose** `li#li` since it hits the "Elements, including their descendent elements, that have host language semantics specifying that the element is not displayed" condition.

That is of course linked to non-interference with the host language (`aria-owns` cannot expose an element that HTML is hiding, but can expose an element that ARIA is trying to hide (`aria-hidden`)). Which is more or less what I was trying to say by putting `display: none` at the DOM level and `aria-hidden` at the accessibility tree level.
Of course, `display: none` affects the accessibility tree, but indirectly because it acts on the DOM tree in way that accessibility tree building is concerned, that's why I was writing that it affects the DOM tree.

adampage:
So, I reacquainted myself with [the spec’s definition of “_hidden_”](https://pr-preview.s3.amazonaws.com/w3c/aria/1839/cd9b5c4...adampage:d414dcf.html#dfn-hidden):

> Indicates that the element is not visible, perceivable, or interactive to _any_ user. An element is considered hidden if it or any one of its ancestor elements is not rendered or is explicitly hidden.

Based on that, I took [another pass at the `aria-hidden` section](https://pr-preview.s3.amazonaws.com/w3c/aria/1839/cd9b5c4...adampage:d414dcf.html#:~:text=An%20element%20will%20be%20excluded%20from%20the%20accessibility%20tree%20if%20it%20or%20any%20of%20its%20owning%20elements%20are%20hidden%20or%20have%20their%20aria%2Dhidden%20attribute%20value%20set%20to%20true.) with the goal of disambiguating its use of the term “_hidden_” by leveraging the existing “_Excluding Elements from the Accessibility Tree_” section.

Beyond that, I wordsmithed the clarifications we discussed earlier in this thread and [added them to the `aria-owns` section](https://pr-preview.s3.amazonaws.com/w3c/aria/1839/cd9b5c4...adampage:d414dcf.html#:~:text=aria%2Downs%20is%20resolved,the%20accessibility%20tree.).

Jym77:
I feel there is still a small clarification needed to say that `aria-owns` breaks DOM parent/child relation in addition to create new ones (i.e. the result is indeed an accessibility tree, not a DAG).

[Owning element](https://w3c.github.io/aria/#dfn-owning-element) states:
> An 'owning element' is any DOM ancestor of the [element](https://dom.spec.whatwg.org/#concept-element), or any element with an [aria-owns](https://w3c.github.io/aria/#aria-owns) attribute which references the ID of the element.

Because of the "or", and no mention of unicity, in case 1 above
```html
<ul aria-owns="li"></ul>
  <div aria-hidden="true">
    <li id="li">Child node</li>
  </div>
```
Both the `ul` and the `div` are "owning element" for the `li` (or at least the definition doesn't really tells which to use), and the definition of `aria-owns` doesn't really answer that question either. (and since any DOM ancestor is a "owning element", it is actually not a problem to have more than one owning element).

So, without disambiguating that, the `li` has an `aria-hidden` "owning element" (the `div`) and should be hidden, which is precisely what we want to avoid.

Should we change the definition of "owning element" to state something like
> An 'owning element' is any element with an [aria-owns](https://w3c.github.io/aria/#aria-owns) attribute which references the ID of the element or, if none, any DOM ancestor of the [element](https://dom.spec.whatwg.org/#concept-element).

Otherwise, I think this is correct and expresses what is needed 👍

spectranaut:
I just want to note that @jym77's recommended definition of owned is nice, and, that there is also a 1.3-blocking issue related to inconsistent use the term "owned by", just like how hidden was just revisited:

https://github.com/w3c/aria/issues/1161

The "owned by inconsistencies" can be addressed before of after this PR, but it might also lead to a term change or term update.

adampage:
Thanks @spectranaut and @Jym77, I’ve just passed along your suggested clarification for “owning element” in PR #1454. I think it will be helpful to address it in that context.

If that goes through, then I’d plan to update this PR by leveraging @spectranaut’s [recently clarified definition of “hidden”](https://github.com/w3c/aria/pull/1856/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R312), the improved definition of “owning element”, and the new “accessibility parent” term.  My first instinct would be to update [the `aria-hidden` state definition](https://pr-preview.s3.amazonaws.com/adampage/aria/pull/1839.html#:~:text=User%20agents%20determine,to%20true.) with something like:

```html
<p>
  ...An element is considered [=element/hidden=] if:
</p>
<ul>
  <li>
    It or any of its <a>accessibility parents</a> are not rendered
  </li>
  <li>
    It or any of its <a>owning elements</a> are not rendered or have their
    <code>aria-hidden</code> attribute value set to <code>true</code>
  </li>
</ul>
```

Could I trouble you to first take a look at [my comment in #1454](https://github.com/w3c/aria/pull/1454#discussion_r1106590105), and then consider my proposal here?

adampage:
I’ve created [a new Codepen to describe & test assorted scenarios involving `aria-owns` and hidden content](https://codepen.io/adampage/full/QWBRPWq/91bf1f14960448d246eed01699a2749a).

We originally [decided to align the spec with the current behavior of Chrome & Firefox](https://www.w3.org/2022/10/20-aria-minutes#x168:~:text=want%20to%20do%20what%20Chrome%20and%20Firefox%20do%2C%20then%20update%20spec), but my testing showed some interesting inconsistencies for Chrome 110 on macOS vs Windows.  Firefox 110 seemed to behave consistently with our expectations everywhere.

Once #1454 is finalized, I’ll take another whack at the spec language for this PR, orienting it around Firefox’s demonstration of our intent.

spectranaut:
> I’ve created [a new Codepen to describe & test assorted scenarios involving `aria-owns` and hidden content](https://codepen.io/adampage/full/QWBRPWq/91bf1f14960448d246eed01699a2749a).

omg this code pen is AMAZING.

index.html
```diff
@@ -11756,7 +11756,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 			<sdef>aria-hidden</sdef>
 			<div class="state-description">
 				<p><a>Indicates</a> whether the <a>element</a> is exposed to an accessibility API. See related <sref>aria-disabled</sref>.</p>
-				<p>User agents determine an element's [=element/hidden=] status based on whether it is rendered, and the rendering is usually controlled by CSS. For example, an element whose <code>display</code> property is set to <code>none</code> is not rendered. An element is considered [=element/hidden=] if it, or any of its ancestors are not rendered or have their <code>aria-hidden</code> attribute value set to <code>true</code>.</p>
+				<p>User agents determine an element's [=element/hidden=] status based on whether it is rendered, and the rendering is usually controlled by CSS. For example, an element whose <code>display</code> property is set to <code>none</code> is not rendered. An element will be <a href="#tree_exclusion">excluded from the accessibility tree</a> if it or any of its <a>owning elements</a> are [=element/hidden=] or have their <code>aria-hidden</code> attribute value set to <code>true</code>.</p>
```
scottaohara, line 11759: i think my only issue with this update is that the clarification about when to take aria-owns into account is so far removed from this initial mention of owning elements.  As prior to reading the new paragraph added on line 12333 of this diff, i would have interpreted the use of "owning elements" to mean that "boo" should also be "hidden" because the hidden div "owns" boo... but for the fact on 12333 it is stated that's not true because the owning element is itself not rendered, so thus it can't own the "boo" element.

```
<div hidden aria-owns=boo>foo</div>
<div id=boo>boo</div>
```
i could see this clarified by either mentioning that if an element is an owning element due to usage of aria-owns, how this works is clarified later.  Or that these bits of text are closer together in the spec, or that this paragraph remains just about ancestors being hidden, and owning elements / aria-owns headaches are talked about in one new paragraph/sets of paragraphs.

cookiecrook:
> > I’ve created [a new Codepen to describe & test assorted scenarios involving `aria-owns` and hidden content](https://codepen.io/adampage/full/QWBRPWq/91bf1f14960448d246eed01699a2749a).
> 
> omg this code pen is AMAZING.

Agreed. Most useful CodePen I think I've ever seen.  

@AdamPage, you are planning to turn some of this into a series of WPT tests for https://github.com/web-platform-tests/interop-2023-accessibility-testing/issues/32 correct? Are you planning to file the implementation bugs you've discovered? Trying to determine how best to cross-reference the trackers from the issue, etc.

index.html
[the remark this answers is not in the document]
cookiecrook: > This current wording excludes that I think.

The below test is terrible but something like:
- if the current element is hidden/unrendered (def tbd), OR
- (if the element does not have aria-hidden=true set AND its nearest ancestor with aria-hidden defined is (both aria-hidden=false and NOT hidden aka rendered))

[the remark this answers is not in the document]
cookiecrook: I acknowledge that distinction is difficult to parse. Perhaps a Deep Dive (w./ Aaron) or Agenda+ for a weekly?

adampage: Heya @scottaohara, now that the accessibility parent/child language has been merged in, I [updated this statement](https://github.com/w3c/aria/pull/1839/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R11607) to reference `accessibility parents` instead of `owning elements`. I think this is now more accurate and also sidesteps the ambiguity you described here, at least until an author digs deeper into the `aria-owns` section? Here’s the full new sentence:

> An element will be excluded from the accessibility tree if it or any of its **accessibility parents** are hidden or have their `aria-hidden` attribute value set to `true`.

@spectranaut, I’m seeing that this would be the spec’s first mention of _accessibility **parents**_ plural. I re-read that term’s definition and it feels okay to pluralize it here to essentially mean _ancestors_. Do you agree?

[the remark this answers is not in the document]
adampage: In anticipation of #2090 getting merged (in which we made `aria-hidden="false"` synonymous with `undefined`), I’m now thinking that this part can be left as is?

adampage:
> @adampage, you are planning to turn some of this into a series of WPT tests for [web-platform-tests/interop-accessibility#32](https://github.com/web-platform-tests/interop-accessibility/issues/32) correct? Are you planning to file the implementation bugs you've discovered? Trying to determine how best to cross-reference the trackers from the issue, etc.

Heya @cookiecrook, I just turned all of these tests into comparable (I hope) WPT tests: [PR #44822](https://github.com/web-platform-tests/wpt/pull/44822). Using your latest tip, I’ve marked that file as `.tentative`, but will definitely plan to file any legit implementation bugs once this ARIA spec PR (and that WPT PR, if you think it deserves to be in the test suite) is accepted and merged.

I’d be grateful for your thoughts on my approach for testing this spec change within WPT’s capabilities. My original Codepen tests weren’t accname-focused; I just rendered a bunch of `generic`s and went looking in the a11y tree to see how they all got exposed as static text nodes.

With WPT, I chose to use the test driver’s `get_computed_label()` function and translated all my test markup scenarios into `<button>` elements relying on a **name from content** accname assembly. I’m hoping this is roughly equivalent, but the ARIA spec clarification I’m attempting to make in this PR is bigger than accname.

index.html
spectranaut: Ahhh actually no, there is only one "accessibility parent" -- I think the appropriate phrase here is "Accessibility ancestor". We don't have it defined, but I think it is a clear enough term? You could link to the accessibility tree section of the document: https://w3c.github.io/aria/#accessibility_tree

spectranaut (changes requested):
I think this looks good now, except for the switch from "accessibility parents" to "accessibility ancestors"! I'll mark as approving once that is done :)

adampage:
Thanks, @spectranaut. I’ve [swapped in “_accessibility ancestors_”](https://github.com/w3c/aria/pull/1839/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R11605) for “_accessibility parents_” and resolved the merge conflict.

spectranaut (approved):
Looks good!

spectranaut:
@scottaohara and @cookiecrook -- this is another oldie that I think we can land soon. You both have taken a look and provided feedback and Adam responded, can you take another review?

adampage:
Heya @cookiecrook / @scottaohara, could I trouble you for a review of this oldie-but-goodie?

I also recently overhauled a supporting WPT test: web-platform-tests/wpt#44822, with [these latest results](https://wpt.fyi/results/accname/aria-owns.html?label=pr_head&max-count=1&pr=44822).

index.html
```diff
@@ -14802,6 +14802,29 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+            </pre>
+            <p>In the following example, “(opens in a new window)” remains <a href="#tree_exclusion">excluded from the accessibility tree</a> since its <abbr title="Document Object Model">DOM</abbr> ancestor is [=element/hidden from all users=] in host language terms.</p>
+            <pre class="example highlight">
+&lt;a href="https://www.w3.org/" aria-owns="new-window-warning"&gt;
+  World Wide Web Consortium
+&lt;/a&gt;
+&lt;div hidden&gt;
+  &lt;span id="new-window-warning"&gt; (opens in a new window)&lt;/span&gt;
+&lt;/div&gt;
+            </pre>
```
scottaohara, line 14827: per my comment in the related WPT, I would very much like this final example added.

I think this PR looks good, otherwise.
```suggestion
            </pre>
             <p>In the following example, the <code>&lt;div id="instructions"&gt;</code> and its content remains <a href="#tree_inclusion">included in the accessibility tree</a> 
		since the element referencing it with the <code>aria-owns</code> attribute is [=element/hidden from all users=].</p>
            <pre class="example highlight">
&lt;div hidden aria-owns="instructions"&gt;
  ...
&lt;/a&gt;
...
&lt;div id="instructions"&gt;
  Instructions go here...
&lt;/div&gt;
            </pre>
```

scottaohara (commented):
I'm good with everything that's been added for this PR, but I would like the example I created merged in as well.

adampage:
Thanks very much, @scottaohara! I committed your example and then [made a couple follow-on editorial tweaks](https://github.com/w3c/aria/pull/1839/commits/e80dc49f6af3260db73afea917fb1ed4d5321b51).

Does this read okay to you?:

> In the following example, `<div id="instructions">` and its text content remain exposed and unmoved [in the accessibility tree](https://github.com/w3c/aria/pull/1839#tree_inclusion) since the element attempting to reparent it with `aria-owns` is _also_ [hidden from all users](https://github.com/w3c/aria/pull/1839#dfn-hide-from-all-users).

scottaohara:
i'm going to approve as is - but i don't think "also" needed to be added.  that made me initially think that the sentence was trying to say that the element that was attempting to be reparented was hidden from the a11y tree - and it made me go back to look if that had been changed

scottaohara (approved):

adampage:
Updated. Thank you, @scottaohara. 🙏🏻

scottaohara (approved):

spectranaut:
Agenda-ing this also for the tests. We should get this landed, and the tests, if we want to move forward with this change.

adampage:
Here are the [latest browser results](https://wpt.fyi/results/accname/aria-owns.html?label=pr_head&max-count=1&pr=44822) for [the companion WPT test](https://github.com/web-platform-tests/wpt/pull/44822).

spectranaut:
Just read the minutes from yesterday's meeting: https://www.w3.org/2025/09/04-aria-minutes.html#9476

Since this is largely in agreement with two implementations (besides minor bugs) I think we should land this change THEN open a bug on webkit and everyone else.

spectranaut:
Actually, I said this is ready for merge, but it would be nice to get an approving review from @cookiecrook first

index.html
```diff
@@ -14857,6 +14857,39 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+             <p>In the following example, <code>&lt;div id="instructions"&gt;</code> and its text content remain exposed and unmoved <a href="#tree_inclusion">in the accessibility tree</a> since the element attempting to reparent it with <code>aria-owns</code> is [=element/hidden from all users=].</p>
```
cookiecrook: Reparent has a particular meaning in psychology, but I am not aware of a formal definition that matches the meaning used here. Rephrasing for clarity.

```suggestion
             <p>In the following example, <code>&lt;div id="instructions"&gt;</code> and its text content remain exposed and unmoved <a href="#tree_inclusion">in the accessibility tree</a> since the would-be accessibility parent element with <code>aria-owns</code> is [=element/hidden from all users=].</p>
```

cookiecrook (approved):
Approving with one editorial clarification.

adampage:
Thanks, @cookiecrook. I’ve committed that suggestion and merged in the latest, @spectranaut 🚀

PR #1852, opened 2022-12-07
editorial: add note prohibited attribs are still global

### LINKED ISSUE #1824: Does aria-label trigger presentation role conflict?
In ARIA 1.2 the concept of prohibited properties and states was introduced. This included prohibiting `aria-label` and `aria-labelledby` on `presentation` and `none`. Reading the description of global attributes then seems to suggest (but not explicitly state) that prohibited props are not global.

> ## 6.5 Global States and Properties
> ... The following global states and properties are supported by all roles and by all base markup elements unless otherwise prohibited... 
> 
> * `aria-label` (Except where prohibited)

If `aria-label` is no longer considered a global property for `presentation`, that means using it wouldn't cause presentational role conflict to occur, and so some browsers have started ignoring `aria-label` when used on presentational elements (implicit and explicit):

```html
<img src="img1.png" alt="" aria-label="Hello world" />
<img src="img2.png" role="none" aria-label="Hello world" />
```

| Browser | img1     | img2     |
|---------|----------|----------|
| Firefox | Included | Included |
| Chrome  | Included | Skipped  |
| Safari  | Skipped  | Included |

It's not clear whether or not this was the intended behavior for prohibited properties. In either case it would be good for the presentational role conflict resolution to clarify what should happen with prohibited attributes.

This issue came up in migrating ACT rules to WAI-ARIA 1.2, and is one of the outstanding questions we have for updating our [Element marked as decorative is not exposed](https://www.w3.org/WAI/standards-guidelines/act/rules/46ca7f/) rule.

jnurthen:
Closes #1824

### DIFF
--- a/index.html
+++ b/index.html
@@ -591,7 +591,7 @@
 			<p>Informative list of properties that are inherited by a <a>role</a> from superclass roles. <a>States</a> and [=ARIA/properties=] are inherited from superclass roles in the Roles Model, not from ancestor [=elements=] in the <abbr title="Document Object Model">DOM</abbr> tree. These properties are not explicitly defined on the role, as the inheritance of properties is automatic. This information is provided to facilitate reading of the specification. The set of supported states and properties combined with inherited states and properties forms the full set of states and properties supported by the role.</p>
 		</section>
 		<section id="prohibitedattributes">
-				<h3>Prohibited States and Properties</h3>
+				<h3><dfn>Prohibited</dfn> States and Properties</h3>
 				<p>List of states and properties that are prohibited on a <a>role</a>. Authors MUST NOT specify a prohibited state or property.</p>
 				<p class="note">A host language attribute with the appropriate <a href="#implicit_semantics">implicit WAI-ARIA semantic</a> would also prohibit a state or property in this section. </p>
 			</section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -13597,8 +13597,9 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
 			<li>If a <a href="#mustContain">required owned element</a> has an explicit non-presentational role, user agents MUST ignore an inherited presentational role and expose the element with its explicit role. If the action of exposing the explicit role causes the accessibility tree to be malformed, the expected results are undefined.</li>
-			<li>If an element has global WAI-ARIA states or properties, user agents MUST ignore the <code>presentation</code> role and instead expose the element's implicit role. However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</li>
+			<li>If an element has <a>global</a> WAI-ARIA states or properties a, user agents MUST ignore the <code>presentation</code> role and instead expose the element's implicit role. However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</li>
```
scottaohara, line 13600: was that "a" after properties supposed to be added?

index.html
```diff
@@ -13597,8 +13597,9 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
 		</ul>
+		<p class="note">Some <a>global</a> WAI-ARIA states and properties are <a>prohibited</a> on some roles. These states and properties are still considered global states and properties for the purposes of Presentational Role Conflict resolution.</p>
```
scottaohara, line 13602: minor wording nit to not repeat "some" twice in the same sentence.  fully aware this is a personal opinion so please ignore if you would like.

```suggestion
		<p class="note">Some <a>global</a> WAI-ARIA states and properties are <a>prohibited</a> on certain roles. These states and properties are still considered global states and properties for the purposes of Presentational Role Conflict resolution.</p>
```

index.html
MelSumner, line 13600:
```suggestion
			<li>If an element has <a>global</a> WAI-ARIA states or properties, user agents MUST ignore the <code>presentation</code> role and instead expose the element's implicit role. However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</li>
```

MelSumner (changes requested):
One small change and +1 to Scott's feedback.

PR #1856, opened 2023-01-13
Editorial: Update hidden term definition to match accname and core-aam

### LINKED ISSUE #1159: Inconsistent use of the term "hidden"
The aria-hidden description says the following:

> source: https://www.w3.org/TR/wai-aria-1.2/#aria-hidden
> An element is considered hidden if it, or any of its ancestors are not rendered or have their aria-hidden attribute value set to true.

This second part, of an element being considered hidden when it has aria-hidden="true" is inconsistent with the definition of hidden, which says: 

> Indicates that the element is not visible, perceivable, or interactive to any user. An element is considered hidden if it or any one of its ancestor elements is not rendered or is explicitly hidden.

"not visible ... to any user" is not necessarily the case for elements with aria-hidden="true". More commonly, aria-hidden="true" is used for elements that are visible but should not be included in the accessibility tree.

spectranaut:
Closes #1159

I introduced a new term, "hidden from all users", to capture the original definition of hidden, and changed the definition of "hidden" to mean what is means in the accname specification. 

The new definition of "hidden" also works in all cases where "hidden" is referenced in core-aam and the most of the aria spec. When necessary, I changed "hidden" to "hidden from all users" -- which is only in the tab panel definition.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1856.html" title="Last updated on Feb 9, 2023, 6:12 AM UTC (710ba4f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1856/88d8616...710ba4f.html" title="Last updated on Feb 9, 2023, 6:12 AM UTC (710ba4f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -309,7 +309,10 @@
 			</dd>
 			<dt><dfn data-dfn-for="element" data-export="">Hidden</dfn></dt>
 			<dd>
-				<p>Indicates that the <a>element</a> is not visible, <a>perceivable</a>, or interactive to <em>any</em> user. An element is considered <em>hidden</em> if it or any one of its ancestor elements is not rendered or is explicitly hidden.</p>
+				<p>Indicates that the <a>element</a> is absent from the accessibility tree and is not exposed to accessibility APIs. An element is considered hidden if it or any one of its ancestor elements has been explicitly hidden, either by CSS properties <code>display:none</code>, <code>visibility:hidden</code> or <code>visibility:collapse</code>, or with the aria attribue <code>aria-hidden</code>.<p>
+			<dt><dfn data-dfn-for="element" data-lt="hide from all users" data-export="">Hidden From All Users</dfn></dt>
+			<dd>
+				<p>Indicates that the <a>element</a> is not visible, <a>perceivable</a>, or interactive to <em>any</em> user. An element is considered <em>hidden from all users</em> if it or any one of its ancestor elements is not rendered or is explicitly hidden from all users. In the ARIA specifications, it is important to understand an <a>element</a> can be [=element/hidden=] by removing it from the accessibility tree but not [=element/hidden from all users=] by the use of <code>aria-hidden</code>.</p>
 			</dd>
 			<dt><dfn>Identifies</dfn></dt>
 			<dd>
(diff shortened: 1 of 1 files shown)


chlane (approved):
I agree with being explicit about addressing "all users" since ARIA does not visually hide elements

MelSumner (approved):
👍 :shipit: 

Thank you for doing this work!

spectranaut:
@pkra, I'd like if an editor could take a look at this, if you have the time!

adampage (approved):
Thanks for this clarification!

I recently encountered the ambiguity of “hidden” in #1839 and ended up [relying on its current meaning as being hidden from _all users_](https://github.com/w3c/aria/pull/1839#issuecomment-1378194801).  As soon as #1856 is merged, I’ll update my PR to use the newer, more explicit definition.

index.html
```diff
@@ -309,7 +309,10 @@ <h1>Important Terms</h1>
...
+				<p>Indicates that the <a>element</a> is absent from the accessibility tree and is not exposed to accessibility APIs. An element is considered hidden if it or any one of its ancestor elements has been explicitly hidden, either by CSS properties <code>display:none</code>, <code>visibility:hidden</code> or <code>visibility:collapse</code>, or with the aria attribue <code>aria-hidden</code>.<p>
+			<dt><dfn data-dfn-for="element" data-lt="hide from all users" data-export="">Hidden From All Users</dfn></dt>
+			<dd>
+				<p>Indicates that the <a>element</a> is not visible, <a>perceivable</a>, or interactive to <em>any</em> user. An element is considered <em>hidden from all users</em> if it or any one of its ancestor elements is not rendered or is explicitly hidden from all users. In the ARIA specifications, it is important to understand an <a>element</a> can be [=element/hidden=] by removing it from the accessibility tree but not [=element/hidden from all users=] by the use of <code>aria-hidden</code>.</p>
```
pkra, line 315: Slight wording suggestion:

```suggestion
				<p>Indicates that the <a>element</a> is not visible, <a>perceivable</a>, or interactive to <em>any</em> user. An element is considered <em>hidden from all users</em> if it or any one of its ancestor elements is not rendered or is explicitly hidden from all users. In the ARIA specifications, it is important to understand that by using <code>aria-hidden</code> an <a>element</a> can be [=element/hidden=] but not [=element/hidden from all users=] .</p>
```

 I'm not feeling too strongly about it but had to read the long sentence twice.

index.html
```diff
@@ -309,7 +309,10 @@ <h1>Important Terms</h1>
 			</dd>
 			<dt><dfn data-dfn-for="element" data-export="">Hidden</dfn></dt>
 			<dd>
-				<p>Indicates that the <a>element</a> is not visible, <a>perceivable</a>, or interactive to <em>any</em> user. An element is considered <em>hidden</em> if it or any one of its ancestor elements is not rendered or is explicitly hidden.</p>
+				<p>Indicates that the <a>element</a> is absent from the accessibility tree and is not exposed to accessibility APIs. An element is considered hidden if it or any one of its ancestor elements has been explicitly hidden, either by CSS properties <code>display:none</code>, <code>visibility:hidden</code> or <code>visibility:collapse</code>, or with the aria attribue <code>aria-hidden</code>.<p>
```
pkra, line 312: I found it somewhat confusing to have the old content here; it feels like most of it belongs with "hidden from all users". I also think it would be good to have short definitions of terms and mostly reference the spec for details (so that we don't have to update several places).

I.e., something like:

"Indicates that the element is excluded from the accessibility tree [link to 7.1]  and therefore not exposed to accessibility APIs. For example, an element is considered hidden if it is `hidden for all users` or by using the aria attribute `aria-hidden` on it or any one of its ancestor elements ."

or even shorter:

"Indicates that the element is excluded from the accessibility tree and therefore not exposed to accessibility APIs. See also [ 7.1 ] , [ hidden for all users ] , [ aria-hidden ]"

index.html
pkra: I wrote this before the comment for `hidden`; that comment might apply here, too.

index.html
pkra: I'm not feeling too strongly about it since I think that there's nothing technically wrong with the text as is.

spectranaut: Thanks for this feedback/perspective, I like the first slightly longer suggestion and will incorporate!

index.html
spectranaut: ok I did edit it!

pkra (approved):

PR #1018, opened 2019-07-22
Issue899 accname from heading

jongund:
Update 2023-Jan-25: This PR abandoned in favor of #1860 and can be closed.

No longer attached to https://github.com/w3c/accname/issues/138 (Note: PR modified by @cookiecrook)

[Proposed definition of accessible name from heading](https://raw.githack.com/w3c/aria/issue899-accname-from-heading/index.html#namecalculation)

[Proposed list of roles that support name from heading](https://raw.githack.com/w3c/aria/issue899-accname-from-heading/index.html#namefromheading)

## Non-Landmark Roles Authoring Techniques
* [Proposed changes to the `alertdialog` role authoring requirements](https://raw.githack.com/w3c/aria/issue899-accname-from-heading/index.html#alertdialog)
* [Proposed changes to the `article` role authoring requirements](https://raw.githack.com/w3c/aria/issue899-accname-from-heading/index.html#article)
* [Proposed changes to the `dialog` role authoring requirements](https://raw.githack.com/w3c/aria/issue899-accname-from-heading/index.html#dialog)

## Landmark Roles Authoring Techniques
* [Proposed changes to the `complementary` role authoring requirements](https://raw.githack.com/w3c/aria/issue899-accname-from-heading/index.html#complementary)
* [Proposed changes to the `navigation` role authoring requirements](https://raw.githack.com/w3c/aria/issue899-accname-from-heading/index.html#navigation)
* [Proposed changes to the `region` role authoring requirements](https://raw.githack.com/w3c/aria/issue899-accname-from-heading/index.html#region)

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jan 20, 2021, 10:59 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Faria%2F487fb4f3d2053c865ff7ccbc93a35c7547e67ce4%2Findex.html%3FisPreview%3Dtrue)

```
waiting for function failed: timeout 30000ms exceeded
```

[...]

PR #1860, opened 2023-01-26
Update from PR #1018 for nameFrom: heading

### LINKED ISSUE #899: ARIA Spec could be more flexible when elements with "nameFrom:author" are left unlabeled by the author
I received the following bug report from a member of my team.

> Summary:
> Web AX, VoiceOver: When browsing list of articles using VO Articles rotor on [site redacted], VO announces each item as "Article" without speaking headline
> 
> Steps To Reproduce:
> 1. Turn on VoiceOver.
> 2. Open Safari and browse to cnn.com.
> 3. Press VO-U to bring up rotors.
> 4. Press VO-Right arrow repeatedly to select “Articles” rotor.
> 5. Explore rotor items by pressing VO-Down arrow.
> 
> Results:
> When using VO’s Articles rotor, VO does show all the articles presented on [site redacted], but only speaks each item as “Article” without providing any additional info, such as the article’s headline. For better clarity, VO should speak some additional information here.

Upon closer inspection, the articles are not labeled but they all have a heading in close proximity to the start of the article element. For the sake of the user, we could probably work around this web page error heuristically in either WebKit or VoiceOver, but the ARIA Spec appears to explicitly discourage this.

The ARIA spec could add some language explicitly allowing user agents to heuristically determine the label for unlabeled elements whose label has not been set by the author. In this case, the most logical choice would be to find the first heading in the article.

Perhaps: "When Authors fails to label elements who role is defined as `nameFrom: author` User Agents MAY use heuristics to determine the element's label. For example, browsers may set the label of an unlabeled dialog to be the first heading in that dialog."

cookiecrook:
Closes #899

Adds `nameFrom: heading` (this overcomes PR #1018)

Will update these along the way after further discussion.

* [ ] Related APG Issue/PR:

# Implementation tracking

* [ ] validator tests
* [ ] WPT tests: [LINK]()
   * WebKit: [ISSUE]()
   * Gecko: [ISSUE]()
   * Blink: [ISSUE]()
* [x] Does this need AT implementations? Unlikely. Should be a engine-specific change.

### DIFF
--- a/index.html
+++ b/index.html
@@ -617,9 +617,11 @@
 				<dd>One of the following values:
 					<ol>
 						<li>author: name comes from values provided by the author in explicit markup features such as the <pref>aria-label</pref> attribute, the <pref>aria-labelledby</pref> attribute, or the host language labeling mechanism, such as the <code>alt</code> or <code>title</code> attributes in <abbr title="Hypertext Markup Language">HTML</abbr>, with HTML <code>title</code> attribute having the lowest precedence for specifying a text alternative.</li>
-						<li>contents: name comes from the text value of the <a>element</a> node. Although this may be allowed in addition to "author" in some <a>roles</a>, this is used in content only if higher priority "author" features are not provided. Priority is defined by the <a href="#mapping_additional_nd_te" class="accname">accessible name and description computation</a> algorithm [[ACCNAME-1.2]].</li>
+						<li>heading: name comes from the text value of the first descendant (i.e., depth first) <a>element</a> node with the role of <code>heading</code>. Although "heading" may be allowed in addition to "author" in some <a>roles</a>, "heading" is used in content only if higher priority "author" features are not provided.</li>
+						<li>contents: name comes from the text value of the <a>element</a> node. Although this may be allowed in addition to "author" in some <a>roles</a>, this is used in content only if higher priority "author" features are not provided.</li>
             <li>prohibited: the element does not support name from author. Authors MUST NOT use the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attributes to name the element.</li>
 					</ol>
+					<p>Priority is defined by the <a href="#mapping_additional_nd_te" class="accname">accessible name and description computation</a> algorithm. [[ACCNAME-1.3]]<p>
 				</dd>
 			</dl>
 			<section id="namecomputation">
@@ -634,21 +636,22 @@
 				<h4>Accessible Name and Description Computation</h4>
 				<p><a href="#mapping_additional_nd_te" class="accname">Accessible Name and Description Computation</a> is defined in the Accessible Name and Description specification.</p>
 			</section>
-            <section id="namefromauthor">
-                <h4>Roles Supporting Name from Author</h4>
-                <div id="index_fromauthor">
-                </div>
+			<section id="namefromauthor">
+				<h4>Roles Supporting Name from Author</h4>
+				<div id="index_fromauthor"></div>
 			</section>
-            <section id="namefromcontent">
-                <h4>Roles Supporting Name from Content</h4>
-                <div id="index_fromcontent">
-                </div>
+			<section id="namefromheading">
+				<h4>Roles Supporting Name from Heading</h4>
+				<div id="index_fromheading"></div>
+			</section>
+			<section id="namefromcontent">
+				<h4>Roles Supporting Name from Content</h4>
+				<div id="index_fromcontent"></div>
 			</section>
 			<section id="namefromprohibited">
-                <h4>Roles which cannot be named (Name prohibited)</h4>
-                <div id="index_fromprohibited">
-                </div>
-            </section>
+				<h4>Roles which cannot be named (Name prohibited)</h4>
+				<div id="index_fromprohibited"></div>
+			</section>
 		</section>
 		<section id="childrenArePresentational">
 			<h3>Presentational Children</h3>
@@ -988,7 +991,12 @@
 					</tr>
 					<tr>
 						<th class="role-namefrom-head" scope="row">Name From:</th>
-						<td class="role-namefrom">author</td>
+						<td class="role-namefrom">
+ 							<ul>
+ 								<li>author</li>
+ 								<li>heading</li>
+ 							</ul>
+ 						</td>
 					</tr>
 					<tr>
 						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
@@ -1168,7 +1176,12 @@
 					</tr>
 					<tr>
 						<th class="role-namefrom-head" scope="row">Name From:</th>
-						<td class="role-namefrom">author</td>
+						<td class="role-namefrom">
+ 							<ul>
+ 								<li>author</li>
+ 								<li>heading</li>
+ 							</ul>
+ 						</td>
 					</tr>
 					<tr>
 						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
@@ -2660,7 +2673,12 @@
 					</tr>
 					<tr>
 						<th class="role-namefrom-head" scope="row">Name From:</th>
-						<td class="role-namefrom">author</td>
+						<td class="role-namefrom">
+ 							<ul>
+ 								<li>author</li>
+ 								<li>heading</li>
+ 							</ul>
+ 						</td>
 					</tr>
 					<tr>
 						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
@@ -3018,7 +3036,6 @@
 			<div class="role-description">
 				<p>A dialog is a descendant window of the primary window of a web application. For <abbr title="Hypertext Markup Language">HTML</abbr> pages, the primary application window is the entire web document, i.e., the <code>body</code> element.</p>
 				<p>Dialogs are most often used to prompt the user to enter or respond to information. A dialog that is designed to interrupt workflow is usually modal. See related <rref>alertdialog</rref>.</p>
-				<p>Authors MUST provide an accessible name for a dialog, which can be done with the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attribute.</p>
 				<p>Authors SHOULD ensure that all dialogs (both modal and non-modal) have at least one focusable descendant element. Authors SHOULD focus an element in the modal dialog when it is displayed, and authors SHOULD manage focus of modal dialogs.</p>
 				<p class="note">In the description of this role, the term "web application" does not refer to the <rref>application</rref> role, which specifies specific assistive technology behaviors.</p>
 			</div>
@@ -3073,7 +3090,12 @@
 					</tr>
 					<tr>
 						<th class="role-namefrom-head" scope="row">Name From:</th>
-						<td class="role-namefrom">author</td>
+						<td class="role-namefrom">
+ 							<ul>
+ 								<li>author</li>
+ 								<li>heading</li>
+ 							</ul>
+ 						</td>
 					</tr>
 					<tr>
 						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
@@ -6802,7 +6824,7 @@
 			<div class="role-description">
 				<p>A <rref>landmark</rref> containing content that is relevant to a specific, author-specified  purpose and sufficiently important that users will likely want to be able to navigate to the section easily and to have it listed in a summary of the page. Such a page summary could be generated dynamically by a user agent or assistive technology.</p>
 				<p>Authors SHOULD limit use of the region role to sections containing content with a purpose that is not accurately described by one of the other <a href="#landmark_roles">landmark roles</a>, such as <rref>main</rref>, <rref>complementary</rref>, or <rref>navigation</rref>.</p>
-				<p>Authors MUST give each element with role region a brief label that describes the purpose of the content in the region. Authors SHOULD reference a visible label with <pref>aria-labelledby</pref> if a visible label is present. Authors SHOULD include the label inside of a heading whenever possible. The heading MAY be an instance of the standard host language heading element or an instance of an element with role <rref>heading</rref>.</p>
+				<p>Authors MUST give each element with a region role a brief label that describes the purpose of the region. For host languages that support elements with the default role of region (e.g., HTML <code>section</code>), authors would need to specify an explicit <code>role="region"</code> on the element to enable the nameFrom: heading technique.</p>
 				<p><a>Assistive technologies</a> SHOULD enable users to quickly navigate to elements with role <code>region</code>.
   				[=User agents=] SHOULD treat elements with role <code>region</code> and have an accessible name as navigational <a>landmarks</a>.
   				[=User agents=] MAY enable users to quickly navigate to elements with role <code>region</code>.</p>
@@ -6858,7 +6880,12 @@
 					</tr>
 					<tr>
 						<th class="role-namefrom-head" scope="row">Name From:</th>
-						<td class="role-namefrom">author</td>
+						<td class="role-namefrom">
+ 							<ul>
+ 								<li>author</li>
+ 								<li>heading</li>
+ 							</ul>
+ 						</td>
 					</tr>
 					<tr>
 						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>


index.html
```diff
@@ -617,9 +617,11 @@ <h3>Accessible Name Calculation</h3>
...
+						<li>heading: name comes from the text value of the first descendant (i.e., depth first) <a>element</a> node with the role of <code>heading</code>. Although "heading" may be allowed in addition to "author" in some <a>roles</a>, "heading" is used in content only if higher priority "author" features are not provided.</li>
+						<li>contents: name comes from the text value of the <a>element</a> node. Although this may be allowed in addition to "author" in some <a>roles</a>, this is used in content only if higher priority "author" features are not provided.</li>
             <li>prohibited: the element does not support name from author. Authors MUST NOT use the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attributes to name the element.</li>
 					</ol>
+					<p>Priority is defined by the <a href="#mapping_additional_nd_te" class="accname">accessible name and description computation</a> algorithm. [[ACCNAME-1.3]]<p>
```
cookiecrook, line 624:
```suggestion
					<p>Priority is defined by the <a href="#mapping_additional_nd_te" class="accname">accessible name and description computation</a> algorithm. [[ACCNAME-1.2]]<p>
```

cookiecrook:
Adding three reviewers that were prior contributors to #1018.

scottaohara (commented):
just a quick list of the roles that will get name from heading 

- alertdialog
- article
- complementary
- dialog
- form
- region (with HTML caveat)

**My initial review:**

`complementary` may need a caveat added per the conditional complementary role for the `aside` element.

similarly, I think `form` would need a host language caveat mentioned as well, as while I think this update works since ARIA requires role=form have an accName, I don't think every HTML `form` element should automatically get an accName.  That'd have the potential to add extra and unnecessary landmark verbosity to many web pages

should we consider adding `group` to this as well?  Esp. since it's more likely for AT to even expose an element with `role=group` if it's named. If there's a heading there, it could be done for authors without the need to use aria-labelling.  But per my comment above, maybe this would have the inverse impact of creating too much verbosity where things were already "fine" from the user's perspective?


additionally, `navigation` should be considered for naming by heading.  many instances of people using headings (visible or not) to name navigation landmarks via `aria-labelledby`.


i'm quite a fan of all other aspects of this PR.

cookiecrook:
@scottaohara wrote:

Thanks.

> `complementary` may need a caveat added per the conditional complementary role for the `aside` element.

To clarify, are you suggesting it be removed from both `<aside>` and `role="complementary"` or only from the contexts where aside is not complementary?

> similarly, I think `form` would need a host language caveat mentioned as well, as while I think this update works since ARIA requires role=form have an accName, I don't think every HTML `form` element should automatically get an accName. That'd have the potential to add extra and unnecessary landmark verbosity to many web pages

I don't have a strong opinion either way, so I'll remove it in the PR unless there are objections. To clarify, are you suggesting it be removed from both `<form>` and `role="form"` or something else?

> should we consider adding `group` to this as well? Esp. since it's more likely for AT to even expose an element with `role=group` if it's named. If there's a heading there, it could be done for authors without the need to use aria-labelling. But per my comment above, maybe this would have the inverse impact of creating too much verbosity where things were already "fine" from the user's perspective?

I think group should remain explicitly named, for the verbosity risk you've outlined.... but I could be convinced if you and others have strong opinions for it.

> additionally, `navigation` should be considered for naming by heading. many instances of people using headings (visible or not) to name navigation landmarks via `aria-labelledby`.

I support this.

jnurthen (changes requested):
common/script/aria.js needs moving to aria-common but otherwise looks good.

I like the scope of the other changes being limited to 
* alertdialog
* article
* complementary
* dialog
* region (with caveats)
as it is now. We can always add more later if folks think that is desired.

scottaohara:
@cookiecrook 
Re: complementary and form, yes would need to be clear that while these ARIA roles can get name from heading, the equivalent HTML elements may not (aside) or would not (form) get name from headings.

re: group, agreed.  i felt icky suggesting it, and your response didn't wash the ick away.  so let's forget it..

cookiecrook:
> Re: complementary and form, yes would need to be clear that while these ARIA roles can get name from heading, the equivalent HTML elements may not (aside) or would not (form) get name from headings.

Maybe we should consider taking `complementary` and `region` out of the list because of these caveats? We can always add them back in with separate issues.

scottaohara:
sounds like a good idea to me.

index.html
```diff
@@ -6802,7 +6819,7 @@ <h5>Presentational Role Inheritance</h5>
...
 				<p>Authors SHOULD limit use of the region role to sections containing content with a purpose that is not accurately described by one of the other <a href="#landmark_roles">landmark roles</a>, such as <rref>main</rref>, <rref>complementary</rref>, or <rref>navigation</rref>.</p>
-				<p>Authors MUST give each element with role region a brief label that describes the purpose of the content in the region. Authors SHOULD reference a visible label with <pref>aria-labelledby</pref> if a visible label is present. Authors SHOULD include the label inside of a heading whenever possible. The heading MAY be an instance of the standard host language heading element or an instance of an element with role <rref>heading</rref>.</p>
+				<p>Authors MUST give each element with a region role a brief label that describes the purpose of the region. For host languages that support elements with the default role of region (e.g., HTML <code>section</code>), authors would need to specify an explicit <code>role="region"</code> on the element to enable the nameFrom: heading technique.</p>
```
scottaohara, line 6822: per https://github.com/w3c/aria/pull/1860#issuecomment-1432724292, this would need to be revised

scottaohara, line 6822: Maybe just remove the last sentence until ready to put it back in?

```suggestion
				<p>Authors MUST give each element with a region role a brief label that describes the purpose of the region.</p>
```

scottaohara (approved):

pkra:
DPUB-aria should probably also consider changes after this is complete. I filed w3c/dpub-aria#50 to track that part.

pkra:
From https://github.com/w3c/aria/blob/main/documentation/process.md

>  At least one implementation is required for merge. Implementation or implementation commitment from two browsers is required for merge.

Do we have that?

cookiecrook:
I believe we have informal commitment from at least two.  @aleventhal @benbeaudry @jcsteh do you have any issues with "name from heading"? 

For clarity, this isn't my proposal or original PR. I just volunteered to clean up the 4-year-old PR after Jon retired. So I'd like to get this merged as soon as possible.

benbeaudry:
@cookiecrook This change is a step in the right direction in my opinion. I don't have any issues with it. Merge it, and we'll build it in Chromium eventually!

pkra:
@spectranaut @jnurthen perhaps we should have labels for the new process requirements, e.g., something for each step in https://github.com/w3c/aria/blob/main/documentation/process.md#normative-changes). 

Then PRs / features like this could also easily be looked up.

index.html
```diff
@@ -662,9 +662,11 @@ <h3>Accessible Name Calculation</h3>
 				<dd>One of the following values:
 					<ol>
 						<li>author: name comes from values provided by the author in explicit markup features such as the <pref>aria-label</pref> attribute, the <pref>aria-labelledby</pref> attribute, or the host language labeling mechanism, such as the <code>alt</code> or <code>title</code> attributes in HTML, with HTML <code>title</code> attribute having the lowest precedence for specifying a text alternative.</li>
+						<li>heading: name comes from the text value of the first descendant (i.e., depth first) <a>element</a> node with the role of <code>heading</code>. Although "heading" is allowed in addition to "author" in some <a>roles</a>, "heading" is used in content only if higher priority "author" features are not provided.</li>
```
aleventhal, line 665: Just to clarify, does this mean:
- Take the name from the first descendant that's a heading?
- Take the name from the the first descendant if it's a heading?

Should we ignore anything in the depth-first search, such as hidden, generic or role="none"|"presentation" elements?

cookiecrook: The first matching... which you phrased above as "Take the name from the first descendant that's a heading?"

I thought that was clear from the existing text, but obvs open to word-smithing the PR if it improves the clarity.

spectranaut:
I added the "waiting for implementations" tag but I was wrong, this isn't ready for implementation -- I didn't look close enough. This change still needs the following before it can be implemented:
* three approving reviews
* accname changes
* tests

This new process is a work in progress, so let me know if there are questions/concerns!

cookiecrook:
@jcsteh wrote:
> Is it reasonable for UAs to impose some limit on how many nodes we check?

Certainly possible, but where to draw the line? Some web dev will hit an arbitrary depth limit. Do we tie it to DOM or rendering heuristics? My [original issue](https://github.com/w3c/accname/issues/138) was intended to allow some spec permissiveness so that implementations could account for this as best they could, which may eventually inform a best path forward... AS of now, some of those heuristics are explicitly disallowed by the spec.

What modification do you propose?

accdc (approved):

index.html
cookiecrook:
```suggestion
						<li>heading: name comes from the text value of the first descendant <a>element</a> node matching the role of <code>heading</code> in an <em>iterative deepening depth-first search.</em> Although "heading" is allowed in addition to "author" in some <a>roles</a>, "heading" is used in content only if higher priority "author" features are not provided.</li>
```

index.html
```diff
@@ -662,9 +662,11 @@ <h3>Accessible Name Calculation</h3>
 				<dd>One of the following values:
 					<ol>
 						<li>author: name comes from values provided by the author in explicit markup features such as the <pref>aria-label</pref> attribute, the <pref>aria-labelledby</pref> attribute, or the host language labeling mechanism, such as the <code>alt</code> or <code>title</code> attributes in HTML, with HTML <code>title</code> attribute having the lowest precedence for specifying a text alternative.</li>
+						<li>heading: name comes from the text value of the first descendant <a>element</a> node matching the role of <code>heading</code> in an <em>iterative deepening depth-first search.</em> Although "heading" is allowed in addition to "author" in some <a>roles</a>, "heading" is used in content only if higher priority "author" features are not provided.</li>
```
cookiecrook, line 665: Moving the more technical part of the definition to AccName.

```suggestion
						<li>heading: name comes from the text alternative of the first descendant <a>element</a> node matching the role of <code>heading</code> as defined in the AccName computation algorithm for <a href="https://w3c.github.io/accname/#comp_name_from_heading"><code>nameFrom: heading</code></a>. Although "heading" is allowed in addition to "author" in some <a>roles</a>, "heading" is used in content only if higher priority "author" features are not provided.</li>
```

index.html
aleventhal: Looks clear to me now.

aleventhal:
I'm no longer blocking this, so @jnurthen you're up next and then this can finally land.

aleventhal:
Ah, this is "waiting for implementations", which means we should land the changes and then this gets merged, sounds like.

index.html
```diff
@@ -662,9 +662,11 @@ <h3>Accessible Name Calculation</h3>
 				<dd>One of the following values:
 					<ol>
 						<li>author: name comes from values provided by the author in explicit markup features such as the <pref>aria-label</pref> attribute, the <pref>aria-labelledby</pref> attribute, or the host language labeling mechanism, such as the <code>alt</code> or <code>title</code> attributes in HTML, with HTML <code>title</code> attribute having the lowest precedence for specifying a text alternative.</li>
+						<li>heading: name comes from the text alternative of the first descendant <a>element</a> node matching the role of <code>heading</code> as defined in the AccName computation algorithm for <a href="https://w3c.github.io/accname/#comp_name_from_heading"><code>nameFrom: heading</code></a>. Although "heading" is allowed in addition to "author" in some <a>roles</a>, "heading" is used in content only if higher priority "author" features are not provided.</li>
```
cookiecrook, line 665:
```suggestion
						<li>heading: name comes from the text alternative of the first descendant <a>element</a> node (depth first search) matching the role of <code>heading</code> as defined in the AccName computation algorithm for <a href="https://w3c.github.io/accname/#comp_name_from_heading"><code>nameFrom: heading</code></a>. Although "heading" is allowed in addition to "author" in some <a>roles</a>, "heading" is used in content only if higher priority "author" features are not provided.</li>
```

index.html
```diff
@@ -662,9 +662,11 @@ <h3>Accessible Name Calculation</h3>
 				<dd>One of the following values:
 					<ol>
 						<li>author: name comes from values provided by the author in explicit markup features such as the <pref>aria-label</pref> attribute, the <pref>aria-labelledby</pref> attribute, or the host language labeling mechanism, such as the <code>alt</code> or <code>title</code> attributes in HTML, with HTML <code>title</code> attribute having the lowest precedence for specifying a text alternative.</li>
+						<li>heading: name comes from the text alternative of the first descendant <a>element</a> node (depth first search) matching the role of <code>heading</code> as defined in the AccName computation algorithm for <a href="https://w3c.github.io/accname/#comp_name_from_heading"><code>nameFrom: heading</code></a>. Although "heading" is allowed in addition to "author" in some <a>roles</a>, "heading" is used in content only if higher priority "author" features are not provided.</li>
```
cookiecrook, line 665: Note for other reviewers, the only recent change is "(depth first search)" to clarify the markup test the WG discussed since I posted the original PR.

cookiecrook:
@aleventhal wrote:
> Ah, this is "waiting for implementations", which means we should land the changes in Blink and not wait for it to be merged into the spec, right?

Sorry I missed this... I think the test was supposed to land first... but there was some discussion of DFS vs IDS which hadn't landed yet... See the last test in:
- https://github.com/web-platform-tests/wpt/pull/50507

cookiecrook:
Oh good... looks like that Chrome change wasn't merged into shipping yet. All browsers still align...

https://wpt.fyi/results/accname/name/comp_name_from_heading.tentative.html?label=pr_head&max-count=1&pr=50507

cookiecrook:
PR has been open for over 2 years... I'm no longer sure how many times I've re-resolved git conflicts that came in later.

cookiecrook:
@aleventhal wrote:

> I'm no longer blocking this

@aleventhal @tranjocelyn See [above comment on Feb 4](https://github.com/w3c/aria/pull/1860#issuecomment-2635621870)... If you've landed this already, there is more to be done.
https://wpt.fyi/results/accname/name/comp_name_from_heading.tentative.html?label=experimental&label=master&aligned

pkra:
@cookiecrook @aleventhal the current process requires 1 implementation and another intent-to-implement. So I think if webkit merges https://github.com/WebKit/WebKit/pull/43080, we should be able to merge here (and in accname).

If that's not in the cards, then I would suggest to split this PR into two: the piece which introduces name-from-heading and the piece that allows name-from-heading on specific roles. 

The first part is the lengthier one but wouldn't technically need implementations, so I think we could merge that. The second part is then just adding things to the name-from table rows.

(I would also split out the the editorial changes - they should be separate.)

cookiecrook:
@pkra wrote:
> we should be able to merge here (and update accname).

Agreed. I will wait for @rahimabdi to merge the WebKit update and then resolve the conflicts here. I don't see a reason to split the PRs into separate algo and attributes PRs. That's an academic distinction that I can't see providing any benefit.

> (I would also split out the the editorial changes - they should be separate.)

Are you talking about the tabs vs space correction that was on either side of my new nameFromHeading section? Normally I'd agree with not introducing whitespace changes in a substantive to different parts of the document, but these are related adjacent sections. That's always been fair-game with ARIA editorial practices as far as I am aware. Otherwise, you're asking me to introduce a new inconsistency to the leading whitespace within my new section. Right? (Please reconsider? 🙏🏼)

cookiecrook:
> Otherwise, you're asking me to introduce a new inconsistency to the leading whitespace within my new section. 

For clarity, my choices are: 
- keep it the way it is, which is what we ultimately want, or…
- Add my new section with leading chars that are inconsistent with the file conventions, or…
- Add my new section with leading chars that are inconsistent with the surrounding text.

pkra:
> > (I would also split out the the editorial changes - they should be separate.)
> 
> Are you talking about the tabs vs space correction that was on either side of my new nameFromHeading section? 

No, I meant the two removals. I primarily suggested them to simplify a merge. (FYI prettier formatting is done automatically by CI).

cookiecrook:
Working on this again... Merge conflicts plus the monorepo change that has happened in the meantime make it impractical to fix the existing PRs in the old repos. 

Working on another. Fourth times the charm. At least the implementations are no longer the bottleneck???

cookiecrook:
This issue superseded by:
- https://github.com/w3c/aria/pull/2650

PR #1862, opened 2023-01-26
Remove aria-expanded from listbox

### LINKED ISSUE #1026: Revisit allowed roles for aria-expanded
After #681, it looks like the current roles that support aria-expanded are:
- application
- button
- checkbox
- combobox
- gridcell
- link
- listbox
- menuitem
- row
- rowheader
- tab
- treeitem

Also inherited into:
- columnheader
- menuitemcheckbox
- menuitemradio
- switch

I was curious about the use cases for the following roles specifically:

**application**
Similar to `document`, I'm not sure when it would make sense to give the application root itself aria-expanded

**listbox**
I often see confusion around how to properly implement a custom `<select>`, and one of the common patterns is a `<div role="listbox" tabindex="0" aria-expanded="true/false">`. I haven't run into any indication that this is a supported pattern, and I've run into some support issues with it (though more investigation would be needed if it is desired). If that's not an accepted pattern, I think it'd make sense to not support listbox. Some examples where you can see the listbox used as `<select>` in the wild:
- https://www.telerik.com/kendo-angular-ui/components/dropdowns/
- https://developer.microsoft.com/en-us/fabric#/controls/web/dropdown

**gridcell**
I'm less sure about this one, but it seems like a better experience to have a button nested in a grid cell when it's desirable to trigger an expanded element. Is there a use case for having `aria-expanded` on a grid cell that isn't covered by a button?

**rowheader / columnheader**
Similar to gridcell, is there a use case for having `aria-expanded` directly on a row or column header rather than having a child button?

**row**
I know this is valid in a treegrid, but I've run into instances of it being added to a row in a table or grid (e.g. for a grid row that has some sort of supplemental content that visually appears below the row). Using `aria-expanded` on a row doesn't seem like the proper approach for those use case, so if treegrid > row is really the only allowed case, perhaps the documentation could be edited to make that more clear?

smhigley:
Closes #1026

TL:DR; I believe the history of this is that it was added when the expandable listbox was proposed as an alternative to the ARIA 1.1 combobox pattern. Now that we have the updated combobox pattern, there doesn't seem to be any benefit to keeping `aria-expanded` supported on listbox.

I wasn't sure if the best approach was to remove the attribute entirely (as in this PR), or:
- keep it in the table, but with "**(deprecated)**" after it, similar to the "(deprecated on this role in ARIA 1.2)" we have for former global attrs.
- add a note in the listbox section about it being deprecated in ARIA 1.X

* [x] Related APG Issue/PR: N/A (expandable listbox example already removed)

# Implementation tracking

* [ ] validator tests
* [ ] WPT tests: [LINK]()
   * WebKit: [ISSUE]()
   * Gecko: [ISSUE]()
   * Blink: [ISSUE]()

### DIFF
--- a/index.html
+++ b/index.html
@@ -4641,7 +4641,6 @@
 						<td class="role-properties">
 							<ul>
 								<li><pref>aria-errormessage</pref></li>
-								<li><sref>aria-expanded</sref></li>
 								<li><sref>aria-invalid</sref></li>
 								<li><pref>aria-multiselectable</pref></li>
 								<li><pref>aria-readonly</pref></li>


JAWS-test:
I think for consistency with other removed attributes, it would be useful to explicitly state "deprecated" for aria-expanded as well

cookiecrook (approved):

scottaohara (approved):

pkra (approved):

spectranaut:
Since this has approving reviews, and we talked about it in a meeting, I'm going to make issues on browsers and validators

spectranaut:
I'm trying to understand if this has browser implications, but I can't find in the spec anything that says user agents shouldn't expose aria attributes that are not list in the "supported states and properties" list. I wonder if I missing something...?

WilcoFiers (changes requested):
Instead of removing this, can it be listed as deprecated? There is nothing bad about people using this, even if there's no major benefit to it.

spectranaut:
Wilco's comment on the issue opened on axe: https://github.com/dequelabs/axe-core/issues/4433#issuecomment-2082333155

scottaohara:
it's not consistently exposed now, and per the linked Webkit issue it seems support will go down.

fwiw, i'm not opposed to marking it as deprecated for 1.3 - but maybe this PR should then be the start of addressing some of the conversation of #1990 - calling out why something was deprecated / is slated for future removal.

spectranaut:
Discussed in https://www.w3.org/2024/05/02-aria-minutes.html#t05

The decision was to create a deprecation section in the spec, and list this attribute there.

spectranaut:
Further discussion of the deprecation issue: https://www.w3.org/2025/01/09-aria-minutes#5464

spectranaut:
@smhigley there are implementations, this just needs the conflicts resolved -- unfortunate conflicts just with prettier. I think it would be easiest if you fixed the conflicts, then we can merge?

pkra:
Rebased on main.

PR #1871, opened 2023-02-09
Remove verbiage regarding combobox values

### LINKED ISSUE #1720: Add explicit value calculations for `textbox` and `searchbox` role

chlane:
[Closes #1720](https://github.com/w3c/aria/issues/1720)

Removed verbiage regarding combobox values which includes a MUST statement.
#https://github.com/w3c/aria/issues/1720

***
<a href="https://pr-preview.s3.amazonaws.com/chlane/aria/pull/1871.html" title="Last updated on Feb 9, 2023, 10:06 PM UTC (4c47628)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1871/f91a328...chlane:4c47628.html" title="Last updated on Feb 9, 2023, 10:06 PM UTC (4c47628)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2928,7 +2928,7 @@
 		<div class="role" id="deletion">
 			<rdef>deletion</rdef>
 			<div class="role-description">
-				<p>A deletion contains content that is marked as removed or content that is being suggested for removal. See related <rref>insertion</rref>.</p>
+				<p>A deletion represents content that is marked as removed, content that is being suggested for removal, or content that is no longer relevant in the context of its accompanying content. See related <rref>insertion</rref>.</p>
 				<p>Deletions are typically used to either mark differences between two versions of content or to designate content suggested for removal in scenarios where multiple people are revising content.</p>
 			</div>
 			<table class="role-features">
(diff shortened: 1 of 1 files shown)


spectranaut:
This is a follow up to closing this issue: https://github.com/w3c/aria/pull/1827

spectranaut (changes requested):
I marked this PR as "draft" because I think we shouldn't change the specification until we have something specific to point to.

Some context: When reviewing the PR #1827 to add non-input value calculations to listbox and searchbox, we realized the value calculation was more complicated than presented. Sarah suggested, since there is some discussion of value already in acc name, to move the value calculation to there. 

Here is the issue for that: https://github.com/w3c/accname/issues/184

I think we should put this PR on hold until the accname work is done, then we can point to accname in this PR.

mcking65:
Thank you @spectranaut, I just caught up on all the discussion and agree this should be on hold.

Even if there are some potential issues with the current language, it is working in practice in Chromium now. For select-only comboboxes to work, it is imperative that the browser calculate a value. The accname algorithm is the best we have at the moment.

In practice, it is not as important for ARIA to say anything about the value of textbox and searchbox. That is not to say that it doesn't matter, it is just not as pressing an issue in practice.

spectranaut:
We are discussing adding combobox values to accname, when we do that we will update this language.

PR #1873, opened 2023-02-16
fix(aria-level): update to specify only level 1-6

### LINKED ISSUE #1050: heading with aria-level > 6
I suggest that https://w3c.github.io/aria/#heading explicitly mentions that aria-level, unlike HTML headings, can be greater than 6.

ariellalgilmore:
Closes #1050 

Related APG Issue/PR: https://github.com/w3c/aria-practices/issues/2620

***
<a href="https://pr-preview.s3.amazonaws.com/ariellalgilmore/aria/pull/1873.html" title="Last updated on Mar 3, 2023, 7:50 AM UTC (9ac5880)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1873/90a154a...ariellalgilmore:9ac5880.html" title="Last updated on Mar 3, 2023, 7:50 AM UTC (9ac5880)">Diff</a>


JAWS-test:
It should be considered whether ARIA headings may also have a level greater than 6 and that AT that does not transmit this is faulty. So far, there are already some browser-screenreader combinations that correctly output a level greater than 6

index.html
```diff
@@ -12021,7 +12021,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 			<pdef>aria-level</pdef>
 			<div class="property-description">
 				<p><a>Defines</a> the hierarchical level of an <a>element</a> within a structure.</p>
-				<p>This can be applied inside trees to tree items, to headings inside a document, to nested grids, nested tablists and to other structural items that may appear inside a container or participate in an ownership hierarchy. The value for <pref>aria-level</pref> is an integer greater than or equal to 1.</p>
+				<p>This can be applied inside trees to tree items, to headings inside a document, to nested grids, nested tablists and to other structural items that may appear inside a container or participate in an ownership hierarchy. The value for <pref>aria-level</pref> is an integer greater than or equal to 1. If <pref>aria-level</pref> is being used on an element with role <pref>heading</pref> then authors MUST set the value to an integer between 1 and 6.</p>
```
pkra, line 12024: What would be the benefits of this change?

From the summary in https://github.com/w3c/aria/issues/1050#issuecomment-796481101 it seems that levels 1-9 on headings are widely supported in user agents, with only Chrome not supporting higher integer values. Similarly only JAWS appears to be failing with levels above 6 while Talkback struggles slightly with levels 7-9.

ariellalgilmore: my thoughts on this change were that as of right now all UA and AT support 1-6 and unless we suggest all UA and AT to support 1-9 then at the moment to ensure authors using `aria-level` with role heading are getting the expected experience across all UA and AT then they should use 1-6

pkra: Thanks for clarifying. That sounds like good authoring advice (like your APG issue https://github.com/w3c/aria-practices/issues/2620). From a spec perspective, it looks like pretty good support (especially with Chrome and JAWS looking like bugs), so I personally would rather file issues with user agents and AT instead of this change.

scottaohara: Tend to agree.  While I’m not itching to use heading level 8s anytime soon, I’m not sure why the spec should be so strict on that?

yes html only has up to level 6, but also, html doesn’t have buttons that natively convey a pressed state…

pkra: @ariellalgilmore I would wait until the WG has a chance to look at the PR next week.

spectranaut:
Discussed at today's meeting: https://www.w3.org/2023/02/23-aria-minutes#t02

index.html
```diff
@@ -12026,6 +12026,8 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
                 <p class="note"> In the case of a <rref>treegrid</rref>, <pref>aria-level</pref> is supported on elements with the role <rref>row</rref>, not elements with role <rref>gridcell</rref>. At first glance, this may seem inconsistent with the application of <pref>aria-level</pref> on <rref>treeitem</rref> elements, but it is consistent in that the <rref>row</rref> acts as the leaf node within the vertical orientation of the <rref>grid</rref>, whereas the <rref>gridcell</rref> is a leaf node within the horizontal orientation of each <rref>row</rref>. Level is not supported on sets of cells within rows, so the <pref>aria-level</pref> attribute is applied to the element with the role <rref>row</rref>.</p>
+								<p class="note"> At the time of this writing, most combinations of user agents and assistive technologies only support <pref>aria-level</pref> integers 1-9 when used on elements with role <pref>heading</pref>.</p>
```
pkra, line 12029: I would lead with role=heading to avoid confusion with other roles. I also think @MelSumner's comment on the call regarding usability in general should be included. Assuming, there is additional APG documentation coming, let's link to APG. 

```suggestion
<p class="note">On elements with role <pref>heading</pref>, values for <pref>aria-level</pref> above 6 can create difficulties for users. Also, at the time of this writing, most combinations of user agents and assistive technologies only support <pref>aria-level</pref> integers 1-9 on headings. See the [WAI-ARIA Authoring Practices](https://w3c.github.io/aria-practices/) for additional guidance on <pref>aria-level</pref>.</p>
```

ariellalgilmore: Thank you @pkra for the review and suggestions!

MelSumner (approved):

pkra (approved):

scottaohara (approved):

Issue #1256, opened 2020-05-01
Consider allowing `aria-hidden=false` on visibly rendered elements inside an ancestor with `aria-hidden=true`


cookiecrook:
In relation to `aria-hidden` error cases #1254 and #1255, @aleventhal suggested that we reconsider allowing `aria-hidden=false` on _visibly rendered_ elements inside an ancestor with `aria-hidden=true` 

The example of this is `aria-hidden=true` on a main section, but `aria-hidden=false` to "unhide" the descendant, such as a dialog or menu. In this case, the WG generally agreed that the author intention is pretty clear.

```html
<main aria-hidden="true">
  ... main content hidden during this state, and usually "grayed out" visually ...
  <div role="menu" aria-hidden="false">
      ... dialog or menu content "unhidden" ...
  </div>
</main>
```

It's worth noting that, years ago, @asurkov was opposed to having `aria-hidden="false"` expose _non-rendered_ elements to the accessibility tree, and I think in most cases that concern is still valid. This issue is proposing that visible, _rendered_ elements could use this pattern... somewhat similar to a non-inert descendant of an inert ancestor.

Opening for discussion. @aleventhal suggested both @asurkov and @jcsteh may have comments.

aleventhal:
If we do this, I think we wouldn't want to implement the proposal in issue 1254, because a valid use case would be to put aria-hidden=true on the html/body, and aria-hidden=false on a descendant.

jongund:
I think we should be limiting rather than expanding the scope of `aria-hidden`, we should look at limited the creativity of this problematic property.  

I have some questions:
1. Is there a place for this property in a website that is engineered for accessibility, or is it primary purpose to support some type of repair of an inaccessible website?
1. Can `aria-hidden` be more like the `hidden` attribute and have no values, if present it removes content from the accessibility tree?

The original goal Tim Burners-Lee had for the web was for people to write something once and have it viewable everywhere.   Maybe modern web development left this principle behind many years ago, but I have mostly heard in this discussion developers using this feature have to do a lot of testing to see if it will have the effect they want, assuming they even understand the effect they are looking for is supporting accessibility.  A feature that requires more testing is something we should be trying to limit, rather than expand the scope. 

Hiding/Showing Options for the web

|  Technology | Property  | Graphical Rendering  | Accessibility Tree  | 
|---|---|---|---|
|  HTML | `hidden attribute`  |  not rendered | not included  |  
|  CSS | `display: none`  |  not rendered | not included  |
[...]

PR #1880, opened 2023-03-03
prevent use of aria-hidden=true on document root elements

### LINKED ISSUE #1254: aria-hidden error case: disallow aria-hidden on root or document element (e.g. <html> or <body>)
aria-hidden error case: disallow aria-hidden on root or document element (e.g. `<html>` or `<body>`)

The WG discussed a variety of author error cases we should formalize to provide a better end-user experience for assistive technology users.

The first anti-pattern we see frequently is authors adding `aria-hidden="true"` to the `<html>` or `<body>` elements. The call attendees agreed this should be universally disallowed. Suggestion was to ignore `aria-hidden` completely on these elements (and the equivalent elements in other host languages, like `<svg>`).

As I understand it, source HTML including `<body aria-hidden="true">` in source HTML would be ignored by the UA, and not reflected in the DOM. Also the following would be no-op statements or throw warnings. These should potentially trigger console messages, too. Suggestion was to use `console.log()`, but I think `console.warn()` might be appropriate, too.

~~document.body.ariaHidden = true; // no-op [and warn] b/c on body element~~
~~document.body.setAttribute('aria-hidden', 'true'); // no-op [and warn] b/c on body element~~

[Update 17 June 2021: See discussion and related issues linked below. If we proceed, aria-hidden on body should not be ignored in all cases.. It should only be ignored if the hidden state results in no remaining accessible content in the page.]

[Update 22 March 2023: element reflection should not affected (difficult or impossible IDL) be so the previously proposed (now stricken) no-ops should map to the accessibility internals and APIs only... e.g. UAs MUST NOT expose…]

scottaohara:
closes #1254

adds user agent and author requirements regarding the use of aria-hidden=true on root-level elements of a document.

@cookiecrook I took a first stab at drafting up the author and UA must nots.  Please revise however you see fit.

* [ ] Related APG Issue/PR:
* [x] [ARIA in HTML 447](https://github.com/w3c/html-aria/pull/447)

# Implementation tracking

* [ ] [validator](https://github.com/validator/validator/issues/1559)
* [ ] WPT tests: [LINK]()
* [x] Browser implementations (link to issue or when done, link to commit):
   * [x] WebKit: [269523](https://bugs.webkit.org/show_bug.cgi?id=269523)
   * [ ] Gecko: [1880578](https://bugzilla.mozilla.org/show_bug.cgi?id=1880578)
   * [x] Blink: [325302202](https://issues.chromium.org/issues/325302202)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1880.html" title="Last updated on Jul 10, 2024, 8:20 AM UTC (f15f533)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1880/97f4517...f15f533.html" title="Last updated on Jul 10, 2024, 8:20 AM UTC (f15f533)">Diff</a>


index.html
```diff
@@ -11771,9 +11771,10 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p>Authors MAY, with caution, use aria-hidden to hide visibly rendered content from assistive technologies <em>only</em> if the act of hiding this content is intended to improve the experience for users of assistive technologies by removing redundant or extraneous content. Authors using aria-hidden to hide visible content from screen readers MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.</p>
+				<p>Authors MUST NOT use <code>aria-hidden</code> to hide root elements of a document. For instance, the <code>html</code> or <code>body</code> elements in an HTML document. Otherwise, authors MAY, with caution, use aria-hidden to hide visibly rendered content from assistive technologies <em>only</em> if the act of hiding this content is intended to improve the experience for users of assistive technologies by removing redundant or extraneous content. Authors using aria-hidden to hide visible content from screen readers MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.</p>
```
cookiecrook, line 11774: Root element is the right term for `<html>` but maybe not `<body>`. I don't recall what the term is but it maybe listed somewhere in an XML or HTML spec.

index.html
```diff
@@ -11771,9 +11771,10 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p class="note">At the time of this writing, <code><sref>aria-hidden</sref>="false"</code> is known to work inconsistently in browsers. As future implementations improve, use caution and test thoroughly before relying on this approach.</p>
+				<p>To prevent authors erroneously hiding entire window-rendered documents only to those using assistive technology, user agents MUST ignore the attribute if it is specified on document root elements. For instance, in HTML the <code>html</code> and <code>body</code> elements. Or the root <code>svg</code> element if it is rendered as its own document. If authors were to specify <code>aria-hidden="true"</code> on the opening tag for an embedded document, for instance on a <code>math</code> or <code>svg</code> embedded within an HTML document, user agents would still be expected to hide these elements from assistive technologies.</p>
```
cookiecrook, line 11777: Need to ask some HTML spec people if "rendered as its own document" is the right way to phrase that. I suspect there may be some other defined term.

In addition to `<math>` and `<svg>` should we consider whether to allow this on docs contained in an iframe? Is "embedded" the right term for those elements and containers? How does this apply to SVG in the context of an ``? Is the `src` reference "embedded" or "rendered as its own document"? Note: I think WebKit may still the only implementation that will try to pull some accessibility info (svg:title?) from the referenced SVG image, so there may be implementation-specific impact to whatever phrasing you land on.

cookiecrook (commented):
Great start. Raising some questions that I don't currently know the answer to. Thanks for taking this one on!

jcsteh (commented):
Thanks for working on this.

Does this conflict with #1256? That proposal has its own set of challenges, but as I understand it, it could be used to emulate dialog.showModal(). That is, an author might set aria-hidden=true on the root and then set aria-hidden=false on a (non-native) modal dialog.

I tend to think the risk of abuse with aria-hidden on the root is higher than the advantages of supporting the theoretical emulation of dialog.showModal(), but I know there's been some push to have parallel support in ARIA for native HTML features, which is why I'm raising it here.

index.html
scottaohara: a comment i made on the ARIA wg call yesterday concerning iframes, is that i didn't think that should factor into whether the html/body elements should respect aria-hidden=true.  And that's because even if the browser detected that _this_ body element should be allowed to be hidden because it's within an iframe, that could still mean that the parent iframe was not hidden, and then there'd be an iframe available but no content within it.   Rather, it'd be better if there was consistency between browsers to completely hide iframes, and then if it / it's rendered content needed to be hidden, then the iframe would be the element to hide, not its nested document's html / body element.

index.html
cookiecrook: The [HTML Spec for the body](https://html.spec.whatwg.org/multipage/sections.html#the-body-element) element includes:
> The [body](https://html.spec.whatwg.org/multipage/sections.html#the-body-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) the contents of the document.

So this may work:
```suggestion
				<p>Authors MUST NOT use <code>aria-hidden</code> to hide the root element or the host language element that <a href="https://html.spec.whatwg.org/multipage/dom.html#represents">represents</a> the contents of a document. For instance, the <code>html</code> or <code>body</code> elements in an HTML document. Authors MAY, with caution, use <code>aria-hidden</code> to hide visibly rendered content from assistive technologies <em>only</em> if the act of hiding this content is intended to improve the experience for users of assistive technologies by removing redundant or extraneous content. Authors using <code>aria-hidden</code> to hide visible content from screen readers MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.</p>
```

index.html
cookiecrook: Same change as above.
```suggestion
				<p>To prevent authors erroneously hiding entire window-rendered documents only to those using assistive technology, user agents MUST not expose the hidden state to assistive technologies if it is specified on the root element or the host language element that <a href="https://html.spec.whatwg.org/multipage/dom.html#represents">represents</a> the contents of the primary document in view. For instance, the <code>html</code> or <code>body</code> elements in an HTML document. Or the root <code>svg</code> element if it is rendered as its own document. If authors were to specify <code>aria-hidden="true"</code> on the opening tag for an embedded document, for instance on a <code>math</code> or <code>svg</code> embedded within an HTML document, user agents would still be expected to hide these elements from assistive technologies.</p>
```

scottaohara:
thanks a ton for those edits, @cookiecrook

index.html
```diff
@@ -11796,9 +11796,10 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p class="note">At the time of this writing, <code><sref>aria-hidden</sref>="false"</code> is known to work inconsistently in browsers. As future implementations improve, use caution and test thoroughly before relying on this approach.</p>
+				<p>To prevent authors erroneously hiding entire window-rendered documents only to those using assistive technology, user agents MUST not expose the hidden state to assistive technologies if it is specified on the root element or the host language element that <a href="https://html.spec.whatwg.org/multipage/dom.html#represents">represents</a> the contents of the primary document in view. For instance, the <code>html</code> or <code>body</code> elements in an HTML document, or the root <code>svg</code> element if it is rendered as its own primary document in the browser window. If authors were to specify <code>aria-hidden="true"</code> on the opening tag for an embedded document, for instance on a <code>math</code> or <code>svg</code> embedded within an HTML document, user agents would still be expected to hide these elements from assistive technologies.</p>
```
scottaohara, line 11802:
```suggestion
				<p>To prevent authors erroneously hiding entire window-rendered documents only to those using assistive technology, user agents MUST not expose the hidden state to assistive technologies if it is specified on the root element or the host language element that <a data-cite="html/dom.html#represents">represents</a> the contents of the primary document in view. For instance, the <code>html</code> or <code>body</code> elements in an HTML document, or the root <code>svg</code> element if it is rendered as its own primary document in the browser window. If authors were to specify <code>aria-hidden="true"</code> on the opening tag for an embedded document, for instance on a <code>math</code> or <code>svg</code> embedded within an HTML document, user agents would still be expected to hide these elements from assistive technologies.</p>
```

index.html
```diff
@@ -11796,9 +11796,10 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p class="note">Authors are advised to use extreme caution and consider a wide range of disabilities when hiding visibly rendered content from assistive technologies. For example, a sighted, dexterity-impaired individual might use voice-controlled assistive technologies to access a visual interface. If an author hides visible link text "Go to checkout" and exposes similar, yet non-identical link text "Check out now" to the accessibility API, the user might be unable to access the interface they perceive using voice control. Similar problems can also arise for screen reader users. For example, a sighted telephone support technician may attempt to have the blind screen reader user click the "Go to checkout" link, which they might be unable to find using a type-ahead item search ("Go to…").</p>
```
scottaohara, line 11800:
```suggestion
				<p class="note">Authors are advised to use extreme caution and consider a wide range of disabilities when hiding visibly rendered content from assistive technologies. For example, a sighted, dexterity-impaired individual might use voice-controlled assistive technologies to access a visual interface. If an author hides visible link text "Go to checkout" and exposes similar, yet non-identical link text "Check out now" to the accessibility API, the user might be unable to access the interface they perceive using voice control. Similar problems can also arise for screen reader users. For example, a sighted telephone support technician might attempt to have the blind screen reader user click the "Go to checkout" link, which they might be unable to find using a type-ahead item search ("Go to…").</p>
```

cookiecrook (approved):

aleventhal (approved):

aleventhal (approved):
Hopefully it is clear to readers what a primary document is. Do we need to define that in the glossary or is it really obvious to the intended audience?

cookiecrook (commented):
@scottaohara, in an earlier comment you said the embedded view (iframe, img, etc) could be hidden by the containing context, so on a per-document level, this rule need not be aware of its embeddedness, or whether it's the "primary" document.

> the root element or the host language element that <a data-cite="html/dom.html#represents">represents</a> the contents of the ~~primary~~ document in view. 

I agree with you, so it's possible this ~"root element or the host language element representing contents" is sufficient without the undefined "primary document" term.

aleventhal:
What if the parent document doesn't have as much context as the iframe's contents about whether it should be hidden or not? They might not even be from the same domain. Is it difficult for any of the implementations to determine whether the document is top level?

aleventhal (commented):
Removing my +1 until we get consensus on whether the top level document's documentElement/body is treated differently from the iframe's.

cookiecrook:
> Removing my +1 until we get consensus on whether the top level document's documentElement/body is treated differently from the iframe's.

If that's a deal-breaker for Aaron, I'd say let's keep "primary." It does allow more permissive authoring flexibility, albeit adding a bit more implementation complexity.

aleventhal:
I think it just adds a single condition to the 'if' statement. Not too complex IMO. I'd rather not deal with the bug that gets filed with a legit use case. But perhaps Scott has a good reason to disallow in iframe html/body elements.

scottaohara:
Allowing an html/body element to have aria-hidden=true on it in an iframe wouldn’t do anything to prevent the exposure of the iframe in the parent doc

However is the iframe in the parent doc could be consistently set to be hidden / taken out of the tab order, then that could prevent the nested iframe content from being accessible 

I fail to understand why there should be any distinction whether aria-hidden is set to the html/body elements within an iframe or not, since it’d be unnecessary if the iframe in the parent doc was properly hidden, and if the iframe was not properly hidden, then respecting the aria-hidden attribute in that context would result in the problem we are trying to solve with this change

spectranaut:
Adding agenda to "get consensus" re: Aaron's concerns so we can move this one forward!

spectranaut:
discussed in https://www.w3.org/2024/05/09-aria-minutes.html#t10

Decision to move forward

index.html
```diff
@@ -11605,10 +11605,11 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p class="note">Authors are advised to use extreme caution and consider a wide range of disabilities when hiding visibly rendered content from assistive technologies. For example, a sighted, dexterity-impaired individual might use voice-controlled assistive technologies to access a visual interface. If an author hides visible link text "Go to checkout" and exposes similar, yet non-identical link text "Check out now" to the accessibility API, the user might be unable to access the interface they perceive using voice control. Similar problems can also arise for screen reader users. For example, a sighted telephone support technician might attempt to have the blind screen reader user click the "Go to checkout" link, which they might be unable to find using a type-ahead item search ("Go to…").</p>
```
scottaohara, line 11609:
```suggestion
				<p class="note">Authors are advised to use extreme caution and consider a wide range of disabilities when hiding visibly rendered content from assistive technologies. For example, a sighted, dexterity-impaired individual might use voice-controlled assistive technologies to access a visual interface. If an author hides visible link text "Go to checkout" and exposes similar, yet non-identical link text "Check out now" to the accessibility API, the user might be unable to access the interface they perceive using voice control software. Similar problems can also arise for screen reader users. For example, a sighted telephone support technician might attempt to have the blind screen reader user click the "Go to checkout" link, which they might be unable to find using a type-ahead item search ("Go to…"), since that text would have been hidden by the use of the attribute.</p>
```

index.html
```diff
@@ -11605,10 +11605,11 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p class="note">The original intent for <code>aria-hidden="false"</code> was to allow user agents to expose content that was otherwise hidden from the accessibility tree. However, due to ambiguity in the specification and inconsistent browser support for the <code>false</code> value, the original intent is no longer supported.</p>
+        <p>To prevent authors erroneously hiding entire window-rendered documents only to those using assistive technology, user agents MUST NOT expose the hidden state to assistive technologies if it is specified on the root element or the host language element that <a data-cite="html/dom.html#represents">represents</a> the contents of the primary document in view. For instance, the <code>html</code> or <code>body</code> elements in an HTML document, or the root <code>svg</code> element if it is rendered as its own primary document in the browser window. If authors were to specify <code>aria-hidden="true"</code> on the opening tag for an embedded document, for instance on a <code>math</code> or <code>svg</code> embedded within an HTML document, user agents would still be expected to hide these elements from assistive technologies.</p>
```
scottaohara, line 11612: maybe if i futz with the white space, the build process will magically work this time.
```suggestion
				<p>To prevent authors erroneously hiding entire window-rendered documents only to those using assistive technology, user agents MUST NOT expose the hidden state to assistive technologies if it is specified on the root element or the host language element that <a data-cite="html/dom.html#represents">represents</a> the contents of the primary document in view. For instance, the <code>html</code> or <code>body</code> elements in an HTML document, or the root <code>svg</code> element if it is rendered as its own primary document in the browser window. If authors were to specify <code>aria-hidden="true"</code> on the opening tag for an embedded document, for instance on a <code>math</code> or <code>svg</code> embedded within an HTML document, user agents would still be expected to hide these elements from assistive technologies.</p>
```

aleventhal (approved):
Excellent. I can't find anything that would be confusing, vague or isn't a clear improvement.

jcsteh (approved):

aleventhal:
In terms of "waiting for implementations". Chrome has implemented this.

pkra:
@scottaohara the WebKit issue also seems to have been resolved, so this might be ready to merge.

scottaohara:
related HTML AAM PR has been merged.

scottaohara:
@jnurthen @spectranaut i think this has been ready to merge for awhile?  at least chromium/webkit have implemented per comments here

PR #1886, opened 2023-03-10
Editorial: move wiki documents to document folder

### LINKED ISSUE #1882: wiki cleanup
As discussed at the editors meeting, the wiki is not in great shape. Most articles appear out of date. Very few have seen a significant update in the past year.

We should decide which pages we want to maintain and which alternative means we have to maintain them. We should also decide how to archive what we don't want to maintain.

spectranaut:
Partially resolves #1882

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ This repository is for the main deliverable of the ARIA Working Group, Accessibl
 
 ## Contributing to this Repository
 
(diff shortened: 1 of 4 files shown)


spectranaut:
When you approve this I'll delete the wiki pages, @pkra !

pkra (approved):

PR #1894, opened 2023-03-23
aria-atomic has no default

annevk:
Its computed value depends on its ancestors.

# PR tracking

* [ ] Related APG Issue/PR:

# Implementation tracking

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:

***
<a href="https://pr-preview.s3.amazonaws.com/annevk/aria/pull/1894.html" title="Last updated on Mar 23, 2023, 10:49 AM UTC (b9e3131)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1894/9033f63...annevk:b9e3131.html" title="Last updated on Mar 23, 2023, 10:49 AM UTC (b9e3131)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5935,7 +5935,7 @@
 					  &lt;p>If you find any variations in results, please let us know!&lt;/p>
 					&lt;/div>
 					&lt;p>...&lt;/p>
-				</pre>		
+				</pre>
 				<p>In cases where an element with role <rref>note</rref> has been determined to need a programmatic association with the content it supplements, authors can use one of the following mechanisms to associate the elements:</p>
 				<ul>
 					<li>If the <code>note</code> contains structured or interactive content (for example, a link, button, list, table, etc.) use <pref>aria-details</pref>.</li>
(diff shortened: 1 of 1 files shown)


annevk:
I had a discussion with @cookiecrook about cases like this and what I think we want to do is distinguish between the DOM attribute and the AT property. Also to make it clear what computations are owned by the AT.

The DOM attribute is "true" or "false", invalid, or missing.

The AT property depends on the DOM attribute, the computed role, ancestor elements, etc.

I will take a stab at demonstrating what that would mean for the specification tomorrow by creating a draft PR for a single property.

spectranaut:
Just to be clear, this proposed change is editorial, correct? This is related to this re-thinking of the "default value" in https://github.com/w3c/aria/pull/1896 but we aren't changing anything normative about the spec, or any behavior of the browser?

cookiecrook (approved):

annevk:
I'm not sure it's editorial (it's not a typo fix and arguably fixes an error), but it shouldn't have normative impact.

> with this change, should we also not change or removing the word "default" in the following sentence:

Reworking that would be a larger change we'd make as part of what #1896 illustrates.

All this is making clear is that it doesn't have an intrinsic default. You need to read the prose to determine the default.

cookiecrook:
@jnurthen @spectranaut Can this be merged? As @annevk said, it's not technically "editorial", but it's clearly "errata" IMO.

jnurthen (approved):

Issue #1895, opened 2023-03-23
Reorganize "implicit value for role" in the spec to clarify differences from IDL defaults


cookiecrook:
@annevk and I were discussing the IDL fix-ups and realized the spec will need to clarify distinctions between:
- the IDL default (including missing/invalid states), which cannot change per role, and 
- the accessibility implementation defaults, which can change per role, but won't reflect in the IDL, as they are an implementation detail of the browser accessibility internals. These are the ones listed in the ARIA spec as "Implicit Value for Role."

Given `aria-orientation` as an example, the attribute characteristics table does not make any mention of these Implicit Values. There is an informative note that mentions the differences for slider (horizontal) and scrollbar (vertical), but there is no single listing anywhere in the spec. There's also an orientation default on listbox, but you have to know that or do an awkward substring search for "Default for aria-orientation is"

We will need to add the default/missing/invalid IDL values to the _attribute's_ characteristics table, so having some seemingly conflict accessibility values in the _role's_ characteristics table may lead to confusion.

I propose all the normative accessibility defaults in the spec be moved to the _attribute's_ characteristics table. So aria-orientation will list its values, including the IDL default/missing/invalid values, as well as the accessibility implementation defaults, which will not be reflected in the IDL. Some explanatory text of this distinction will need to be added in one or more places... 

The roles with non-default values should also have informative notes cross-referencing these normative defaults defined in the attribute. Ideally we could do this in a ReSpec function to reduce duplication.

PR #1896, opened 2023-03-24
Illustrate how attributes and properties could be distinct concepts

annevk:
This could be a potential solution to #1895. The idea here would be to define all the ARIA attributes in terms of DOM/HTML types. This example shows what aria-orientation would look like defined as an enumerated attribute.

And then separately define how identically-named ARIA properties end up using those attributes to compute their own state for use by AT.

I think applying this throughout would resolve a lot of ambiguity and would help end up solving the reflect problem as well.

Closes #????

Describe Change Here!

# PR tracking

* [ ] Related APG Issue/PR:

# Implementation tracking

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:

### DIFF
--- a/index.html
+++ b/index.html
@@ -5935,7 +5935,7 @@
 					  &lt;p>If you find any variations in results, please let us know!&lt;/p>
 					&lt;/div>
 					&lt;p>...&lt;/p>
-				</pre>		
+				</pre>
 				<p>In cases where an element with role <rref>note</rref> has been determined to need a programmatic association with the content it supplements, authors can use one of the following mechanisms to associate the elements:</p>
 				<ul>
 					<li>If the <code>note</code> contains structured or interactive content (for example, a link, button, list, table, etc.) use <pref>aria-details</pref>.</li>
@@ -5947,7 +5947,7 @@
 					&lt;button aria-details="info-note">Get Started&lt;/button>
 					...
 					&lt;div role="note" id="info-note">
-					  &lt;p>Need more information before you get started?&lt;/p> 
+					  &lt;p>Need more information before you get started?&lt;/p>
 					  &lt;p>Visit our &lt;a href="...">product description page&lt;/a> to get all the information you need.&lt;/p>
 					&lt;/div>
 				</pre>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -12310,8 +12310,51 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+							<th class="value-name" scope="row"><code>vertical</code></th>
+							<td>vertical</td>
+							<td class="value-description">The element is oriented vertically.</td>
+						</tr>
+					</tbody>
+				</table>
+
+			    <p>The <pref>aria-orientation</pref> attribute's <a>invalid value default</a> and <a>missing value default</a> are the undefined state.
+
+				<p class="note">In ARIA 1.1, the default value for <pref>aria-orientation</pref> changed from <code>horizontal</code> to <code>undefined</code>.</p>
```
cookiecrook, line 12345: These statements are seemingly in conflict unless we clarify the two different value concepts like "IDL value" and "accessibility internal value" (not particular to this name) ....

index.html
```diff
@@ -12340,29 +12383,6 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 					</tr>
 				</tbody>
 			</table>
-			<table class="value-descriptions">
-				<caption>Values:</caption>
```
cookiecrook, line 12344: Curious how the other editors feel about removing the values table.

cookiecrook (commented):
Thanks for the start on this @annevk. I'd like to hear @spectranaut and @jnurthen's comments before progressing much further.

index.html
annevk: Right, I was thinking we'd have a general distinction between "attribute" and "property" where the property is what AT computes the value of based on the "attribute". And I guess this note was about the "property" so should be moved a bit and clarified.

index.html
spectranaut: looks like this table was not removed, just moved up and expanded?

index.html
spectranaut: I'm having trouble wrapping my head around this a bit, even though I do understand the problem you are trying to solve, IDL value/defaults vs "accessibility internal value"/defaults.

maybe we should use "implicit value" for the "Accessibility internal values"/defaults, as that is already what is in the role tables?

Maybe we should introduce some terms definitions in the PR so that it will be easier to understand where we are moving?

index.html
```diff
@@ -12310,8 +12310,51 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+							<td class="value-description">The element is oriented vertically.</td>
+						</tr>
+					</tbody>
+				</table>
+
+			    <p>The <pref>aria-orientation</pref> attribute's <a>invalid value default</a> and <a>missing value default</a> are the undefined state.
+
+				<p class="note">In ARIA 1.1, the default value for <pref>aria-orientation</pref> changed from <code>horizontal</code> to <code>undefined</code>.</p>
+
+				<p>The state of the <pref>aria-orientation</pref> property of an element <var>element</var> is computed as follows:
```
spectranaut, line 12347: Now that you added the algorithm the "explicit value" for aria-orientation is in two places, the algorithm and the characteristic/value table for the relevant roles (see https://w3c.github.io/aria/#listbox). Maybe it doesn't need to be in two places? Or maybe for spec understandablity, it should be?

spectranaut (commented):
This is definitely editorial, correct?

I really like this direction, though! We just really need to lock down the terms we are using here.

index.html
annevk: I was using that to encompass things that the browser specifically does for AT when AT is enabled, which is somewhat separate implementation-wise and also in terms of what we're comfortable exposing. Whereas reflecting is more general and will also function if AT isn't enabled.

index.html
annevk: You mean "Implicit Value for Role"? Yeah, I think we could remove that if we go down this path.

index.html
annevk: I suspect that for some attributes we might end up removing it if they are a simpler type. We only need a table for enumerated attributes when defining them in terms of HTML attribute types.

index.html
cookiecrook: Yes, what we'd be talking about is not the default value of the DOM property, where el.ariaOrientation would always return `undefined`, but something like a new webdriver getter like `computedAriaOrientation`, or `computedAccessibleNode.orientation` where the default value on a range could be `horizontal`

pkra:
As per editors' call today, asking @spectranaut to make a suggestion on moving this forward.

PR #1906, opened 2023-04-01
Editorial: Normalize HTML spec references using xref

### LINKED ISSUE #1903: Cleanup links to HTML spec (maybe others?)
Follow on to https://github.com/w3c/aria/pull/1900.  

Quoting James's last comment, we should clean up the other references to the HTML spec (and might as well clean up links to other specs if necessary):
>..use either xref (if possible) or data-cite (if no reasonable xref exists) would be great!

adampage:
Closes #1903

A little editorial cleanup to normalize the way we reference HTML spec definitions for elements and attributes.  I’ll leave a few notes in context below.

***
<a href="https://pr-preview.s3.amazonaws.com/adampage/aria/pull/1906.html" title="Last updated on Apr 17, 2023, 9:15 PM UTC (4f8ea7e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1906/6207a61...adampage:4f8ea7e.html" title="Last updated on Apr 17, 2023, 9:15 PM UTC (4f8ea7e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3626,7 +3626,10 @@
 					</tr>
 					<tr>
 						<th class="role-related-head" scope="row">Related Concepts:</th>
-						<td class="role-related"><a href="https://www.w3.org/TR/html/grouping-content.html#the-div-element">HTML <code>div</code></a>, <a href="https://www.w3.org/TR/html/textlevel-semantics.html#the-span-element">HTML <code>span</code></a></td>
+						<td class="role-related">
+							HTML <code>[^div^]</code>,
+							HTML <code>[^span^]</code>
+						</td>
 					</tr>
 					<tr>
 						<th class="role-scope-head" scope="row">Required Context Role:</th>
--- a/index.html
+++ b/index.html
@@ -836,7 +836,7 @@
 			</ul>
 		</section>
 	</section>
-	<section id="role_definitions" data-cite="dom">
+	<section id="role_definitions" data-cite="dom html">
 		<h2>Definition of Roles</h2>
 		<p>Below is an alphabetical list of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> <a>roles</a> to be used by authors.</p>
 		<p>Abstract roles are used for the ontology. Authors MUST NOT use abstract roles in content.</p>
@@ -10181,7 +10181,7 @@
 		</div>
 	</section>
 </section>
-<section class="normative" id="states_and_properties" data-cite="dom">
+<section class="normative" id="states_and_properties" data-cite="html dom">
 	<h1>Supported States and Properties</h1>
 	<section id="statevsprop">
 		<h2>Clarification of States versus Properties</h2>
--- a/index.html
+++ b/index.html
@@ -10181,7 +10181,7 @@
 		</div>
 	</section>
 </section>
-<section class="normative" id="states_and_properties" data-cite="html dom">
+<section class="normative" id="states_and_properties" data-cite="dom html">
 	<h1>Supported States and Properties</h1>
 	<section id="statevsprop">
 		<h2>Clarification of States versus Properties</h2>
--- a/index.html
+++ b/index.html
@@ -157,7 +157,7 @@
 <section id="sotd">
 	<p>The Accessible Rich Internet Applications Working Group seeks feedback on any aspect of the specification. When submitting feedback, please consider issues in the context of the companion documents. To comment, <a href="https://github.com/w3c/aria/issues/new">file an issue in the <abbr title="World Wide Web Consortium">W3C</abbr> ARIA GitHub repository</a>. If this is not feasible, send email to <a href="mailto:public-aria@w3.org?subject=Comment%20on%20WAI-ARIA%201.2">public-aria@w3.org</a> (<a href="http://lists.w3.org/Archives/Public/public-aria/">comment archive</a>). In-progress updates to the document can be viewed in the <a href="http://w3c.github.io/aria/">publicly visible editors' draft</a>.</p>
 </section>
-<section class="informative" id="introduction" data-cite="dom">
+<section class="informative" id="introduction">
 	<h1>Introduction</h1>
 	<p>The goals of this specification include:</p>
 	<ul>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1232,9 +1232,7 @@ <h2>Definition of Roles</h2>
           </tr>
           <tr>
             <th class="role-base-head" scope="row">Base Concept:</th>
-            <td class="role-base">
-                <a href="https://www.w3.org/TR/html5/grouping-content.html#the-dl-element"><abbr title="Hypertext Markup Language">HTML</abbr> <code>dl</code></a>
-            </td>
+            <td class="role-base"><code>&lt;[^dl^]&gt;</code> in HTML</td>
```
adampage, line 1235: I settled on a syntax convention of “`<element-name> in HTML`”, where just the element name is linked using its xref term.  I kept (or added, where needed) the `&lt;` and `&gt;` brackets to aid in quick understandability/differentiation when referencing an **HTML element** versus a plain text role string or attribute name.

index.html
```diff
@@ -2360,9 +2358,7 @@ <h2>Definition of Roles</h2>
 					</tr>
 					<tr>
 						<th class="role-related-head" scope="row">Related Concepts:</th>
-						<td class="role-related">
-							<code>&lt;select&gt;</code> in [[HTML]]
-						</td>
+						<td class="role-related"><code>&lt;[^select^]&gt;</code> in HTML</td>
```
adampage, line 2361: ReSpec is failing to resolve `[^select^]` here, even though [xref claims to translate it](https://respec.org/xref/?term=select&types=element). @jnurthen, do you know whether I might be doing something wrong here?  Maybe I need to update the xref configuration on [line 146](https://github.com/w3c/aria/blob/06b266f0570c6319c1b5ea4dee316940062ea5ce/index.html#L146)?  If not, do you think it’d be appropriate for me to file a bug at [w3c/respec](https://github.com/w3c/respec/issues)?

The same is true for:

- [`[^input/type/range^]`](https://respec.org/xref/?term=range&types=attr-value)
- [`[^input/value^]`](https://respec.org/xref/?term=value&types=element-attr)

Even though ReSpec is reporting errors, I’ve gone ahead and committed my changes for review for now — I’ll be sure to back these out if there’s no quick resolution.

index.html
```diff
@@ -12702,7 +12703,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				<tbody>
 					<tr>
 						<th class="property-related-head" scope="row">Related Concepts:</th>
-						<td class="property-related"><code>required</code> attribute in [[HTML]]</td>
+						<td class="property-related"><code>[^input/required^]</code> attribute in HTML</td>
```
adampage, line 12706: We’ve got a few references to form-related attributes like `required` that could be misinterpreted as _global_ attributes when they actually have multiple distinct HTML spec definitions, like `[^input/required^]` vs. `[^textarea/required^]` vs. `[^select/required^]`.  For now, I just linked to the `input`-scoped definition and didn’t add any additional language to disambiguate.  What do you think?  Here are the applicable attributes I encountered:

- `placeholder`
- `readonly`
- `required`

index.html
```diff
@@ -13435,7 +13436,7 @@ <h3>State and Property Attributes</h3>
 	</section>
 	<section id="host_general_focus">
 		<h2>Focus Navigation</h2>
-		<p>An implementing host language MUST provide support for the author to make all interactive elements focusable, that is, any renderable or event-receiving elements. An implementing host language MUST provide a facility to allow web authors to define whether these focusable, interactive elements appear in the default tab navigation order. The <code>tabindex</code> <a>attribute</a> in <abbr title="Hypertext Markup Language">HTML</abbr> is an example of one implementation.</p>
+		<p>An implementing host language MUST provide support for the author to make all interactive elements focusable, that is, any renderable or event-receiving elements. An implementing host language MUST provide a facility to allow web authors to define whether these focusable, interactive elements appear in the default tab navigation order. The <code>tabindex</code> <a>attribute</a> in HTML is an example of one implementation.</p>
```
adampage, line 13439: <abbr title="Today I learned">TIL</abbr> that when ReSpec encounters a unique `<abbr>` pattern in the document, it [automatically generates identical `<abbr>` markup for all subsequent occurrences](https://respec.org/docs/#abbreviations:~:text=ReSpec%20will%20then%20wrap%20all%20matching%20instances%20abbreviations%20with%20%3Cabbr%3E.) of the plain text abbreviation string. 🤩 

Because I encountered a few explicit abbreviation markups for “HTML” during this PR, I found and left the first occurrence of “`<abbr title="Hypertext Markup Language">HTML</abbr>`” [on line 475](https://github.com/w3c/aria/blob/06b266f0570c6319c1b5ea4dee316940062ea5ce/index.html#L475), then simplified all later occurrences to just plain “HTML” to reduce noise in the document.

<abbr title="For what it’s worth">FWIW</abbr>, it looks like we’ve got a ton more abbreviations we could simplify if that were desirable?  E.g., “WAI-ARIA”, “DOM”, “API”.  If anyone think that’s a good idea, please say the word — I’d be happy to do a separate PR for that.

*[TIL]: Today I learned
*[FWIW]: For what it’s worth

index.html
jnurthen: I worked out that this was due to the local data-cite on the sections. This is an override of the document level data-cite. For now I've added html to that but we might want to look at why we have this in the first place (it is to disambiguate the definition of attribute from [dom] and [webidl]. The fact that both these core specs define the same term seems problematic. They should probably use different reference types rather than every spec author having to disambiguate.)

adampage: I see, thanks for sleuthing that out and explaining.  Where does the document-level configuration for `webidl` live?  I did a little experimentation after your commits to see if I could at least reduce the total number of overrides by favoring `dom html` by default (e.g., [on `<body>`](https://github.com/w3c/respec/wiki/Auto-linking-external-references#:~:text=It%20is%20recommended%20to%20add%20data%2Dcite%20attribute%20on%20the%20%3Cbody%3E.)) and then adding in a lesser number of overrides for `webidl`, but I don’t see where `webidl` is being specified to begin with?

jnurthen: I had to chase how this works for webidl and that involved looking at the respec code....

The xref reference in the respecConfig var at the top of the document is where this is normally done BUT webidl doesn't appear there.

It seems that when respec comes across an IDL reference (I didn't look to see what triggers this but guessed it was class="idl") then it traverses back to the closest data-cite and adds webidl to the references. If there is no data-cite then webidl gets added to the document level. 

So - we can resolve this by doing the following:
* Add data-cite="webidl" to the 2 sections of the document that have class="idl"
* fix the 4 other places in the document which are using webidl for references by adding data-cite="webidl" to those places.
We could then remove all instances of data-cite="html" or data-cite="dom" in the document.

I don't know whether this is more or less ugly than the current status.

adampage: Mmm, thanks, that’s exactly what I had in mind.  I’ll give it a shot and compare its ugliness.  😅

adampage: In 7d9d11d, I traded 10 `html/dom` overrides for 3 `webidl` ones.  Slightly less ugly?

adampage:
@spectranaut and @jnurthen, what do you think of [this spot in the readme](https://github.com/w3c/aria/pull/1906/commits/4f8ea7ed05c08af8eaa3171a50aaf6f70f952db0?short_path=b335630#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) to add some `xref` authoring guidance?

jnurthen (approved):

pkra:
@adampage I've opened https://github.com/w3c/aria/issues/1918 for your suggestion

pkra:
Oh sorry -- I misread the comment. You already added what you wanted here, correct?

adampage:
Thanks, @pkra — yep, I’ve already added [an update to the readme in this PR](https://github.com/w3c/aria/pull/1906/commits/4f8ea7ed05c08af8eaa3171a50aaf6f70f952db0).

PR #1913, opened 2023-04-11
RFC-2119 github workflow

### LINKED ISSUE #1902: add github action to prevent errant RFC-2119 terms sneaking back in
# Describe the change

We should prevent lower case RFC-2119 normative words. Any uses of RFC-2119 MUST be uppercase in the ARIA spec.

Link to documentation:

ariellalgilmore:
Closes #1902 

worked with @andreancardona!

Adds a new github workflow for new PR's that checks if the PR is introducing any lower case versions of RFC-2119 terminology

### DIFF
--- a/.github/workflows/RFC-2119-usage.yml
+++ b/.github/workflows/RFC-2119-usage.yml
@@ -13,4 +13,4 @@ jobs:
       uses: JJ/github-pr-contains-action@releases/v10
       with:
         github-token: ${{github.token}}
-        diffDoesNotContain: " must | must not | required | shall | shall not | should | should not | recommended | may | optional "
+        diffDoesNotContain: " must | must not | shall | shall not | should | should not | may"
--- a/.github/workflows/RFC-2119-usage.yml
+++ b/.github/workflows/RFC-2119-usage.yml
@@ -14,3 +14,4 @@ jobs:
       with:
         github-token: ${{github.token}}
         diffDoesNotContain: " must | must not | should | should not | may "
+        
--- a/.github/workflows/RFC-2119-usage.yml
+++ b/.github/workflows/RFC-2119-usage.yml
@@ -14,4 +14,3 @@ jobs:
       with:
         github-token: ${{github.token}}
         diffDoesNotContain: " must | must not | should | should not | may "
-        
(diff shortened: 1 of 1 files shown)


pkra:
This looks really helpful! Thank you @ariellalgilmore & @andreancardona!

@daniel-montalvo is there a W3C policy on third party github actions?

daniel-montalvo (approved):

.github/workflows/RFC-2119-usage.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  check_pr:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check PR
+      uses: JJ/github-pr-contains-action@releases/v10
+      with:
+        github-token: ${{github.token}}
+        diffDoesNotContain: " must | must not | required | shall | shall not | should | should not | recommended | may | optional "
```
spectranaut, line 16: remove "required", "optional", "recommended" and make follow up issue to discuss removing these from the spec.

andreancardona, line 16:
```suggestion
        diffDoesNotContain: " must | must not | shall | shall not | should | should not | may"
```

andreancardona: ✅  here is the issue to track this: https://github.com/w3c/aria/issues/1916

.github/workflows/RFC-2119-usage.yml
```diff
@@ -0,0 +1,16 @@
+name: "Check if PR introduces errant RFC-2119"
+on: 
+  pull_request:
+      types: [opened, edited, assigned, closed, review_requested, ready_for_review]
```
spectranaut, line 4: I'm a little confused about this list, I'm not super familiar with github workflows, but does this check need to be run in all these scenarios? shouldn't it only run if the PR is updated/edited? Or am I'm miss understanding this list..

spectranaut (changes requested):
This is exciting! but I have one question.

.github/workflows/RFC-2119-usage.yml
ariellalgilmore: agree! i updated it just to include `opened/edited`

spectranaut (approved):
looks great!

pkra (approved):

PR #1931, opened 2023-05-09
add proposed header and footer roles

### LINKED ISSUE #1915: Consider adding a 'header' and 'footer' role
A similar topic came up when we had been discussing `hgroup` and what that could map to (just a `group` right now), but in working on other updates to the HTML AAM spec today, I thought I'd check the temperature of the working group for roles like these.  Especially since the header and footer elements are only exposed as landmarks under certain conditions, and the rest of the time these "semantic" HTML elements are no different than `div` elements.

Could it not be potentially beneficial to allow AT to expose _something_ beyond a generic or group role for these elements, even if we put a stipulation that the role only be overtly exposed to the user if provided an accessible name?

Note that there are potentially similar existing roles with dpub `doc-pageheader` and `doc-pagefooter` 

Edit: I should have mentioned that I _am_ aware this is retreading an older issue - https://github.com/w3c/aria/issues/963  - but as that was closed in 2019 before the `generic` concept was implemented, and now we're also working on defining minimum role to _combat_ generics not being useful, again, maybe we should _think about if we should start thinking about rethinking this maybe_?

scottaohara:
adds proposed header and footer roles to allow authors to identify areas of the page that would fit the more general HTML definition for the elements of the same name (and would not be considered landmark elements).

These would also allow for the respective HTML elements to expose these new roles - essentially 'renamed' group roles - if they are given properties important to accessibility (focusable, named) but are not meeting the conditions to be exposed as banner / contentinfo landmarks.  The alternative is to just expose the elements as groups, which is doable, but arguably also overloads the use of the group role.

Closes #1915

Quick draft of potential mappings for these, which would go into core aam as a new PR if we want to include these roles in ARIA.  Or, if we do not want to include new roles but _still_ want to surface these elements as more than just 'groups' per my above comment, then these mappings could go into HTML AAM for the scenario where they'd apply.

```
MSAA IA2
    Role: ROLE_SYSTEM_GROUPING
    Object Attribute: xml-roles:header

UIA
    Control Type: Group
    Localized Control Type: header

ATK
    Role: ROLE_PANEL
    Object Attribute: xml-roles:header

AX API
    AXRole: AXGroup
    AXSubrole: <nil>
    AXRoleDescription: header

=====

MSAA IA2
    Role: ROLE_SYSTEM_GROUPING
    Object Attribute: xml-roles:footer

UIA
    Control Type: Group
    Localized Control Type: footer

ATK
    Role: ROLE_PANEL
    Object Attribute: xml-roles:footer

AX API
    AXRole: AXGroup
    AXSubrole: <nil>
    AXRoleDescription: footer
```

# PR tracking

* [ ] Related APG Issue/PR:

# Implementation tracking

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:

### DIFF
--- a/index.html
+++ b/index.html
@@ -3549,6 +3549,84 @@
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="footer">
+			<rdef>footer</rdef>
+			<div class="role-description">
+				<p>A set of user interface objects and information representing information about its closest ancestral content group. For instance, a <code>footer</code> can include information about who wrote the specific section of content, such as an <rref>article</rref>. It can contain links to related documents, copyright information or other indices and colophon specific to the current section of the page.</p>
+				<p>A <code>footer</code> does not represent information about the parent document, or globally repeating content found across multiple pages related to the website. For such content, the <rref>contentinfo</rref> role would be more appropriate.</p>
+			</div>
+			<table class="role-features">
+				<caption>Characteristics:</caption>
+				<thead>
+					<tr>
+						<th scope="col">Characteristic</th>
+						<th scope="col">Value</th>
+					</tr>
+				</thead>
+				<tbody>
+					<tr>
+						<th class="role-abstract-head" scope="row">Is Abstract:</th>
+						<td class="role-abstract"> </td>
+					</tr>
+					<tr>
+						<th class="role-parent-head" scope="row">Superclass Role:</th>
+						<td class="role-parent"><rref>section</rref></td>
+					</tr>
+					<tr>
+						<th class="role-children-head" scope="row">Subclass Roles:</th>
+						<td class="role-children">Placeholder</td>
+					</tr>
+					<tr>
+						<th class="role-base-head" scope="row">Base Concept:</th>
+						<td class="role-base"><code>&lt;[^footer^]&gt;</code> in HTML</td>
+					</tr>
+					<tr>
+						<th class="role-related-head" scope="row">Related Concepts:</th>
+						<td class="role-related"><code>&lt;[^fieldset^]&gt;</code> in HTML</td>
+					</tr>
+					<tr>
+						<th class="role-scope-head" scope="row">Required Accessibility Parent Roles:</th>
+						<td class="role-scope"> </td>
+					</tr>
+					<tr>
+						<th class="role-mustcontain-head" scope="row">Allowed Accessibility Child Roles:</th>
+						<td class="role-mustcontain"> </td>
+					</tr>
+					<tr>
+						<th class="role-required-properties-head">Required States and Properties:</th>
+						<td class="role-required-properties"> </td>
+					</tr>
+					<tr>
+						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
+						<td class="role-properties"> </td>
+					</tr>
+					<tr>
+						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
+						<td class="role-inherited">Placeholder</td>
+					</tr>
+					<tr>
+						<th class="role-namefrom-head" scope="row">Name From:</th>
+						<td class="role-namefrom">author</td>
+					</tr>
+					<tr>
+						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
+						<td class="role-namerequired"> </td>
+					</tr>
+					<tr>
+						<th class="role-namerequired-inherited-head" scope="row">Inherits Name Required:</th>
+						<td class="role-namerequired-inherited"> </td>
+					</tr>
+					<tr>
+						<th class="role-childpresentational-head" scope="row">Children Presentational:</th>
+						<td class="role-childpresentational"> </td>
+					</tr>
+					<tr>
+						<th class="role-presentational-inherited-head" scope="row">Inherits Presentational:</th>
+						<td class="role-presentational-inherited"> </td>
+					</tr>
+				</tbody>
+			</table>
+		</div>
 		<div class="role" id="form">
 			<rdef>form</rdef>
 			<div class="role-description">
@@ -3999,6 +4077,84 @@
 								<li><sref>aria-disabled</sref></li>
 							</ul>
 						</td>
+					</tr>
+					<tr>
+						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>
+						<td class="role-inherited">Placeholder</td>
+					</tr>
+					<tr>
+						<th class="role-namefrom-head" scope="row">Name From:</th>
+						<td class="role-namefrom">author</td>
+					</tr>
+					<tr>
+						<th class="role-namerequired-head" scope="row">Accessible Name Required:</th>
+						<td class="role-namerequired"> </td>
+					</tr>
+					<tr>
+						<th class="role-namerequired-inherited-head" scope="row">Inherits Name Required:</th>
+						<td class="role-namerequired-inherited"> </td>
+					</tr>
+					<tr>
+						<th class="role-childpresentational-head" scope="row">Children Presentational:</th>
+						<td class="role-childpresentational"> </td>
+					</tr>
+					<tr>
+						<th class="role-presentational-inherited-head" scope="row">Inherits Presentational:</th>
+						<td class="role-presentational-inherited"> </td>
+					</tr>
+				</tbody>
+			</table>
+		</div>
+		<div class="role" id="header">
+			<rdef>header</rdef>
+			<div class="role-description">
+				<p>A set of user interface objects and information that represents a collection of introductory items for the element's closest ancestral content group. For instance, a <code>header</code> can include the heading, introductory statement and related meta data for a section of content, for instance a <rref>region</rref> or <rref>article</rref>, within a web page.</p>
+				<p>A <code>header</code> does not represent site-oriented or globally repeating content found across multiple pages of a website. For such content, the <rref>banner</rref> role would be more appropriate.</p>
+			</div>
+			<table class="role-features">
+				<caption>Characteristics:</caption>
+				<thead>
+					<tr>
+						<th scope="col">Characteristic</th>
+						<th scope="col">Value</th>
+					</tr>
+				</thead>
+				<tbody>
+					<tr>
+						<th class="role-abstract-head" scope="row">Is Abstract:</th>
+						<td class="role-abstract"> </td>
+					</tr>
+					<tr>
+						<th class="role-parent-head" scope="row">Superclass Role:</th>
+						<td class="role-parent"><rref>section</rref></td>
+					</tr>
+					<tr>
+						<th class="role-children-head" scope="row">Subclass Roles:</th>
+						<td class="role-children">Placeholder</td>
+					</tr>
+					<tr>
+						<th class="role-base-head" scope="row">Base Concept:</th>
+						<td class="role-base"><code>&lt;[^header^]&gt;</code> in HTML</td>
+					</tr>
+					<tr>
+						<th class="role-related-head" scope="row">Related Concepts:</th>
+						<td class="role-related"><code>&lt;[^fieldset^]&gt;</code> in HTML</td>
+					</tr>
+					<tr>
+						<th class="role-scope-head" scope="row">Required Accessibility Parent Roles:</th>
+						<td class="role-scope"> </td>
+					</tr>
+					<tr>
+						<th class="role-mustcontain-head" scope="row">Allowed Accessibility Child Roles:</th>
+						<td class="role-mustcontain"> </td>
+					</tr>
+					<tr>
+						<th class="role-required-properties-head">Required States and Properties:</th>
+						<td class="role-required-properties"> </td>
+					</tr>
+					<tr>
+						<th class="role-properties-head" scope="row">Supported States and Properties:</th>
+						<td class="role-properties"> </td>
 					</tr>
 					<tr>
 						<th class="role-inherited-head" scope="row">Inherited States and Properties:</th>


index.html
```diff
@@ -4027,6 +4105,84 @@ <h2>Definition of Roles</h2>
...
+						<th class="role-children-head" scope="row">Subclass Roles:</th>
+						<td class="role-children">Placeholder</td>
+					</tr>
+					<tr>
+						<th class="role-base-head" scope="row">Base Concept:</th>
+						<td class="role-base"><code>&lt;[^header^]&gt;</code> in HTML</td>
+					</tr>
+					<tr>
+						<th class="role-related-head" scope="row">Related Concepts:</th>
+						<td class="role-related"><code>&lt;[^fieldset^]&gt;</code> in HTML</td>
```
scottaohara, line 4141:
```suggestion
						<td class="role-related"></td>
```

scottaohara (commented):
need to remove these bits that got copied over accidentally.

cookiecrook:
@scottaohara wrote:
```
AX API
    AXRole: AXGroup
    AXSubrole: <nil>
    AXRoleDescription: header/footer
```

Seems reasonable. Checking with others.

cookiecrook:
> @scottaohara wrote:
> Seems reasonable. Checking with others before confirming.

No comments in 24h. Ship it.

index.html
```diff
@@ -4027,6 +4105,84 @@ <h2>Definition of Roles</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="header">
+			<rdef>header</rdef>
+			<div class="role-description">
+				<p>A set of user interface objects and information that represents a collection of introductory items for the element's closest ancestral content group. For instance, a <code>header</code> can include the heading, introductory statement and related meta data for a section of content, for instance a <rref>region</rref> or <rref>article</rref>, within a web page.</p>
+				<p>A <code>header</code> does not represent site-oriented or globally repeating content found across multiple pages of a website. For such content, the <rref>banner</rref> role would be more appropriate.</p>
```
cookiecrook, line 4112: Maybe "sectionheader" to be more explicit that this is about sub-sections rather than page headers?

Ditto for footer.

spectranaut (commented):
Bear with me as I work out my thoughts on this one because I'm not sure where I land -- whether to introduce these new roles or to leave the distinction between landmark header/footer and not-landmark header/footers in html-aam entirely.

Just to be clear, if this PR lands, the corresponding change in core-aam would be that https://w3c.github.io/html-aam/#el-header and https://w3c.github.io/html-aam/#el-footer point to these new roles and say use these mappings?

I guess one reason why we should introduce these roles is because it is a place to document the concept of a "section header" and "section footer" as apposed to the landmark header/banner and landmaker footer/contentinfo. This makes me feel like there should be some very author-specific guidance about this, and makes me realize we aren't tracking changes to MDN with changes in the spec. Should we be tracking MDN changes? Alternatively, it seems almost too basic, but should there be an APG example for this?

Also, where is the current author facing documentation that says a html `header` and `footer` are considered landmarks in these cases and not in these other cases? I just check mdn and it is not there.

Also, would we want a "screen reader guidance" section as well? Essentially explaining you can expose this information when you feel like it's necessary, as @cookiecrook mentioned in the VoiceOver case?

cookiecrook:
> Also, would we want a "screen reader guidance" section as well? Essentially explaining you can expose this information when you feel like it's necessary, as @cookiecrook mentioned in the VoiceOver case?

No AT changes are needed here. This is only an engine change that results in a different role description. Nothing more.

scottaohara:
@spectranaut probably not a bad idea to add a check for has MDN documentation been updated - these docs are often behind the curve with a11y info/updates.  header mentions the 'scoped' nature of the header element - lightly, now.  footer calls out an old webkit bug in its accessibility section.

scottaohara:
@jnurthen @cookiecrook @spectranaut any other thoughts on this preventing it from merging? got reminded of it when reviewing the (now linked) WPT.

re-author guidance on when a header/footer is a section header/footer vs the landmarks.  that can be added to MDN, but that's blocked on this change being made.  One could argue the info could be more overtly stated with the current reality of how these items work. But I would rather make one change than two.

cookiecrook:
> @jnurthen @cookiecrook @spectranaut any other thoughts on this preventing it from merging? got reminded of it when reviewing the (now linked) WPT.
> 
> re: author guidance on when a header/footer is a section header/footer vs the landmarks. that can be added to MDN, but that's blocked on this change being made. One could argue the info could be more overtly stated with the current reality of how these elements work. But I would rather make one change than two.

I was suggesting the role name be changed from `header`/`footer` to `sectionheader`/`sectionfooter` to prevent author confusion with the existing role `banner`…

index.html
```diff
@@ -3549,6 +3549,84 @@ <h2>Definition of Roles</h2>
 				</tbody>
 			</table>
 		</div>
+		<div class="role" id="footer">
+			<rdef>footer</rdef>
```
scottaohara, line 3553:
```suggestion
			<rdef>sectionfooter</rdef>
```

scottaohara:
thanks @cookiecrook.  for some reason i had a memory of already doing this, but clearly my memory was faulty.  i've actually done it now.

spectranaut:
Per our new process, changes to the ARIA spec can't land until there are implementations or implementation commitments, and related changes to the AAMs/Accname, so I think you need to open all those PRs and bugs on browsers.

But I just realized we sort of need an intermediary, because we shouldn't open bugs on browsers until we have "consensus" from the working group... we we need a way of keeping track of that some how.  I think it used to be that once we had consensus we merged, and that is how we "kept track".

spectranaut:
Ah good point, @scottaohara -- three reviews = consensus. So, we can open issues on implementations after there are three reviews, and we can merge after there is at least one implementation and implementation commitment

spectranaut (approved):
I think these sections look good! Very clear description.

cookiecrook (approved):

jnurthen (approved):

spectranaut:
Assigning to myself to make tests and issues on browsers.

jcsteh:
Sorry if I've missed something, but I'm a bit unclear: will this result in changing HTML `<header>`/`<footer>` so that they are no longer landmarks?

pkra:
>will this result in changing HTML `<header>`/`<footer>` so that they are no longer landmarks?

@jcsteh no, these roles are in addition to those. 

To quote @scottaohara's starting comment:

> These would also allow for the respective HTML elements to expose these new roles - essentially 'renamed' group roles - if they are given properties important to accessibility (focusable, named) but are not meeting the conditions to be exposed as banner / contentinfo landmarks.

jcsteh:
Ah, I get it. I think I was (and am still a little) confused by the "if they are given properties important to accessibility (focusable, named)" condition. If we're creating new roles for these, why impose that condition only for implicit roles on HTML elements? As an example, a `<p>` is always a paragraph, even if it isn't focusable or has no name. Generics are different, but generics are special because they're effectively non-semantic. These new roles are either semantic... or they're not.

scottaohara:
that's the problem that these new roles are trying to solve.  right now header and footer, if not landmarks, are treated as generic.  but in HTML header and footer have semantic meaning regardless of their usage.

while these elements always have semantic meaning, we talked alot about how we don't to necessarily add new announcements of these roles for situations where the content alone is understandable.  hence why we wanted to treat these similar to the group role - where if the group is unnamed or not focusable, then the group (and these roles) wouldn't have to be overtly announced to users.

and then on the flip side, the other reason for these specific roles is because authors do provide header and footer elements accessible names, or make them focusable (for better or worse) but if these elements aren't scoped to be landmarks, then they are named/focusable generics and we want to avoid that.  sometimes having these html elements be nameable and sometimes not.

jcsteh:
> hence why we wanted to treat these similar to the group role - where if the group is unnamed or not focusable, then the group (and these roles) wouldn't have to be overtly announced to users.

That's reasonable, but there's an important distinction between "not announced by screen readers" and "not exposing the roles to accessibility APIs". With the group role, it's always exposed; screen readers just choose not to report it if there's no significant property. If I read your description correctly (and maybe I'm misunderstanding it), you're suggesting that `<header>` wouldn't be exposed at all unless it has a significant property (name, focusable, etc.). Aside from anything else, this would diverge from ARIA, since role="section header" would always be exposed but `<header>` might not be.

jcsteh:
> but in HTML header and footer have semantic meaning regardless of their usage.

Do they? HTML-AAM mostly maps them to "generic", with the exception of ATK. (IMO, UIA doesn't count because it basically overloads/abuses LocalizedControlType for everything instead of having proper native semantics. 😒)

> so `<header>` would be exposed as sectionheader, when not exposed as a banner

For the sake of absolute clarity, "when not  exposed as a banner" would be the only exposure condition as far as HTML-AAM and thus browsers are concerned? (I understand the intended screen reader behaviour here; I'm just trying to be sure I fully understand the intended API mappings.)

scottaohara:
>Do they? HTML-AAM mostly maps them to "generic",

Yes. The header and footer elements are meant to be used for more than just for `banner` and `contentinfo` purposes if going by their definitions / indented "semantic" (not just a11y mappings) usage in the HTML spec.  

>For the sake of absolute clarity, "when not exposed as a banner" would be the only exposure condition as far as HTML-AAM and thus browsers are concerned?

yes. Essentially header and footer would _not_ be exposed as generics anymore.  So the conditions for them to be exposed as banner/contentinfo, respectively remain the same, and the instances of where they would instead be generic would instead map to these proposed 'group' roles with implicit aria-roledescriptions.

core-aam/index.html
```diff
@@ -4339,6 +4339,109 @@ <h4 id="role-map-searchbox"><code>searchbox</code></h4>
...
+              <tr>
+                <th>
+                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+                </th>
+                <td>
+                  <span class="property">AXRole: <code>AXGroup</code></span
+                  ><br />
+                  <span class="property">AXSubrole: <code>&lt;nil&gt;</code></span
+                  ><br />
+                  <span class="property">AXRoleDescription: <code>sectionheader</code></span>
```
cookiecrook, line 4440: Missed this first time around. AXRoleDescription is a user-presented string, so should include the space. Presumably the same is true for UIA's _**Localized**_ Control Type".

I've also reached out to the API owners on whether they want a new AXSubrole.

cookiecrook (changes requested):

core-aam/index.html
scottaohara, line 4440:
```suggestion
                  <span class="property">AXRoleDescription: <code>section header</code></span>
```

core-aam/index.html
```diff
@@ -4339,6 +4339,109 @@ <h4 id="role-map-searchbox"><code>searchbox</code></h4>
...
+              <tr>
+                <th>
+                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+                </th>
+                <td>
+                  <span class="property">AXRole: <code>AXGroup</code></span
+                  ><br />
+                  <span class="property">AXSubrole: <code>&lt;nil&gt;</code></span
+                  ><br />
+                  <span class="property">AXRoleDescription: <code>sectionfooter</code></span>
```
scottaohara, line 4389:
```suggestion
                  <span class="property">AXRoleDescription: <code>section footer</code></span>
```

core-aam/index.html
```diff
@@ -4339,6 +4339,109 @@ <h4 id="role-map-searchbox"><code>searchbox</code></h4>
...
+                  ><br />
+                  <span class="property">Object Attribute: <code>xml-roles:sectionfooter</code></span>
+                </td>
+              </tr>
+              <tr>
+                <th><abbr title="User Interface Automation">UIA</abbr></th>
+                <td>
+                  <span class="property">Control Type: <code>Group</code></span
+                  ><br />
+                  <span class="property">Localized Control Type: <code>sectionfooter</code></span>
```
scottaohara, line 4370:
```suggestion
                  <span class="property">Localized Control Type: <code>section footer</code></span>
```

core-aam/index.html
```diff
@@ -4339,6 +4339,109 @@ <h4 id="role-map-searchbox"><code>searchbox</code></h4>
...
+                  ><br />
+                  <span class="property">Object Attribute: <code>xml-roles:sectionheader</code></span>
+                </td>
+              </tr>
+              <tr>
+                <th><abbr title="User Interface Automation">UIA</abbr></th>
+                <td>
+                  <span class="property">Control Type: <code>Group</code></span
+                  ><br />
+                  <span class="property">Localized Control Type: <code>sectionheader</code></span>
```
scottaohara, line 4422:
```suggestion
                  <span class="property">Localized Control Type: <code>section header</code></span>
```

core-aam/index.html
scottaohara: thank you for catching this!  i've updated the instances

core-aam/index.html
```diff
@@ -4339,6 +4339,109 @@ <h4 id="role-map-searchbox"><code>searchbox</code></h4>
...
+                </td>
+              </tr>
+              <tr>
+                <th>
+                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+                </th>
+                <td>
+                  <span class="property">AXRole: <code>AXGroup</code></span
+                  ><br />
+                  <span class="property">AXSubrole: <code>&lt;nil&gt;</code></span
```
cookiecrook, line 4387:
```suggestion
                  <span class="property">AXSubrole: <code>AXSectionFooter</code></span
```

core-aam/index.html
```diff
@@ -4339,6 +4339,109 @@ <h4 id="role-map-searchbox"><code>searchbox</code></h4>
...
```
cookiecrook, line 4438:
```suggestion
                  <span class="property">AXSubrole: <code>AXSectionHeader</code></span
```

cookiecrook (approved):
approved with the new subroles

cookiecrook (approved):

scottaohara:
confirming that Edge/Chrome (both msaa/ia2 and uia trees) expose the sectionheader and sectionfooter roles.

scottaohara:
> Ah good point, @scottaohara -- three reviews = consensus. So, we can open issues on implementations after there are three reviews, and we can merge after there is at least one implementation and implementation commitment

@spectranaut since this has been implemented in chromium, and the other browsers have their accepted bugs on them, seems we're good to merge per your last comment?

PR #1949, opened 2023-06-06
Clarification for exposure of generic elements

scottaohara:
closes https://github.com/w3c/html-aam/issues/489

reworded the last paragraph of the generic definition to indicate that it can be ignored when not providing information important to the a11y tree, but if it does provide such information, then the generic element should be exposed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3637,7 +3637,8 @@
 			<div class="role-description">
 				<p>A nameless container <a>element</a> that has no semantic meaning on its own.</p>
 				<p>The <code>generic</code> role is intended for use as the implicit role of generic elements in host languages (such as HTML <code>div</code> or <code>span</code>), so is primarily for implementors of user agents. Authors SHOULD NOT use this role in content. Authors MAY use <rref>presentation</rref> or <rref>none</rref> to remove implicit accessibility semantics, or a semantic container role such as <rref>group</rref> to semantically group descendants in a named container.</p>
-				<p>Like an element with role <rref>presentation</rref>, an element with role <code>generic</code> can provide a limited number of accessible states and properties for its descendants, such as <pref>aria-live</pref> attributes. However, unlike elements with role <rref>presentation</rref>, <code>generic</code> elements are exposed in <a>accessibility APIs</a> so that assistive technologies can gather certain properties such as layout and bounds.</p>
+				<p>Like an element with role <rref>presentation</rref>, an element with role <code>generic</code> can provide a limited number of accessible states and properties for its descendants, such as <pref>aria-live</pref> attributes.</p>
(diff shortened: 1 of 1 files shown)


benbeaudry (approved):
Great clarification. LGTM.

index.html
```diff
@@ -3637,7 +3637,8 @@ <h2>Definition of Roles</h2>
...
-				<p>Like an element with role <rref>presentation</rref>, an element with role <code>generic</code> can provide a limited number of accessible states and properties for its descendants, such as <pref>aria-live</pref> attributes. However, unlike elements with role <rref>presentation</rref>, <code>generic</code> elements are exposed in <a>accessibility APIs</a> so that assistive technologies can gather certain properties such as layout and bounds.</p>
+				<p>Like an element with role <rref>presentation</rref>, an element with role <code>generic</code> can provide a limited number of accessible states and properties for its descendants, such as <pref>aria-live</pref> attributes.</p>
+<p>User agents MAY ignore <code>generic</code> elements which do not provide such useful state or properties which assistive technology would find useful to expose. However, unlike elements with role <rref>presentation</rref>, <code>generic</code> elements are exposed in <a>accessibility APIs</a> when such attributes have been specified.</p>
```
mcking65, line 3641: Even though it is only a MAY statement, The language should be more specific regarding when it is permissible for UA to ignore the generic. We don't define "useful" anywhere.

Also, we should use "that" instead of "which" where the language is intended to be restrictive.

```suggestion
<p>unlike elements with role <rref>presentation</rref>, user agents expose <code>generic</code> elements in <a>accessibility APIs</a> when the <code>generic</code> elements specify permitted accessibility attributes. User agents MAY ignore a <code>generic</code> element if the element does not provide any states or properties that are not prohibited on <code>generic</code> elements.</p>
```

adampage (approved):
Ditto. 🚀

mcking65 (changes requested):
As a normative statement, I think it should be more specific. I provided a suggestion.

index.html
scottaohara, line 3641: Thanks @mcking65 

I took what you had suggested and did some more revisions to make this a bit easier to read.  I was having a tough time following your revised User agents MAY sentence... but I think what I have here achieves the same.

```suggestion
<p>However, unlike elements with role <code>presentation</code>, user agents expose <code>generic</code> elements in <a>accessibility APIs</a> when permitted accessibility attributes have been specified. User agents MAY otherwise ignore <code>generic</code> elements if such permitted attributes have not been specified.</p>
```

scottaohara: i have merged in my revision of what you have proposed here.  so resolving this.

mcking65 (approved):

mcking65:
Thank you  @scottaohara. I think it's good to ship now.

PR #1965, opened 2023-06-30
Accessibility child/parent clean up

### LINKED ISSUE #1930: Accessibility Parent/Child follow ups
Follow up from F2F meeting discussion: https://www.w3.org/2023/05/04-aria-minutes#t01

1. Add clarification on "intervening" or revisit the word
2. Editorial: turn definition into bulleted list
3. Editorial: add note that accessibility child/parent is not the same as the browser/platform accessibility tree parent/child
4. Figure out where examples should go relative to the definition (i.e. that the definition would benefit by being adjacent to HTML example snippets)
5. Add "Accessibility descendant" definition & use in place of unclear use of descendant (#1150)
6. Review/remove remaining use of owned/owning elements (#1161)
7. Follow up on #1817

spectranaut:
Closes #1930

Attempt to update definition of "accessibility child" and "accessibility parent" and add "accessibility descendant"

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1965.html" title="Last updated on Jul 28, 2023, 11:38 PM UTC (2c76e79)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1965/051ed84...2c76e79.html" title="Last updated on Jul 28, 2023, 11:38 PM UTC (2c76e79)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -259,6 +259,25 @@
 			  <p>Operating systems and other platforms provide a set of interfaces that expose information about <a class="termref">objects</a> and <a class="termref">events</a> to <a>assistive technologies</a>. Assistive technologies use these interfaces to get information about and interact with those <a class="termref">widgets</a>. Examples of accessibility APIs are <a href="https://docs.microsoft.com/en-us/windows/win32/winauto/microsoft-active-accessibility">Microsoft Active Accessibility</a> [[MSAA]], <a href="https://docs.microsoft.com/en-us/windows/win32/winauto/entry-uiauto-win32">Microsoft User Interface Automation</a> [[UI-AUTOMATION]], <abbr title="Microsoft Active Accessibility">MSAA</abbr> with <cite><a href="https://docs.microsoft.com/en-us/windows/win32/winauto/iaccessibleex"><abbr title="User Interface Automation">UIA</abbr> Express</a></cite> [[UIA-EXPRESS]], the
 				  <a href="https://developer.apple.com/documentation/appkit/nsaccessibility">Mac <abbr title="OS Ten">OS X</abbr> Accessibility Protocol</a> [[AXAPI]], the <cite><a href="https://developer.gnome.org/atk/unstable/">Linux/Unix Accessibility Toolkit</a></cite> [[ATK]] and <cite><a href="https://developer.gnome.org/libatspi/stable/">Assistive Technology Service Provider Interface</a></cite> [[AT-SPI]], and <a href="https://wiki.linuxfoundation.org/accessibility/iaccessible2/start">IAccessible2</a> [[IAccessible2]].</p>
 			</dd>
+			<dt><dfn data-export="" data-lt="accessibility child|accessibility children|child element|child|children|child elements">Accessibility child</dfn></dt>
+			<dd>
+				<p>To calculate the <a>accessibility children</a> of an element `X`, first considered a modified <abbr title="Document Object Model">DOM</abbr> tree where elements (and their subtrees) specified by `aria-owns` are reparented to the element with `aria-owns` set.
+				<p>Now, an <a>accessibility child</a> of an <a>element</a> `X` is considered to be any of the following:
+				<ul>
+					<li>A DOM child of `X` in that modified DOM tree.</li>
+					<li>A DOM descendant of `X` in that modified DOM tree where the only intervening elements between the descendant and `X` have role <rref>generic</rref> or <rref>none</rref>.</li>
+				</ul>
+				</p>
+                                <p class="note">Accessibility child, accessibility descendant and accessibility parents are terms used to defined required relationships between different DOM elements with ARIA roles or properties. The terms do not refer to a platform accessibility API trees and the relationship between platform accessiblity nodes.</p>
+			</dd>
+			<dt><dfn data-export="" data-lt="accessibility descendant|accessibility descendants">Accessibility descendant</dfn></dt>
+			<dd>
+				<p>To calculate the <a>accessibility descendants</a> of an element `X`, first considered a modified <abbr title="Document Object Model">DOM</abbr> tree where elements (and their subtrees) specified by `aria-owns` are reparented to the element with `aria-owns` set. The <a>accessibility descendants</a> of `X` are all DOM descendants of `X` in that modified DOM tree.</a></p>
+			</dd>
+			<dt><dfn data-export="" data-lt="accessibility parent|parent element|parent">Accessibility parent</dfn></dt>
+			<dd>
+				<p>To calculate the <a>accessibility parent</a> of an element `X`, first considered a modified <abbr title="Document Object Model">DOM</abbr> tree where elements (and their subtrees) specified by `aria-owns` are reparented to the element with `aria-owns` set. The <a>accessibility parent</a> of `X` is the closest DOM ancestor of the `X` in that modified tree that does not have role <rref>generic</rref> or <rref>none</rref>.</p>
+			</dd>
 			<dt><dfn data-export="">Accessible object</dfn></dt>
 			<dd>
 			  <p>A [=nodes|node=] in the <a class="termref">accessibility tree</a> of a platform <a>accessibility <abbr title="application programming interface">API</abbr></a>. Accessible objects expose various <a class="termref">states</a>, [=ARIA/properties=], and <a class="termref">events</a> for use by <a>assistive technologies</a>.  In the context of markup languages (e.g., HTML and SVG) in general, and of WAI-ARIA in particular, markup [=elements=] and their [=attributes=] are represented as accessible objects.  </p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -259,6 +259,25 @@ <h1>Important Terms</h1>
...
+			</dd>
+			<dt><dfn data-export="" data-lt="accessibility descendant|accessibility descendants">Accessibility descendant</dfn></dt>
+			<dd>
+				<p>To calculate the <a>accessibility descendants</a> of an element `X`, first considered a modified <abbr title="Document Object Model">DOM</abbr> tree where elements (and their subtrees) specified by `aria-owns` are reparented to the element with `aria-owns` set. The <a>accessibility descendants</a> of `X` are all DOM descendants of `X` in that modified DOM tree.</a></p>
+			</dd>
+			<dt><dfn data-export="" data-lt="accessibility parent|parent element|parent">Accessibility parent</dfn></dt>
+			<dd>
+				<p>To calculate the <a>accessibility parent</a> of an element `X`, first considered a modified <abbr title="Document Object Model">DOM</abbr> tree where elements (and their subtrees) specified by `aria-owns` are reparented to the element with `aria-owns` set. The <a>accessibility parent</a> of `X` is the closest DOM ancestor of the `X` in that modified tree that does not have role <rref>generic</rref> or <rref>none</rref>.</p>
+			</dd>
```
spectranaut, line 280: I moved this section to be more alphabetical. We talked about moving this out of the "important terms" section of this document, @jnurthen had the strongest opinions on this but he is out for 3 weeks! (the PR can be open that long I'd be happy to move this).

index.html
```diff
@@ -13567,10 +13571,10 @@ <h3>States and Properties</h3>
 			<p>In general, [=user agents=] do not do much validation of <abbr title="Accessible Rich Internet Application">WAI-ARIA</abbr> [=ARIA/properties=]. User agents MAY do some minor validation on request and enforce things like <pref>aria-posinset</pref> being within 1 and <pref>aria-setsize</pref>, inclusive. User agents are not  responsible for logical validation, such as the following:</p>
 				<ol>
 					<li>Circular references created by relations, such as specifying that two [=elements=] own each other. </li>
-					<li>Correct usage with regard to <abbr title="Document Object Model">DOM</abbr> tree structure, such as an <a>element</a> being owned by more than one other element.</li>
+					<li>Correct usage with regard to <abbr title="Document Object Model">DOM</abbr> tree structure.</li>
```
spectranaut, line 13574: I just didn't understand the point of the second half of this sentence so I removed it.

smhigley: My reading of this was that it's saying user agents are not responsible for validating something like this:

```html
<div aria-owns="child">content</div>
<div aria-owns="child">content 2<div>
<div id="child">I have two parents</div>
```

spectranaut: ohhh I see ok got it. Well, I will put it back, also, it seems like "owns" make sense here since it actually means the reference of `aria-owns`

index.html
```diff
@@ -259,6 +259,25 @@ <h1>Important Terms</h1>
...
 				  <a href="https://developer.apple.com/documentation/appkit/nsaccessibility">Mac <abbr title="OS Ten">OS X</abbr> Accessibility Protocol</a> [[AXAPI]], the <cite><a href="https://developer.gnome.org/atk/unstable/">Linux/Unix Accessibility Toolkit</a></cite> [[ATK]] and <cite><a href="https://developer.gnome.org/libatspi/stable/">Assistive Technology Service Provider Interface</a></cite> [[AT-SPI]], and <a href="https://wiki.linuxfoundation.org/accessibility/iaccessible2/start">IAccessible2</a> [[IAccessible2]].</p>
 			</dd>
+			<dt><dfn data-export="" data-lt="accessibility child|accessibility children|child element|child|children|child elements">Accessibility child</dfn></dt>
+			<dd>
+				<p>To calculate the <a>accessibility children</a> of an element `X`, first considered a modified <abbr title="Document Object Model">DOM</abbr> tree where elements (and their subtrees) specified by `aria-owns` are reparented to the element with `aria-owns` set.
+				<p>Now, an <a>accessibility child</a> of an <a>element</a> `X` is considered to be any of the following:
```
scottaohara, line 265: can't nest lists within paragraphs

```suggestion
				<p>Now, an <a>accessibility child</a> of an <a>element</a> `X` is considered to be any of the following:</p>
```

index.html
```diff
@@ -259,6 +259,25 @@ <h1>Important Terms</h1>
...
+				<ul>
+					<li>A DOM child of `X` in that modified DOM tree.</li>
+					<li>A DOM descendant of `X` in that modified DOM tree where the only intervening elements between the descendant and `X` have role <rref>generic</rref> or <rref>none</rref>.</li>
+				</ul>
+				</p>
```
scottaohara, line 270:
```suggestion
```

index.html
```diff
@@ -259,6 +259,25 @@ <h1>Important Terms</h1>
...
+                                <p class="note">Accessibility child, accessibility descendant and accessibility parents are terms used to defined required relationships between different DOM elements with ARIA roles or properties. The terms do not refer to a platform accessibility API trees and the relationship between platform accessiblity nodes.</p>
```
scottaohara, line 271: should these _always_ be terms used to defined required relationships?  I'd very much like to use these terms (and had started to in a draft pr) to communicate relationships for scoping.  I guess the way i'm looking at it there is that the relationships are not _required_ but in the situations where applicable, being a child of X would give the child element Y a contextual role, or not.

scottaohara (commented):
tbh i'm having a really hard time following these changes.  the prior wording made sense to me and now instead of defining what the terms are for people unfamiliar with them, they've been replaced with implementor algorithms heavily referring to `aria-owns`.  I think we need both.

index.html
```diff
@@ -259,6 +259,25 @@ <h1>Important Terms</h1>
...
 				  <a href="https://developer.apple.com/documentation/appkit/nsaccessibility">Mac <abbr title="OS Ten">OS X</abbr> Accessibility Protocol</a> [[AXAPI]], the <cite><a href="https://developer.gnome.org/atk/unstable/">Linux/Unix Accessibility Toolkit</a></cite> [[ATK]] and <cite><a href="https://developer.gnome.org/libatspi/stable/">Assistive Technology Service Provider Interface</a></cite> [[AT-SPI]], and <a href="https://wiki.linuxfoundation.org/accessibility/iaccessible2/start">IAccessible2</a> [[IAccessible2]].</p>
 			</dd>
+			<dt><dfn data-export="" data-lt="accessibility child|accessibility children|child element|child|children|child elements">Accessibility child</dfn></dt>
+			<dd>
+				<p>To calculate the <a>accessibility children</a> of an element `X`, first considered a modified <abbr title="Document Object Model">DOM</abbr> tree where elements (and their subtrees) specified by `aria-owns` are reparented to the element with `aria-owns` set.
```
adampage, line 264: The description of “a modified DOM tree” threw me for a loop here. Would it be more accurate to name the “[accessibility tree](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1965.html#accessibility_tree)” instead, since it’s generated from the DOM but then modified through features like `aria-owns`, `aria-hidden`, thereby diverging from the actual DOM? That would also feel harmonious with the “accessibility child/descendant/parent” terminology, IMO.

index.html
```diff
@@ -259,6 +259,25 @@ <h1>Important Terms</h1>
...
+				</ul>
+				</p>
+                                <p class="note">Accessibility child, accessibility descendant and accessibility parents are terms used to defined required relationships between different DOM elements with ARIA roles or properties. The terms do not refer to a platform accessibility API trees and the relationship between platform accessiblity nodes.</p>
+			</dd>
+			<dt><dfn data-export="" data-lt="accessibility descendant|accessibility descendants">Accessibility descendant</dfn></dt>
+			<dd>
+				<p>To calculate the <a>accessibility descendants</a> of an element `X`, first considered a modified <abbr title="Document Object Model">DOM</abbr> tree where elements (and their subtrees) specified by `aria-owns` are reparented to the element with `aria-owns` set. The <a>accessibility descendants</a> of `X` are all DOM descendants of `X` in that modified DOM tree.</a></p>
+			</dd>
+			<dt><dfn data-export="" data-lt="accessibility parent|parent element|parent">Accessibility parent</dfn></dt>
```
adampage, line 277: If we’re introducing a definition for “accessibility descendant” that is distinct from “accessibility child”, should we also add a definition for “accessibility ancestor” that is distinct from “accessibility parent”, for completeness?

index.html
adampage, line 264: I like the establishing of `X` as a shorthand, that’s handy. 👌🏻

index.html
```diff
@@ -259,6 +259,25 @@ <h1>Important Terms</h1>
...
 				  <a href="https://developer.apple.com/documentation/appkit/nsaccessibility">Mac <abbr title="OS Ten">OS X</abbr> Accessibility Protocol</a> [[AXAPI]], the <cite><a href="https://developer.gnome.org/atk/unstable/">Linux/Unix Accessibility Toolkit</a></cite> [[ATK]] and <cite><a href="https://developer.gnome.org/libatspi/stable/">Assistive Technology Service Provider Interface</a></cite> [[AT-SPI]], and <a href="https://wiki.linuxfoundation.org/accessibility/iaccessible2/start">IAccessible2</a> [[IAccessible2]].</p>
 			</dd>
+			<dt><dfn data-export="" data-lt="accessibility child|accessibility children|child element|child|children|child elements">Accessibility child</dfn></dt>
+			<dd>
+				<p>To calculate the <a>accessibility children</a> of an element `X`, first considered a modified <abbr title="Document Object Model">DOM</abbr> tree where elements (and their subtrees) specified by `aria-owns` are reparented to the element with `aria-owns` set.
+				<p>Now, an <a>accessibility child</a> of an <a>element</a> `X` is considered to be any of the following:
+				<ul>
+					<li>A DOM child of `X` in that modified DOM tree.</li>
```
adampage, line 267: As a quick follow-up to my comment above, the use of DOM here muddies things a bit for me. In following code for example, I’d like to be able to simultaneously refer to `#treat` as both an accessibility child of `#cat` and a DOM child of `#jar`, so that I can cleanly distinguish between different usage scenarios involving it (like in [#1839](https://github.com/w3c/aria/pull/1839)):

```html
<div id="cat" aria-owns="treat"></div>
<div id="jar">
    <div id="treat"></div>
</div>
```

index.html
spectranaut: hmmm yeah honestly I thought of that but then @smhigley thought we were building up to define the accessibility tree in terms of accessibility child/parent. we have a bit of a chicken and an egg situation

index.html
spectranaut: Ahhhh good point...

index.html
spectranaut: Good point. I'll revise this.

spectranaut:
> tbh i'm having a really hard time following these changes. the prior wording made sense to me and now instead of defining what the terms are for people unfamiliar with them, they've been replaced with implementor algorithms heavily referring to `aria-owns`. I think we need both.

Hmm yeah I'm really not married to this new language... But as to your point about liking the original language -- in the face to face, people have a lot of confusion about the original language from @smhigley  and to be honest in some ways it was incorrect -- it didn't make clear that for dom parent, there was an "exclusive or" situation (there is only one accessibility parent) or which one is the actual accessibility parent. I think it makes sense if you already know what the phrase means.

Also, I had a lot of trouble thinking of a way to define "accessibility descendant" with those previous definitions.

I can think about it more. But can you clarify one thing for me: why is it a problem to use "aria-owns" in this definition? Some how we need to explain what the "accessibility tree is".

spectranaut:
Ok actually I feel this direction was confusing, given you were both confused, and foolish because there is actually a definition of the accessibility tree. Trying something different.

Issue #1956, opened 2023-06-15
`aria-controls` with late-mounted elements


romgrk:
## Describe your concern

I've been investigating the output reports of tools such as lighthouse that present ARIA issues. One common issue we have is with buttons that have `aria-haspopup="menu"` and `aria-controls={elementId}`, where we don't actually mount the corresponding menu DOM element until the point where the button is clicked. This is done for performance reasons for cases where there are a lot of elements, which is particularly important for lower-end devices.

My question is, is there a way to mark late-mounted elements with ARIA? If so, how?

### Link to the version of the specification or documentation you were looking at at.

Link to documentation: https://w3c.github.io/aria/#aria-controls

Does the issue exists in [the editors draft](https://w3c.github.io/aria/)? Unsure/NA

jnurthen:
IMO this is a false positive from the tools as the behaviour is defined.

`aria-controls` is not required on a button (even for a combobox `aria-controls` is only required when the combobox is expanded.
https://w3c.github.io/aria/#document-handling_author-errors_states-properties states that 

> User agents MUST NOT expose WAI-ARIA attributes that reference unresolved IDs. For example:
> 
> When the state or property has only one ID reference that cannot be resolved, treat as if the state or property is not present.
> When the state or property has a list of ID references, ignore any that can't be resolved. If none in the list can be resolved, treat as if the state or property is not present.

Maybe we should move this from the Handling Author Errors section (9.2) to the States and Property Attribute Processing Section (8.6) to indicate better than this is supported behaviour not an error to be flagged by validators.

romgrk:
> aria-controls is not required on a button (even for a combobox aria-controls is only required when the combobox is expanded.)

Sounds good, that answers my question.

> ...to indicate better than this is supported behaviour not an error to be flagged by validators.

It could be nice to attach a note about that in the spec.

Unless you want to keep this open to track moving it in a new section, this issue can be closed.

mpaiva:
@mcking65  - I'm looking forward to pairing with you on this issue

mpaiva:
@mcking65 - I'll be taking a look at this issue tomorrow.

scottaohara:
related https://github.com/w3c/aria/issues/1970

romgrk:
[...]

Issue #1404, opened 2021-02-18
Extend support for aria-expanded to the radio role


36degrees:
This issue follows on from [a previous discussion in #1277](https://github.com/w3c/aria/issues/1277#issuecomment-689478291).

The current working draft of ARIA 1.2 extends support for `aria-expanded` to the checkbox role, but not the radio role. We'd like to propose that it should be extended to the radio role as well.

## Overview

The [radio](https://design-system.service.gov.uk/components/radios/#conditionally-revealing-content) and [checkbox](https://design-system.service.gov.uk/components/checkboxes/#conditionally-revealing-content) components in the GOV.UK Design System both support conditionally revealing additional controls, for example a text input, based on the checked state of a radio or checkbox.

For example, take the question "How do you want to be contacted?" If a user selects **Email**, you can ask for their email address.

![Screenshot. A fieldset with the legend 'How do you want to be contacted?', containing three radio buttons labelled 'Email', 'Phone' and 'Text message'. The 'Email' radio button is selected, which has conditionally revealed an additional text input labelled 'Email address'](https://user-images.githubusercontent.com/121939/108375252-b1e06280-71f9-11eb-985e-8876a0f99b54.png)

As [noted in the ARIA documentation for `aria-expanded`](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded), 'simplifying the user interface by collapsing sections may improve usability for all, including those with cognitive or developmental disabilities.'

We use `aria-expanded` as part of the implementation for both radio buttons and checkboxes. This is because, despite being [invalid according to ARIA 1.0](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-expanded), it still has good support in current assistive technologies (as documented below). We also believe that it improves the experience for screen reader users because it:

- makes the behaviour more predictable, as the user is notified that interacting with the control will reveal, or expand, something else on the page
- notifies the user of the change in the page when it happens, as the user interacts with the control

## Current support for `aria-expanded` on radios and checkboxes

To find out what support currently exists for `aria-expanded`, we tested different combinations of assistive technologies and browsers.

Almost all of the combinations either partly or fully supported `aria-expanded` on checkboxes. The only exception was Voiceover / Safari on macOS.

[...]

PR #1996, opened 2023-08-16
Editorial: aria-controls spec update

### LINKED ISSUE #995: What to do about aria-controls
`aria-controls` did not have the widest support, but with the [JAWS release notes (April 2019)](https://support.freedomscientific.com/Downloads/JAWS/JAWSWhatsNew?utm_content=93956937&utm_medium=social&utm_source=twitter&hss_channel=tw-2382446636#ChangesApril2019) it will no longer be announced by default.

>If you encounter an element on a web page with a defined ARIA controls relationship, JAWS will no longer say “use JAWSKEY+ALT+M to move to controlled element” by default. In most cases, the target of the controls relationship is adjacent to the element or does not provide any useful information.

This update appears to have been made specifically to address "In most cases, the target of the controls relationship is adjacent to the element or does not provide any useful information."  But testing should be done to verify.

Additional reading from past `aria-controls` discussion:
[2017 Ticket about `aria-controls` should only point to visible elements](https://github.com/w3c/aria/issues/531).

scottaohara:
resolves #995 

The intent with this update is, 4 years later, to propose a draft solution for the issue I opened.

With the progress made on aria-details, and the overlaps in how this attribute could have been useful, I did take great liberty in pulling in further explanation/examples for aria-controls from the existing aria-details spec text.

Things i didn't overly call out in this pr, but should probably make their way in there concern recommendations on whether or not to expose the controlling relationship depending on DOM/a11y API relationship.  E.g., we probably don't need an explicit user exposed controlling relationship if a controlled element is directly after the controlling element in the dom (e.g., a disclosure widget) but we _would_ want AT to know they could make this association clear if the elements were located in different portions of the a11y tree - e.g., a button opening a non-modal dialog that happens to be at the bottom of the DOM, but the button is somewhere else entirely.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1996.html" title="Last updated on May 21, 2025, 1:52 PM UTC (b91f88e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/1996/b085915...b91f88e.html" title="Last updated on May 21, 2025, 1:52 PM UTC (b91f88e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -13120,7 +13120,7 @@ button.ariaPressed; // null</pre
             <p>Instance where an <code>aria-controls</code> association could be made:</p>
             <ul>
               <li>A tree view representing a table of contents where choosing a treeitem updates content of a neighboring document pane.</li>
-              <li>A group of checkboxes can control what commodity prices are tracked live in a table or graph.</li>
+              <li>A series of checkboxes can each control what commodity prices are tracked live in a table or graph.</li>
               <li>A tab controls the display of its associated tab panel.</li>
               <li>A radio button group applies filter choices to a listing of search results.</li>
             </ul>
--- a/index.html
+++ b/index.html
@@ -13113,7 +13113,9 @@ button.ariaPressed; // null</pre
             </p>
             <p>
               The <code>aria-controls</code> property is for referencing elements that are modified by the user interacting with the currently focused element or composite widget. The presence of
-              <code>aria-controls</code> enables <a>assistive technologies</a> to programmatically associate the currently focused element with the element or elements it controls.  For instance, it can be used to inform users that by interacting with the controlling element they have revealed an element or elements that were previously in the hidden state. Or, by interacting with an element, they caused the selection or value of a controlled element to change.
+              <code>aria-controls</code> enables <a>assistive technologies</a> to programmatically associate the currently focused element with the element or elements it controls. For instance, it
+              can be used to inform users that by interacting with the controlling element they have revealed an element or elements that were previously in the hidden state. Or, by interacting with
+              an element, they caused the selection or value of a controlled element to change.
             </p>
             <p>Instance where an <code>aria-controls</code> association could be made:</p>
             <ul>
--- a/index.html
+++ b/index.html
@@ -13107,7 +13107,7 @@ button.ariaPressed; // null</pre
         <div class="property" id="aria-controls">
           <pdef>aria-controls</pdef>
           <div class="property-description">
-            <p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the focused element or composite widget. See related <pref>aria-details</pref>.</p>
+            <p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the focused element or composite widget. See related <pref>aria-details</pref> and <pref>aria-owns</pref>.</p>
             <p>The <code>aria-controls</code> property is for referencing elements which are modified through user interaction of the current element or composite widget the user is interacting with. The presence of <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element's association with the controlled elements, or that an association exists for instances where the controlled element was previously in the hidden state, but through user interaction of the controlling element, the associated element has become accessible.</p>
             <p>Instance where an <code>aria-controls</code> association could be made:</p>
             <ul>
--- a/index.html
+++ b/index.html
@@ -13094,10 +13094,16 @@ button.ariaPressed; // null</pre
             </p>
             <p>Instance where an <code>aria-controls</code> association could be made:</p>
             <ul>
-              <li>Interacting with a text field or editable combobox results in the display of a listbox popup. Upon entering text, the associated listbox is filtered, or the selected option changes to match the text value entered by the user.
+              <li>
+                Interacting with a text field or editable combobox results in the display of a listbox popup. Upon entering text, the associated listbox is filtered, or the selected option changes to
+                match the text value entered by the user.
+              </li>
               <li>A tree view representing a table of contents where choosing a treeitem updates content of a neighboring document pane.</li>
               <li>A series of checkboxes can each control what commodity prices are tracked live in a table or graph.</li>
-              <li>An interactive element reveals associated content when selected. For instance, selecting a tab control reveals its associated tab panel. Or checking a radio button reveals additional information or form controls related to the chosen radio button.</li>
+              <li>
+                An interactive element reveals associated content when selected. For instance, selecting a tab control reveals its associated tab panel. Or checking a radio button reveals additional
+                information or form controls related to the chosen radio button.
+              </li>
               <li>Radio buttons allow for filtering to a listing of search results.</li>
             </ul>
             <p>
@@ -13696,9 +13702,15 @@ button.ariaPressed; // null</pre
           <sdef>aria-expanded</sdef>
           <div class="state-description">
             <p><a>Indicates</a> whether a related element is expanded (shown) or collapsed (hidden).</p>
-            <p>The <sref>aria-expanded</sref> attribute is applied to a focusable, interactive element that toggles visibility of content of a different element. If the element with <sref>aria-expanded</sref> is also a <rref>treeitem</rref> in a <rref>tree</rref> or a <rref>row</rref> in a <rref>treegrid</rref>, then it SHOULD also be the <a>accessibility parent</a> of the content it expands and collapses. Otherwise, the element with <sref>aria-expanded</sref> SHOULD NOT be the <a>accessibility parent</a> of the content that is expanding or collapsing. Rather, identify that relationship between the interactive element and the element being controlled using <pref>aria-controls</pref>.</p>
+            <p>
+              The <sref>aria-expanded</sref> attribute is applied to a focusable, interactive element that toggles visibility of content of a different element. If the element with
+              <sref>aria-expanded</sref> is also a <rref>treeitem</rref> in a <rref>tree</rref> or a <rref>row</rref> in a <rref>treegrid</rref>, then it SHOULD also be the
+              <a>accessibility parent</a> of the content it expands and collapses. Otherwise, the element with <sref>aria-expanded</sref> SHOULD NOT be the <a>accessibility parent</a> of the content
+              that is expanding or collapsing. Rather, identify that relationship between the interactive element and the element being controlled using <pref>aria-controls</pref>.
+            </p>
             <p>For example, <sref>aria-expanded</sref> is applied to a parent <rref>treeitem</rref> to indicate whether its child branch of the tree is shown.</p>
-            <pre class="example highlight">&lt;ul role="tree"&gt;
+            <pre class="example highlight">
+&lt;ul role="tree"&gt;
   &lt;li role="treeitem" aria-expanded="false" aria-selected="false"&gt;
     &lt;span&gt;Fruits&lt;/span&gt;
     &lt;ul role="group" hidden&gt;
@@ -13707,12 +13719,15 @@ button.ariaPressed; // null</pre
       &lt;li role="treeitem" aria-selected="false"&gt;Yuzu&lt;/li&gt;
     &lt;/ul&gt;
   &lt;/li&gt;
-&lt;/ul&gt;</pre>
+&lt;/ul&gt;</pre
+            >
             <p>Similarly, it can be applied to a <rref>button</rref> to control the visibility of another element and its content on the current page.</p>
-            <pre class="example highlight">&lt;button type="button" aria-controls="mangosteen" aria-expanded="false"&gt;Mangosteen&lt;/button&gt;
+            <pre class="example highlight">
+&lt;button type="button" aria-controls="mangosteen" aria-expanded="false"&gt;Mangosteen&lt;/button&gt;
 &lt;div id="mangosteen" hidden&gt;
   An edible fruit native to tropical lands surrounding the Indian Ocean.
-&lt;/div&gt;</pre>
+&lt;/div&gt;</pre
+            >
           </div>
           <table class="def">
             <caption>
--- a/index.html
+++ b/index.html
@@ -13112,10 +13112,10 @@ button.ariaPressed; // null</pre
               <pref>aria-owns</pref>.
             </p>
             <p>
-              The <code>aria-controls</code> property is for referencing elements that are modified by the user interacting with the currently focused element or composite widget.
-              The presence of <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element’s association with the controlled element or elements. It can also inform them that an
-              association exists for instances where the controlled element was previously in the hidden state and through user interaction of the controlling element, the associated element has
-              become accessible.
+              The <code>aria-controls</code> property is for referencing elements that are modified by the user interacting with the currently focused element or composite widget. The presence of
+              <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element’s association with the controlled element or elements. It can also inform them
+              that an association exists for instances where the controlled element was previously in the hidden state and through user interaction of the controlling element, the associated element
+              has become accessible.
             </p>
             <p>Instance where an <code>aria-controls</code> association could be made:</p>
             <ul>
--- a/index.html
+++ b/index.html
@@ -13127,7 +13127,7 @@ button.ariaPressed; // null</pre
             </ul>
             <p>
               Additionally, the <code>aria-controls</code> property supports multiple ID references. For example, a control can be used to highlight different instances of spelling errors. A user
-              agent MAY convey to a user that there are a number of related controlled elements (the misspellings) and/or allow the user to navigate to them in sequence.
+              agent MAY convey to a user that there are a number of related controlled elements (the misspellings), allow the user to navigate to the controlled elements in sequence, or both.
             </p>
           </div>
           <table class="def">
--- a/index.html
+++ b/index.html
@@ -13107,13 +13107,16 @@ button.ariaPressed; // null</pre
         <div class="property" id="aria-controls">
           <pdef>aria-controls</pdef>
           <div class="property-description">
-            <p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the current element. See related <pref>aria-owns</pref>.</p>
-            <p>For example:</p>
+            <p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the current element or composite widget. See related <pref>aria-details</pref>.</p>
+            <p>The <code>aria-controls</code> property is for referencing elements which are modified through user interaction of the current element or composite widget the user is interacting with. The presense of <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element's association with the controlled elements, or that an association exists for instances where the controlled element was previously in the hidden state, but through user interaction of the controlling element, the associated element has become accessible.</p>
+            <p>Instance where an <code>aria-controls</code> association could be made:</p>
             <ul>
-              <li>A table of contents tree view can control the content of a neighboring document pane.</li>
+              <li>A tree view representing a table of contents where choosing a treeitem updates content of a neighboring document pane.</li>
               <li>A group of checkboxes can control what commodity prices are tracked live in a table or graph.</li>
               <li>A tab controls the display of its associated tab panel.</li>
+              <li>A radio button group applies filter choices to a listing of search results.</li>
             </ul>
+            <p>Additionally, the <code>aria-controls</code> property supports referring to multiple elements. For example, a button is pressed to highlight different instances of spelling errors in a document. A user agent MAY expose to a user that there are X number of controlled elements and allow for a mechanism to cylcle through those elements, and returning to the controlling element when the user chooses to do so.</p>
           </div>
           <table class="def">
             <caption>
--- a/index.html
+++ b/index.html
@@ -13116,7 +13116,7 @@ button.ariaPressed; // null</pre
               <li>A tab controls the display of its associated tab panel.</li>
               <li>A radio button group applies filter choices to a listing of search results.</li>
             </ul>
-            <p>Additionally, the <code>aria-controls</code> property supports referring to multiple elements. For example, a button is pressed to highlight different instances of spelling errors in a document. A user agent MAY expose to a user that there are X number of controlled elements and allow for a mechanism to cylcle through those elements, and returning to the controlling element when the user chooses to do so.</p>
+            <p>Additionally, the <code>aria-controls</code> property supports referring to multiple elements. For example, a button is pressed to highlight different instances of spelling errors in a document. A user agent MAY expose to a user that there are X number of controlled elements and allow for a mechanism to cycle through those elements, and returning to the controlling element when the user chooses to do so.</p>
           </div>
           <table class="def">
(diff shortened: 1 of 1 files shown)


scottaohara:
linking this issue to make sure any resolution to it is taken into account here **or** necessary updates to resolve are contained in this PR https://github.com/w3c/aria/issues/1956#issuecomment-1672079393

spectranaut:
@scottaohara did you want this to be reviewed in draft stage?

scottaohara:
@spectranaut absolutely.  i need feedback to move forward.

spectranaut:
@jnurthen thought he was waiting to review until it wasn't in draft, so here is your answer james :)

index.html
```diff
@@ -11175,13 +11175,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 		<div class="property" id="aria-controls">
 			<pdef>aria-controls</pdef>
 			<div class="property-description">
-				<p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the current element. See related <pref>aria-owns</pref>.</p>
-				<p>For example:</p>
+				<p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the current element or composite widget. See related <pref>aria-details</pref>.</p>
```
smhigley, line 11178: Am I right in reading this as saying that `aria-controls` should sometimes be added to the composite widget container role, and not to the specific interactable/focusable child element? I know it's technically allowed, my two thoughts on it are just:
1. Does it do anything in practice? (Or if it doesn't now, does that mapping exist as a pattern on native platforms & is that the pattern we want to try to convince AT to support in the future?)
2. Would it make sense to be explicit here about which element `aria-controls` should go on? It's allowed on all elements of base markup so that's not super helpful, and the new wording is a bit more ambiguous than the current wording, at least to me.

accdc (approved):

index.html
```diff
@@ -11175,13 +11175,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Instance where an <code>aria-controls</code> association could be made:</p>
 				<ul>
-					<li>A table of contents tree view can control the content of a neighboring document pane.</li>
+					<li>A tree view representing a table of contents where choosing a treeitem updates content of a neighboring document pane.</li>
 					<li>A group of checkboxes can control what commodity prices are tracked live in a table or graph.</li>
 					<li>A tab controls the display of its associated tab panel.</li>
+					<li>A radio button group applies filter choices to a listing of search results.</li>
 				</ul>
+				<p>
+				<p>Additionally, the <code>aria-controls</code> property supports referring to multiple elements. For example, a button is pressed to highlight different instances of spelling errors in a document. A user agent MAY expose to a user that there are X number of controlled elements and allow for a mechanism to cylcle through those elements, and returning to the controlling element when the user chooses to do so.</p>
```
muan, line 11188:
```suggestion
				<p>Additionally, the <code>aria-controls</code> property supports referring to multiple elements. For example, a button is pressed to highlight different instances of spelling errors in a document. A user agent MAY expose to a user that there are X number of controlled elements and allow for a mechanism to cycle through those elements, and returning to the controlling element when the user chooses to do so.</p>
```

index.html
```diff
@@ -11175,13 +11175,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 		<div class="property" id="aria-controls">
 			<pdef>aria-controls</pdef>
 			<div class="property-description">
-				<p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the current element. See related <pref>aria-owns</pref>.</p>
-				<p>For example:</p>
+				<p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the current element or composite widget. See related <pref>aria-details</pref>.</p>
+				<p>The <code>aria-controls</code> property is for referencing elements which are modified through user interaction of the current element or composite widget the user is interacting with. The presense of <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element's association with the controlled elements, or that an association exists for instances where the controlled element was previously in the hidden state, but through user interaction of the controlling element, the associated element has become accessible.</p>
```
muan, line 11179:
```suggestion
				<p>The <code>aria-controls</code> property is for referencing elements which are modified through user interaction of the current element or composite widget the user is interacting with. The presence of <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element's association with the controlled elements, or that an association exists for instances where the controlled element was previously in the hidden state, but through user interaction of the controlling element, the associated element has become accessible.</p>
```

index.html
muan: And do we mean "user agent" here or "assistive technologies"?

For example: 

> Assistive technologies SHOULD enable users to quickly navigate to elements with role banner

muan (commented):
I think my review was requested because I had voiced that I care about `aria-controls`.. My question after reading the initial issue and this PR is,... pardon my directness for a lack of better phrasing, what does this PR resolves?

Problems raised in the issue:

1. JAWS tried to support `aria-controls` but failed to provide a good experience so it was removed/hidden.
2. It's unclear if `aria-controls` is supposed provide navigation mechanism.

Both the intro to Heydon's post and developers confusion as to when and if we should use it remain:

> It's poorly supported, _does very little_, and _does what it does when it does badly_. 

* does very little: in terms of "forms an association", with unknown behavior.
* what it does badly: navigation.

---

And so with this PR

1. I don’t think this provides a clearer explanation as to how AT should implement `aria-controls`.
2. It is still debatable whether a navigation mechanism should to be implemented ("user agent MAY ... allow for a mechanism to cycle through those elements"), but that was a MAY example for when it controls multiple elements, specifically.

It seems to me from the initial issue that the most valuable use case of `aria-controls` is navigation ([NVDA issue](https://github.com/nvaccess/nvda/issues/8983), [`@carmacleod`'s comment](https://github.com/w3c/aria/issues/995#issuecomment-506903946) and Léonie's agreement, [Heydon's post suggesting "what we think it does"](https://www.heydonworks.com/article/aria-controls-is-poop)) WHEN the controlled element(s) are not the immediate siblings of the controller (or widget).

Is is possible to spell the ideal behaviors out with some *SHOULD* behaviors?

To quote @patrickhlauke's [comment](https://github.com/w3c/aria/issues/995#issuecomment-552025686):

> if it has no actual real-world benefit, and other AT have been loathe to implement any useful behavior for it...it feels like it should be removed perhaps. it's certainly not something i would advise people use, because it's more of an empty gesture at this point.

which I am afraid may still be the case after this PR.

index.html
scottaohara: yes.  thank you.  i had started to write the sentence based on what use agents should expose, but then drifted into what AT could provide.  I'll get this corrected.

index.html
scottaohara: the composite widget bit is kinda important for things like listboxes, radiogroups, menus - where focus could be on the composite widget and never reach a child element.  maybe if that's just clarified its meant to be for those use cases?

scottaohara:
Hey @muan, yeah.... I can completely understand your response to this PR, especially since I see you were added as a reviewer without any context as to the 'why'.  So, to back up a bit - this PR is not done and was meant as a starting point to get the ball rolling of ways to extend/improve the currently lacking definition into something that would provide more guidance for how AT could expose it to users.  This really should have a "not ready for merge" or "needs help" label.  The PR is not ready, nor expected to be ready for any level of merging.

The initial take I made with this PR was to incorporate some aspects of `aria-details` which have been getting traction with AT - and do some of the things that people initially thought aria-controls might do. E.g., the navigational aspect of allowing a user to navigate to/from the associated elements.

Talking with @smhigley about this some more, another way this could go would be to double down on what aria-controls does well now - and that is how it is used to associated comboboxes with their popups, or even with how a text field can be associated with a listbox for a type-ahead/filtering component.  For instance, NVDA and Narrator - when an aria-controls association is made between a searchbox and a listbox - will announce the newly selected option as the listbox options are filtered.  

Some very rough example text pertaining to the above:
>aria-controls MAY (SHOULD?) allow for AT to identify the updated content, or that the content has updated - depending on the controlled element.  For instance, a listbox filtered by a search field could announce the newly selected option based on the user filtering. Or, a user typing into a search field could dynamically update a listing of search results - where the assistive technology could inform the user that the results were updated, but not directly announcing the changed content.

So there are a couple of ways this could go.  I think it's probably important to acknowledge that any past assumptions/criticisms of the attribute need to be taken into account, but may also not be exactly what needs to be done here - especially since aria-details has been introduced and arguably serves the need that some people were looking at aria-controls to solve. But there is also potential overlap here. Or, per the direction sarah and i were discussing, we potentially forgo that overlap and go a different direction... or maybe it's a combination of the two... something else entirely?


I hope all that helps, and again i'm sorry you didn't have that context before you spent time reviewing. Really appreciate any additional thoughts you may have on this.

index.html
scottaohara, line 11178: maybe this?  
```suggestion
				<p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the focused element or composite widget. See related <pref>aria-details</pref>.</p>
```

muan:
@scottaohara No problem. I understand now. Thanks for the explanation.

jnurthen (commented):
All of these are small improvements to the prose around aria-controls without really solving any of the real problems (I know you're aware of this as you have already commented similarly in the PR comments)

In order to get something useful I think we really do need to work out how to add the suggestion that AT allow navigation to/from the controlled region(s) when useful (and how to work out when it is useful). Some language around when not to use aria-controls might also be good so we can try to avoid many of the cases which caused the noise and JAWS to disable their implementation by default.

I don't have any good suggestions as to how to do this though.

index.html
```diff
@@ -11175,13 +11175,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 		<div class="property" id="aria-controls">
 			<pdef>aria-controls</pdef>
 			<div class="property-description">
-				<p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the current element. See related <pref>aria-owns</pref>.</p>
-				<p>For example:</p>
+				<p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the focused element or composite widget. See related <pref>aria-details</pref>.</p>
```
mcking65, line 11178: This implies that the element must be focused in order to have aria-controls on it

index.html
```diff
@@ -11175,13 +11175,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Instance where an <code>aria-controls</code> association could be made:</p>
 				<ul>
-					<li>A table of contents tree view can control the content of a neighboring document pane.</li>
+					<li>A tree view representing a table of contents where choosing a treeitem updates content of a neighboring document pane.</li>
 					<li>A group of checkboxes can control what commodity prices are tracked live in a table or graph.</li>
 					<li>A tab controls the display of its associated tab panel.</li>
+					<li>A radio button group applies filter choices to a listing of search results.</li>
 				</ul>
+				<p>
+				<p>Additionally, the <code>aria-controls</code> property supports referring to multiple elements. For example, a button is pressed to highlight different instances of spelling errors in a document. A user agent MAY expose to a user that there are X number of controlled elements and allow for a mechanism to cycle through those elements, and returning to the controlling element when the user chooses to do so.</p>
```
mcking65, line 11188: Do we already have sufficient consensus on this use case? Have we sufficiently defined and discussed AT expectations?

index.html
patrickhlauke:
```suggestion
				<p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by an element or composite widget. See related <pref>aria-details</pref>.</p>
```

(though admittedly this might make it more difficult to understand

scottaohara: using 'focused' was intential here.  i'm not sure i'd want to change that.

index.html
scottaohara: nope.  this was to prompt discussion

scottaohara:
fromn today's call - the scope of this is going to be reduced to be an editorial update to expand on the current definition of aria-controls.   we _don't_ have the ability to do adequate testing for any updates to improve/change the way aria-controls should work right now, so if we move forward with this PR at all, it will be limited to just editorial description of the attribute, rather than changes to it.

spectranaut:
Link to minutes from today's meeting: https://www.w3.org/2024/01/04-aria-minutes.html#t06

index.html
cookiecrook, line 11178: I understand the desire to add `aria-details` to the related list, but it's not clear why `aria-controls` was removed. I think it should remain. 
```suggestion
				<p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the focused element or composite widget. See related <pref>aria-details</pref> and <pref>aria-owns</pref>.</p>
```

Or if there's a good reason to remove it, the reciprocal relationship (from `aria-owns` back to `aria-controls`) should be removed, too.

index.html
```diff
@@ -11175,13 +11175,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 		<div class="property" id="aria-controls">
 			<pdef>aria-controls</pdef>
 			<div class="property-description">
-				<p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the current element. See related <pref>aria-owns</pref>.</p>
-				<p>For example:</p>
+				<p><a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the focused element or composite widget. See related <pref>aria-details</pref>.</p>
+				<p>The <code>aria-controls</code> property is for referencing elements which are modified through user interaction of the current element or composite widget the user is interacting with. The presence of <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element's association with the controlled elements, or that an association exists for instances where the controlled element was previously in the hidden state, but through user interaction of the controlling element, the associated element has become accessible.</p>
```
cookiecrook, line 11179: Editorial Nit: Second sentence is run-on. Can you break it up into multiple or turn it into an example list?

index.html
cookiecrook, line 11188: Edited for brevity/clarity.

```suggestion
				<p>Additionally, the <code>aria-controls</code> property supports multiple ID references. For example, a control may be used to highlight different instances of spelling errors. A user agent MAY convey to a user that there are a number of related controlled elements (the misspellings) and/or allow the user to navigate to them in sequence.</p>
```

cookiecrook (approved):

pkra (approved):

index.html
```diff
@@ -11175,13 +11175,17 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Instance where an <code>aria-controls</code> association could be made:</p>
 				<ul>
-					<li>A table of contents tree view can control the content of a neighboring document pane.</li>
+					<li>A tree view representing a table of contents where choosing a treeitem updates content of a neighboring document pane.</li>
 					<li>A group of checkboxes can control what commodity prices are tracked live in a table or graph.</li>
 					<li>A tab controls the display of its associated tab panel.</li>
+					<li>A radio button group applies filter choices to a listing of search results.</li>
 				</ul>
+				<p>
+				<p>Additionally, the <code>aria-controls</code> property supports multiple ID references. For example, a control may be used to highlight different instances of spelling errors. A user agent MAY convey to a user that there are a number of related controlled elements (the misspellings) and/or allow the user to navigate to them in sequence.</p>
```
jnurthen, line 11188:
```suggestion
				<p>Additionally, the <code>aria-controls</code> property supports multiple ID references. For example, a control can be used to highlight different instances of spelling errors. A user agent MAY convey to a user that there are a number of related controlled elements (the misspellings) and/or allow the user to navigate to them in sequence.</p>
```

jnurthen:
@scottaohara I think we can merge this once the conflicts are resolved

scottaohara:
thanks @jnurthen 

i'll try to get through this soon

pkra:
Rebased on main.

pkra:
This disconnects the suggestion from @jnurthen at https://github.com/w3c/aria/pull/1996#discussion_r2023382643 -- sorry about that, @scottaohara, but I was hoping this still saves you some time.

pkra:
I took one liberty while rebasing: I removed an errand lonely opening `<p>` tag.

index.html
```diff
@@ -13107,13 +13107,27 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-            <p>For example:</p>
+            <p>
+              <a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the focused element or composite widget. See related <pref>aria-details</pref> and
+              <pref>aria-owns</pref>.
+            </p>
+            <p>
+              The <code>aria-controls</code> property is for referencing elements which are modified through user interaction of the current element or composite widget the user is interacting with.
+              The presence of <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element's association with the controlled elements, or that an
+              association exists for instances where the controlled element was previously in the hidden state, but through user interaction of the controlling element, the associated element has
+              become accessible.
```
scottaohara, line 13118: adding in @jnurthen's suggestion. tried simplifying the first sentence further.
```suggestion
              The <code>aria-controls</code> property is for referencing elements that are modified by the user interacting with the currently focused element or composite widget.
              The presence of <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element’s association with the controlled element or elements. It can also inform them that an
              association exists for instances where the controlled element was previously in the hidden state and through user interaction of the controlling element, the associated element has
              become accessible.
```

index.html
```diff
@@ -13107,13 +13107,27 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+              The <code>aria-controls</code> property is for referencing elements that are modified by the user interacting with the currently focused element or composite widget.
+              The presence of <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element’s association with the controlled element or elements. It can also inform them that an
+              association exists for instances where the controlled element was previously in the hidden state and through user interaction of the controlling element, the associated element has
+              become accessible.
+            </p>
+            <p>Instance where an <code>aria-controls</code> association could be made:</p>
             <ul>
-              <li>A table of contents tree view can control the content of a neighboring document pane.</li>
+              <li>A tree view representing a table of contents where choosing a treeitem updates content of a neighboring document pane.</li>
               <li>A group of checkboxes can control what commodity prices are tracked live in a table or graph.</li>
```
scottaohara, line 13123:
```suggestion
              <li>A series of checkboxes can each control what commodity prices are tracked live in a table or graph.</li>
```

index.html
```diff
@@ -13107,13 +13107,27 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
               <li>A tab controls the display of its associated tab panel.</li>
+              <li>A radio button group applies filter choices to a listing of search results.</li>
```
scottaohara, line 13125:
```suggestion
              <li>Radio buttons allow for filtering to a listing of search results.</li>
```

index.html
```diff
@@ -13107,13 +13107,27 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+              The presence of <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element’s association with the controlled element or elements. It can also inform them that an
+              association exists for instances where the controlled element was previously in the hidden state and through user interaction of the controlling element, the associated element has
+              become accessible.
+            </p>
+            <p>Instance where an <code>aria-controls</code> association could be made:</p>
             <ul>
-              <li>A table of contents tree view can control the content of a neighboring document pane.</li>
+              <li>A tree view representing a table of contents where choosing a treeitem updates content of a neighboring document pane.</li>
               <li>A group of checkboxes can control what commodity prices are tracked live in a table or graph.</li>
               <li>A tab controls the display of its associated tab panel.</li>
```
scottaohara, line 13124: as there is still mixed feelings about adding aria-expanded to the radio role - maybe this example better plays to what https://github.com/w3c/aria/issues/1404 is asking for.  e.g., since the radio cannot have its state toggled without making another selection - the expanded state is a bit awkward.  But, like the tab to tabpanel association example referenced here, that's arguably what people have been looking for - a programmatic way to convey that other content has rendered due to selection (or checking in the radio's case)

```suggestion
              <li>An interactive element reveals associated content when selected. For instance, selecting a tab control reveals its associated tab panel. Or checking a radio button reveals additional information or form controls related to the chosen radio button.</li>
```

index.html
```diff
@@ -13107,13 +13107,27 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+            <p>
+              The <code>aria-controls</code> property is for referencing elements that are modified by the user interacting with the currently focused element or composite widget.
+              The presence of <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element’s association with the controlled element or elements. It can also inform them that an
+              association exists for instances where the controlled element was previously in the hidden state and through user interaction of the controlling element, the associated element has
+              become accessible.
+            </p>
+            <p>Instance where an <code>aria-controls</code> association could be made:</p>
             <ul>
-              <li>A table of contents tree view can control the content of a neighboring document pane.</li>
+              <li>A tree view representing a table of contents where choosing a treeitem updates content of a neighboring document pane.</li>
```
scottaohara, line 13122:
```suggestion
              <li>Interacting with a text field or editable combobox results in the display of a listbox popup. Upon entering text, the associated listbox is filtered, or the selected option changes to match the text value entered by the user.
              <li>A tree view representing a table of contents where choosing a treeitem updates content of a neighboring document pane.</li>
```

index.html
```diff
@@ -13107,13 +13107,27 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-            <p>For example:</p>
+            <p>
+              <a>Identifies</a> the <a>element</a> (or elements) whose contents or presence are controlled by the focused element or composite widget. See related <pref>aria-details</pref> and
+              <pref>aria-owns</pref>.
+            </p>
+            <p>
+              The <code>aria-controls</code> property is for referencing elements that are modified by the user interacting with the currently focused element or composite widget. The presence of
+              <code>aria-controls</code> enables <a>assistive technologies</a> to make users aware of the current element’s association with the controlled element or elements. It can also inform them
+              that an association exists for instances where the controlled element was previously in the hidden state and through user interaction of the controlling element, the associated element
+              has become accessible.
```
scottaohara, line 13118:
```suggestion
              The <code>aria-controls</code> property is for referencing elements that are modified by the user interacting with the currently focused element or composite widget. The presence of
              <code>aria-controls</code> enables <a>assistive technologies</a> to programmatically associate the currently focused element with the element or elements it controls.  For instance, it can be used to inform users that by interacting with the controlling element they have revealed an element or elements that were previously in the hidden state. Or, by interacting with an element, they caused the selection or value of a controlled element to change.
```

scottaohara:
@pkra, @jnurthen, @cookiecrook being the 3 approving reviewers, i have made further updates based on your feedback / lingering updates that I've been ridiculously sitting on for the past year and half...  

i have not committed those updates as I want to make it clear exactly what I'm proposing to change vs what you have previously approved.  

if you're fine with the proposed changes, then we can commit and merge this.

if you do not care for the further clarifications/updates I wanted to add, then we can leave those out and merge what you had previously approved.

thanks

pkra:
I've left a :+1: on each suggestion to re-confirm my review.

PR #2027, opened 2023-09-05
Editorial: Updated accessibility children example 44 to clarify accessibility children exclusions

### LINKED ISSUE #2021: Clarification about accessibility children definition (2nd exclusion point)
**Concern description:**

In the process of reviewing the definition of "accessibility children," I undertook a comparison with the definition of "accessibility parent". While the outlined points are consistent for both concepts, there are disparities in the exclusions.

The second exclusion for "accessibility children" is somewhat intricate to me. We define "accessibility children" as:

> - The DOM children of the [element](https://dom.spec.whatwg.org/#concept-element).
> - All DOM descendants of the [element](https://dom.spec.whatwg.org/#concept-element) with only elements of role [generic](https://w3c.github.io/aria/#generic) or [none](https://w3c.github.io/aria/#none) intervening.
> - All DOM elements specified via an [aria-owns](https://w3c.github.io/aria/#aria-owns) relationship to the element.
> - All DOM descendants of an element with role [generic](https://w3c.github.io/aria/#generic) or [none](https://w3c.github.io/aria/#none) specified via [aria-owns](https://w3c.github.io/aria/#aria-owns) with only elements of role [generic](https://w3c.github.io/aria/#generic) or [none](https://w3c.github.io/aria/#none) intervening.

However, we then proceed to exclude:

> - All DOM elements whose corresponding [accessible object](https://w3c.github.io/aria/#dfn-accessible-object) have been reparented in the [accessibility tree](https://w3c.github.io/aria/#dfn-accessibility-tree) via [aria-owns](https://w3c.github.io/aria/#aria-owns)

This exclusion seems partially in contrast with points #3 and #4 of the definition above. Consider the following example:

```
<div role="list">
  <div role="listitem" aria-hidden="true">Excluded element</div>
  <div role="listitem" id="reparented">Reparented element</div>
</div>
<div aria-owns="reparented" role="list"></div>
```

According to the current definition, and please correct me if I'm mistaken, we seem to assert that the reparented list item, which, in reality, is an accessibility child of the second list with aria-owns="reparented," is an exclusion and not considered an "accessibility child" of it.

In my view, a potential resolution could involve rewording the definition to provide a sequenced set of steps to determine what qualifies as an accessibility child. This would ensure that the aria-owns relationship takes precedence over the DOM structure. For instance:

1. If the DOM element is referenced by an aria-owns attribute, then the element is an accessibility child of the owner element (I know I'm simplifying but it's just to propose an idea);
2. Otherwise, if ...

**Link to documentation:**
https://w3c.github.io/aria/#dfn-accessibility-child

giacomo-petri:
Closes #2021

This PR aims to improve clarity regarding the 2nd exclusion related to the "accessibility children" definition

> All DOM elements whose corresponding [accessible object](https://w3c.github.io/aria/#dfn-accessible-object) have been reparented in the [accessibility tree](https://w3c.github.io/aria/#dfn-accessibility-tree) via [aria-owns](https://w3c.github.io/aria/#aria-owns).

by updating the relevant example, and explicitly confirming that the second list indeed contains one accessibility child.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 7, 2023, 5:27 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fgiacomo-petri%2Faria%2Fa015c7faea5f5181dbadc4175e286719590944cc%2Findex.html%3FisPreview%3Dtrue)

```

Specification: http://labs.w3.org/spec-generator/uploads/8f0tnv/index.html?isPreview=true%3FisPreview%3Dtrue&publishDate=2023-09-07
ReSpec version: 34.1.6
File a bug: https://github.com/w3c/respec/
Error: undefined

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria%232027.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -13505,7 +13505,7 @@ button.ariaPressed; // null</pre>
   &lt;div role="listitem"&gt;Accessibility Child 4&lt;/div&gt;
 &lt;/div&gt;
           </pre>
-          <p>In the following example, the first <rref>list</rref> element has no accessibility children, where as the second <rref>list</rref> element has one accessibility child, specifically the listitem with ID value "reparented".</p>
+          <p>In the following example, the first <rref>list</rref> element has no accessibility children, where as the second <rref>list</rref> element has one accessibility child, specifically the <rref>listitem</rref> with ID value "reparented".</p>
 	  <pre class="example highlight">
 &lt;div role="list"&gt;
   &lt;div role="listitem" aria-hidden="true"&gt;Excluded element&lt;/div&gt;
--- a/index.html
+++ b/index.html
@@ -13505,7 +13505,7 @@ button.ariaPressed; // null</pre>
   &lt;div role="listitem"&gt;Accessibility Child 4&lt;/div&gt;
 &lt;/div&gt;
           </pre>
-          <p>In the following example, the <rref>list</rref> element has no accessibility children:</p>
+          <p>In the following example, the first <rref>list</rref> element has no accessibility children, where as the second <rref>list</rref> element has one accessibility child, specifically the listitem with ID value "reparented".</p>
 	  <pre class="example highlight">
 &lt;div role="list"&gt;
   &lt;div role="listitem" aria-hidden="true"&gt;Excluded element&lt;/div&gt;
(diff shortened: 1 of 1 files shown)


spectranaut (changes requested):
thank you for this!! :)

Just one small fix.

pkra (approved):
Thank you!

scottaohara (approved):

spectranaut (approved):

PR #2037, opened 2023-09-13
Updates to aria-hidden

MelSumner:
If merged, this PR reflects discussions around how `aria-hidden` could be improved to better serve end-users. It provides improved author and user agent guidance.

The text content of this PR change were discussed during TPAC meetings and concept was drafted by @aleventhal and me. 

# PR tracking

* [ ] Related APG Issue/PR:

# Implementation tracking

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:


index.html
```diff
@@ -11867,7 +11867,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p>Authors MUST NOT place focusable elements within <code>aria-hidden</code> content. This is because the AT will have no information about the content, so it cannot be presented to the user with AT. User agents MUST repair this situation by ignoring the <code>aria-hidden</code> attribute on the ancestor of the focused element.</p>
+				<p>Authors MUST NOT place focusable elements within <code>aria-hidden</code> content. This is because the AT will have no information about the content, so it cannot be presented to the user with AT. User agents MUST repair this situation by ignoring the <code>aria-hidden</code> attribute on the ancestor of the focused element. When a repair is made, user agents should inform authors through developer tools (if available).</p>
```
scottaohara: uppercase to should.
remove awkward/recursive second mention of AT from the "this is because the at..." sentence.

```suggestion
				<p>Authors MUST NOT place focusable elements within <code>aria-hidden</code> content. This is because the AT will have no information about the content, so it cannot be presented to the user. User agents MUST repair this situation by ignoring the <code>aria-hidden</code> attribute on the ancestor of the focused element. When a repair is made, user agents SHOULD inform authors through developer tools (if available).</p>
```

index.html
```diff
@@ -11861,9 +11861,14 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p class="note">Authors are advised to use extreme caution and consider a wide range of disabilities when hiding visibly rendered content from assistive technologies. For example, a sighted, dexterity-impaired individual might use voice-controlled assistive technologies to access a visual interface. If an author hides visible link text "Go to checkout" and exposes similar, yet non-identical link text "Check out now" to the accessibility API, the user might be unable to access the interface they perceive using voice control. Similar problems can also arise for screen reader users. For example, a sighted telephone support technician might attempt to have the blind screen reader user click the "Go to checkout" link, which they might be unable to find using a type-ahead item search ("Go to…").</p>
```
aleventhal, line 11872: Maybe we could improve this paragraph. I find it cumbersome to understand everything but the first 7 words.

Authors are advised to use extreme caution with aria-hidden. Misuse easily and often leads to inaccessible content. Starting with ARIA 1.3, several mitigations have been added to reduce the likelihood of this. However, aria-hidden is too powerful to make any mistakes with: many users have no way to perceive content that uses it, and therefore authors MUST avoid using aria-hidden on any content required for successful user journeys.

aleventhal (changes requested):

index.html
```diff
@@ -11861,9 +11861,14 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 		<div class="state" id="aria-hidden">
 			<sdef>aria-hidden</sdef>
 			<div class="state-description">
-				<p><a>Indicates</a> whether the <a>element</a> is exposed to an accessibility API. See related <sref>aria-disabled</sref>.</p>
+				<p><a>Indicates</a> whether the <a>element</a> can be included in the accessible tree as a visible accessibility node.</p>
```
scottaohara, line 11864: not an objection, but not sure i understand the changing of this line.   what is the difference between being included in the a11y tree (or a11y api) vs being included as a "visible" node.  are these different things, or is this redundancy?

aleventhal: We want the freedom, on platforms that can express visibility as a boolean state (non-Apple platforms) to use that rather than pruning. This gives some power back to the AT to repair.

index.html
```diff
@@ -11861,9 +11861,14 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>While authors do not need to use <code>aria-hidden</code> when visibly hiding content with CSS, </code>authors MAY, with caution, use <code>aria-hidden="true"</code> to hide visibly rendered content from assistive technologies. However, authors should <em>only</em> do so if the act of hiding this content is intended to improve the experience for users of assistive technologies (e.g., to remove redundant or extraneous content). Authors using <code>aria-hidden="true"</code> to hide visible content from screen readers MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.</p>
+				<p>User agents MUST either prune <code>aria-hidden</code> nodes from the accessibility tree, or mark all <code>aria-hidden</code> nodes as hidden.</p>
```
michaelDCurran, line 11867: I'm assuming this paragraph is referring to aria-hidden nodes plus their descendants? If so I think it would be good to clearly call out that all descendants of an aria-hidden node need to be marked as hidden, not just the root node itself.

aleventhal: Yes, correct. The value of aria-hidden (both true and false) inherits into descendants. An aria-hidden node is one where aria-hidden=true was set on the node itself, or it was inherited by a descendant.

ggordon-vispero:
Great to see that   the aria-hidden problem has gotten some traction. JAWS users have been prevented from accessing key content because of aria-hidden misuse. Can't wait to be able to try this change in canary builds of Chrome.

index.html
scottaohara: yeh i get that.  i dunno, maybe i'll have a better way of framing the question tomorrow or something, i've been trying to re-ask the question for far too long now...

scottaohara: i _get_ that the end goal here is to have more control over the 'visible' state of the node in the a11y tree.  i question if this is the right way to introduce the attribute's purpose, particularly when the term 'visible' (and variants) are heavily used for whether something is visually rendered or not.

or another way to think about it, is aria-hidden identifying what nodes are visible in the a11y tree, or is it used to hide nodes (which can be re-made visible)?  or is it being used to mark nodes as hidden OR visible (aria-hidden=false).

i hope that makes more sense?  i'm not sure how else to describe this in text right now.  the previous sentence made sense, this one doesn't (right now? maybe more changes are coming - it is a draft afer all) because it's unclear what the changing in words means since the rest of the text talks about hiding from the a11y tree.

index.html
```diff
@@ -11861,9 +11861,14 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p>Authors MAY, with caution, use aria-hidden to hide visibly rendered content from assistive technologies <em>only</em> if the act of hiding this content is intended to improve the experience for users of assistive technologies by removing redundant or extraneous content. Authors using aria-hidden to hide visible content from screen readers MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.</p>
+				<p>While authors do not need to use <code>aria-hidden</code> when visibly hiding content with CSS, </code>authors MAY, with caution, use <code>aria-hidden="true"</code> to hide visibly rendered content from assistive technologies. However, authors should <em>only</em> do so if the act of hiding this content is intended to improve the experience for users of assistive technologies (e.g., to remove redundant or extraneous content). Authors using <code>aria-hidden="true"</code> to hide visible content from screen readers MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.</p>
```
scottaohara, line 11866: >While authors do not need to use <code>aria-hidden</code> when visibly hiding content with CSS

this should probably be more specific of where this is true, and/or allude to the fact that not all methods of visually hiding content with css will result in the content also being hidden from the accessibility tree.

index.html
```diff
@@ -11861,9 +11861,14 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Authors MUST NOT place focusable elements within <code>aria-hidden</code> content. This is because the AT will have no information about the content, so it cannot be presented to the user with AT. User agents MUST repair this situation by ignoring the <code>aria-hidden</code> attribute on the ancestor of the focused element from that point forward. When this repair is made, user agents SHOULD inform authors through developer tools (if available).</p>
+				<p>Authors MUST NOT use <code>aria-hidden="true"</code> on the document element or body of the main document (e.g., <code>window.top</code>), where that would hide the entire accessibility tree. User agents MUST repair this situation by ignoring <code>aria-hidden="true"</code> when it would prevent all visible content from being available to ATs. Again, when this repair is made, user agents SHOULD inform authors through developer tools (if available). Notably, this situation does not occur when <code>aria-hidden="false"</code> is provided on a visible descendant, e.g. on a <code>role="dialog"</code>element, because that would generate some accessible content.</p>
```
jcsteh, line 11871: This raises a couple of mutation edge cases that I think need to be clearly specified:

1. aria-hidden="true" is set on the document, aria-hidden="false" is set only on a descendant dialog. Later, the dialog is removed. Should the UA now ignore aria-hidden="true" on the document, even though it wasn't ignoring it before?
2. aria-hidden="true" is set on the document, no descendants have aria-hidden="false". The UA ignores aria-hidden on the document. Later, a diaog with aria-hidden="false" is added. Should the UA now stop ignoring aria-hidden on the document so that anything other than the dialog disappears?

If the answer to both of these is "yes", we might be able to cover that with a single sentence like "This repair should be re-evaluated whenever aria-hidden="false" descendants are added or removed from the document."

jcsteh (changes requested):

index.html
aleventhal: Jamie, for #1, I would say that this is a realistic scenario that occurs. Basically, the author is trying to do a modal dialog and forgets to remove aria-hidden on the document. Since it's a realistic scenario, I think we should ask UAs to track that scenario and now ignore the aria-hidden. 

For #2, once any aria-hidden is ignored on a given node for any of the repair reasons, it can just be ignored from there on out. Personally I think that makes implementation simpler in that we can keep a list of bad aria-hidden roots. Also, once a repair occurs, a warning of some kind should be emitted through developer tools -- this could take a number of different forms, where a console warning is the simplest.

index.html
```diff
@@ -11861,9 +11861,14 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>User agents MUST either prune <code>aria-hidden</code> nodes from the accessibility tree, or mark all <code>aria-hidden</code> nodes as hidden.</p>
+				<p>Authors MAY use <code>aria-hidden="false"</code> within an <code>aria-hidden="true"</code> subtree to unhide the visibly rendered content within the <code>aria-hidden="false"</code> subtree.</p>
+				<p class="note">This is a change from ARIA 1.2 and earlier versions in several ways: 1) <code>aria-hidden="false"</code> no longer un-hides content that is also hidden in CSS, as there are other techniques to present content only to AT users that works more consistently (e.g. positioning offscreen), and 2) <code>aria-hidden="false"</code> previously was a <code>noop</code> on (e.g., did not affect) visible content, and 3) <code>aria-hidden="false"</code> previously operated only on a single element. However, the <code>aria-hidden="false"</code> behavior of previous versions was never consistently implemented, understood or commonly used.</p>
```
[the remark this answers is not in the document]
aleventhal: Ok, thank you. We should fix this @MelSumner.

index.html
aleventhal: These are good questions and will lead to more clarity. We would love your help with specific rewording suggestions to get there. I think we should start with what it does for the author, and then later discuss how the UA should apply that.

index.html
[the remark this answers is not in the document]
aleventhal: Maybe we should chat about this one. The platforms that stand out are Mac/iOS, which don't provide a means of marking a node hidden. They've discussed it in the past and wish to remain an outlier rather than make this change, which would also involve changing their ATs. @cookiecrook, given James' concerns, is there a chance to reconsider?

Jamie, I was leaning toward it being a UA choice -- but if Firefox were to do the same thing, I suppose we could make it a platform choice based on availability of the hidden state. For Chrome, we would like to mark hidden on all platforms other than Apple platforms right now, because this gives us the best experience everywhere. This is what we think is best for our users. I understand what you're saying about interop, but on platforms that already have a hidden state, ATs already that by skipping those nodes just like the obviously must do for nodes that are there. But maybe you have a specific example of what might not be good for interop — I'd be interested to know if the problem created is bigger than the improvement we get by allowing ATs to do repairs on bad content.

I don't think hit testing should return hidden nodes.

Regarding responsibility of repairs all being on the UA, we don't agree about this. I think we should pass some responsibility to ATs, who are more nimble and better understand their users and use cases, and can apply more logic. ATs are never going to work exactly the same. So I guess we just disagree on this point. @ggordon-vispero has convinced me that they should have a chance at repairing content when their phone lines are lighting up about it, and they can't get traction anywhere else. Not every user is going to run with a bunch of bookmarklets or extensions to fix things.

index.html
```diff
@@ -11861,9 +11861,14 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Authors MUST NOT place focusable elements within <code>aria-hidden</code> content. This is because the AT will have no information about the content, so it cannot be presented to the user with AT. User agents MUST repair this situation by ignoring the <code>aria-hidden</code> attribute on the ancestor of the focused element from that point forward. When this repair is made, user agents SHOULD inform authors through developer tools (if available).</p>
```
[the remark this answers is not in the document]
aleventhal: I think the wording got messed up on this one. It's intended to say that Authors MUST not allow elements within <code>aria-hidden</code> content to obtain focus, because there is more than one way of doing that, right? And it makes it clear that scrollable focusable areas are included.

For #2, it is meant to apply to elements that actually ended up receiving focus. At that point forward, aria-hidden="true" will be ignored on the entire subtree that has it from that point forward, even if the element within it loses focus. I agree it's not going to solve everything, but that's why we have multiple mitigations. There are definitely scenarios where something gets focus when the user has a screen reader and is navigating with a virtual buffer (e.g. the author places it there because the user clicked an open dialog button).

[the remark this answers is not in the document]
aleventhal: @MelSumner in our initial draft, I have different text. I have:

Authors MUST prevent elements from receiving focus within aria-hidden content, because when AT users reach this content, the AT can provide the user with no information about the content, leading to a disorienting and inaccessible experience. User agents MUST repair this situation by ignoring aria-hidden on the ancestor of the focused element.

I think it's better because it may be very difficult to avoid having focusable elements. But if the user can't actually focus the element via input means, maybe that's good enough. E.g. it's not tabbable, not clickable and hidden from AT, that prevents it from being focused.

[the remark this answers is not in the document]
aleventhal: Thinking about it more, I guess it depends what is meant by focusable. I would thing that an <input> will be considered focusable, but if I have code that prevents it from being focused, or redirects focus if it does occur — either seems fine, right? I don't think we want to ask authors to have to go through their aria-hidden content and apply tabindex=-1. Interested in other opinions on this wording.

index.html
tranjocelyn: I actually liked the examples here! Would prefer if they were kept.

index.html
scottaohara: re: our talk today and looping in @MelSumner - i'm going to noodle on some ways to reword this and will submit that suggestion(s) asap

tranjocelyn:
Can examples of good and bad use cases be included? I think that would be helpful for authors.

index.html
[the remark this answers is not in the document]
jcsteh: Making it a UA choice is a huge interop issue. Certain things (like how we render video controls or a date picker) are reasonably UA choices because they don't have a material impact on what a user can access that the site provides, even if a user might prefer one over the other. The different behaviour is surprising and confusing for both authors and users.

Especially having done this from both sides, I do understand the pragmatic need for repair, but we should still be doing our utmost to maximise interop here. Having significant differences in terms of what a user can access depending on what combination of software they use in multiple different dimensions doesn't really help anyone. This is why I'm more in favour of browser repairs here. That said, if we can get agreement that this is desirable across the board, we don't have inconsistencies across platforms and we iron out hit testing behaviour, etc., I can live with this.

If hit testing doesn't return hidden nodes, that solves one problem - an AT can easily get to visible content - but it also creates another. Your AT repair use case is now biased towards screen reader users who use a keyboard, where hit testing doesn't matter much. What about low vision screen reader users who use a touch screen or mouse? There's no way they could ever access this hidden content, repair or not, because explore by touch and mouse tracking require hit testing. The AT can't even work around that because the AT has no idea of visual layers created by z-order, stacking contexts, etc.

Finally, do Android and UI Automation have a way to expose hidden nodes in the tree? I didn't think they did.

index.html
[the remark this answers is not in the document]
jcsteh: Ah, I misunderstood the intent here. The repair is only intended to apply to focused elements, so the text is correct there. However, the mention of focusable elements earlier (which is effectively author requirement, not UA repair) confused me.

Perhaps we could say:

`If an element inside an aria-hidden ancestor receives focus, user agents MUST repair this situation by ignoring the <code>aria-hidden</code> attribute on the ancestor of the focused element from that point forward.`

I didn't think of the "open dialog" style example. Thanks for that. As a side note, this is one reason I'd really like to see solid use cases being included alongside ARIA spec changes (probably not in the spec itself, but at least in the PR). This makes it a lot easier as a reviewer to understand why a change is being made and would avoid a lot of misunderstandings like this. It would also make it a lot easier when looking at historical changes to figure out why the change was made.

index.html
jcsteh: Fair enough. In that case, the spec wording needs to be clarified accordingly. Something like this:

`User agents MUST repair this situation by ignoring <code>aria-hidden="true"</code> when it would prevent all visible content from being available to ATs. This repair should be re-evaluated whenever a descendant with <code>aria-hidden="false"></code> is removed. Once <code>aria-hidden="true"></code> is ignored on an element, it should remain ignored henceforth, regardless of changes to descendants.`

jcsteh (changes requested):

index.html
MelSumner: > While authors do not typically need to use `aria-hidden` when using CSS methods like `display: none` to visibly hide content...

@scottaohara would that address your concern?

index.html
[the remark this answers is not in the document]
aleventhal: UI Automation only has a hidden text attribute, looks like: UIA_IsHiddenAttributeId. Possibly not enough, I'll have to look into it.

I should note Chrome actually already uses STATE_SYSTEM_INVISIBLE and hidden:true on focusable nodes inside of aria-hidden, and has done so for a while. I'm not aware of it causing interoperability issues. We don't expose any other nodes in aria-hidden subtrees currently. If there are demonstrable issues with doing it then we would reconsider this part of the proposal.

The reason I like this mitigation overall is that the other mitigations aren't enough, say if the user is visiting via virtual buffer and not focusing the aria-hidden content, aria-hidden=false wasn't used, and it wasn't on the body or html. Those other mitigations are great but it won't get everything, and there are some mitigations that aren't appropriate in the browser. For example, some screen readers probably want to allow scripts or settings files to add aria-hidden content back to the virtual buffer by listing specific bad urls. I don't think it's a big interoperability issue so much, because the default is still to not put the aria-hidden content into the virtual buffer. So the default will essentially the same between screen readers -- to not present it. It's true I'm most concerned about screen readers and I'm glad for your help to find potential issues in other ATs.

@ggordon-vispero would appeciate your thoughts on potential uses for the aria-hidden nodes if you are willing to share, and on any potential interoperability issues.

index.html
BenDMyers: Maybe "methods like `display: none` or `visibility: hidden`"?

WilcoFiers, line 11866: Instead of "screen readers" use "assistive technologies". These are not interchangeable terms.

index.html
```diff
@@ -11861,9 +11861,14 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>While authors do not need to use <code>aria-hidden</code> when visibly hiding content with CSS, </code>authors MAY, with caution, use <code>aria-hidden="true"</code> to hide visibly rendered content from assistive technologies. However, authors should <em>only</em> do so if the act of hiding this content is intended to improve the experience for users of assistive technologies (e.g., to remove redundant or extraneous content). Authors using <code>aria-hidden="true"</code> to hide visible content from screen readers MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.</p>
+				<p>User agents MUST either prune <code>aria-hidden</code> nodes from the accessibility tree, or mark all <code>aria-hidden</code> nodes as hidden.</p>
+				<p>Authors MAY use <code>aria-hidden="false"</code> within an <code>aria-hidden="true"</code> subtree to unhide the visibly rendered content within the <code>aria-hidden="false"</code> subtree.</p>
```
WilcoFiers, line 11868: I don't think this is a good idea. It has become incredibly common for devs to put things like `aria-hidden={!open}` in their code. Just look at a quick search for aria-hidden on GitHub:
https://github.com/search?q=aria-hidden%3D%7B+language:JavaScript&type=code&l=JavaScript

It is a rarity that devs set aria-hidden to null. That doesn't suggest to me that it is a common misunderstanding that putting aria-hidden="false" in an element with aria-hidden="true" will unhide the element. This is relied on a in a lot of places. Now not all of these might be real problems, since CSS might be at play there too, but assuming even 10% of those cases actually required aria-hidden, that is a lot of components that break because they can no longer be hidden with aria-hidden="true".

Very strong :-1: from me on this change.

WilcoFiers (changes requested):

index.html
WilcoFiers: I think if you wanted an option to unhide with `aria-hidden`, a new value should be introduced like `aria-hidden="reset"`, something explicit like that. Even done that way I think is very worrying. While that does add a new feature that may have some uses, it also removes the capability for a node to completely hide itself with aria-hidden="true". There is also some worrying interaction with inert, which cannot be unset and may result in inactive components that are still in the accessibility tree.

WilcoFiers: > This would only have any affect in a subtree that's both already aria-hidden="true" (marked on an ancestor) and visible to sighted users. Does that help?

That's how most accessible modals work: aria-hidden="true" on the container wrapping all non-modal content. It is essential for modals that aria-hidden="true" hides everything, and that stuff doesn't "leak" out because a dev on some third party component library just happened to put `aria-hidden={!open}` instead of `aria-hidden={!open ? true : null}`.

smhigley: I agree with @WilcoFiers, this change would have potentially huge consequences — I personally would need to go have a mild panic about all the places this could potentially break our UI 😅.

The caveat that this would only apply to nodes that are visible to sighted users doesn’t help at all — all our modal background content is still technically visible.

alice:
This issue caught my attention as it floated by my inbox. I felt like I might have some useful thoughts, given my earlier work on `inert` and involvement in implementing other accessibility tree pruning mechanisms like `aria-modal`.

I tried to get my head around everything that's happening here, and it seems like there are several fairly distinct (albeit related) changes bundled together:
1. A change in the _semantics_ of `aria-hidden`, from (generally) accessibility tree pruning behaviour to adding a "hidden" state (where available);
2. Adding guidance for authors around interactions between `aria-hidden` and CSS properties which make content non-rendered;
3. Adding normative requirements for UAs to "repair" when elements which are ARIA hidden are focused;
4. Adding normative requirements for UAs to ignore `aria-hidden` on a root element under certain circumstances, with a caveat added by the point below; and
5. Adding an entirely new mechanism to use `aria-hidden=false` to "un-hide" content.

That's quite a lot of substantial and potentially controversial changes to put in one PR. 

There's already been a fair amount of conversation on several of these points. I also have thoughts on several, as well as on the actual wording that’s been proposed, but I'd like to especially highlight a [comment](https://github.com/w3c/aria/pull/2037#discussion_r1326741684) from @jcsteh:

> this is one reason I'd really like to see solid use cases being included alongside ARIA spec changes (probably not in the spec itself, but at least in the PR). This makes it a lot easier as a reviewer to understand why a change is being made and would avoid a lot of misunderstandings like this. It would also make it a lot easier when looking at historical changes to figure out why the change was made.

I spent ... longer than I should probably admit trying to understand what changes are being proposed here and, more importantly, what problems they're trying to solve. Some of these changes are substantial enough that they could use a mini-[explainer](https://tag.w3.org/explainers/). At a minimum, there should be some top-level explanation of the problems being solved, and links out to previous discussion in github threads and minuted meetings. (It needn't be a separate document necessarily, but at least a substantial PR description.)

I realise this is a drive-by comment, but is it too late to split this change out into discrete PRs? Those could add the necessary context which would help avoid the conversation becoming overwhelming, help the less controversial aspects getting blocked, and with finding agreement on the more controversial ones.

index.html
smhigley: Another consideration -- what if the ignored `aria-hidden` attribute is removed, then re-added to the same node? Does it continue to be ignored, or does that reset it?

smhigley:
I'd like to +1 @alice's comment about this being quite a lot of pretty independent and substantial changes being bundled together in one PR. Would it be at all possible to split them out into a few separate PRs? The UA repair guidance seems the least controversial, and I'd hate to see it held up by other controversial changes that it doesn't depend upon.

I was also wondering if the removed vs. hidden distinction might make more sense to spec out in Core AAM, since it seems more like a mapping/implementation detail and not related to the intent & functionality of the attribute. Then perhaps we could update the ARIA spec to use more generic language instead of "prune from the accessibility tree".

aleventhal:
Makes sense. We shall break it up when there's time again!

MelSumner:
@aleventhal do we want to make some time to go through this?

aleventhal:
Yes, I think @Scott O'Hara ***@***.***> wanted to work on it
too.

scottaohara:
there is the PR for adjusting expectations for aria-hidden=false - which was done to resolve a related issue(s).  But it definitely has an impact on where this PR was headed.

PR #2041, opened 2023-09-17
Add statemment about translating in keyshortcuts

### LINKED ISSUE #1910: Consider adding aria-keyshortcuts to the list of Translatable States and Properties
# Describe your concern
Since keyboards may be localized, it seems that the aria-keyshortcuts attribute should be added to the list of [Translatable States and Properties](https://www.w3.org/TR/wai-aria-1.2/#translatable-states-and-properties). Adding it to the list may make it more apparent that the string value for the aria-keyshortcuts attribute should also be localized.
Link to documentation:
[ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/)

Does the issue exist in [the editor's draft](https://w3c.github.io/aria/) (the editor's draft is the most recent draft of the specification)?
Yes, this issue also exists in the editor's draft

jnurthen:
Fixes #1910

Add statemment about translating in keyshortcuts

Not editorial but doesn't require any implementation

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2041.html" title="Last updated on Sep 29, 2023, 11:06 PM UTC (941e380)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2041/bb56584...941e380.html" title="Last updated on Sep 29, 2023, 11:06 PM UTC (941e380)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -12007,7 +12007,7 @@ button.ariaPressed; // null</pre>
 				<p>User agents MUST NOT change keyboard behavior in response to the <code>aria-keyshortcuts</code> attribute. Authors MUST handle scripted keyboard events to process <code>aria-keyshortcuts</code>. The <code>aria-keyshortcuts</code> attribute exposes the existence of these shortcuts so that assistive technologies can communicate this information to users.</p>
 				<p>Authors SHOULD provide a way to expose keyboard shortcuts so that all users can discover them, such as through the use of a tooltip. Authors MUST ensure that <code>aria-keyshortcuts</code> applied to disabled elements are unavailable.</p>
 				<p>Authors SHOULD avoid implementing shortcut keys that inhibit operating system, user agent, or assistive technology functionality. This requires the author to carefully consider both which keys to assign and the contexts and conditions in which the keys are available to the user. For guidance, see the keyboard shortcuts section of the <cite><a href="" class="practices"><abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> Authoring Practices</a></cite>.</p>
-				<p>If the page is translated authors SHOULD consider whether the keyboard shortcut is valid in the translated language or whether it needs to be changed.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -12007,6 +12007,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p>Authors SHOULD provide a way to expose keyboard shortcuts so that all users can discover them, such as through the use of a tooltip. Authors MUST ensure that <code>aria-keyshortcuts</code> applied to disabled elements are unavailable.</p>
 				<p>Authors SHOULD avoid implementing shortcut keys that inhibit operating system, user agent, or assistive technology functionality. This requires the author to carefully consider both which keys to assign and the contexts and conditions in which the keys are available to the user. For guidance, see the keyboard shortcuts section of the <cite><a href="" class="practices"><abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> Authoring Practices</a></cite>.</p>
+				<p>If the page is translated authors SHOULD consider whether the keyboard shortcut is valid in the translated language or whether it needs to be changed.</p>
```
adampage, line 12010:
```suggestion
				<p>If the page is translated authors SHOULD consider whether the keyboard shortcut will be valid in each translated language or whether it needs to be changed.</p>
```

Just a soft editorial suggestion here, to account for multiple localizations?

adampage (approved):

index.html
cookiecrook, line 12010: Rather than "[only] If the page is translated, consider…" I think this should be something for all authors to consider, especially since many attributes are "translatable" by the UA...

I raised this as a concern when the attribute was proposed... For example `META`+numbers can't be achieved in some layouts because the numbers are all on the `SHIFT` plane.

Aaron's response was that Gmail and the Google Docs suite already localizes the shortcut. I think the implication was that no major application could ship internationally _without_ localizing them.

Note: This edit incorporates Adam's prior editorial suggestion.

```suggestion
				<p>Authors SHOULD consider whether the keyboard shortcut will be valid in each language and physical keyboard layout, and consider localizing the shortcut in languages, locales, and common hardware keyboard configurations.</p>
```

cookiecrook (approved):
👍

PR #2043, opened 2023-09-17
Editorial: adds new 1.3 translatable attributes

### LINKED ISSUE #1316: Expand translatable properties section
Looks like the [translatable](https://w3c.github.io/aria/#translatable-states-and-properties) section is missing a few properties. It defines:
- aria-label
- aria-placeholder
- aria-roledescription
- aria-valuetext

I think it's missing:
- aria-colindextext
- aria-description
- aria-rowindextext

The braille ones are a bit more complex but my understanding is that they should also be translated since they won't be translated by assistive technologies:
- aria-braillelabel
- aria-brailleroledescription

I can do a PR if those additions make sense.

jnurthen:
Closes #1316

Adds the new translatable attributes in 1.3 to the section

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2043.html" title="Last updated on Nov 4, 2023, 3:44 AM UTC (0de440c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2043/bb56584...0de440c.html" title="Last updated on Nov 4, 2023, 3:44 AM UTC (0de440c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -10401,8 +10401,11 @@ button.ariaPressed; // null</pre>
 		<p>To be understandable by assistive technology users, the values of the following <a>states</a> and [=ARIA/properties=] are <a data-cite="html/dom.html#translatable-attributes">translatable attributes</a> and should be translated when a page is localized:</p>
 		<ul>
 			<li><pref>aria-label</pref></li>
+			<li><pref>aria-colindextext</pref></li>
+			<li><pref>aria-description</pref></li>
 			<li><pref>aria-placeholder</pref></li>
 			<li><pref>aria-roledescription</pref></li>
+			<li><pref>aria-rowindextext</pref></li>
 			<li><pref>aria-valuetext</pref></li>
 		</ul>
 	</section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -10401,8 +10401,11 @@ <h2>Translatable Attributes</h2>
 		<p>To be understandable by assistive technology users, the values of the following <a>states</a> and [=ARIA/properties=] are <a data-cite="html/dom.html#translatable-attributes">translatable attributes</a> and should be translated when a page is localized:</p>
 		<ul>
 			<li><pref>aria-label</pref></li>
+			<li><pref>aria-colindextext</pref></li>
+			<li><pref>aria-description</pref></li>
 			<li><pref>aria-placeholder</pref></li>
 			<li><pref>aria-roledescription</pref></li>
+			<li><pref>aria-rowindextext</pref></li>
 			<li><pref>aria-valuetext</pref></li>
```
pkra, line 10409: I'd add the braille properties. For strings of non-braille patterns, it seems useful. I doubt automated translation will try to handle braille patterns, but why not let them try if they want.

cookiecrook: It's probably fine to add those, because the auto-localizing probably won't work on the Braille unicode range.

However, since we have some author requirements in the braille attributes about using _either braille or non-braille chars, but not both_, there may need to be a user agent note or should req.

cookiecrook (approved):

PR #2040, opened 2023-09-15
Fix broken link in aria-level note

ariellalgilmore:
Closes #2038

Switch markdown for html format and update to current APG website

***
<a href="https://pr-preview.s3.amazonaws.com/ariellalgilmore/aria/pull/2040.html" title="Last updated on Sep 26, 2023, 6:05 PM UTC (5d5ae37)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2040/bb56584...ariellalgilmore:5d5ae37.html" title="Last updated on Sep 26, 2023, 6:05 PM UTC (5d5ae37)">Diff</a>

PR #2044, opened 2023-09-19
Removal of apg markdown link

scottaohara:
the link to the APG was written in markdown syntax and that didn't convert over to a hyperlink when published.  so came here to fix that... but then i tried following the link which just took me to a page saying that content had moved... and then when i tried to find a suitable link to replace, there didn't appear to be any actual content on the APG about aria-level in context of headings... so I removed the link all together since it didn't make sense to have a link to more information, where such information didnt' actually exist.

but if i'm wrong and I just didn't look hard enough for a relevant link on the apg, then please let me know and i'll add it back in


pkra:
I think this is a duplicate of #2040

pkra:
Or perhaps "conflicting" is more appropriate.

scottaohara:
thanks @pkra, missed that one.  yah, 'conflicting' is the better word :)

I can close this and comment on the other one then.

PR #2056, opened 2023-10-06
Editorial: Provide examples of types of menus

### LINKED ISSUE #1084: Make the difference between menu and menubar clearer
In my opinion, the difference between `menu` and `menubar` is not sufficiently explained in the specification because it is only explained in `menubar`:

> A presentation of menu that usually remains visible and is usually presented horizontally.

But the definition of `menu` does not say that it is not permanently visible. Therefore I suggest to add that to `menu`: 

"A menu is not permanently visible, but is only displayed after user action. It can be, for example, a context menu, a submenu within a menu bar or a menu which is called up via a button."

In addition, I suggest removing the first "usually" in the description of `menubar`, since it is not clear why this (" usually remains visible") does not always apply. 

See https://github.com/w3c/aria-practices/issues/1195

scottaohara:
Closes #1084

Updated menu content to attempt to help differentiate `menu` and `menubar`, per the request of the original issue.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2056.html" title="Last updated on Feb 6, 2025, 6:48 PM UTC (5635635)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2056/92b4e8d...5635635.html" title="Last updated on Feb 6, 2025, 6:48 PM UTC (5635635)">Diff</a>


index.html
```diff
@@ -5289,7 +5289,7 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
 			<rdef>menu</rdef>
 			<div class="role-description">
 				<p>A type of <a>widget</a> that offers a list of choices to the user.</p>
-				<p>A menu is often a list of common actions or functions that the user can invoke. The <code>menu</code> <a>role</a> is appropriate when a list of menu items is presented in a manner similar to a menu on a desktop application.</p>
+				<p>A menu is often presented as a popup list of items, representing common actions or functions a user can invoke, related to the item that the user interacted with to reveal the menu. The <code>menu</code> <a>role</a> is appropriate when a list of menu items is presented in a manner similar to a menu on a desktop application. For instance, a menu could be used to represent a context menu for its invoking element, or it would be used to render sub-menus for items of a <rref>menubar</rref> or another <code>menu</code> popup.</p>
```
spectranaut, line 5292:
```suggestion
				<p>A menu is a popup list of items that represent common actions or functions a user can invoke. The menu and list of <code>menuitems</code> are almost always directly related to the item that the user interacted with to reveal the menu. The <code>menu</code> <a>role</a> is appropriate when a list of menu items is presented in a manner similar to a menu on a desktop application. For instance, a menu could be used to represent a context menu for its invoking element, or it would be used to render sub-menus for items of a <rref>menubar</rref> or another <code>menu</code> popup.</p>
```

I mainly think the first sentence could be split in two, that parenthetical makes it a bit hard to read, so here is my attempt! But I don't have a strong preference, if you prefer your sentence, so I'm otherwise approving.

Also I would remove the word "common" from the first sentence because it seems to me unnecessary.

scottaohara: thanks @spectranaut 

pulling your change in.  I had just kept 'common' there since it was in the original text.  but yeh, unless someone thinks otherwise, i don't see the downside of removing it.

pkra (approved):

index.html
```diff
@@ -5289,7 +5289,7 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
 			<rdef>menu</rdef>
 			<div class="role-description">
 				<p>A type of <a>widget</a> that offers a list of choices to the user.</p>
-				<p>A menu is often a list of common actions or functions that the user can invoke. The <code>menu</code> <a>role</a> is appropriate when a list of menu items is presented in a manner similar to a menu on a desktop application.</p>
+				<p>A menu is a popup list of items that represent a set of actions or functions a user can invoke. The menu and list of <code>menuitems</code> are almost always directly related to the item that the user interacted with to reveal the menu. The <code>menu</code> <a>role</a> is appropriate when a list of menu items is presented in a manner similar to a menu on a desktop application. For instance, a menu could be used to represent a context menu for its invoking element, or it would be used to render sub-menus for items of a <rref>menubar</rref> or another <code>menu</code> popup.</p>
```
mcking65, line 5292: I think we can further simplify and clarify this language. I believe it is also important to specify that activating menuitems close menus,  to enumerate how menus are triggered, and provide a link to APG.

```suggestion
				<p>A menu is a popup containing a set of <code>menuitem</code> elements that can be invoked to perform an action or function. The actions are nearly always directly related to the element that the user activated to popup the menu. The <code>menu</code> <a>role</a> is appropriate when a popup is presented and behaves in a manner similar to a menu on a desktop application. activating a <code>menuitem</code> both performs the action and causes the popup to close.</p>
				<p>Menu popups can be triggered by:</p>
				<ul>
					<li>Activating a <code>menuitem</code> in a parent <code>menu</code> or <rref>menubar</rref>.
					<li>Activating a <rref>button</rref> that has the <pref>aria-haspopup</pref> set to <code>true</code>.
					<li>Executing a command that triggers a context menu for an element.</li>
				</ul>
				<p>See the <a href="https://www.w3.org/WAI/ARIA/apg/patterns/menubar/">WAI-ARIA Authoring Practices</a> for more details about implementing menus.</p>
```

mcking65 (changes requested):
I think this should address more of the most misunderstood aspects of appropriate menu usage and reference the APG. I provided a specific suggestion.

index.html
scottaohara: Is it odd to link to the APG page that only has menubar examples, from the menu definition?

not part of my previous PR, but this suggestion made me realize do we need to change the instance of `<code>menuitem</code>` to just say "menu item" - so as to not imply that only `menuitem` elements are allowed, and not also `menuitemradio` and `menuitemcheckbox`?

For the activation of menuitems, does that belong in the `menuitem` section?  In other words, taking `menuitemcheckbox` into consideration, is it _always_ true that activating one of those should cause a menu popup to close?

Pending those above questions, I'm otherwise good with these suggestions.  I just don't know if they all belong in just this `menu` update?

scottaohara: @mcking65 re-ping on my questions concerning your suggested updates, so we can move this along

scottaohara: so as to move this forward again, i used some of the provided feedback, but not all.

index.html
```diff
@@ -5289,7 +5289,8 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
 			<rdef>menu</rdef>
 			<div class="role-description">
 				<p>A type of <a>widget</a> that offers a list of choices to the user.</p>
-				<p>A menu is often a list of common actions or functions that the user can invoke. The <code>menu</code> <a>role</a> is appropriate when a list of menu items is presented in a manner similar to a menu on a desktop application.</p>
+				<p>A menu is container, generally visually rendered as a popup or overlay, for a set of menu items that can be invoked to perform an action or function. The function is almost always directly related to the element, or a close relation to the element that the user activated to invoke the menu. Activating a menu item both performs the function, and results in the automatic dismissal of the menu.</p>
```
scottaohara, line 5292:
```suggestion
				<p>A menu is a container, generally visually rendered as a popup or overlay, for a set of menu items that can be invoked to perform an action or function. The function is almost always directly related to the element, or a close relation to the element that the user activated to invoke the menu. Activating a menu item both performs the function, and results in the automatic dismissal of the menu.</p>
```

index.html
```diff
@@ -5289,7 +5289,8 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
 			<rdef>menu</rdef>
 			<div class="role-description">
 				<p>A type of <a>widget</a> that offers a list of choices to the user.</p>
-				<p>A menu is often a list of common actions or functions that the user can invoke. The <code>menu</code> <a>role</a> is appropriate when a list of menu items is presented in a manner similar to a menu on a desktop application.</p>
+				<p>A menu is a container, generally visually rendered as a popup or overlay, for a set of menu items that can be invoked to perform an action or function. The function is almost always directly related to the element, or a close relation to the element that the user activated to invoke the menu. Activating a menu item both performs the function, and results in the automatic dismissal of the menu.</p>
```
cookiecrook, line 5292: Editorial nits
```suggestion
				<p>A menu is a container, generally rendered as a popup or overlay, for a set of menu items that can be invoked to perform an action or function. The function is almost always closely related or directly related to the element that the user activated to invoke the menu. Activating a menu item both performs the associated function of the menu item, and results in the automatic dismissal of the menu.</p>
```

index.html
```diff
@@ -5289,7 +5289,8 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
...
+				<p>The <code>menu</code> <a>role</a> is appropriate when a set of menu items is presented in a manner similar to a popup menu for a desktop application. For instance, a menu could be used to represent a context menu for its invoking element, or it would be used to render sub-menus for items of a <rref>menubar</rref> or another <code>menu</code> popup.</p>
```
cookiecrook, line 5293:
```suggestion
				<p>The <code>menu</code> <a>role</a> is appropriate when a set of menu items is presented in a manner similar to a popup menu. For instance, a menu could be used to represent a context menu for its invoking element, or it would be used to render sub-menus for items of a <rref>menubar</rref> or another <code>menu</code> popup.</p>
```

cookiecrook (approved):
approving with minor comments

PR #2066, opened 2023-10-19
Editorial: Created new entry: "focusable".

### LINKED ISSUE #1936: do we need a focusable definition
Followup from editors call: do we need a focusable definition

Defined in HTML and SVG - should we point to those definitions and how?

MarioBatusic:
Closes #1936

Crated a minimal entry with a link to the definition in the HTML Spec.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2066.html" title="Last updated on Aug 21, 2025, 10:11 AM UTC (2a688f1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2066/5b53575...2a688f1.html" title="Last updated on Aug 21, 2025, 10:11 AM UTC (2a688f1)">Diff</a>


common/terms.html
```diff
@@ -79,7 +79,19 @@
     <dd>
       <p>Translated to platform-specific <a class="termref" data-lt="accessibility api">accessibility APIs</a> as defined in the <a href="" class="core-mapping">Core Accessibility API Mappings</a>.</p>
     </dd>
-	<dt><dfn data-lt="graphical document|graphical documents">Graphical Document</dfn></dt>
+  </dd>
+  <dt><dfn>Focusable</dfn></dt>
+  <dd>
+    <p>An element or an area in a document is focusable, if it can get the system focus. In different circumstances one can use one or more of following methods to set focus:</p>
```
rahimabdi, line 85: To remain consistent with other references to "focus", I would suggest "receive focus" rather than "get". For examples where focus is referenced:

- https://w3c.github.io/aria/#alert
- https://w3c.github.io/aria/#status
- https://w3c.github.io/aria/#aria-disabled

cookiecrook (changes requested):
Current PR doesn't include a lot of the items discussed as necessary in the original PR attempt: https://github.com/w3c/aria-common/pull/39

In particular, this comment and follow-on disucssion https://github.com/w3c/aria-common/pull/39#issuecomment-616901209

cookiecrook:
HTML's "focusable area" is defined here: https://html.spec.whatwg.org/multipage/interaction.html#focusable-area

common/terms.html
```diff
@@ -79,7 +79,19 @@
...
+  <dt><dfn>Focusable</dfn></dt>
+  <dd>
+    <p>An element or an area in a document is focusable, if it can receive focus. In different circumstances one can use one or more of the following methods to set focus:</p>
+    <ul>
+      <li>by scripting</li>
+      <li>using a mouse</li>
+      <li>using keyboard</li>
+      <li>by touch.</li>
+    </ul>
+    <p>For a detailed description see <a href="https://html.spec.whatwg.org/#focusable-area">Focusable area</a> in the HTML Specification.</p>
```
cookiecrook, line 92:
```suggestion
    <p>An element or area matching the definition of <a href="https://html.spec.whatwg.org/#focusable-area">focusable area</a> in the HTML Specification.</p>
```

cookiecrook: Replying to @MarioBatusic's email response question with a specific edit suggestion. If the term is to be defined in ARIA, it should be a direct cross-reference to the other spec's definition, rather than a terse-but-incomplete variant.

cookiecrook (changes requested):

MarioBatusic:
@cookiecrook Thanx for your help.

common/terms.html
cookiecrook: Seems like this change has been incorporated into a different commit. Closing this thread and I will re-review.

index.html
```diff
@@ -13692,12 +13702,15 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+            >
             <p>Similarly, it can be applied to a <rref>button</rref> to control the visibility of another element and its content on the current page.</p>
-            <pre class="example highlight">&lt;button type="button" aria-controls="mangosteen" aria-expanded="false"&gt;Mangosteen&lt;/button&gt;
+            <pre class="example highlight">
+&lt;button type="button" aria-controls="mangosteen" aria-expanded="false"&gt;Mangosteen&lt;/button&gt;
 &lt;div id="mangosteen" hidden&gt;
   An edible fruit native to tropical lands surrounding the Indian Ocean.
-&lt;/div&gt;</pre>
+&lt;/div&gt;</pre
+            >
```
cookiecrook: was this intentional? If not, merge this diff revert? 
```suggestion
&lt;/div&gt;</pre>
```

index.html
```diff
@@ -13692,12 +13702,15 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
       &lt;li role="treeitem" aria-selected="false"&gt;Yuzu&lt;/li&gt;
     &lt;/ul&gt;
   &lt;/li&gt;
-&lt;/ul&gt;</pre>
+&lt;/ul&gt;</pre
+            >
```
cookiecrook: was this intentional? If not, merge this diff revert?
```suggestion
&lt;/ul&gt;</pre>
```

index.html
```diff
@@ -13681,9 +13685,15 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+            <p>
+              The <sref>aria-expanded</sref> attribute is applied to a focusable, interactive element that toggles visibility of content of a different element. If the element with
+              <sref>aria-expanded</sref> is also a <rref>treeitem</rref> in a <rref>tree</rref> or a <rref>row</rref> in a <rref>treegrid</rref>, then it SHOULD also be the
+              <a>accessibility parent</a> of the content it expands and collapses. Otherwise, the element with <sref>aria-expanded</sref> SHOULD NOT be the <a>accessibility parent</a> of the content
+              that is expanding or collapsing. Rather, identify that relationship between the interactive element and the element being controlled using <pref>aria-controls</pref>.
+            </p>
```
cookiecrook: These RFC-2119 reqs are passively phrased as requirements on the element, rather than an Author or User Agent... 

I presume these are both AUTHOR requirements, so the sentences should be rephrased to include some variant of "Authors SHOULD" and "Authors SHOULD NOT" rather than the ambiguously phrased "the element SHOULD NOT be…"

cookiecrook:
```suggestion
            <p>
              The <sref>aria-expanded</sref> attribute is applied to a focusable, interactive element that toggles visibility of content of a different element. If the element with
              <sref>aria-expanded</sref> is also a <rref>treeitem</rref> in a <rref>tree</rref> or a <rref>row</rref> in a <rref>treegrid</rref>, then the Author SHOULD ensure the element is also the
              <a>accessibility parent</a> of the content it expands and collapses. Otherwise, the Author SHOULD ensure the element with <sref>aria-expanded</sref> is not the <a>accessibility parent</a> of the content
              that is expanding or collapsing. Rather, identify that relationship between the interactive element and the element being controlled using <pref>aria-controls</pref>.
            </p>
```

common/terms.html
pkra: This seems to have been resolved.

index.html
pkra: This is prettier formatting. Tweaking it here will only be changed again upon merge.

index.html
pkra: Ditto

index.html
pkra: There was no real change here, just prettier formatting. I've created an issue at https://github.com/w3c/aria/issues/2603/ to follow up on this.

pkra:
@jnurthen I think this PR needs some actual references (if only to avoid respec complaining). What do you think?

pkra:
Continued on https://github.com/w3c/aria/pull/2605

PR #2062, opened 2023-10-12
Editorial: Update lone reference of "alternative text" to "text alternative"

rahimabdi:
Closes [#2059](https://github.com/w3c/aria/issues/2059) 

In ARIA 1.2, there is a single reference to "alternative text" in the `presentation` role description section however, all other instances in the spec use the term "text alternative". For consistency, this PR updates the lone instance of "alternative text" to "text alternative".

**NOTE**: This aligns with WCAG which has no instances of the term "alternative text".

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2062.html" title="Last updated on Oct 19, 2023, 5:27 PM UTC (d50c04d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2062/75f7ad7...d50c04d.html" title="Last updated on Oct 19, 2023, 5:27 PM UTC (d50c04d)">Diff</a>

PR #2069, opened 2023-10-20
editorial: remove vague "meaningful" and parenthetical example

### LINKED ISSUE #2068: What is non-meaningful alternative text?
Not clear what "should not provide meaningful text alternative" means.

[editors draft](https://w3c.github.io/aria/) ?
Yes.

I think we would be good with just "should not provide alternative text as @pkra suggested in #2062

I am interested in whether there are use cases that suggest keeping current formulation.

pkra:
Following up on #2062.
Closes #2068

In this negation, it seems very vague (what is non-meaningful-but-not-empty alt?) and I think a clear "SHOULD NOT add alt" is better.

Otherwise, if there's a use case for "non-meaningful text alternative" then perhaps this can be added to the parenthetical example.

To further reduce confusion, the parenthical example using an empty alt attribute was also removed.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2069.html" title="Last updated on Jan 30, 2025, 8:00 AM UTC (4a7c3c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2069/a2d34fe...4a7c3c7.html" title="Last updated on Jan 30, 2025, 8:00 AM UTC (4a7c3c7)">Diff</a>


index.html
```diff
@@ -5987,7 +5987,7 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
 &lt;span&gt; Sample Content &lt;a href="...">let's go!&lt;/a> &lt;/span&gt;
 </pre>
 				<p>In HTML, the <code>&lt;img&gt;</code> <a>element</a> is treated as a single entity regardless of the type of image file. Consequently, using <code>role="none"</code> or <code>role="presentation"</code> on an HTML <code>img</code> is equivalent to using <code>aria-hidden="true"</code>. In order to make the image contents accessible, authors can embed the object using an <code>&lt;object&gt;</code> or <code>&lt;iframe&gt;</code> <a>element</a>, or use inline <abbr title="Scalable Vector Graphics">SVG</abbr> code, and follow the accessibility guidelines for the image content.</p>
-				<p>Authors SHOULD NOT provide a meaningful text alternative (for example, use <code>alt=""</code> in HTML) when the <code>none</code>/<code>presentation</code> role is applied to an image.</p>
+				<p>Authors SHOULD NOT provide a text alternative (for example, use <code>alt=""</code> in HTML) when the <code>none</code>/<code>presentation</code> role is applied to an image.</p>
```
scottaohara, line 5990: it would be good to provide rational as to why someone would use `alt=""` and `role=none` on an image, since that's redundantly redundant.

pkra: Ugh, right. Should I just remove that parenthetical example, too? I suspect even if it was negated, it would still cause confusion.

scottaohara: sometimes some of the best fixes are just removing the problematic content :)

pkra: That was all the motivation I needed. :smile:

scottaohara (approved):

index.html
```diff
@@ -5987,7 +5987,7 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
 &lt;span&gt; Sample Content &lt;a href="...">let's go!&lt;/a> &lt;/span&gt;
 </pre>
 				<p>In HTML, the <code>&lt;img&gt;</code> <a>element</a> is treated as a single entity regardless of the type of image file. Consequently, using <code>role="none"</code> or <code>role="presentation"</code> on an HTML <code>img</code> is equivalent to using <code>aria-hidden="true"</code>. In order to make the image contents accessible, authors can embed the object using an <code>&lt;object&gt;</code> or <code>&lt;iframe&gt;</code> <a>element</a>, or use inline <abbr title="Scalable Vector Graphics">SVG</abbr> code, and follow the accessibility guidelines for the image content.</p>
-				<p>Authors SHOULD NOT provide a meaningful text alternative (for example, use <code>alt=""</code> in HTML) when the <code>none</code>/<code>presentation</code> role is applied to an image.</p>
+				<p>Authors SHOULD NOT provide a text alternative when the <code>none</code>/<code>presentation</code> role is applied to an image.</p>
```
jnurthen, line 5990: I'm not sure that this was trying to say that. 
I think it was saying don't use alt="something"
But it is fine (although redundant) to use alt="" with role="none" as they essentially do the same thing. Indeed the example below does exactly that.

```suggestion
				<p>Authors SHOULD NOT provide a non empty text alternative when the <code>none</code>/<code>presentation</code> role is applied to an image.</p>
```

jnurthen (changes requested):
if we really don't intend to allow alt="" on role="none" then fine - but I'm not sure that is the intent. If it is then we need to change the example too.

index.html
pkra: I take the point regarding the following example but since this will get flagged by nu validator already, we should perhaps remove it.

I find the double negation unnecessarily complicated. Since empty aria-label cause validation errors, I think it wouldn't reduce confusion as much so I think it's better as is. But it's not really that important to me - happy to merge the suggestion.

pkra:
> if we really don't intend to allow alt="" on role="none" then fine - but I'm not sure that is the intent. If it is then we need to change the example too.

E.g., nu validator flags it as error "An img element which has an alt attribute whose value is the empty string must not have a role attribute."

index.html
MelSumner: FWIW I don't think `alt="" role="none"` should fail the validator. I think it should be okay to do this. 

It should also be fine to do `role="none"` or `alt=""`

pkra:
@jnurthen I've added your suggestion. Can I merge this?

pkra:
@jnurthen another ping.

@scottaohara are you still ok with this adjustment?

jnurthen (approved):

spectranaut:
couldn't fix the merge conflicts so I added a new PR: https://github.com/w3c/aria/pull/2505

in order to close/merge this :)

spectranaut:
closing, merged replace PR

PR #2090, opened 2023-12-14
Change aria-hidden=false to synonym of undefined

### LINKED ISSUE #1256: Consider allowing `aria-hidden=false` on visibly rendered elements inside an ancestor with `aria-hidden=true`

scottaohara:
resolves #1256

This PR modifies the previous note about aria-hidden=false to instead indicate that as of ARIA 1.3 is serves as a synonym for the 'undefined' value.

It also includes a minor editorial change to remove the mention of 'screen readers' from the last paragraph of the normative text.  It seemed to state, especially when the sentence ends with referring to 'assistive technologies' in general.

See [last comment](https://github.com/w3c/aria/issues/1256#issuecomment-1799271506)

Question: assuming this PR is approved, would we also want to add the following sentence to the note?
>In future versions of ARIA, the undefined value will be deprecated in favor of the false value.

# PR tracking

* [X] Related Core AAM Issue/PR: N/A
* [X] Related AccName Issue/PR: N/A
* [X] Related APG Issue/PR: N/A
* [X] Any other dependent changes? N/A
  * [X] WPT tests (multiple implementations passing before merge)

# Implementation tracking

* [X] "author MUST" tests: N/A
* [X] "user agent MUST" tests
  * [X] https://github.com/web-platform-tests/interop-accessibility/issues/88
* [X] Browser implementations (link to issue or when done, link to commit):
   * [x] WebKit: https://webkit.org/b/267150
   * [x] Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1880579
   * [x] Blink:  https://bugs.chromium.org/p/chromium/issues/detail?id=1500299
* [X] Does this need AT implementations? No

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2090.html" title="Last updated on May 2, 2024, 8:19 PM UTC (6b47058)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2090/b5a3d7f...6b47058.html" title="Last updated on May 2, 2024, 8:19 PM UTC (6b47058)">Diff</a>


aleventhal (approved):

jcsteh (approved):

index.html
```diff
@@ -11853,9 +11853,9 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p class="note">At the time of this writing, <code><sref>aria-hidden</sref>="false"</code> is known to work inconsistently in browsers. As future implementations improve, use caution and test thoroughly before relying on this approach.</p>
+				<p class="note">As of ARIA 1.3, due to inconsistent support in browsers, and the pervasiveness of user interfaces that rely on assuming <code><sref>aria-hidden</sref>="false"</code> does not expose hidden content to the accessibility tree, the <code>false</code> value is now synonymous with <code>undefined</code>.</p>
```
scottaohara, line 11858:
```suggestion
				<p class="note">As of ARIA 1.3, due to inconsistent support in browsers, and the pervasiveness of user interfaces that rely on assuming <code><sref>aria-hidden</sref>="false"</code> does not expose hidden content to the accessibility tree, though that was not the value's original intent, the <code>false</code> value is now synonymous with <code>undefined</code>.</p>
```

cookiecrook (changes requested):
@aleventhal approved this PR, but he was also [proposing that `aria-hidden=false` be allowed on visibly rendered items](https://github.com/w3c/aria/issues/1256#issuecomment-1478532654), a partial equivalent to the `inert` attribute. What changed Aaron?

If there's a possibility we'll ever want to bring this value back, we should not change it now, and instead wait to change it once https://github.com/WICG/aom/issues/203 (or some other testing approach) gives us access to inspect whether the node is "hidden from accessibility"

cookiecrook:
@jnurthen there were other related issues in the [ARIA-hidden project](https://github.com/orgs/w3c/projects/63) that have been moved elsewhere. Should a keyword be added instead? Is there an easy way to find those issues? I'm not sure why they were put in a 1.3/1.4 project when we had 1.3/1.4 milestones. Apologies if that explanation was covered in one of the meetings I missed.

aleventhal:
> @aleventhal approved this PR, but he was also [proposing that `aria-hidden=false` be allowed on visibly rendered items](https://github.com/w3c/aria/issues/1256#issuecomment-1478532654), a partial equivalent to the `inert` attribute. What changed, Aaron?

I wanted aria-hidden="false" to be useful, but unfortunately, as was pointed out to me, it's misused by a lot of JS toolkits which treat it as a boolean, and they end up setting "false" all over the place where they really should have removed the attribute.

> If there's a possibility we'll ever want to bring this value back, we should not change it now, and instead wait to change it once [WICG/aom#203](https://github.com/WICG/aom/issues/203) (or some other testing approach) gives us access to inspect whether the node is "hidden from accessibility"

Because of the problem in js toolkits, and because browsers all implement it differently, I think we should go ahead and deprecate it as a possible value.

cookiecrook:
> Because of the problem in js toolkits, and because browsers all implement it differently, I think we should go ahead and deprecate [`aria-hidden=false`] as a possible value.

Okay. I'm going to ask @twilco and @minorninth to review as well in case there are implementation concerns with its removal.

I'm hopeful it goes without saying that this should not be merged until we have extensively passing WPT tests. Since this one wasn't in the initial Interop 2024 Focus Area proposal, that may not happen this year, but we can treat it as a tentative area for inclusion if there is time to fix it before the Fall releases.

cookiecrook (commented):
Updating my "Changes Requested" review to "Comment" tentatively pending review from others.

cookiecrook (commented):
Second time's the charm.

index.html
```diff
@@ -11853,9 +11853,9 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 			<div class="state-description">
 				<p><a>Indicates</a> whether the <a>element</a> is exposed to an accessibility API. See related <sref>aria-disabled</sref>.</p>
```
minorninth, line 11854: What if we changed this to:

<p>If set to true, hides an element and its entire subtree from assistive technology, regardless of whether it's visibly rendered.</p>

In particular:
* The current wording ("indicates whether the element is exposed to an accessibility API") sort of implies that both true and false could have meaning. Let's make it more clear that only aria-hidden=true means anything.
* I think it's confusing to say that aria-disabled is related. That seems like a stretch to me. I think it'd be more clear to leave that out.

index.html
```diff
@@ -11853,9 +11853,9 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p class="note">At the time of this writing, <code><sref>aria-hidden</sref>="false"</code> is known to work inconsistently in browsers. As future implementations improve, use caution and test thoroughly before relying on this approach.</p>
+				<p class="note">As of ARIA 1.3, <code><sref>aria-hidden</sref>="false"</code> is now synonymous with <code>aria=hidden="undefined"</code>. This change was made due to inconsistent browser support for <code>false</code>, and the pervasiveness of user interfaces that rely on assuming <code>aria-hidden="false"</code> would not expose hidden content to the accessibility tree, though that was not the value's original intent.</p>
```
minorninth, line 11858: I think it's a bit confusing to put the new definition, the original intent, and the history all in one paragraph like this. How about:

1. One paragraph that explains that (as of ARIA 1.3), aria-hidden=false and aria-hidden=undefined are the same. Then it should clarify WHAT happens if aria-hidden is false or unspecified - namely that the user agent will expose the element to assistive technology if it's visibly rendered, and hide it if not. I know it's slightly redundant, but I'd rather be super clear here.
2. A second paragraph marked as a Note, saying that originally there was an idea that aria-hidden=false could be used to expose content to assistive technology even if it's rendered, however it wasn't clearly specified, and it's no longer supported because it was never consistently implemented by browsers anyway.

minorninth:
I support removing the semantics of aria-hidden=false, but I'd like to improve the wording. I left some comments. Thanks.

index.html
cookiecrook: Indicates is one of the primary reserved terms (indicates, identifies, etc) that allude to the value type. How about this:
```suggestion
				<p><a>Indicates</a>, when set to true, that an <a>element</a> and its entire subtree are hidden from assistive technology, regardless of whether it's visibly rendered. See related <sref>aria-disabled</sref>.</p>
```

minorninth: Yep, I like that better. Thanks

index.html
```diff
@@ -11851,11 +11851,11 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 		<div class="state" id="aria-hidden">
 			<sdef>aria-hidden</sdef>
 			<div class="state-description">
-				<p><a>Indicates</a> whether the <a>element</a> is exposed to an accessibility API. See related <sref>aria-disabled</sref>.</p>
+				<p><a>Indicates</a>, when set to true, that an <a>element</a> and its entire subtree are hidden from assistive technology, regardless of whether it's visibly rendered. See related <sref>aria-disabled</sref>.</p>
```
scottaohara, line 11854: while the comparision to aria-disabled was pre-existing, I tend to agree with Dominic's comment for its removal.
```suggestion
				<p><a>Indicates</a>, when set to <code>true</code>, that an <a>element</a> and its entire subtree are hidden from assistive technology, regardless of whether it is visibly rendered.</p>
```

cookiecrook (approved):

index.html
```diff
@@ -11851,11 +11851,11 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p class="note">At the time of this writing, <code><sref>aria-hidden</sref>="false"</code> is known to work inconsistently in browsers. As future implementations improve, use caution and test thoroughly before relying on this approach.</p>
+				<p class="note">As of ARIA 1.3, <code><sref>aria-hidden</sref>="false"</code> is now synonymous with <code>aria=hidden="undefined"</code>. This change was made due to inconsistent browser support for <code>false</code>, and the pervasiveness of user interfaces that rely on assuming <code>aria-hidden="false"</code> would not expose hidden content to the accessibility tree, though that was not the value's original intent.</p>
```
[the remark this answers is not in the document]
scottaohara: @minorninth @cookiecrook ping again to try and close out the lingering comments on this.  if alternate text is still desired, please provide what you'd want this to say.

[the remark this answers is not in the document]
minorninth: Sorry for the delay. The improved wording works for me, thanks.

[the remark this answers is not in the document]
scottaohara: thanks!

PR #2091, opened 2023-12-14
Allow treeitems to omit groups by specifiying aria-level, aria-posinset, and aria-setsize

### LINKED ISSUE #2065: Hierarchy of treeitems per aria level only?

[treeitem](https://w3c.github.io/aria/#treeitem):

> An expandable collection of treeitem elements are enclosed in an element with the [group](https://w3c.github.io/aria/#group) [role](https://w3c.github.io/aria/#dfn-role).

In this case, the hierarchy of the treeitems results from the hierarchy of the groups.

In addition, aria-level can be used with the treeitem.

Two problems arise from this:
* What if the hierarchy of the groups contradicts with the hierarchy in aria-level. Which hierarchy wins? This is only explained in [Core AAM](https://w3c.github.io/core-aam/#mapping_additional_position), but a corresponding reference in the ARIA specification is missing.
* Why should groups be used when aria-level is used. Wouldn't it be sufficient to omit group if aria-level is used? 

My suggestion:

> The hierarchy of treeitems is mapped either via proper nesting in groups (role=group) or via aria-level.

## Will this require a change to [CORE-AAM](https://w3c.github.io/core-aam/)?

I do not think so

## Will this require a change to the [ARIA authoring guide](https://w3c.github.io/aria-practices/)?

Not necessarily, but the [aria-level example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1b/) could be modified to remove the hierarchy by group

smhigley:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2091--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2091--wai-aria.netlify.app%2Findex.html)

Closes #2065

Make it clear that authors may create a tree with a flat DOM hierarchy of `treeitem` nodes inside a `tree`, as long as all treeitems have explicit values for `aria-level`, `aria-setsize`, and `aria-posinset`.

Open question: should we specify how `aria-setsize` and `aria-posinset` should be calculated?

# PR tracking

* [ ] Related APG Issue/PR:

# Implementation tracking

* [ ] "author MUST" tests:
* [ ] "user agent MUST" tests: https://github.com/web-platform-tests/wpt/pull/60503
* [x] Browser implementations (link to issue or when done, link to commit):
   * WebKit:
   * Gecko:
   * Blink:
* [x] Does this need AT implementations?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2091.html" title="Last updated on Mar 17, 2025, 11:06 AM UTC (15a8fcd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2091/bf8e158...15a8fcd.html" title="Last updated on Mar 17, 2025, 11:06 AM UTC (15a8fcd)">Diff</a>


index.html
```diff
@@ -9997,6 +9997,12 @@ <h5>Presentational Role Inheritance</h5>
           <li>The user interface makes the meaning and purpose of each state apparent.</li>
           <li>The user interface provides a separate method for controlling each state.</li>
         </ul>
+		<p>
+			User agents MUST calculate an implicit value for <sref>aria-level</sref> from the document structure if the author does not explicitly specify a value.
+		</p>
+		<p>
+			Authors MAY omit using nested <rref>group</rref> elements to indicate level hierarchy. When doing so, authors MUST ensure all <rref>treeitem</rref> elements in the <rref>tree</rref> have explicit values for <sref>aria-level</sref>, <sref>aria-posinset</sref>, and <sref>aria-setsize</sref>.
```
adampage, line 10004:
```suggestion
			Authors MAY use nested <rref>group</rref> elements to implicitly indicate level hierarchy. Otherwise, authors MUST ensure all <rref>treeitem</rref> elements in the <rref>tree</rref> have explicit values for <sref>aria-level</sref>, <sref>aria-posinset</sref>, and <sref>aria-setsize</sref>.
```

Just a subjective editorial suggestion here. The first sentence was a little fuzzy for me on first read, so I wonder if this logic flip might be clearer but still accurate?

adampage (approved):

rahimabdi (approved):

index.html
smhigley: Yeah, I like the rewording to positive over negative better as well 👍

spectranaut:
@mcking65 is this a change we would want to reflect in the APG?

index.html
```diff
@@ -9997,6 +9997,12 @@ <h5>Presentational Role Inheritance</h5>
           <li>The user interface makes the meaning and purpose of each state apparent.</li>
           <li>The user interface provides a separate method for controlling each state.</li>
         </ul>
+		<p>
+			User agents MUST calculate an implicit value for <sref>aria-level</sref> from the document structure if the author does not explicitly specify a value.
+		</p>
+		<p>
+			Authors MAY use nested <rref>group</rref> elements to implicitly indicate level hierarchy. Otherwise, authors MUST ensure all <rref>treeitem</rref> elements in the <rref>tree</rref> have explicit values for <sref>aria-level</sref>, <sref>aria-posinset</sref>, and <sref>aria-setsize</sref>.
+		</p>
```
spectranaut, line 10005: This is really a nit, but it seems to me these two paragraphs (which are related to the structure of the dom/accessibility tree) should be below the other line that is edited, because we also mention `group` and structural stuff up there.

Otherwise it's "structure stuff" then "selected stuff" then "structural stuff".

index.html
```diff
@@ -9997,6 +9997,12 @@ <h5>Presentational Role Inheritance</h5>
           <li>The user interface makes the meaning and purpose of each state apparent.</li>
           <li>The user interface provides a separate method for controlling each state.</li>
         </ul>
+		<p>
+			User agents MUST calculate an implicit value for <sref>aria-level</sref> from the document structure if the author does not explicitly specify a value.
```
spectranaut, line 10001: Oh I guess this is related to your open question. Honestly, i think we might need to explicitly say how it is calculated, because "from the document structure" seems a little to vague to me.

For example, if the tree items are group with divs that don't have role group, I think we should ignore that structure, right? This sentence is not clear about that.

spectranaut (changes requested):
I do think we need to be a bit more specific about how to calculate aria-level, if we think the only relevant "structure" is nested `group` elements.

index.html
smhigley: @spectranaut I actually already addressed my open question in this PR 😅: https://github.com/w3c/aria/pull/2104

Which I think might also address your comment? Let me know if you think it needs different wording (though those comments probably belong on the other PR)

index.html
smhigley: That makes sense, I think I was initially hesitant to put it towards the beginning b/c it seemed like it was addressing a relatively unusual use case. Though I think your point about grouping (heh) the relative structure info holds more weight.

smhigley:
@spectranaut I think I've addressed your comments -- would you mind taking another look when you get a chance?

spectranaut (approved):

pkra:
Rebased on main.

spectranaut:
@smhigley do you remember if any browsers do this already? Or should we open issues on browsers?

Or, do you want to discuss the question you have in the description first:
> Open question: should we specify how aria-setsize and aria-posinset should be calculated?

Personally I think that could be a separate PR, if we wanted to do that.

HaTheo:
If I understand the issue today, there are two primary test cases:

1. **Flat Structure:** If I have a list of fruits with children, such as Apple and Pear, we would expect the browser to infer the relationship if it’s defined in a flat list with aria-levels implying nesting. This inference should be possible as long as the Author provides aria-levels that allow the browser to calculate it.

      ```
      <div role="tree" id="flat-tree">
          <div role="treeitem"aria-level="1">Fruits</div>
          <div role="treeitem"  aria-level="2">Apples</div>
          <div role="treeitem"  aria-level="2">Pears</div>
          <div role="treeitem"  aria-level="2">Oranges</div>
          <div role="treeitem"  aria-level="1">Vegitables</div>
          <div role="treeitem"  aria-level="2">Carrots</div>
          <div role="treeitem"  aria-level="2">Broccoli</div>
          <div role="treeitem"  aria-level="2">Spinach</div>
      </div>
      ```

2. **Hierarchical Structure:** Similarly, we would assume that any browser with a tree structure, as indicated by nested elements within the list, would automatically calculate and provide the aria-level.

    ```
    <div role="tree" id="flat-tree">
      <div role="treeitem"aria-level="1">Fruits
        <div role="group">
          <div role="treeitem" >Apples</div>
          <div role="treeitem">Pears</div>
          <div role="treeitem">Oranges</div>
         </div>
      </div>
      <div role="treeitem"  aria-level="1">Vegitables
         <div role="group">
          <div role="treeitem">Carrots</div>
          <div role="treeitem">Broccoli</div>
          <div role="treeitem">Spinach</div>
         </div>
      </div>
    </div>
    ```
    
However, as verified through manual testing, I haven’t encountered any browser that effectively exposes the hierarchy to Platform Accessibility APIs with Flat Structure (at least on my Mac OS machine), the Hierarchical Structure seems to work fine in Chrome, and I will test others later. Therefore, I propose two actions:

  1. Write a WPT (Web Page Test): We should create a WPT for this scenario.
  
  2. File Bugs: We should also file bugs for each browser that doesn’t meet these requirements.


@spectranaut I am happy to take both of these tasks on....

HaTheo:
@smhigley, and @scottaohara (as this is sort of Core-AAM related). I want to confirm that I understand this issue before filing bugs against browsers. There are two parts to this:

1.  Is a flat DOM structure valid? _<- this one is obviously a yes, I think, based on what you said_

2. Should the computed Web Accessibility Tree be hierarchical? _<- Less obvious but maybe implied as this might be required to translate to platform A11y APIs_

I would like to clarify my understanding of the primary motivation behind this. A flatter DOM is mainly designed to support performant UI constraints. It achieves this by creating shallower trees (trivially) and simplifying virtualization (by allowing stacking of elements). Currently, the traditional structure of a Tree requires subtrees and leaf nodes to be physically grouped together under the parent tree item.

Therefore, I understand that # 1 is the immediate goal (it is valid to define three items flatly using aria-level, aria-setsize, and aria-posinset). However, regarding # 2, I want to confirm the expected browser behavior: By providing these exact coordinates (level, posinset, setsize), is the expectation that the browser will compute a hierarchical Accessibility Tree from the DOM without aria-owns or nested DOM elements? I ask this for clarification, as it makes this relatively novel; I can't easily think of anywhere outside aria-owns (which is really its intent) where aria properties will modify the accessibility tree structure.

In other words is the expectation that this HTML

```
<div role="tree" id="flat-tree-with-explicit-attributes">
  <div role="treeitem" aria-level="1" aria-posinset="1" aria-setsize="2">Fruits</div>
  <div role="treeitem" aria-level="2" aria-posinset="1" aria-setsize="3">Apples</div>
  <div role="treeitem" aria-level="2" aria-posinset="2" aria-setsize="3">Pears</div>
  <div role="treeitem" aria-level="2" aria-posinset="3" aria-setsize="3">Oranges</div>
  
  <div role="treeitem" aria-level="1" aria-posinset="2" aria-setsize="2">Vegetables</div>
  <div role="treeitem" aria-level="2" aria-posinset="1" aria-setsize="3">Carrots</div>
  <div role="treeitem" aria-level="2" aria-posinset="2" aria-setsize="3">Broccoli</div>
  <div role="treeitem" aria-level="2" aria-posinset="3" aria-setsize="3">Spinach</div>
</div>

```

Becomes this computed accessibility tree:

-  Tree
    - treeitem "Fruit"
      - treeitem "Apples"
      - treeitem "Pears"
      - treeitem "Oranges"
    - treeitem "Vegitables"
      - treeitem "Carrots"
      - treeitem "Broccoli"
      - treeitem "Spinach"
   
I ask this because, when I think about how this would translate to platform Accessibility trees, I can see how it would become an issue, as I imagine many would expect a hierarchy, and having rules to do this consistently might make sense and align with the intention. Alternatively, would it be expected that a browser simply creates a flat Tree but includes the appropriate grouping properties?

HaTheo:
@smhigley I've tested this across major browsers (Chromium, Firefox, Edge, Safari) and across ATs/Platforms (NVDA, JAWS, Narrator, Orca, and VoiceOver). I confirmed this within the Platform and Browser accessibility trees; the level is properly represented. If the intention here is simply to read the level and ensure it's included in platform/browser accessibility trees, regardless of whether it's written with a flat or hierarchical structure (nested groups), Flat Trees with defined levels seem to be working fine today.

https://codepen.io/HaTheo/pen/myRRKBK

spectranaut:
The spec matches implementations, I believe we can merge this, @pkra ?

pkra:
@spectranaut :+1:  

I've resolved the merge conflict (by reverting and older prettier commit).

Issue #1480, opened 2021-05-06
Consider a flag warning new developers to avoid some aria features (role desc, braille props, etc)


cookiecrook:
In the role description deep dive this morning, we discussed ways to mitigate author misuse of `aria-roledescription`, and potentially other properties that 1) aren't commonly required and 2) are easy for inexperienced devs to misuse. 

One possibility is to include a flag for "uncommon usage" (name TBD) or something else that indicates:

- It's **_NOT_** expected that most web developers would need to use these features.
- Even if they are tempted to use one of these, take additional time to seriously understand why you need it and how to use it.
- Seriously web devs, **_DO NOT_** use this unless you have a really good reason.

IMO, the warning text should be bright and bold, immediately following the short description with sufficient use of warning icons or ⚠️ emoji. We might encourage validators or other checker tools to flag these for author review, too.

The braille properties, particularly those that allow characters in the Braille Unicode range, are other good examples. The ARIA WG added these for advanced usage by testing or other educational orgs. No only are they not required for most web apps and pages, chances are that lay usage will actually be more harmful than helpful. In other words, it's dangerous to use these without a full understanding.

In some ways, `aria-label` and `aria-hidden` could be considered dangerous; you can misuse them to develop a very inaccessible interface. However, I don't think common features like those should be included in the list due to the widespread use and the generally-more-understandable nature of the features. If we had the warning on `aria-hidden`, web developers would be more likely to ignore the warning when/if they saw it on `aria-braillelabel`…

cookiecrook:
These three are the obvious ones:
```
aria-roledescription
aria-braillelabel
aria-brailleroledescription
```

I'm not as sure about:
```
aria-keyshortcuts
```

@joanmarie @aleventhal @minorninth ??

aleventhal:
```role```

JAWS-test:
Why aria-keyshortcuts?

minorninth:
Are you just talking about language in the spec discouraging certain attributes? No arguments there. You're not proposing a javascript console warning or something like that, right?

Maybe we should add the same warning for aria-owns?

cookiecrook:
> Are you just talking about language in the spec discouraging certain attributes? No arguments there. 

Yes

> You're not proposing a javascript console warning or something like that, right?

[...]

PR #2104, opened 2024-01-12
Add normative text on calculating aria-level, aria-posinset, and aria-setsize

### LINKED ISSUE #1194: Consider requiring user agents to calculate aria-level, aria-posinset, and aria-setsize when not declared by authors
The proposed comment role includes the following language:

>If the author has not explicitly declared aria-level, aria-posinset, or aria-setsize for a comment element, user agents MUST automatically compute the missing values and expose them to assistive technologies.

Currently, ARIA does not include a normative requirement for computing these properties when they are used on elements with other roles that support them. For example, the definition of aria-level includes the following language:

>If the DOM ancestry accurately represents the level, the user agent can calculate the level of an item from the document structure. This attribute can be used to provide an explicit indication of the level when that is not possible to calculate from the document structure or the aria-owns  attribute. User agent support for automatic calculation of level may vary; authors SHOULD test with user agents and assistive technologies to determine whether this attribute is needed. If the author intends for the user agent to calculate the level, the author SHOULD omit this attribute.

So, user agents are explicitly given a pass here. Let's consider changing this language to require calculation.

smhigley:
Closes #1194

As discussed in the issue, this PR adds normative wording on requiring user agents to calculate `aria-level`, `aria-posinset`, and `aria-setsize`. It also adds a section outlining how to compute each attribute, based on the current [Core AAM wording](https://w3c.github.io/core-aam/#mapping_additional_position) that will be removed after this is merged.

# PR tracking

* Related AccName Issue/PR: N/A
* Any other dependent changes? N/A

Once this PR and all related PRs have been been approved by the working group, tests
should be written and issues should be opened on browsers. Add N/A and check when not
applicable.

* Related APG Issue/PR:  N/A
* MDN Issue/PR:  N/A
* "author MUST" tests: N/A
   * WebKit:
   * Gecko:
   * Blink:
* Does this need AT implementations?  N/A

### DIFF
--- a/index.html
+++ b/index.html
@@ -11846,7 +11846,7 @@ button.ariaPressed; // null</pre>
 				<p>This can be applied inside trees to tree items, to headings inside a document, to nested grids, nested tablists and to other structural items that might appear inside a container or participate in an ownership hierarchy. The value for <pref>aria-level</pref> is an integer greater than or equal to 1.</p>
 				<p>Levels increase with depth. If the <abbr title="Document Object Model">DOM</abbr> ancestry does not accurately represent the level, authors SHOULD explicitly define the <pref>aria-level</pref> <a>attribute</a>.</p>
 				<p>This attribute is applied to elements that act as leaf nodes within the orientation of the set, for example, on elements with role <rref>treeitem</rref> rather than elements with role <rref>group</rref>. This means that multiple elements in a set can have the same value for this attribute. Although it would be less repetitive to provide a single value on the container, restricting this to leaf nodes ensures that there is a single way for <a>assistive technologies</a> to use the attribute.</p>
-				<p>If the <abbr title="Document Object Model">DOM</abbr> ancestry accurately represents the level, the <a>user agent</a> can calculate the level of an item from the document structure. This attribute can be used to provide an explicit indication of the level when that is not possible to calculate from the document structure or the <pref>aria-owns</pref> attribute. User agent support for automatic calculation of level might vary; authors SHOULD test with [=user agents=] and assistive technologies to determine whether this attribute is needed. If the author intends for the user agent to calculate the level, the author SHOULD omit this attribute.</p>
+				<p>If the <abbr title="Document Object Model">DOM</abbr> ancestry accurately represents the level, the <a>user agent</a> can calculate the level of an item from the document structure as defined in the <a href="#tree_groups">Calculating Group Position</a> section. This attribute can be used to provide an explicit indication of the level when that is not possible to calculate from the document structure or the <pref>aria-owns</pref> attribute. User agent support for automatic calculation of level might vary; authors SHOULD test with [=user agents=] and assistive technologies to determine whether this attribute is needed. If the author intends for the user agent to calculate the level, the author SHOULD omit this attribute.</p>
                 <p class="note"> In the case of a <rref>treegrid</rref>, <pref>aria-level</pref> is supported on elements with the role <rref>row</rref>, not elements with role <rref>gridcell</rref>. At first glance, this might seem inconsistent with the application of <pref>aria-level</pref> on <rref>treeitem</rref> elements, but it is consistent in that the <rref>row</rref> acts as the leaf node within the vertical orientation of the <rref>grid</rref>, whereas the <rref>gridcell</rref> is a leaf node within the horizontal orientation of each <rref>row</rref>. Level is not supported on sets of cells within rows, so the <pref>aria-level</pref> attribute is applied to the element with the role <rref>row</rref>.</p>
 <p class="note">On elements with role <pref>heading</pref>, values for <pref>aria-level</pref> above 6 can create difficulties for users. Also, at the time of this writing, most combinations of user agents and assistive technologies only support <pref>aria-level</pref> integers 1-9 on headings.</p>
 			</div>
@@ -12258,6 +12258,7 @@ button.ariaPressed; // null</pre>
 &lt;/ul&gt;</pre>
 				<p>When specifying <pref>aria-posinset</pref>, authors MUST specify a value that is an integer greater than or equal to 1, and less than or equal to the size of the set when that size is known. If authors specify <pref>aria-posinset</pref>,  authors MUST also specify a value for <pref>aria-setsize</pref>.</p>
 				<p>When specifying <code>aria-posinset</code> on a <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, or <rref>menuitemradio</rref>, authors SHOULD set the value of <code>aria-posinset</code> with respect to the total number of items in the <rref>menu</rref>, excluding any separators.</p>
+				<p>If <code>aria-posinset</code> is not specified by the author, user agents MUST compute it as defined in the <a href="#tree_groups">Calculating Group Position</a> section.</p>
 			</div>
 			<table class="property-features">
 				<caption>Characteristics:</caption>
@@ -12897,6 +12898,7 @@ button.ariaPressed; // null</pre>
 				<p>If all items up to the current item in a set are present in the document structure, it is not necessary to set this <a>attribute</a>, as the <a>user agent</a> can automatically calculate the position for these items. However, if all previous items in the set are not present in the document structure at a given moment, the author MUST set this <a>attribute</a>to provide an explicit indication of an element's position.</p>
 				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.</p>
 				<p>When specifying <code>aria-setsize</code> on a <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, or <rref>menuitemradio</rref>, authors SHOULD set the value of <code>aria-setsize</code> based on the total number of items in the <rref>menu</rref>, excluding any separators.</p>
+				<p>If <code>aria-posinset</code> is not specified by the author, user agents MUST compute it as defined in the <a href="#tree_groups">Calculating Group Position</a> section.</p>
 				<p>The following example shows items 5 through 8 in a set of 16.</p>
 				<pre class="example highlight">&lt;h2 id="label_fruit"&gt; Available Fruit &lt;/h2&gt;
 &lt;ul role="listbox" aria-labelledby="label_fruit"&gt;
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -11846,7 +11846,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>If the <abbr title="Document Object Model">DOM</abbr> ancestry accurately represents the level, the <a>user agent</a> can calculate the level of an item from the document structure as defined in the <a href="#tree_groups">Calculating Group Position</a> section. This attribute can be used to provide an explicit indication of the level when that is not possible to calculate from the document structure or the <pref>aria-owns</pref> attribute. User agent support for automatic calculation of level might vary; authors SHOULD test with [=user agents=] and assistive technologies to determine whether this attribute is needed. If the author intends for the user agent to calculate the level, the author SHOULD omit this attribute.</p>
```
cookiecrook, line 11849: VoiceOver assumes no level on headings when not supplied by an author, and just speaks "heading" not the more verbose "heading level 0" or an arbitrary "heading level 2." I worry this change will advise AT to verbosely convey info that the author may not have intended.

pkra: A quick test indicates that Chrome and Firefox appear to calculate a level of 2 for a role=heading with no level. 

While NVDA and JAWS expose that level in speech and braille, Orca does not.

pkra, line 11849:
```suggestion
				<p>Unless this property is required by , If the <abbr title="Document Object Model">DOM</abbr> ancestry accurately represents the level, the <a>user agent</a> can calculate the level of an item from the document structure as defined in the  section. This attribute can be used to provide an explicit indication of the level when that is not possible to calculate from the document structure or the <pref>aria-owns</pref> attribute. User agent support for automatic calculation of level might vary; authors SHOULD test with [=user agents=] and assistive technologies to determine whether this attribute is needed. If the author intends for the user agent to calculate the level, the author SHOULD omit this attribute.</p>
```

The DOM order is mentioned and then aria-owns as an example of things being problematic - but the algorithm only operates on the accessibility tree.  (The other properties also reference the DOM which seems to me should change.)

Something like 

> Unless required by a role, authors SHOULD not use this property; instead authors SHOULD ensure that <a href="#tree_groups">Calculating Group Position</a> works correctly. If that is not the case, this property MAY be used to provide explicit information of the level.

These properties nowadays almost feel like a case for #1480 to me.

index.html
```diff
@@ -11846,7 +11846,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				<p>This can be applied inside trees to tree items, to headings inside a document, to nested grids, nested tablists and to other structural items that might appear inside a container or participate in an ownership hierarchy. The value for <pref>aria-level</pref> is an integer greater than or equal to 1.</p>
 				<p>Levels increase with depth. If the <abbr title="Document Object Model">DOM</abbr> ancestry does not accurately represent the level, authors SHOULD explicitly define the <pref>aria-level</pref> <a>attribute</a>.</p>
```
pkra, line 11847: I think we should remove the second sentence (see my other comment).

index.html
```diff
@@ -12258,6 +12258,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 &lt;/ul&gt;</pre>
 				<p>When specifying <pref>aria-posinset</pref>, authors MUST specify a value that is an integer greater than or equal to 1, and less than or equal to the size of the set when that size is known. If authors specify <pref>aria-posinset</pref>,  authors MUST also specify a value for <pref>aria-setsize</pref>.</p>
 				<p>When specifying <code>aria-posinset</code> on a <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, or <rref>menuitemradio</rref>, authors SHOULD set the value of <code>aria-posinset</code> with respect to the total number of items in the <rref>menu</rref>, excluding any separators.</p>
+				<p>If <code>aria-posinset</code> is not specified by the author, user agents MUST compute it as defined in the <a href="#tree_groups">Calculating Group Position</a> section.</p>
```
pkra, line 12261: Could we make replace the second sentence of the paragraph of this section with this? And then update the second paragraph similarly to my comment on aria-level?

index.html
```diff
@@ -12897,6 +12898,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 				<p>If all items up to the current item in a set are present in the document structure, it is not necessary to set this <a>attribute</a>, as the <a>user agent</a> can automatically calculate the position for these items. However, if all previous items in the set are not present in the document structure at a given moment, the author MUST set this <a>attribute</a>to provide an explicit indication of an element's position.</p>
 				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.</p>
 				<p>When specifying <code>aria-setsize</code> on a <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, or <rref>menuitemradio</rref>, authors SHOULD set the value of <code>aria-setsize</code> based on the total number of items in the <rref>menu</rref>, excluding any separators.</p>
+				<p>If <code>aria-posinset</code> is not specified by the author, user agents MUST compute it as defined in the <a href="#tree_groups">Calculating Group Position</a> section.</p>
```
pkra, line 12901: I would suggest the same as with posinset.

index.html
```diff
@@ -13278,6 +13280,36 @@ <h2>Relationships in the Accessibility Tree</h2>
   &lt;div role="listitem"&gt;The "list" is my accessibility parent.&lt;/div&gt;
 &lt;/div&gt;
           </pre>
+
+		  <section id="tree_groups">
+			<h3>Calculating Group Position in the Accessibility Tree</h3>
+			<p>When <pref>aria-level</pref>, <pref>aria-posinset</pref>, or <pref>aria-setsize</pref> are defined on supported roles, they allow authors to override the default group position and level calculation. When those attributes are not specified by the author, user agents MUST compute them following the rules in this section.</p>
+
+			<p>If <pref>aria-level</pref> is not specified on <rref>comment</rref>, <rref>row</rref> within a <rref>treegrid</rref>, or <rref>treeitem</rref>, user agents MUST compute the level as follows:</p>
```
pkra, line 13288: I don't like listing roles but I don't see a good alternative. The best I can come up with "when not required by the role".

index.html
```diff
@@ -13278,6 +13280,36 @@ <h2>Relationships in the Accessibility Tree</h2>
...
+
+			<p>If <pref>aria-posinset</pref> or <pref>aria-setsize</pref> is not specified on any role the supports it with the exception of <rref>article</rref>, user agents MUST compute them as follows:</p>
+			<ul>
+				<li>For <rref>treeitem</rref> and <rref>comment</rref>: walk the accessibility tree backward and forward until the explicit or computed level becomes less than the current item's level. Count items only if they are at the same level as the current item.</li>
+				<li>For all other supported roles: process the accessibility parent, counting all accessibility children that have the same role.</li>
+				<li>For <pref>aria-posinset</pref>, return 1 + the number of group items before the current item in the accessibility tree.</li>
+				<li>For <pref>aria-setsize</pref>, return the total number of group items, including the current item.</li>
+			</ul>
+
+			<p>If the author provides one or more of <pref>aria-setsize</pref> and <pref>aria-posinset</pref>, it is the author's responsibility to supply them for all elements in the set. User agent correction of missing values in this case is not defined.</p>
```
pkra, line 13309: > "it is the author's responsibility to"

Maybe: "authors MUST"?

> User agent correction of missing values in this case is not defined.

Are UAs ok with this?

index.html
pkra: Also: copy&paste error - "aria-posinset" should be "aria-setsize" here.

pkra (changes requested):
I like the new section on calculating. I'm just not certain how to integrate it into the rest of the spec.

Reviewing this as "request changes" for just the two minor issues (the copy&paste error and the -1 issue at the very end.)

index.html
cookiecrook: I think it's unfortunate the didn't catch that this ARIA 1.1 spec change increased SR verbosity unnecessarily, but since we didn't, and the implementations have moved on, I will forego my worry. It's not so problematic a user change that we should break the IDL reflection and force multiple implementations to revert it.

spectranaut:
Discussed in today's ARIA meeting: https://www.w3.org/2025/02/06-aria-minutes#3ad8

pkra:
@smhigley unfortunately, this disconnects the old comments from the lines in the rebased commit. I hope the thread is sufficient to reconstruct.

spectranaut:
@smhigley I think you never got to @pkra review on this is one?

and @pkra, you added agenda but then we didn't end up discussing this. Do we still need to discuss?

pkra:
> and @pkra, you added agenda but then we didn't end up discussing this. Do we still need to discuss?

Not from my end.

smhigley:
@rahimabdi wanted your review here to make sure this doesn't add any snags for IDL mappings (based on what was said in the call I think this may not affect IDL, but would be nice to get your review to confirm)

spectranaut:
Link to discussion from last weeks meeting: https://www.w3.org/2025/06/26-aria-minutes.html#b9b5

index.html
```diff
@@ -13278,6 +13280,36 @@ <h2>Relationships in the Accessibility Tree</h2>
...
+
+			<p><pref>aria-level</pref>, <pref>aria-posinset</pref>, and <pref>aria-setsize</pref> are all 1-based. When the property is not present or is "0", it indicates the property is not computed or not supported. If any of these properties are specified by the author as either "0" or a negative number, user agents SHOULD use "1" instead.</p>
```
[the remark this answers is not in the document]
rahimabdi: @scottaohara's PR for addressing a value of `-1` (indeterminate): https://github.com/w3c/aria/pull/2341

index.html
```diff
@@ -16230,6 +16232,49 @@ <h2>Relationships in the Accessibility Tree</h2>
...
+            <li>For all other supported roles: process the accessibility parent, counting all accessibility children that have the same role.</li>
+            <li>For <pref>aria-posinset</pref>, return 1 + the number of group items before the current item in the accessibility tree.</li>
+            <li>For <pref>aria-setsize</pref>, return the total number of group items, including the current item.</li>
+          </ul>
+          <p>
+            If the author provides one or more of <pref>aria-setsize</pref> and <pref>aria-posinset</pref>, it is the author's responsibility to supply them for all elements in the set. User agent
+            correction of missing values in this case is not defined.
+          </p>
+          <p>
+            <pref>aria-level</pref>, <pref>aria-posinset</pref>, and <pref>aria-setsize</pref> are all 1-based. When the property is not present or is "0", it indicates the property is not computed or
```
rahimabdi, line 16274: I raised a core-aam issue that this isn't necessarily the case for WebKit (for the reasons @cookiecrook mentioned on the [June 26 2025 ARIA WG](https://www.w3.org/2025/06/26-aria-minutes.html#b9b5) meeting: https://github.com/w3c/core-aam/issues/249

rahimabdi (commented):
Agreed with @keithamus, I don't anticipate any IDL impact however, it does raise the question of whether ARIA should add additional numeric IDL counterparts for these attributes. Similar to the `class` content attribute supporting dual IDL attributes (`className`, `classList`), `aria-level`/`aria-posinset`/`aria-setsize` could be supplied with an additional, author-friendly IDL attribute that is truly numeric:

For example, each attribute could get a secondary IDL attribute but the original one would remain. Numeric IDL attributes also have a "[default value](https://html.spec.whatwg.org/#default-value)" (similar to a missing value default) that could be specified for the numeric IDL counterparts:
- `ariaLevel`
    - currently string reflects as DOMString?
    - 🆕 `ariaLevelNumeric` could reflect as an unsigned Long (default value = 0)
- `ariaPosInSet`
    - currently string reflects as nullable DOMString?
    - 🆕 `ariaPosInSetNumeric` could reflect as unsigned Long (default value = 1)
- `ariaSetSize`
    - currently string reflects as nullable  DOMString?  
    - 🆕 `ariaSetSizeNumeric` could reflect as Long (default value = -1)

PR #2114, opened 2024-01-30
Remove abbr elements

### LINKED ISSUE #2113: Remove abbr elements from site
Devtools suggests there are over 700 `abbr` elements in the current 1.2 spec. [Recent research](https://adrianroselli.com/2024/01/using-abbr-element-with-title-attribute.html) shows that these have little-to-no benefit to the user, with most assistive technology ignoring the element. Some assistive technology does announce the contents of the `abbr` element's title attribute, which then, with so many elements on the page, creates a very verbose-sounding document.

The best practice for expanding abbreviations is to expand them in text the first time they are used in a document. For example: "BBC (British Broadcasting Corporation)".

This PR removes all the `abbr` elements I could find and expands each abbreviation the first time I found it on the page.

I was looking at the current 1.3 [Editor's Draft](https://w3c.github.io/aria/).

fstrr:
Closes #2113

This PR removes `abbr` elements from the site. The linked issue contains a description of the issue.


pkra:
Closing this, primarily because of https://github.com/w3c/aria/issues/2113#issuecomment-1931854063 

But also because it already has merge conflicts and will only get more as it sits. And finally, because files in `./common` must be changed in the aria- common repository.

Issue #2158, opened 2024-04-15
Allow posinset/setsize for toolbars


scottaohara:
This is a follow on issue to #1721, specifically [this comment/discussion in the minutes](https://github.com/w3c/aria/issues/1721#issuecomment-1883679522) about allowing toolbars to either implicitly add an X of Y to controls within - or to allow authors to add an X of Y to controls within a toolbar via aria-posinset and aria-setsize.

PR #2159, opened 2024-04-15
Editorial: note to expand radiogroup content 'allowances'

### LINKED ISSUE #1721: Consider a mechanism to associate controls without an explicit grouping
This is semi-related to my previous issue #1442 as the impetus for this issue is the fact that HTML allows for radio buttons to be associated with each other without the need of a parent grouping element (e.g., via the `name` attribute having the same value).

The reason this is related to the previously mentioned issue is that with ARIA, the only way to indicate that choosing a radio button is required in a set of radio buttons is for that to be declared on the radio group parent. 

However, consider the following markup:

```html
<label><input type=radio name=c required>Choice 1</label>
content / additional form fields that are related to choice 1
...
<label><input type=radio name=c>Choice 2</label>
content / additional form fields that are related to choice 2
...
<label><input type=radio name=c>Choice 3</label>
content / additional form fields that are related to choice 3
```

The above wouldn't neatly fit into a `radiogroup`, as the layout requires additional content and controls that don't make sense (? or at least not to me/it isn't clear per the spec that they'd make sense either) to be children of a group that is specifically defined as being for radios.  The suggestion that ARIA makes to use `aria-owns` to revise the DOM hierarchy to force-include radio buttons within a `radiogroup` would be problematic here as the programmatic reading order of the above markup is important, and if that were altered then it'd be exposed as if it was marked up as:

```html
<fieldset role=radiogroup aria-label=foo aria-required=true>
  <label><input type=radio name=c>Choice 1</label>
  <label><input type=radio name=c>Choice 2</label>
  <label><input type=radio name=c>Choice 3</label>
</fieldste>

content / additional form fields that are related to choice 1
...
content / additional form fields that are related to choice 2
...
content / additional form fields that are related to choice 3
```
where now the radio buttons are all programmatically exposed in a single grouping (for the browsers that support `aria-owns`), but now the content that previously relied on the radio buttons being in the programmatic reading order prior to the content they introduced is lost. Additional remediation efforts would be needed to re-introduced to delineate the boundaries of each choice's content/further controls.

---

Sure, we could just not use ARIA radios / radiogroup here and rely on HTML alone... but this seems like something that ARIA should have a mechanism for (attribute parity), and arguably it could be useful in other situations too.  E.g., identifying that a series of radio buttons or standard buttons part of an implicit group by announcing an x of y numeration to them.

scottaohara:
This PR attempts to close #1721

Per the discussion we had around the issue, the potential usability and wcag issues this could cause, there were few pros to relaxing how one could create a grouping of radio buttons.  The issue of how would one name such a group if there is no actual "group" container to name was a very good point.  Though HTML allows for such loose grouping of their radio buttons, they do not have a solution for all the potential a11y problems it can solve.

So, this PR takes a different approach to resolving the issue, and instead adds a note to clarify that a radiogroup does not have to be limited to just radio buttons, while ALSO pointing out the potential UX issues of adding other arbitrary/focusable elements.

Maybe something could even be added to overtly state that "hey if you want to group radio buttons like you can in HTML, then use HTML radio buttons!"

Happy to wordsmith this further (examples might be good?)  Or entertain other solutions (we could close the issue as won't fix, but i recall being told we should do _something_ rather than just that).

I created https://github.com/w3c/aria/issues/2158 as a separate issue to resolve, since a lot of the conversation about this concept touched on the topic now logged in that issue.  and I didn't want resolving this issue/pr to lose that separate though related topic.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6358,6 +6358,15 @@
 			<div class="role-description">
 				<p>A checkable input in a group of elements with the same role, only one of which can be checked at a time.</p>
 				<p>Authors SHOULD ensure that [=elements=] with role <code>radio</code> are explicitly grouped in order to indicate which ones affect the same value. This is achieved by enclosing the radio elements in an element with role <rref>radiogroup</rref>. If it is not possible to make the radio buttons <abbr title="Document Object Model">DOM</abbr> children of the <rref>radiogroup</rref>, authors SHOULD use the <pref>aria-owns</pref> <a>attribute</a> on the <rref>radiogroup</rref> element to indicate the <a>relationship</a> to its children.</p>
+				<div class="note" title="Radiogroups containing other non-radio content">
+					<p>While a <code>radiogroup</code> is primarily meant to group and thus associate related <code>radio</code> buttons, a <code>radiogroup</code> can contain other
+					non-<code>radio</code> button elements. For instance, descriptive text beyond what serves as the radio button's label, or even in some cases, form controls 
+					that are enabled only when a specific radio button has been chosen (checked).</p>
+					<p>However, including non-radio button content into a <code>radiogroup</code> is not without its potential UX drawbacks. As it is generally expected
+					to use arrow keys to navigate between radio buttons, any non-radio button content could go unnoticed by some users, as arrowing through the radio buttons
+					would mean that any other tabbable elements (e.g., hyperlinks or other form fields) would be skipped. When creating a <code>radiogroup</code> with other
+					arbitrary content, consider if the non-radio content could be just as impactful if it was presented as an immediate sibling <em>after</em> the <code>radiogroup</code>.</p>
+				</div>
 			</div>
 			<table class="role-features">
 				<caption>Characteristics:</caption>


index.html
```diff
@@ -6455,6 +6455,15 @@ <h5>Presentational Role Inheritance</h5>
...
 				<!-- keep previous sentence synced with menuitemradio, etc. -->
+				<div class="note" title="Radiogroups containing other non-radio content">
+					<p>While a <code>radiogroup</code> is primarily meant to group and thus associate related <code>radio</code> buttons, a <code>radiogroup</code> can contain other
+					non-<code>radio</code> button elements. For instance, descriptive text beyond what serves as the radio button's label, or even in some cases, form controls 
+					that are enabled only when a specific radio button has been chosen (checked).</p>
+					<p>However, including non-radio button content into a <code>radiogroup</code> is not without its potential UX drawbacks. As it is generally expected
+					to use arrow keys to navigate between radio buttons, any non-radio button content could go unnoticed by some users, as arrowing through the radio buttons
+					would mean that any other tabbable elements (e.g., hyperlinks or other form fields) would be skipped. When creating a <code>radiogroup</code> with other
```
rahimabdi, line 6464: > As it is generally expected to use arrow keys to navigate between radio buttons, any non-radio button content could go unnoticed by some users, as arrowing through the radio buttons would mean that any other tabbable elements (e.g., hyperlinks or other form fields) would be skipped. 

A couple questions/comments:

- If interactive, non-radio button content (e.g., an additional field that appears on radio button selection) is also nested within the radiogroup, would this require extra work to keep track of focus? It sounds like this results in additional complexity for keyboard operability, i.e., when a radio button is focused, focus should move to the next tabbable element inside the radiogroup (if present) and not the next focusable element immediately after the radio group content
- For additional non-radio content that appears on focus of a radio button, it sounds like this is OK provided the content appears after the current point of focus?
- I'm also curious if this could break cardinality and AT reporting of how many items are in the radiogroup. Perhaps ATs can compensate for additional non-radio button content by adding the extra non-radio button stuff to ensure accurate setsize/posinset? I agree this should be discouraged for the reasons you've mentioned (and that it would be jarring for AT users to get kicked out of forms/focus mode when navigating non-interactive, non-radio button content)

pkra:
Rebased on main

daniel-montalvo:
Conflicts resolved!

index.html
```diff
@@ -7815,6 +7815,19 @@ <h5>Presentational Role Inheritance</h5>
               Authors SHOULD provide an accessible name for a <code>radiogroup</code>, which can be done with the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attribute. Authors SHOULD
               reference a visible label with <pref>aria-labelledby</pref> if a visible label is present for the <code>radiogroup</code>.
             </p>
+            <div class="note" title="Radiogroups containing other non-radio content">
+              <p>
+                While a <code>radiogroup</code> is primarily meant to group and thus associate related <code>radio</code> buttons, a <code>radiogroup</code> can contain other non-<code>radio</code>
+                button elements. For instance, descriptive text beyond what serves as the radio button's label, or even in some cases, form controls that are enabled only when a specific radio button
```
scottaohara:
```suggestion
                button elements. For instance, descriptive text beyond what serves as the radio button's label, or even in some cases, form controls or other information that are enabled, revealed or both when a specific radio button
```

index.html
```diff
@@ -7815,6 +7815,19 @@ <h5>Presentational Role Inheritance</h5>
...
+              <p>
+                While a <code>radiogroup</code> is primarily meant to group and thus associate related <code>radio</code> buttons, a <code>radiogroup</code> can contain other non-<code>radio</code>
+                button elements. For instance, descriptive text beyond what serves as the radio button's label, or even in some cases, form controls or other information that are enabled, revealed or both when a specific radio button
+                has been chosen (checked).
+              </p>
+              <p>
+                However, including non-radio button content into a <code>radiogroup</code> is not without its potential UX drawbacks. As it is generally expected to use arrow keys to navigate between
+                radio buttons, any non-radio button content could go unnoticed by some users, as arrowing through the radio buttons would mean that any other tabbable elements (e.g., hyperlinks or
+                other form fields) would be skipped. When creating a <code>radiogroup</code> with otherarbitrary content, consider if the non-radio content could be just as impactful if it was
```
```suggestion
                other form fields) would be skipped. When creating a <code>radiogroup</code> with other arbitrary content, consider if the non-radio content could be just as impactful if it was
```

Issue #2156, opened 2024-04-12
Definition discrepancies for the "undefined" value across ARIA states and properties


rahimabdi:
# Describe the change

As of ARIA 1.3, `aria-hidden="false"` will be a synonym of `aria-hidden="undefined"`. However, I noticed that the ["undefined"](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2090.html#valuetype_true-false-undefined) value is now differently defined for various states and properties:

- `aria-checked="undefined"` means that "The element does not support being checked."
-  `aria-expanded="undefined"` means that "The element does not own or control a grouping element that is expandable."
- `aria-orientation="undefined"` means that "The element's orientation is unknown/ambiguous." 

For `aria-hidden="undefined"`, it equates to "The element's [hidden](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2090.html#dfn-hidden) state is determined by the user agent based on whether it is rendered."

With the changes to aria-hidden, should we update section [6.2.4 Value](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2090.html#valuetype_true-false-undefined) and provide a broad definition of what "undefined" means, or perhaps note that it is property-dependent?

# Link to the version of the specification or documentation you were looking at

Link to documentation: https://w3c.github.io/aria/

Does the issue exists in [the editors draft](https://w3c.github.io/aria/) (the editors draft is the most recent draft of the specification)?

Yes

KateZhaoTR:
Maybe the "not applicable" in the parenthesis could be replaced with the wording and examples you've provided, something like:

true/false/undefined
Value representing true, false, or undefined (The meaning of 'undefined' is attribute dependent. For example, aria-checked="undefined" means that "The element does not support being checked", while aria-hidden="undefined" is a synonym for aria-hidden="false".). The default value for this value type is undefined unless otherwise specified. For example, an element with [aria-expanded](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2090.html#aria-expanded) set to false is not currently expanded; an element with [aria-expanded](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2090.html#aria-expanded) set to undefined is not expandable.

spectranaut:
@rahimabdi are you interested in drafting a PR with the definition of undefined you suggest?

rahimabdi:
@spectranaut Yes, happy to do so!

rahimabdi:
[...]

PR #2172, opened 2024-04-28
Update glossary definition for "undefined"

rahimabdi:
Closes https://github.com/w3c/aria/issues/2156

This PR clarifies ambiguity in the varying behaviors of the `undefined` value for ARIA states and properties by providing a broader definition in the glossary.

### DIFF
--- a/index.html
+++ b/index.html
@@ -10013,7 +10013,7 @@
 				<dt id="valuetype_tristate">tristate</dt>
 				<dd>Value representing <code>true</code>, <code>false</code>, <code>mixed</code>, or <code>undefined</code> values. The default value for this value type is <code>undefined</code> unless otherwise specified.</dd>
 				<dt id="valuetype_true-false-undefined">true/false/undefined</dt>
-				<dd>Value representing <code>true</code>, <code>false</code>, or <code>undefined</code> (not applicable). The default value for this value type is <code>undefined</code> unless otherwise specified. For example, an element with <sref>aria-expanded</sref> set to <code>false</code> is not currently expanded; an element with <sref>aria-expanded</sref> set to <code>undefined</code> is not expandable.</dd>
+				<dd>Value representing <code>true</code>, <code>false</code>, or <code>undefined</code> (not applicable). Unless otherwise specified, the default value for this value type is <code>undefined</code>. Note that the value <code>undefined</code> conveys a state- or property-dependent behavior including but not limited to lack of support (e.g., <sref>aria-checked</sref>, <sref>aria-pressed</sref>), equivalence with the value <code>false</code> (e.g., <sref>aria-multiselectable</sref>), behavior determined by the user agent (e.g., <sref>aria-hidden</sref>) or non-applicability (e.g., <sref>aria-selected</sref>).
 				<dt id="valuetype_idref">ID reference</dt>
 				<dd>Reference to the ID of another <a>element</a> in the same document</dd>
 				<dt id="valuetype_idref_list">ID reference list</dt>


index.html
```diff
@@ -10013,7 +10013,7 @@ <h3>Value</h3>
...
 				<dt id="valuetype_true-false-undefined">true/false/undefined</dt>
-				<dd>Value representing <code>true</code>, <code>false</code>, or <code>undefined</code> (not applicable). The default value for this value type is <code>undefined</code> unless otherwise specified. For example, an element with <sref>aria-expanded</sref> set to <code>false</code> is not currently expanded; an element with <sref>aria-expanded</sref> set to <code>undefined</code> is not expandable.</dd>
+				<dd>Value representing <code>true</code>, <code>false</code>, or <code>undefined</code> (not applicable). Unless otherwise specified, the default value for this value type is <code>undefined</code>. Note that the value <code>undefined</code> conveys a state- or property-dependent behavior including but not limited to lack of support (e.g., <sref>aria-checked</sref>, <sref>aria-pressed</sref>), equivalence with the value <code>false</code> (e.g., <sref>aria-multiselectable</sref>), behavior determined by the user agent (e.g., <sref>aria-hidden</sref>) or non-applicability (e.g., <sref>aria-selected</sref>).
```
spectranaut, line 10016: Wow this really shows how inconsistent we have been with the use of undefined :sweat_smile: 

I wonder if we should include the examples, or leave it at `Note that the value undefined conveys a state- or property-dependent behavior, consult the values table of the relevant state or property to understand the meaning of an undefined state or property.`

spectranaut (commented):
I also wonder if we need to make more clear that an "undefined" value does not mean the string "undefined", like you brought up in the meeting today (during new PR triage: https://www.w3.org/2024/05/02-aria-minutes.html#t02). Are you making a separate issue for this specifically? Depending on the solution there, it might affect this PR.

Commenting for now because I haven't made up my mind about whether it's ok to explicitly list the examples, or if we should just point people to the tables, I wonder what the other reviewers think.

index.html
smhigley, line 10016: I think it's actually even more complex than this :/. For example, `el.ariaChecked = undefined` is going to default to `false` for most roles, only excluding `option` and `treeitem`. I think it may be better to not reference specific attributes here. Even for something like `aria-pressed`, if HTML ever ends up creating a toggle button element, then `aria-pressed` might also conditionally default to false (and yeah, an element like that shouldn't use ARIA for that state, but if undefined it would still default to a boolean). It might be simpler to stick with only one relatively safe example (e.g. the `aria-expanded` one in the original text), then clarify that for specific behavior, authors should check the tables for that attribute?

I do also think I agree with Valerie that it'd be helpful to clarify that `undefined` here refers to the value, which is set by either leaving the attribute off entirely or explicitly setting e.g. `element.ariaExpanded = undefined`, and is not equivalent to the string `"undefined"` in e.g. `aria-expanded="undefined"`. What do you think?

aardrian: I think "undefined" needs to be yanked out of the `<code>` unless it explicitly means the string value "undefined".

As a _user_ of the spec, this is what I might expect (fake example) so I could make sense of it:
> Values include <code>yeah</code>, <code>nah</code>, <code>undefined</code>, or undefined (by setting no value).

aardrian (commented):
Sarah and Valerie hit the points I wanted to hit.

index.html
smhigley: @aardrian would it help if this section had a link to the IDL reflection section just after it, in addition to text clarifying that `undefined` is not the string?

It does get weirder, in that while you can explicitly set e.g. `el.ariaThing = undefined`, getting it returns `null` 🤷🏻‍♀️

rahimabdi: Thank you @aardrian @smhigley. As @spectranaut noted, it's currently ambiguous if the value undefined is the string "undefined" or undefined as in not present. For example, see Note 2 under [aria-hidden](https://w3c.github.io/aria/#aria-hidden):

As of ARIA 1.3, [aria-hidden](https://w3c.github.io/aria/#aria-hidden)="false" is now synonymous with aria-hidden="undefined".

rahimabdi:
Awaiting resolution on https://github.com/w3c/aria/issues/2177 before updating PR with new `undefined` definition.

index.html
smhigley: @rahimabdi yeah, I believe that's a pretty straightforward mistake in the spec. I made a test page to output the browser-mapped values when ARIA attributes are set to the string `"undefined"`, not set at all, or set programmatically to `undefined`: https://jsfiddle.net/pobwvgs9/

I checked it in Chrome, Edge, Safari, and Firefox on macOS and Windows (Safari only on mac), and all of them had different results for the string `"undefined"`. The results for not-defined and programmatic `= undefined` values are what I'd expect & what matches the spec, while the string `"undefined"` are not.

My proposal would be to go through the spec and remove any instances accidentally using string `="undefined"`, and add the clarification of the string not being the same as the value.

As an aside, I do get an error for `="undefined"` in the W3C Validator, maybe they have some bugs in the attributes you tested?
Screenshot of the HTML validator having been run on a button with aria-hidden=undefined and aria-pressed=foo. The validator shows one error for each attribute.

rahimabdi:
Almost there!

@smhigley @aardrian you mentioned adding a link to the IDL reflection section ([section 10. IDL Interface](https://w3c.github.io/aria/#idl-interface)) although I'm not seeing any immediate relation to `undefined` assignment. [Section 6.3 ARIA Attributes](https://deploy-preview-2172--wai-aria.netlify.app/#aria-attributes) does provide examples of IDL assignment but please let me know which link you prefer I include. Thank you!

rahimabdi:
@smockle Yes; I plan to make additional changes to this PR to align with a set of incoming, related ARIA IDL updates.

PR #2181, opened 2024-05-20
Expectations for aria-hidden and focused elements

### LINKED ISSUE #1765: Consider further calling out aria-hidden expectations for focusable elements
the definition of `aria-hidden` does not make it clear that content - e.g., focusable elements - can/should/are still exposed to AT when focused.

https://www.w3.org/TR/wai-aria-1.2/#tree_inclusion DOES call out that elements can be exposed even if they are set to aria-hidden=true / focused

That's really it though, in regards to informing authors that `aria-hidden=true` has been modified over time to purposefully mitigate against author error.

To that point, should there be additional text added to the `aria-hidden` definition to call out this reality?  Should this be explicitly specified in the author error section?  

It would sure help make the reality of the situation a lot easier to explain.

scottaohara:
This PR closes #1765 and is related to work that was done in #2037, but scoped only to the original issue I filed.

The intent of this PR is to identify not only how user agents would need to handle focusable elements that are aria-hidden (explicitly or due to being a descendant of an aria-hidden container) - but for the case where a focusable element is within an aria-hidden container, that the entire subtree would need to be re-exposed so that any other relevant information to the user could be made available.  (e.g., so as to not just expose a "learn more" link, with no way to determine what someone would be learning about)

a simple example being like:

```
<div aria-hidden=true>
  <h3>Something or other</h3>
   some details about said something, or other.
  <a href=#>Learn more!</a>
</div>
```

Once this PR and all related PRs have been  approved by the working group, tests
should be written and issues should be opened on browsers. Add N/A and check when not
applicable.

* [ ] MDN Issue/PR: could use?
* [X] "author MUST" tests: n/a, no new testable authors requirements 
* [ ] "user agent MUST" tests: Todo.
   * WebKit: Todo.
   * Gecko: Todo.
   * Blink: Partial, see https://github.com/w3c/aria/pull/2181#issuecomment-4041559160
* [X] Does this need AT implementations? n/a

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2181.html" title="Last updated on Jan 30, 2025, 6:55 PM UTC (ac39838)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2181/a2d34fe...ac39838.html" title="Last updated on Jan 30, 2025, 6:55 PM UTC (ac39838)">Diff</a>


index.html
```diff
@@ -11604,6 +11604,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p>Authors MAY, with caution, use aria-hidden to hide visibly rendered content from assistive technologies <em>only</em> if the act of hiding this content is intended to improve the experience for users of assistive technologies by removing redundant or extraneous content. Authors using aria-hidden to hide visible content MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.</p>
+				<p>Additionally, authors SHOULD ensure that any elements that are accessibility descendants of an aria-hidden element, or any elements which have been marked as aria-hidden themselves, are prevented from receiving focus. If an aria-hidden element or a descendant of an aria-hidden ancestor receives focus, User Agents MUST ignore the aria-hidden state of any ancestor to the focused element, resulting in the entire subtree to be exposed to assistive technologies from that point forward, even once the element loses focus or is no longer focusable.</p>
```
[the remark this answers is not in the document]
scottaohara: i know this section is already note-heavy, but i was debating on whether it was worth mentioning that authors SHOULD consider using host language features, like `inert` or `display: none` to properly hide content and thus mitigate needing to worry about this at all.

aleventhal:
That seems like a good idea, and also, to not use aria-hidden if something
is already inert / display-none etc.

rahimabdi:
This may be WPT-testable; should I create a separate issue if you'd like tests for this (the MUST around the subtree being re-exposed)?

index.html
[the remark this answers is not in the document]
rahimabdi: +1, section 7.1 lists these techniques as well: https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion:

> Elements, including their descendent elements, that have host language semantics specifying that the element is not displayed, such as CSS display:none, visibility:hidden, or the HTML hidden attribute.

scottaohara:
talking with @smhigley about this... but should this clarify that if an element is already focused and its container becomes aria-hidden, then the aria-hidden state should remain unless focus is then moved to another element within that hidden subtree.

for instance, think of a custom modal dialog that was built where the order of events are:
- button invokes dialog (focus is on button)
- element containing button set to aria-hidden=true
- dialog is rendered (outside aria-hidden=true container)
- focus moves to dialog

since focus moving is the last thing that occurs, we do not want browsers to immediately undo the aria-hidden that was dynamically added (though focus is still on the button inside the aria-hidden container)

scottaohara:
need to add to this that user agents need to make this repair if the element focused is in the same document - but say if the element that is focused is within an iframe, but the iframe is hidden from an aria-hidden surrounding it in the parent doc, then browsers may not be able to reasonably make this repair.

smhigley:
I have one other question, but mostly for browser implementors; I'm not sure if this amount of detail can be in the spec --

Would the removal of `aria-hidden` occur after the `focus` event on an element, or after `document.activeElement` is set to point to a hidden element? I'm making the distinction because in a lot of our focus management logic for things like dialogs, we look for a focus event on a hidden element (either a focus bumper, or looking for focus moving to the background page), and immediately send focus back into the dialog in the event handler. The result is that `document.activeElement` never points to a hidden element, but the `focus` event does fire. If the change happened as a result of the event, it would break quite a lot of things at least in our library, and IIRC in others as well 😅.

If that's not the case, and if the scenario that @scottaohara already mentioned is also handled, I think I wouldn't have any more concerns about this.

aleventhal:
We aren't actually changing the DOM, aka doing a removeAttribute(). We are
just adding that node to a list of nodes where we will ignore it's
aria-hidden markup from now on.
Because of this, there shouldn't be any ordering questions. It means that
in the a11y update that contains the focus event, the subtree's content
will be there unhidden.

smhigley:
@aleventhal I think I may not have communicated my use case well enough 😅.

The distinction between doing a `.removeAttribute()` and ignoring the attribute through some other method wasn't really material to the use case I had in mind, though bringing it up does make me wonder whether there will be any way for developers to debug what's going on.

For my original question, I made this JS Fiddle to help: https://jsfiddle.net/sjc89htb/. It uses logic similar to what we use at Microsoft, and what I've seen for focus traps elsewhere, where focus is immediately sent elsewhere when it lands on a hidden element. In practice this works perfectly in all ATs I've tested up till now. Would you be able to clarify whether it will continue working with this change, or whether the change will result in the hidden content being exposed?

Thanks for taking the time to investigate and weigh in!

index.html
adampage, line 11607:
```suggestion
				<p>Additionally, authors SHOULD prevent a focusable element from receiving focus if it or any of its accessibility ancestors have been <a href="#tree_exclusion">excluded from the accessibility tree</a>. If a focusable element receives focus while [=element/hidden=], User Agents MUST re-expose it and each of its [=element/hidden=] ancestors to the accessibility tree, and persist that exposure even after the triggering element loses focus or is no longer focusable.</p>
```

For the first sentence, here’s an editorial suggestion that reads a little more clearly to me:

> Additionally, authors SHOULD prevent a focusable element from receiving focus if it or any of its accessibility ancestors have been [excluded from the accessibility tree](https://www.w3.org/TR/wai-aria-1.3/#tree_exclusion).

In the second sentence, “_entire subtree_” sounds greedy. In this contrived example, if `#link` receives focus, I’m supposing we’d expect `#a` to get exposed (and therefore also `#b` and `#link`), but not `#c`:

```html
<div id="a" aria-hidden="true">
  <p id="b">Some stuff that is only hidden by virtue of its hidden parent.</p>
  <a id="link" href="#">I’m hidden but focusable</button>
  <p id="c" aria-hidden="true">A sibling to #link that is coincidentally but explicitly hidden.</p>
</div>
```

If that’s true, then how about this rephrasing?

> If a focusable element receives focus while [hidden](https://www.w3.org/TR/wai-aria-1.3/#dfn-hidden), User Agents MUST re-expose it and each of its [hidden](https://www.w3.org/TR/wai-aria-1.3/#dfn-hidden) ancestors to the accessibility tree, and persist that exposure even after the triggering element loses focus or is no longer focusable.

adampage (changes requested):

index.html
aleventhal: We don't just expose the ancestor chain, but rather the entire subtree under the aria-hidden (basically the aria-hidden markup on the ancestor is marked as problematic and ignored by the browser, which can affect the entire subtree). Of course, there could be additional markup in some parts of the subtree that also hide things (display:none, etc.).

aleventhal:
> @aleventhal I think I may not have communicated my use case well enough 😅.

Thanks for explaining... good chance the issue was on the receiver's end :)

> The distinction between doing a `.removeAttribute()` and ignoring the attribute through some other method wasn't really material to the question I had in mind, though bringing it up does make me worry that this may make it difficult for developers to debug when an element hidden in markup is exposed.

We could log something like an error to the JS console, warning the developer that bad aria-hidden markup had been discarded. I thought we had similar guidance for other user agent repairs, but I can't find that in the spec after a quick search.

> For my original question, I made this JS Fiddle to help: https://jsfiddle.net/sjc89htb/. It uses logic similar to what we use at Microsoft, and what I've seen for focus traps elsewhere, where focus is immediately sent elsewhere when it lands on a hidden element. In practice this works perfectly in all ATs I've tested up till now. Would you be able to clarify whether it will continue working with this change, or whether the change will result in the hidden content being exposed?

It seems to work! Our proposed change is currently implemented in Chrome Canary from this morning, version 127.0.6508.0.  I ran with and without the focus trap. In the version without the focus trap, once you tab into the button, the button is revealed in the a11y tree (I used the devtools Inspector full page a11y view for the test). In the version with the focus trap, focus never goes to the button, and the aria-hidden is never discarded. I think this is an important test that we should add to WPT tests. @scottaohara maybe we can update the spec text to say if focus is not forwarded somewhere else? Perhaps it should mentioned as a valuable technique, although maybe authors should be using inert here.

Anyway, go ahead and play with Canary if you like. I'll wait to hear back what people think of a console message to developers. Maybe something like, console.error("The following element had aria-hidden markup that was discarded, because focus went to it or an element within it. See WAI-ARIA rules on aria-hidden:", bad_aria_hidden_element);

smhigley:
@aleventhal awesome! I love both the idea of putting that scenario in WPT, and the idea of logging the error to the console.

I'm all in on this now, thanks again for investigating!

aleventhal:
I have a patch ready for Chromium to add the following error message to the console:
          "Blocked aria-hidden on a [tag_q_name] element because the element's "
          "descendant received focus. See WAI-ARIA specification for "
          "aria-hidden at https://w3c.github.io/aria/#aria-hidden.",
          element);

Should we add text to this PR saying that user agents may provide a helpful console error message?
Also, can we update this PR to say how developers can fix the issue? (Either by forwarding focus via focusin, by not putting aria-hidden on focusable objects, by using inert etc.)

index.html
scottaohara: i'm not a fan of switching between saying "excluded from the accessibility tree" and then switching to "hidden" (though i realize 'hidden' in this spec doesn't necessarily mean "completely hidden").

beyond what aaron mentioned (though maybe you updated the text since then) i think that's my only thing i'm not sure about with this rewording.  it uses the terminology correctly, but i had to read it a number of times to figure out if it still meant the same thing.  maybe that's just my problem though.

jnurthen (approved):

index.html
scottaohara, line 11607:
```suggestion
				<p>Additionally, authors SHOULD ensure that any elements that are accessibility descendants of an aria-hidden element, or any elements which have been marked as aria-hidden themselves, are prevented from receiving focus. If an aria-hidden element or a descendant of an aria-hidden ancestor receives focus, User Agents MUST ignore the aria-hidden state of any ancestor to the focused element from that point forward, resulting in the entire subtree to be exposed to assistive technologies. The ancestor aria-hidden markup will continue to be ignored, even once the element loses focus or is no longer focusable.</p>
```

scottaohara:
made updates per the provided feedback, so re-requesting reviews.

index.html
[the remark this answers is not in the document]
scottaohara: pulled in the requested changes and added a reference to the tree exclusion section.  so resolving this particular thread.

index.html
```diff
@@ -11603,7 +11603,8 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Additionally, authors SHOULD ensure that any elements that are accessibility descendants of an aria-hidden element, or any elements which have been marked as aria-hidden themselves, are prevented from receiving focus. If an aria-hidden element or a descendant of an aria-hidden ancestor receives focus, User Agents MUST ignore the aria-hidden state of any ancestor to the focused element from that point forward, resulting in the entire subtree to be exposed to assistive technologies. The ancestor aria-hidden markup will continue to be ignored, even once the element loses focus or is no longer focusable.</p>
```
scottaohara, line 11607: cc @aleventhal user agent may like this?   reword as you see fit.

```suggestion
				<p>Additionally, authors SHOULD ensure that any elements that are accessibility descendants of an aria-hidden element, or any elements which have been marked as aria-hidden themselves, are prevented from receiving focus. If an aria-hidden element or a descendant of an aria-hidden ancestor receives focus, User Agents MUST ignore the aria-hidden state of any ancestor to the focused element from that point forward, resulting in the entire subtree to be exposed to assistive technologies. The ancestor aria-hidden markup will continue to be ignored, even once the element loses focus or is no longer focusable.  User agents MAY provide console messages to warn or flag aria-hidden errors to developers.</p>
```

aleventhal:
> Landing a patch to Chromium that adds this console error:
> 
> ```
>   element.AddConsoleMessage(
>       mojom::blink::ConsoleMessageSource::kRendering,
>       mojom::blink::ConsoleMessageLevel::kError,
>       String::Format(
>           "Blocked aria-hidden on a <%s> element because the element that just "
>           "received focus must not be hidden from assistive technology users. "
>           "Avoid using aria-hidden on a focused element or its ancestor. "
>           "Consider using the inert attribute instead, which will also prevent "
>           "focus. For more details, see the aria-hidden section of the "
>           "WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.",
>           element.TagQName().ToString().Ascii().c_str()));
> ```
> 
> Should we add text to this PR saying that user agents may provide a helpful console error message? Also, can we update this PR to say how developers can fix the issue? (Either by forwarding focus via focusin, by not putting aria-hidden on focusable objects, by using inert etc.)

@scottaohara

index.html
```diff
@@ -13171,12 +13173,12 @@ <h1><dfn class="export">Accessibility Tree</dfn></h1>
 	  <h2>Excluding Elements from the Accessibility Tree</h2>
 	  <p>The following [=elements=] are not exposed via the <a>accessibility <abbr title="Application Programing Interfaces">API</abbr></a> and user agents MUST NOT include them in the <a class="termref">accessibility tree</a>:</p>
 	  <ul>
-	    <li>Elements, including their descendent elements, that have host language semantics specifying that the element is not displayed, such as CSS <code>display:none</code>, <code>visibility:hidden</code>, or the HTML <code>hidden</code> attribute.</li>
+	    <li>Elements, including their descendent elements, that have host language semantics specifying that the element is not displayed or is meant to be treated as inert. For insance, elements with CSS <code>display:none</code>, <code>visibility:hidden</code>, the HTML <code>hidden</code> attribute in its default state, or HTML's <code>inert</code> attribute.</li>
```
aleventhal: For insance

aleventhal (approved):

index.html
scottaohara:
```suggestion
	    <li>Elements, including their descendent elements, that have host language semantics specifying that the element is not displayed or is meant to be treated as inert. For instance, elements with CSS <code>display: none</code>, <code>visibility: hidden</code>, the HTML <code>hidden</code> attribute in its default state, or HTML's <code>inert</code> attribute.</li>
```

scottaohara:
per my last commit - it looks like @giacomo-petri had already tried to fit the errant ending p tag (https://github.com/w3c/aria/pull/2244) but since this PR was made before that, it was still here.... and since the errant ending tag is still in the current spec - it clearly got re-introduced by a commit since Giacomo's.

giacomo-petri:
I believe I understand the reason behind this:

> _...omissis... are either prevented from receiving focus, **or upon receiving focus use scripting to immediately move focus to another element.**,_

However, in the context of the following sentence:

> _The ancestor aria-hidden markup will continue to be ignored, even once the element loses focus or is no longer focusable. User agents MAY provide console messages to warn or flag aria-hidden errors to developers._

this could introduce issues, potentially causing a discrepancy between the experience of assistive technologies (where everything is exposed) and keyboard navigation (where the operable control is only used for functional reasons for moving focus somewhere else).

Should we be more strict by removing "either" and stating that authors should ensure these aria-hidden true elements do not receive focus at all? The sentence already begins with "authors SHOULD," which is less strict than "MUST NOT," and we are not prohibiting focus management via JavaScript. However, without "either," I think the guidance would likely be clearer.

scottaohara:
@giacomo-petri that text was deliberately added in because there are component libraries in use that do this without issue.

having `div tabindex=0 aria-hidden=true` bumpers at the start/end of a component that needs to 'trap' tab key focus.  from what i've been told in talking about this with at least one component library (ahem @smhigley) is that this is far more performant than querying all the elements within the container, and needing to make dynamic adjustments for the instances where the focusable elements (particularly first/last) might change, and having to re-determine what elements should loop focus again/where the new focus should go to.  

I am not aware of where this actually hurts anything - but automated checkers flag this up and down and I can tell you that I've seen a report of over 2500 instances of _this "issue"_ across a select few products, and all 2500 are considered a "false positive" because of the effort that went into making sure that there would be no negative user impact.

**That said**, I have also done recent testing to this pattern and haven't found much in the way of a difference to whether the `aria-hidden` exists on that div or not....  but I also have not had the time to go into the deep level of testing that I know others have on this in support of adding the attribute in the first place.  (please note this is not just used for ARIA modal dialog purposes, but other instances where trapping tab key might be necessary, while implementing other keyboard navigation behaviors).   i'm just hesitant at this point to do anything else with this text, since it's already been word smithed into its current state to account for this scenario.  If others have suggestions though that can maintain the intent without undoing what people wanted in there, then please, have at it :)

aleventhal:
What's the status of landing this?
I occasionally get asked about the message in the Chrome dev tools console, and would like to be able to link into the new wording right in the spec, so there is no doubt :)

rahimabdi (approved):

aleventhal:
One change I'm making to Chrome's implementation is to wait a tiny bit before logging the error and discarding the aria-hidden. This is helpful in cases where the author uses setTimeout(0)/requestAnimationFrame() before deciding on the new focus somewhere appropriate. This was requested by Angular Material, who thought it was common practice. For example, the script marks the main content as aria-hidden and adds the dialog, but waits for the dialog to be rendered before determining what the first focusable object in it is that should be focused.
Here's the issue where that's discussed: https://issues.chromium.org/issues/392121909
We can provide an example in a code pen if anyone needs it.
This might be considered an implementation detail — I'm not sure how we would tweak the wording in this PR.

index.html
```diff
@@ -11602,7 +11602,9 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
-				<p>Authors MAY, with caution, use aria-hidden to hide visibly rendered content from assistive technologies <em>only</em> if the act of hiding this content is intended to improve the experience for users of assistive technologies by removing redundant or extraneous content. Authors using aria-hidden to hide visible content MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.</p>
+				<p>Authors MAY, with caution, use aria-hidden to hide visibly rendered content from assistive technologies <em>only</em> if the act of hiding this content is intended to improve the experience for users of assistive technologies by removing redundant or extraneous content. Authors using aria-hidden to hide visible content MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies. Whenever possibe it's strongly recommended authors use host language features to <a href="#tree_exclusion">exclude elements from the accessibility tree</a>.</p>
```
scottaohara, line 11605:
```suggestion
				<p>Authors MAY, with caution, use <sref>aria-hidden</sref> to hide visibly rendered content from assistive technologies <em>only</em> if the act of hiding this content is intended to improve the experience for users of assistive technologies by removing redundant or extraneous content. Authors using <sref>aria-hidden</sref> to hide visible content MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies. Whenever possible it's strongly recommended authors use host language features to <a href="#tree_exclusion">exclude elements from the accessibility tree</a>.</p>
```

smhigley:
There does seem to be a one-time timing mismatch between setting `aria-hidden` and setting focus when I tried to code a pretty bare-bones standard dialog approach in React: https://stackblitz.com/edit/react-template-fb5ctn9x?file=App.js,package.json,index.js

(For a bit of explanation for anyone who tries the stackblitz: Chrome throws the console error when the custom dialog is opened and focused for the first time, but not subsequent times. The difference is due to internal logic in React, not Chrome, if I remember correctly. This issue disappears with changes in React 18.)

I tried to write the logic based on what I believe would be the most common way to code the combination between `aria-hidden` updates and `element.focus()` calls. Those results are in line with my gut feeling, which is that it's unreliable to depend on the timing of whether or not focus has left a region before `aria-hidden` is added. Often authors/framework patterns are structured so that the static DOM updates have finished before `element.focus()` is called -- the intent of doing so is to ensure that the new to-be-focused element actually exists when you attempt to focus it, and that element creation often happens in the same JS cycle as the `aria-hidden` change.

What do folks (and by folks I mostly mean @aleventhal) think of the idea of checking element hidden status when focus moves, rather than when `aria-hidden` is applied? I believe it's a much safer author requirement to say that focus must never enter an `aria-hidden` element, and I think that requirement is more in line with what actually causes problems for screen reader users.

aleventhal:
Hi @smhigley Let's try again after I land my fix. Hopefully this week.

pkra:
@scottaohara I tried and failed to update to main branch (merge and rebase). Sorry for failing on my promise :pensive:  

I suspect it'll be easier time to start with a fresh branch and apply the final changes once review is complete.

index.html
```diff
@@ -11774,8 +11774,10 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p>Additionally, authors SHOULD ensure that any elements that are accessibility descendants of an aria-hidden element, or any elements which have been marked as aria-hidden themselves, are either prevented from receiving focus, or upon receiving focus use scripting to immediately move focus to another element.</p>	
+				<p>User Agents MUST ignore the aria-hidden state of any ancestor to the focused element from that point forward, resulting in the entire subtree to be exposed to assistive technologies. The ancestor aria-hidden markup will continue to be ignored, even once the element loses focus or is no longer focusable.  User agents MAY provide console messages to warn or flag aria-hidden errors to developers.</p>
```
minorninth, line 11779: The previous paragraph says that authors are allowed to let focus land inside aria-hidden and immediately move it outside.

But this paragraph says that user agents must expose the entire aria-hidden subtree if focus lands inside. Is the intent that user agents should not do this if focus temporarily lands inside a hidden subtree and quickly moves out?

That doesn't sound easy to implement.

aleventhal: That's the intention. Not super easy but not super hard either? I think it's worth it. I've heard many a tale of aria-hidden breaking things for screen reader users, but we don't want to be over-repairing either. WDYT?

aleventhal:
Personally I was going to give it one animation frame and then checking
again before performing the repair.
I can't think of a reason a script should need more time than that.

minorninth:
If we go this route, I think the details need to be part of the specification.

Something like: if an element gets focus and it is effectively aria-hidden (by itself or on an ancestor), the user agent MUST treat the entire aria-hidden subtree as not hidden. Furthermore, if the element gets focus and after running all installed event listeners, focus remains on that element, then the user agent should mark the ancestor as permanently not hidden.

I still have some concerns, though. What happens if focus lands on the element while JS is still loading and installing event listeners?

I think it'd probably be safer to just only make it not hidden while focus is inside, rather than making permanent changes to the page.

aleventhal:
> I think it'd probably be safer to just only make it not hidden while focus is inside, rather than making permanent changes to the page.

Let's try prototyping that in each of our browsers. I originally thought it would be easier to just put the aria-hidden markup on the naughty list, but your idea sounds like it would get around some things.  I'd still plan on logging a warning or error to the console after 100ms or something like that, if the focus is still inside the aria-hidden. WDYT?

index.html
rahimabdi, line 11779: @scottaohara Here's a proposed re-wording for clarity (please feel free to tweak):

```suggestion
				<p>When an element receives focus, user agents MUST ignore the aria-hidden state of any ancestor of the focused element, thus re-exposing its entire descendent subtree to assistive technologies. The aria-hidden subtree MUST be permanently ignored, even once the element loses focus or is no longer focusable, if focus persists on the element (i.e., after all event listeners have been processed). User agents MAY provide console messages to warn or flag aria-hidden errors to developers.</p>
```

adampage (approved):

jcsteh:
> I think it'd probably be safer to just only make it not hidden while focus is inside, rather than making permanent changes to the page.

I have a couple of concerns with this if a user is navigating all over the place. Imagine a form which is incorrectly aria-hidden, but the user is also referring to other parts of the page which aren't aria-hidden.

1. It could result in a lot of spurious tree mutation. The user moves focus inside the form, the form gets added to the tree. The user moves out, the form gets removed from the tree. etc. etc. That could cause performance problems in certain situations depending on the client, OS and size of the aria-hidden subtree.
2. It might result in a situation where focus goes inside the aria-hidden content, the user moves focus out of it to look at something else, and then they can't get back to the aria-hidden content they were previously working with. It isn't intuitive that the user can stab in the dark with the tab key until they find the mysteriously auto-vanishing content. They'll just think it's been removed.

Once we make a determination (per spec) that the author has misused aria-hidden, I'd argue we should treat it as misuse full stop and ignore it. That results in a more consistent user experience.

jcsteh:
> What happens if focus lands on the element while JS is still loading and installing event listeners?

The animation frame proposal, while a little obscure, does avoid this problem. Rather than tying this to event listeners (which can be somewhat indeterminate in some cases as you point out), an animation frame is a clearly defined unit and thus easier to reason about. Hopefully. 😰

jcsteh:
> If we go this route, I think the details need to be part of the specification.

I agree; this should not be implementation specific, as I think this could be a pretty significant cause of hard-to-understand interop problems. Aside from that, the changes here look good to me.

pkra:
Discussed at today's editors meeting https://www.w3.org/2026/03/11-aria-editors-minutes.html#7e86

spectranaut:
I'm trying to understand the status of implementation in Chrome by looking at the code and commit history (because Aaron is no longer working on accessibility there). I see [Don't discard aria-hidden for bad focus until after a delay](https://chromium-review.googlesource.com/c/chromium/src/+/6215572) was reverted, and not relanded, so there might still be work to do related to the delay. I think otherwise it is implemented -- that is, if focus lands on an element within an aria-hidden subtree, the subtree is exposed (and I think continues to be exposed after focus moves).

It also seems like implementers might still disagree on whether to keep things exposed after focus moves @minorninth, did you see @jcsteh comments above? Are you ok to keep things permanently exposed? If so it seems like we should move ahead with this proposed change to the spec.

minorninth:
I'm fine with the idea of "permanently" exposing the aria-hidden region once focus lands there and stays there. I just discussed with the rest of the WebKit engineering team and we're all happy with that.

In terms of the spec language, it feels a bit too prescriptive to say that the user agent must check after exactly one animation frame. Could we loosen that language a little bit, and say something along the lines of - if focus lands in an aria-hidden region, the user agent must wait for at least one animation frame to give any event handlers time to respond and move focus to somewhere not hidden, but if focus remains on the same element inside an aria-hidden region after that delay, the user agent must 

That gives us a little bit of flexibility to possibly do the check inside an existing accessibility timer, or wait until clean layout, for example - but still ensure that we expose the aria-hidden region after only a very short delay.

cookiecrook:
WPT Interop Accessibility issue tracker for test writing once unblocked: 
- https://github.com/web-platform-tests/interop-accessibility/issues/232

jnurthen:
Linking to the FF issue https://bugzilla.mozilla.org/show_bug.cgi?id=1944935

PR #2183, opened 2024-05-20
monorepo: add accname history

pkra:
- [x] add accname history
- [x] remove extraneous files
- [x] adjust links to aria-common


pkra:
I've added a missing closing pre and missing closing div to allow prettier to succeed.

PR #2185, opened 2024-05-20
monorepo: add html-aria history

pkra:
- [x] add html-aria history
- [x] remove extraneous files
- [x] adjust links to aria-common
- [ ] consider whether or not we should move html-aria right now


spectranaut:
Do we actually own this document, yet?

PR #2200, opened 2024-05-22
 [Monorepo] [core-aam PR 221]  Update: UIA mappings for elements that had control type text

pkra:
Imported from https://github.com/w3c/core-aam/pull/221


benbeaudry:
This PR has more changes than just the ones described in the original comment. I was expecting to only review changes related to the UIA control type. Is this expected? If so, LGTM.

pkra:
> This PR has more changes than just the ones described in the original comment. 

Ugh - sorry! That's my fault. I clearly overlooked this while resolved the merge conflicts.

@scottaohara  let me know if you'd rather fix it yourself. Otherwise I'll try to find time before tomorrow's call.

pkra:
I forced pushed a fixed commit. @scottaohara  @benbeaudry would you mind taking another look?

benbeaudry (approved):
LGTM

spectranaut (approved):

scottaohara (approved):
so good to merge then?

jcsteh:
Is there a reason deletion is being changed, but not insertion?

jcsteh:
Chromium still seems to expose heading as UIA_TextControlTypeId, so does this really match reality? See https://github.com/w3c/core-aam/issues/220#issuecomment-2641961667.

spectranaut:
Just a note for myself, it seems like the conclusion from the issue discussion is that Chrome will try out this change: https://github.com/w3c/core-aam/issues/220#issuecomment-2648611666

And maybe Firefox is waiting to make sure the Chrome change doesn't break a screen reader before rolling out the same change.

jcsteh:
If someone from Microsoft can assure me that this won't break narrator, I'm happy to just make the change in Firefox, since it's only Narrator that's impacted practically speaking. But I don't currently have that assurance and I don't have any contacts on the Narrator team to ask.

pkra:
@benbeaudry would you be able to help with @jcsteh's comment asking about the impact on Narrator?

PR #2203, opened 2024-05-22
[Monorepo] [core-aam PR 117] Generic Role Section

pkra:
Moved from https://github.com/w3c/core-aam/pull/117/


spectranaut:
@cookiecrook, you left some comments that are unresolved on the original PR (https://github.com/w3c/core-aam/pull/117), would you like to resolve those or even move them before I add reviewers to this superseding PR?

PR #2207, opened 2024-05-23
[accname PR 236]  Editorial: Renaming LabelledBy Recursion since it's not recursive. 

pkra:
Moved from https://github.com/w3c/accname/pull/236/

Closes [accname issue 222](https://github.com/w3c/accname/issues/222)

Renaming "LabelledBy Recursion" to "LabelledBy Traversal" since this step is not actually recursive.


MelSumner (approved):

accdc (approved):

accname/index.html
```diff
@@ -589,8 +589,8 @@ <h4>Computation steps</h4>
                   whose value is not undefined, not the empty string, nor, when trimmed of [=ascii whitespace|whitespace=], is not the empty string:
                   <ol>
                     <li>
-                      If traversal of the <code>current node</code> is due to recursion <strong>and</strong> the <code>current node</code> is an embedded control, ignore <code>aria-label</code> and
-                      skip to rule <a href="#comp_embedded_control">Embedded Control</a>.
+                      If traversal of the <code>current node</code> is due to <a href="#comp_name_from_content_for_each_child">Name From Content descendant node recursion</a> <strong>and</strong> the
+                      <code>current node</code> is an embedded control, ignore <code>aria-label</code> and skip to rule <a href="#comp_embedded_control">Embedded Control</a>.
```
smhigley, line 593: This issue wasn't introduced in this PR, but it appears that Embedded Control actually precedes this step and will be processed first, so this check seems unnecessary and could maybe be removed. I'm guessing that in the past, that wasn't the case 😅.

I don't think this PR should be held up by this, but maybe it's something to fix later.

smhigley (approved):

PR #1805, opened 2022-09-16
feat: aria-actions addition to the ARIA spec

smhigley:
🚀 **Netlify Preview**:
🔄 **this PR updates the following sspecs**:
- [core-aam preview](https://deploy-preview-1805--wai-aria.netlify.app/core-aam/index.html) &mdash; [core-aam diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fcore-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-1805--wai-aria.netlify.app%2Fcore-aam%2Findex.html)
- [ARIA preview](https://deploy-preview-1805--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-1805--wai-aria.netlify.app%2Findex.html)

Resolves #1440 by adding the `aria-actions` attribute

This has a dependency on the changes in https://github.com/w3c/aria/pull/1454.

# PR tracking
Check these when the relevant issue or PR has been made, OR after you have confirmed the
related change is not necessary (add N/A). Leave unchecked if you are unsure. Read the
[Process Document](https://github.com/w3c/aria/wiki/ARIA-WG-Process-Document/_edit) or
[Test Overview](https://github.com/w3c/aria/wiki/ARIA-Test-Overview) for more information.

* [x] ~Related Core AAM Issue/PR~: included in this PR
* [x] Related AccName Issue/PR: https://github.com/w3c/accname/pull/235: 
* Any other dependent changes?

# Test, Documentation and Implementation tracking
Once this PR and all related PRs have been been approved by the working group, tests
should be written and issues should be opened on browsers. Add N/A and check when not
applicable.

* [x] Related APG Issue/PR: w3c/aria-practices#3070, w3c/aria-practices#3071
* [ ] MDN Issue/PR:  
* [ ] "author MUST" tests: 
* [ ] "user agent MUST" tests
* [ ] Browser implementations (link to issue or when done, link to commit):
   * WebKit: [Bug 306476](https://bugs.webkit.org/show_bug.cgi?id=306476)
   * Gecko: [Bug 1992029](https://bugzilla.mozilla.org/show_bug.cgi?id=1992029)
   * Blink: [Bug 514751946](https://issues.chromium.org/issues/514751946)
* Does this need AT implementations? - yes
[...]

PR #2208, opened 2024-05-23
[Monorepo] [accname PR 235] feat: update step 2F to exclude nodes referenced by aria-actions

pkra:
Moved from https://github.com/w3c/accname/pull/235/

### DIFF
--- a/accname/index.html
+++ b/accname/index.html
@@ -671,6 +671,13 @@
                       <span id="step2F.iii"><!-- Don't link to this legacy numbered ID. --></span><em>Name From Each Child:</em> For each <code>rendered child node</code> of the
                       <code>current node</code>:
                       <ol>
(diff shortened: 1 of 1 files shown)


accname/index.html
```diff
@@ -671,6 +671,13 @@ <h4>Computation steps</h4>
                       <span id="step2F.iii"><!-- Don't link to this legacy numbered ID. --></span><em>Name From Each Child:</em> For each <code>rendered child node</code> of the
                       <code>current node</code>:
                       <ol>
+                        <li id="comp_name_from_content_find_child_actions">
```
aleventhal, line 674: Do we need to say both of these things? They seem like they're essentially saying the same as each other.

spectranaut:
@smhigley can we merge this change with the aria actions PR: https://github.com/w3c/aria/pull/1805 ?

PR #2212, opened 2024-05-23
[html-aam PR 540] Clarify li element role mapping

### LINKED ISSUE #2173: Is the accessibility tree built on the DOM tree, or the flat tree (incl. shadow DOMs)?
# Describe your concern

The spec is a bit unclear on how content in the shadow DOM or slots should be handled when building the accessibility tree, notably impacting the "allowed children". Browsers are currently (April 2024) implementing this differently.

```html
<div id="list-host"><li>One</li><li>Two</li><li>Three</li></div>

<script>
  const listHost = document.getElementById('list-host');
  const listRoot = listHost.attachShadow({mode: 'open'});
  listRoot.innerHTML = "<ul><slot></slot></ul>"
</script>
```

This builds a DOM tree, with shadow included, looking like:
```
#shadow-tree
  <ul>
    <slot>
<li> // slotted
<li> // slotted
<li> // slotted
```

And a [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) (i.e. result of flattening the shadow trees and slotting the elements) looking like:
```
<ul>
  <li>
  <li>
  <li>
```

Chrome builds that and then gives the `<li>` elements a role of `listitem`, i.e. look for a parent `<ul>` **in the flat tree**.

On the other hand, Firefox turns the DOM tree into an accessibility tree like:
```
list
  generic // first <li>
  generic
  generic
```
I.e., it gives the `<li>` a role of `generic`, likely looking for a parent `<ul>` **in the DOM tree** only.

ARIA 1.3 has an [accessibility tree Section](https://w3c.github.io/aria/#accessibility_tree) where the text and examples in the [relationship Subsection](https://w3c.github.io/aria/#tree_relationships) feel pretty much geared for "DOM tree", without mentioning shadow DOM. So it sounds from it like an "accessibility child" cannot cross shadow boundaries 🤔 

We recently had [a discussion in name computation](https://github.com/w3c/accname/pull/167) to include shadow and slotted descendants. I think it would make sense to have the similar move for the accessibility tree.

Link to documentation: [the editors draft](https://w3c.github.io/aria/)

> Does the issue exists in [the editors draft](https://w3c.github.io/aria/) (the editors draft is the most recent draft of the specification)?
Yes.

pkra:
Moved from https://github.com/w3c/html-aam/pull/540/

edit:
closes https://github.com/w3c/aria/issues/2173
approving reviewers from migrated HTML AAM PR: @smhigley, @cookiecrook, @spectranaut 

edit again:
pointing back to the original HTML AAM PR that this is one is clarifying - https://github.com/w3c/html-aam/pull/476

* [X] "author MUST" tests: n/a
* [X] "user agent MUST" tests: https://github.com/web-platform-tests/wpt/pull/49643
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=315766
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=2043336
   * Blink: https://issues.chromium.org/issues/517413737
* [X] Does this need AT implementations? no
* [X] MDN Issue/PR: n/a

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -5104,7 +5104,7 @@
     <tr>
       <th>Comments</th>
       <td>
-        If `li` element is not a child of <a data-cite="html">`ol`</a> , <a data-cite="html">`menu`</a> or <a data-cite="html">`ul`</a>, or if the containing
+        If `li` element is not an accessibility child of an <a data-cite="html">`ol`</a> , <a data-cite="html">`menu`</a> or <a data-cite="html">`ul`</a>, or if the containing
         list element is no longer exposed with a `list` role, then expose the `li` element with a `generic` role.
       </td>
     </tr>
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -5104,7 +5104,7 @@
     <tr>
       <th>Comments</th>
       <td>
-        If `li` element is not an accessibility child of an <a data-cite="html">`ol`</a> , <a data-cite="html">`menu`</a> or <a data-cite="html">`ul`</a>, or if the containing
+        If `li` element is not an [=accessibility child=] of an <a data-cite="html">`ol`</a> , <a data-cite="html">`menu`</a> or <a data-cite="html">`ul`</a>, or if the containing
         list element is no longer exposed with a `list` role, then expose the `li` element with a `generic` role.
       </td>
     </tr>


scottaohara:
test results -  https://codepen.io/scottohara/pen/LEPRRNd

cookiecrook (commented):
(Though I have no problem with this minor editorial change) I recall a conversation (perhaps with @twilco, @nmlapre, and @jcsteh) that downgrading the orphaned roles had no user benefit, and had implementation cost, so the related WPT tests were rolled back or moved to tentative. I think that means this line should be removed from the spec, too.

scottaohara:
@cookiecrook that convo definitely happened [regarding the use of explicit aria roles](https://github.com/web-platform-tests/interop-accessibility/issues/109) that weren’t children of their required parents, but html’s li element contextual role was separate to that. Firefox has been treating an orphan li element as generic for [12 years now](https://github.com/web-platform-tests/wpt/issues/45715#issuecomment-2060226458).

If Firefox is willing to undo this and then all browsers consistently expose the li element as a list item, even when not within a required ul, ol, menu or role=list parent - then it can become a screen reader’s decision on how to handle these author errors, or not.  Maybe this is even the way Firefox can resolve the failing tests for `<ul><div><li>not a list item in firefox</div></ul>` and `<ul aria-owns=f></ul><li>not a list item in firefox`

jcsteh:
I came across [a case yesterday which would break if we removed the rule for orphaned `<li>` elements](https://bugzilla.mozilla.org/show_bug.cgi?id=1720273#c0). This also led to the realisation thta Chrome *sometimes* does this too.

`data:text/html,<ul style="list-style: none;" role="tablist"><li><button role="tab">foo</button></li><li><button role="tab">bar`

If the `<li>` elements got a role of listitem, this would break the posinset/setsize calculation for the tabs. What's interesting here is that in this case, Chrome excludes the `<li>` element from the tree. Conversely, if you do this:

`data:text/html,<li>foo`

Chrome exposes listitem. But if you do this:

`data:text/html,<ul role="none"><li>foo`

Chrome excludes the `<li>`. So now I don't know what to think. 😕

cookiecrook:
> html’s li element contextual role was separate to that.

cookiecrook (approved):
Approving with the caveat this would change long-standing implementation support, so we need broad engine stakeholder buy-in.

spectranaut (approved):
Moving over approval from previous PR.

spectranaut:
@scottaohara this has three positive reviews (including @smhigley from the old PR), and tests, I think that we should open bugs on browsers for this change?

scottaohara:
do wpts need to be merged before bugs can be filed?  the wpt for this one still needs to be merged.  is that a before or after activity?  @spectranaut @cookiecrook ??

spectranaut:
I don't think WPT PRs need to be merged before bugs are filed. Probably the spec PR and the test PR should land at the same time.

iirc if the WPT PRs land first, they should be marked as "tentative" which sounds like a hassle to do every time.

spectranaut:
Seems like the next step is to make bugs on browsers.

rahimabdi (approved):

rahimabdi:
> @scottaohara this has three positive reviews (including @smhigley from the old PR), and tests, I think that we should open bugs on browsers for this change?

Done. I've also reached out to @zcorpan to help get the WPTs merged.

rahimabdi:
@scottaohara and I have discussed; all browser bugs are filed and all engines behaving in alignment with this spec update.

Merging.

PR #2214, opened 2024-05-23
[Monorepo] [html-aam PR 533] Clarify hidden elements do not return name

pkra:
Moved from https://github.com/w3c/html-aam/pull/533/

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -1341,7 +1341,9 @@
             <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
             <tr>
               <th>Comments</th>
-              <td>If a descendant of a `table`, the first instance of a `caption` element will provide the `table` its accessible name.</td>
+              <td>
+                <p class="note">If a `caption` element is hidden from the accessibility tree, then it will not provide an accessible name to its parent `table` element.</p>
+              </td>
             </tr>
           </tbody>
         </table>
@@ -2194,7 +2196,7 @@
               </th>
               <td>
                 <div class="role"><span class="type">Role:</span> Use WAI-ARIA mapping</div>
-                <div class="relations"><span class="type">Relations:</span> `IA2_RELATION_LABELLED_BY` with child <a href="#el-legend">`legend`</a> element</div>
+                <div class="relations"><span class="type">Relations:</span> `IA2_RELATION_LABELLED_BY` with first instance of child <a href="#el-legend">`legend`</a> element</div>
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


html-aam/index.html
```diff
@@ -4599,7 +4603,9 @@ <h4 id="el-label">`label`</h4>
             <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
             <tr>
               <th>Comments</th>
-              <td></td>
+              <td>
+                <p class="note">If a `label` element is hidden from the accessibility tree, then it will not provide an accessible name to the labelable element is is associated with.</p>
```
scottaohara, line 4607:
```suggestion
                <p class="note">If a `label` element is hidden from the accessibility tree, then it will not provide an accessible name to the <a data-cite="html//forms.html#category-label">labelable element</a> it is associated with.</p>
```

html-aam/index.html
```diff
@@ -1341,7 +1341,9 @@ <h4 id="el-caption">`caption`</h4>
             <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
             <tr>
               <th>Comments</th>
-              <td>If a descendant of a `table`, the first instance of a `caption` element will provide the `table` its accessible name.</td>
+              <td>
```
jnurthen, line 1344: Did we mean to remove the part about the first instance of caption?

jnurthen: I see now this is moved to table. Marking as resolved.

jnurthen (approved):

adampage (approved):

html-aam/index.html
```diff
@@ -6860,19 +6869,30 @@ <h4 id="el-table">`table`</h4>
...
+                <div class="role">Use WAI-ARIA mapping</div>
+                <div class="relations"><span class="type">Relations:</span> `ATK_RELATION_LABELLED_BY` with first instance of child <a href="#el-caption">`caption`</a> element</div>
+              </td>
             </tr>
             <tr>
               <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
-              <td>Use WAI-ARIA mapping</td>
+              <td>
+                <div class="role">Use WAI-ARIA mapping</div>
+                <div class="role-namefrom"><strong>AXDescription:</strong> value from child <a href="#el-caption">`caption`</a> subtree</div>
```
cookiecrook, line 6894: Subtree seems to imply DOM subtree such as `element.innerText`, but I think what you want here is the computed name of the first rendered caption element that is a child(?) of the table. I did not provide an inline diff because I think some of these suggestions also apply to the above changes.. For example, the "hidden" stipulation is presumably a normative requirement, that is listed in an informative note.... I think adding "rendered" (as I did here) in all normative contexts may resolve that, but there are other ways to make the change too.

cookiecrook (changes requested):

html-aam/index.html
scottaohara: thanks for the catch on the normative requirement in a note.  i'll work on that asap.

i did however intend the subtree to be where the name was derived from, rather than just stating name in general, since authors shouldn't be using things like aria-label to name a legend or caption. 

html aam uses 'subtree' in a lot of the name calculation steps that were written far before i started editing the spec.  so if we want to change the way that's referenced, i think we need to do that in many places (though we can start with this PR)

rahimabdi:
Discussed at the [Aug 27 2025 ARIA Editors meeting](https://www.w3.org/2025/08/27-aria-editors-minutes.html).

I followed up with @scottaohara and it looks like this one is close to being merged after resolution of merge conflicts. I've also left a review.

pkra:
Discussed during today's editors' call  https://www.w3.org/2025/10/22-aria-editors-minutes.html#9319

PR #2217, opened 2024-05-23
[html-aam PR 510] switch attribute for checkbox inputs

pkra:
Moved from https://github.com/w3c/html-aam/pull/510/


pkra:
This one was very messy. @scottaohara if you have the time, please take a closer look to make sure I didn't mess anything up.

rahimabdi:
Re-adding reviewers @annevk @cookiecrook.

jnurthen:
> Re-adding reviewer @cookiecrook.
> 
> @annevk should also be a reviewer but unable to add him at the moment (unsure why), investigating.

We can only assign folks who are in the ARIA contributors group. @annevk can still review but we can't assign him.

html-aam/index.html
```diff
@@ -14589,6 +14642,58 @@ <h4 id="att-step">`step`</h4>
...
+        <h4 id="att-switch">`switch`</h4>
+        <table aria-labelledby="att-switch">
+          <tbody>
+            <tr>
+              <th>HTML Specification</th>
+              <td>`switch`</td>
+            </tr>
+            <tr>
+              <th>Element(s)</th>
+              <td><a data-cite="html/input.html#attr-input-checked">`input`</a> `type=checkbox`</td>
```
cookiecrook, line 14654: isn't this missing the switch attr?

html-aam/index.html
```diff
@@ -14589,6 +14642,58 @@ <h4 id="att-step">`step`</h4>
...
+            <tr>
+              <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
+              <td>
+                <div class="general">Not mapped</div>
+              </td>
+            </tr>
+            <tr>
+              <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
+              <td>
+                <div class="general">Not mapped</div>
```
cookiecrook, line 14685: cross link the WebKit AX PR for the relevant change. @annevk worked on it IIRC.

cookiecrook (changes requested):
Missing most platform mappings... Should this be marked as a draft PR?

html-aam/index.html
```diff
@@ -3225,6 +3225,59 @@ <h4 id="el-input-checkbox">`input` <span class="el-context">(`type` attribute in
...
+              </td>
+            </tr>
+            <tr>
+              <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
+              <td>
+                <div class="general">Use WAI-ARIA mapping</div>
+              </td>
+            </tr>
+            <tr>
+              <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
```
cookiecrook, line 3269: The full name is the "The macOS Accessibility Protocol" or "AX API" for short. I don't recall this even being shortened to just "AX", and in either case, the NSAccessibility link is just one subset portion. This might be the best link in general. https://developer.apple.com/documentation/accessibility Should this be changed elsewhere? HAppy to discuss offline since this may be OT for this PR.

html-aam/index.html
annevk: Let's see, there's AAM tests here:

- https://github.com/web-platform-tests/wpt/blob/master/html-aam/roles-dynamic-switch.tentative.window.js
- https://github.com/web-platform-tests/wpt/blob/master/html-aam/roles-generic.tentative.html

We made several changes to WebKit:

- Made it map to `role=switch`: https://github.com/WebKit/WebKit/pull/18897.
- Made `role=switch` (which `<input type=checkbox switch>` builds on) support the notion of being required, similar to `role=checkbox`. That happened in https://github.com/WebKit/WebKit/pull/18938.
- Calculating rectangles correctly in the a11y code: https://github.com/WebKit/WebKit/pull/19048.

We also made a change to VoiceOver (change itself not public): https://bugs.webkit.org/show_bug.cgi?id=268120.

Hope that helps!

html-aam/index.html
scottaohara: From what I can see looking back to some of the original PRs, this spec has always referred to it as just "AX".  i notice though that it is not referred to consistently across AAMs

in the accessibility api sections, HTML AAM refers to it as macOS Accessibility Protocol [AXAPI].  

in other many other specs that list out the full name (often in api / important term sections), it's referred to by it's older name - The [Mac OS X Accessibility Protocol](https://developer.apple.com/documentation/appkit/accessibility/nsaccessibility) [[AXAPI](https://www.w3.org/TR/accname-1.1/#bib-axapi)] - if it's even directly mentioned at all.


in the mapping tables, it's handled like:
dpub AAM - Mac AX API
html aam - AX
core aam, mathml aam - AX API
graphics aam - AXAPI


i have another issue this can be addressed with, at least for html aam

html-aam/index.html
scottaohara: yeah, the link doesn't exist in the published spec yet.  but looks like it would be `input.html#attr-input-switch` when it does.

scottaohara, line 14654:
```suggestion
              <td><a data-cite="html/input.html#attr-input-switch">`input`</a> `type=checkbox`</td>
```

scottaohara: aw nope, that was just a miss.  already had that link in there for the previous instance.  thanks!

html-aam/index.html
```diff
@@ -3225,6 +3225,59 @@ <h4 id="el-input-checkbox">`input` <span class="el-context">(`type` attribute in
...
+                <span class="el-context"
+                  >(<a data-cite="html/input.html#attr-input-type">`type`</a> attribute in the <a data-cite="html/input.html#checkbox-state-(type=checkbox)">Checkbox</a> state)</span
+                >
+                with the <a data-cite="html/input.html#attr-input-switch">`switch`</a> attribute
+              </td>
+            </tr>
+            <tr>
+              <th>[[wai-aria-1.2]]</th>
+              <td>
+                <a class="core-mapping" href="#role-map-checkbox">`switch`</a> role, with the <a class="core-mapping" href="#ariaCheckedMixed">`aria-checked`</a> state set to "true" if the element's
```
scottaohara, line 3244:
```suggestion
                <a class="core-mapping" href="#role-map-switch">`switch`</a> role, with the <a class="core-mapping" href="#ariaCheckedMixed">`aria-checked`</a> state set to "true" if the element's
```

scottaohara:
> Missing most platform mappings... Should this be marked as a draft PR?

@cookiecrook  yes, this is an incomplete PR since the attribute isn't in the spec yet.  The previous html-aam pr was labeled 'not ready for merge'.

html-aam/index.html
scottaohara: @cookiecrook this PR is not missing the platform mappings?   They all say to use the wai-aria mapping for the switch role (though that was incorrectly linking to checkbox and not switch role, just fixed that) in the input (type attribute in the Checkbox state) with the switch attribute mapping table.

the switch attribute itself just points up to that role table (this spec does similar for the `list` attribute on `input` elements where the existence of the attribute changes the input's role to `combobox`).  "Not mapped" is probably not the right text for the switch attribute table though, and instead each row should just indicate the attribute changes the role of the input.  i'll make that change.

html-aam/index.html
rahimabdi: @scottaohara Would you like me to do raise a separate issue/PR for updating this throughout html-aam?

rahimabdi:
Adding this one for Agenda+ to determine next steps. 

As present, the switch attribute hasn't landed in HTML spec yet, and only WebKit supports the switch attribute without a flag ([as of Safari 17.4](https://webkit.org/blog/15054/an-html-switch-control/)).

Issue #2166, opened 2024-04-23
Implementation concerns over Orphaned Role requirements in Core-AAM, HTML-AAM… and possibly elsewhere.


cookiecrook:
### Core-AAM Spec prose:

> When an element has a role but is not contained in the required context (for example, an orphaned listitem without the required accessible parent of role list), User Agents MUST ignore the role token, and return the computedrole as if the ignored role token had not been included.

```html
<div role="listitem"> <!-- Author error: orphaned listitem. computedrole returns "generic" -->

<div role="list"> <!-- computedrole returns "list" -->
  <div role="listitem"> <!-- computedrole returns "listitem" in the required context. -->
```

Cite: https://w3c.github.io/core-aam/#roleMappingComputedRole

While working through WPT tests for Interop 2024, we've received some additional implementation concerns about the "orphaned roles" prose. Seems like one or more implementations may be working around this in the test context primarily, which defeats the purpose of testing… I'm opening this new ARIA issue and possibly [moving some of these tests to .tentative files](https://github.com/web-platform-tests/interop-accessibility/issues/120).

### Some Concerns

- Engines tend to leave some of these orphaned contexts (list items, cells, etc) for the AT to resolve, since the role distinction does not actually matter until such a time as the AT tries to access it. AT can detect whether the structure is valid, and ignore irrelevancies at user-access time. Requiring the engines to manage these author inconsistencies in real-time (even when not being accesses) puts burden on the implementor and it can be passed to the user in the form of a performance hit.
- Instead of penalizing the user with a perf hit, at least one implementor considered "passing" this requirement in the test context only, which defeats the purpose of testing and the AAMs. After discussion, we agreed that is not ideal and should raise the issue instead.
- The ARIA spec acknowledges that most roles should not change, but there are scenarios (like `display:contents`), where a parent's implicit role may may change due a responsive layout update, causing a cascade of role and style dependencies below. Sometimes a descendant may be disconnected from it's parent context inadvertently, putting additional work on the web engine's accessibility implementation to invalidate the role of an item unexpectedly... 
- There are also some edge cases (but real cases) I can think of where reverse lookups are necessary to comply with this requirement: As one example, and orphaned list item that is reference-included into a list via `aria-owns` from elsewhere in the DOM.
- I anticipate adding more examples to this list…

[...]

PR #2218, opened 2024-05-24
[Monorepo] [html-aam PR 484] contextual role concept proposal

pkra:
Moved from https://github.com/w3c/html-aam/pull/484/

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -161,6 +161,71 @@
         <p>There are currently no deprecated requirements.</p>
       </section>
     </section>
+    <section id="mapping_contextual">
+      <h3>Exposing HTML Features With Contextual Roles</h3>
+      <p>
+        An element with a <STRONG>contextual role</STRONG> is one whose computed role semantics are based on its ancestral <a>accessibility tree</a> relationship to 
+        another element or elements. Or, an element whose computed role is dependent upon whether it has other associated accessibility information, such as an author-provided accessible name. 
+      </p>
+      <p>
+        For instance, an element that is an <STRONG>accessibility child</STRONG> of its required <STRONG>accessibility parent</STRONG> would expose a computed role for its given context, 
+        such as a list item (`li` element) within an unordered list (`ul` element). However, the same element might be used within a different markup pattern and could therefore expose 
+        different contextually-dependent computed roles, such as a `section` element with an author-provided accessible name (role `region`) versus an unnamed `section` element (role `generic`).
+      </p>
+      <p>
+        When the conditions for an element's contextual role are not met, it will commonly expose a computed role of `generic` or with a <STRONG>minimum role</STRONG>, if applicable. 
+      </p>
+      <div class="note">
+        <p>Previously, the concept of a contextual role was loosely referred to as an element's "scoped" relationship to another element.</p>
+      </div>
+      <div class="example">
+        <p>
+          A <code>li</code> element has an implicit WAI-ARIA role mapping of <code>listitem</code> only when the element is an <STRONG>accessibility child</STRONG> of a 
+          <code>menu</code>, <code>ol</code> or <code>ul</code> element that has not had its `list` role mapping suppressed.
+        </p>
+        <pre><code>&lt;ul>
+    &lt;li>...&lt;/li> &lt;!-- computed role `listitem` -->
+  &lt;/ul> 
+  &lt;article>
+    ...
+    &lt;li>...&lt;/li> &lt;!-- computed role `generic`, not `listitem` -->
+    ...</code></pre>
+      </div>
+      <div class="example">
+        <p>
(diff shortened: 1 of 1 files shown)


pkra:
Strangely enough, this merged without conflict -- which makes me doubt the result.

pkra:
@scottaohara I hope I moved the section to the correct position.

rahimabdi:
Re-adding reviewers @cookiecrook @spectranaut @aleventhal and myself.

html-aam/index.html
```diff
@@ -191,6 +191,69 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+        <p>
+          An element with a <strong>contextual role</strong> is one whose computed role semantics are based on its ancestral <a>accessibility tree</a> relationship to another element or elements. Or,
+          an element whose computed role is dependent upon whether it has other associated accessibility information, such as an author-provided accessible name.
+        </p>
+        <p>
+          For instance, an element that is an <strong>accessibility child</strong> of its required <strong>accessibility parent</strong> would expose a computed role for its given context, such as a
+          list item (`li` element) within an unordered list (`ul` element). However, the same element might be used within a different markup pattern and could therefore expose different
+          contextually-dependent computed roles, such as a `section` element with an author-provided accessible name (role `region`) versus an unnamed `section` element (role `generic`).
+        </p>
+        <p>When the conditions for an element's contextual role are not met, it will commonly expose a computed role of `generic` or with a <strong>minimum role</strong>, if applicable.</p>
```
spectranaut, line 205:
```suggestion
        <p>When the conditions for an element's contextual role are not met, it will commonly be exposed with a computed role of `generic` or with a <strong>minimum role</strong>, if applicable.</p>
```

Grammar ^

html-aam/index.html
```diff
@@ -191,6 +191,69 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+            A <code>li</code> element has an implicit WAI-ARIA role mapping of <code>listitem</code> only when the element is an <strong>accessibility child</strong> of a <code>menu</code>,
+            <code>ol</code> or <code>ul</code> element that has not had its `list` role mapping suppressed.
+          </p>
+          <pre><code>&lt;ul>
+      &lt;li>...&lt;/li> &lt;!-- computed role `listitem` -->
+    &lt;/ul> 
+    &lt;article>
+      ...
+      &lt;li>...&lt;/li> &lt;!-- computed role `generic`, not `listitem` -->
+      ...</code></pre>
```
spectranaut, line 220: Maybe we should use a different example than listitem, here and above, given the push back on the orphan roles tests which is partially being discussed here: https://github.com/w3c/aria/issues/2166

But started with the roll back of orphan `<li>` tests in WPT.

spectranaut (changes requested):

html-aam/index.html
scottaohara: orphan `li` elements was not where the pushback was.   it was orphan `element role=listitem` that had the pushback.

cookiecrook: Any time you're tempted to add an orphaned role test, please add a comment that points to the appropriate orphaned role-specific test file.

html-aam/index.html
```diff
@@ -191,6 +191,69 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+          <p>
+            A <code>li</code> element has an implicit WAI-ARIA role mapping of <code>listitem</code> only when the element is an <strong>accessibility child</strong> of a <code>menu</code>,
+            <code>ol</code> or <code>ul</code> element that has not had its `list` role mapping suppressed.
+          </p>
+          <pre><code>&lt;ul>
+      &lt;li>...&lt;/li> &lt;!-- computed role `listitem` -->
+    &lt;/ul> 
+    &lt;article>
+      ...
+      &lt;li>...&lt;/li> &lt;!-- computed role `generic`, not `listitem` -->
```
cookiecrook, line 219: I recall we agreed to get rid of the expectation to computed orphaned generics, since doing it in the engine was sometimes costly and always irrelevant to the user.

html-aam/index.html
cookiecrook: > orphan li elements was not where the pushback was. it was orphan <element role=listitem> that had the pushback.

As I understand it, it's still the same issue whether the role is implicit or explicit.

scottaohara: tried to message you this in slack, but i'm thinking of closing this PR down, and instead taking the approach of just clarifying the few elements that do have contextual roles mapping tables

pkra:
@daniel-montalvo was this closed by accident? If not, should https://github.com/w3c/html-aam/issues/375 be closed as well?

scottaohara:
it's still a todo for me...

not sure if this PR is the form this should be taking after all this time... but don't close the issue.

pkra:
@scottaohara :+1: thanks for clarifying!

PR #2221, opened 2024-05-24
[html-aam PR 451] clarify img mappings

pkra:
Moved from https://github.com/w3c/html-aam/pull/451/

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2221.html" title="Last updated on May 12, 2025, 5:36 PM UTC (46a72de)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2221/b085915...46a72de.html" title="Last updated on May 12, 2025, 5:36 PM UTC (46a72de)">Diff</a>


pkra:
Now pointing to main.

rahimabdi:
Re-assigning reviewers @cookiecrook @smhigley @jnurthen.

html-aam/index.html
```diff
@@ -3113,7 +3122,13 @@ <h4 id="el-img-empty-alt">`img` <span class="el-context">(`alt` attribute value
             <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
             <tr>
               <th>Comments</th>
-              <td></td>
+              <td class="comments">
+                <p>Authors can specify the `alt` attribute to be the empty string by either marking it up as `alt=""` or `alt` with no value.</p>
+                <p>
+                  If an `img` has been provided an `alt=""`, but also has an `aria-label` or `aria-labelledby` attribute specified which returns an accessible name, then the element maps to the `img`
```
scottaohara, line 3128:
```suggestion
                  If an `img` has been provided an `alt=""`, but also has an `aria-label` or `aria-labelledby` attribute specified which returns an accessible name, then the element maps to the `image`
```

html-aam/index.html
```diff
@@ -3016,8 +3016,14 @@ <h4 id="el-img">`img`</h4>
             <tr>
               <th>[[wai-aria-1.2]]</th>
               <td>
-                <a class="core-mapping" href="#role-map-image">`image`</a>
-                or <a class="core-mapping" href="#role-map-img">`img`</a> role
+                <div>
+                  If element has a non-empty `src` or `srcset` attribute, an <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>, or both, then map to the
+                  <a class="core-mapping" href="#role-map-image">`image`</a> or <a class="core-mapping" href="#role-map-img">`img`</a> role.
```
scottaohara, line 3021:
```suggestion
                  <a class="core-mapping" href="#role-map-image">`image`</a> or <a class="core-mapping" href="#role-map-img">`image`</a> role.
```

scottaohara: refer to the `image` role vs the original `img` role, as image has become the primary name for these synonyms
https://w3c.github.io/aria/#image

html-aam/index.html
```diff
@@ -3016,8 +3016,14 @@ <h4 id="el-img">`img`</h4>
             <tr>
               <th>[[wai-aria-1.2]]</th>
               <td>
-                <a class="core-mapping" href="#role-map-image">`image`</a>
-                or <a class="core-mapping" href="#role-map-img">`img`</a> role
+                <div>
+                  If element has a non-empty `src` or `srcset` attribute, an <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>, or both, then map to the
+                  <a class="core-mapping" href="#role-map-image">`image`</a> or <a class="core-mapping" href="#role-map-img">`image`</a> role.
```
scottaohara, line 3021:
```suggestion
                  <a class="core-mapping" href="#role-map-image">`img`</a> or <a class="core-mapping" href="#role-map-img">`image`</a> role.
```

scottaohara:
need to come back to this as https://github.com/whatwg/html/pull/11300 was merged into HTML

smhigley (approved):

html-aam/index.html
```diff
@@ -3125,8 +3125,14 @@ <h4 id="el-img">`img`</h4>
             <tr>
               <th>[[wai-aria-1.2]]</th>
               <td>
-                <a class="core-mapping" href="#role-map-image">`image`</a>
-                or <a class="core-mapping" href="#role-map-img">`img`</a> role
+                <div>
+                  If element has a non-empty `src` or `srcset` attribute, an <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>, or both, then map to the
+                  <a class="core-mapping" href="#role-map-image">`img`</a> or <a class="core-mapping" href="#role-map-img">`image`</a> role.
```
cookiecrook, line 3125: Are these typos? Link text for `#…image` is "img", and vice versa link text for `#…img` is "image"…

Regardless, these should both map to the primary synonym: "image"
```suggestion
                  <a class="core-mapping" href="#role-map-image">`image`</a> role.
```

html-aam/index.html
```diff
@@ -3125,8 +3125,14 @@ <h4 id="el-img">`img`</h4>
...
+                </div>
+                <div>
+                  Otherwise, if the element has no `src` or `srcset` attribute, or if the attribute(s) are set to the empty string, and the `img` has no `alt` attribute or accessible name, then map to
+                  the <a class="core-mapping" href="#role-map-none">`none`</a> or <a class="core-mapping" href="#role-map-presentation">`presentation`</a> role.
```
cookiecrook, line 3129: `none`, not `presentation`. 

```suggestion
                  the <a class="core-mapping" href="#role-map-none">`none`</a> or <a class="core-mapping" href="#role-map-none">`none`</a> role.
```

html-aam/index.html
```diff
@@ -3163,6 +3170,10 @@ <h4 id="el-img">`img`</h4>
             <tr>
               <th>Comments</th>
               <td>
+                <p>
+                  An `img` with a non-empty `src`, `srcset`, or both and no `alt` attribute or <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> will still map to the `image` role,
```
cookiecrook, line 3169: editorial
```suggestion
                  An `img` with a non-empty `src`, `srcset`, or both and no `alt` attribute or <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> will map to the `image` role,
```

index.html
```diff
@@ -13681,9 +13681,15 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+            <p>
+              The <sref>aria-expanded</sref> attribute is applied to a focusable, interactive element that toggles visibility of content of a different element. If the element with
+              <sref>aria-expanded</sref> is also a <rref>treeitem</rref> in a <rref>tree</rref> or a <rref>row</rref> in a <rref>treegrid</rref>, then it SHOULD also be the
+              <a>accessibility parent</a> of the content it expands and collapses. Otherwise, the element with <sref>aria-expanded</sref> SHOULD NOT be the <a>accessibility parent</a> of the content
+              that is expanding or collapsing. Rather, identify that relationship between the interactive element and the element being controlled using <pref>aria-controls</pref>.
+            </p>
```
cookiecrook: Can't use passive voice in an RFC-2119. Are these author requirements or user agent requirements? Most in HTML-AAM are UA reqs, but this looks like an author req, so:

1. It should start with: "Authors SHOULD" or "Authors SHOULD NOT" in order to disambiguate.
2. It should be moved to from HTML-AAM to ARIA. Isn't this authoring rule more broadly applicable than HTML? Hooray for the monorepo!

cookiecrook (changes requested):

johannesodland:
Should the mapping rely on the result of the [select an image source](https://html.spec.whatwg.org/#selecting-an-image-source) in stead of the attributes on the `img` element itself?

The `img` element can have a `picture` parent and more `source` siblings, and all browsers render correctly if `src` and `srcset` on the `img` element is omitted, as long as there is a source element that is selected.

Take the following example (alt attribute omitted on purpose):  

```html
<picture>
  <source srcset="./image.jpg">
  <img>
</picture>
```

Even though it is contrived example without an alt attribute, it will be visible as an image in every browser visiting the page, and should be mapped as such?

For the record, I am hoping we can loosen the requirements for `src` and `srcset` attributes on the `img` element: 
https://github.com/whatwg/html/issues/11313

rahimabdi:
> Should the mapping rely on the result of the [select an image source](https://html.spec.whatwg.org/#selecting-an-image-source) in stead of the attributes on the img element itself?

TIL! My main concern here is the complexity of this algorithm for the purposes of accessibility mappings; for just [determining the source set](https://html.spec.whatwg.org/#update-the-source-set):

```
Let default source be the empty string.

Let srcset be the empty string.

Let sizes be the empty string.

If el is an [img](https://html.spec.whatwg.org/#the-img-element) element that has a [srcset](https://html.spec.whatwg.org/#attr-img-srcset) attribute, then set srcset to that attribute's value.

Otherwise, if el is a [link](https://html.spec.whatwg.org/#the-link-element) element that has an [imagesrcset](https://html.spec.whatwg.org/#attr-link-imagesrcset) attribute, then set srcset to that attribute's value.

If el is an [img](https://html.spec.whatwg.org/#the-img-element) element that has a [sizes](https://html.spec.whatwg.org/#attr-img-sizes) attribute, then set sizes to that attribute's value.

Otherwise, if el is a [link](https://html.spec.whatwg.org/#the-link-element) element that has an [imagesizes](https://html.spec.whatwg.org/#attr-link-imagesizes) attribute, then set sizes to that attribute's value.

If el is an [img](https://html.spec.whatwg.org/#the-img-element) element that has a [src](https://html.spec.whatwg.org/#attr-img-src) attribute, then set default source to that attribute's value.

Otherwise, if el is a [link](https://html.spec.whatwg.org/#the-link-element) element that has an [href](https://html.spec.whatwg.org/#attr-link-href) attribute, then set default source to that attribute's value.

Set el's [source set](https://html.spec.whatwg.org/#source-set) to the result of [creating a source set](https://html.spec.whatwg.org/#create-a-source-set) given default source, srcset, sizes, and img.
```

In this example:

```
<picture>
  <source srcset="./image.jpg">
  <img>
</picture>
```

Perhaps the `<picture>` element should have a mapping (currently not mapped in html-aam), rather than the image as it is decorative?

scottaohara:
if other browsers would agree to shift the `role=image` mapping to the parent `picture` element, rather than the child `img` element, that would seem far less complicated than having to have the `img` mapping determined on whether or not it is within a `picture` and then in that case only, make it ok for it to not have one of the otherwise necessary `src` or `srcset` attributes.  

if the `picture` gets mapped to `role=image` then it could be more of a question of "did a valid source render, or is the element named (via child img naming mechanism, or now even an aria-label/aria-labelledby/title on the picture element)? then role=image.  otherwise, role=none/generic if no source, or if alt text is defined as empty string"

picture naming algorithm would then need to be worked out... but seems to me it'd be something like

picture gets its name from:
1. aria-labelledby or aria-label on picture element itself
2. img element within picture has an accName (following its own naming steps) 
3. if accname still empty, then `title` attribute of picture element
4. otherwise, no accname

marking a picture as decorative would require using `alt=""` on the child `img`, or if the child `img` had `role=none` or `presentation` or `aria-hidden=true`, or one of those aria roles/attributes on the picture element itself.

feels like trying to make it easier to declare an image source within a picture will make it a bit messier on this end, either in having to work with the complicated algorithm @rahimabdi mentioned, or if `picture` were to become the element that exposed the mapping

johannesodland:
> TIL! My main concern here is the complexity of this algorithm for the purposes of accessibility mappings; for just [determining the source set](https://html.spec.whatwg.org/#update-the-source-set):

I was hoping it was possible to reference the outcome of the procedure, so that the mapping would stay in sync with the HTML spec (and the visual rendered result).

> Perhaps the <picture> element should have a mapping (currently not mapped in html-aam), rather than the image as it is decorative?

The way I understand it, the `img` in the example is not decorative. It is the `img` element that is rendered, while the `picture` only holds the `source` elements. The picture element might even have `display: contents`, so that it doesn't generate a box.

I think browsers have supported rendering `img` elements without its own `src` and `srcset` since the introduction of responsive images more than a decade ago, as long as it has a `srcset` on a sibling `source` element. (Even though the standard requires one of them to be set.)

> if other browsers would agree to shift the role=image mapping to the parent picture element, rather than the child img element, that would seem far less complicated than having to have the img mapping determined on whether or not it is within a picture and then in that case only, make it ok for it to not have one of the otherwise necessary src or srcset attributes.

I am no expert here, but since it is the `img` element that is visually rendered as a replaced element, wouldn't that put the mapping on an element that might not be visible/generate a box?

I'm just a front-end developer, and not a browser developer/web platform engineer. So I think smarter people than me might have better answers :)

I've just noticed that all browsers support skipping the `src` and `srcset` on the `img` element, and I was hoping the spec could reflect that. Otherwise we might end up in a situation where we have to set a one-char `src` attribute on the `img` element, just for the specs and mapping to be simpler.

```html
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="./dark.png">
  <source srcset="./light.png">
  <img src="🖼️">
</picture>
```

PR #2223, opened 2024-05-24
[Monorepo] [html-aam PR 395] add validity IDL

pkra:
Moved from https://github.com/w3c/html-aam/pull/395/

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -15435,6 +15435,19 @@
               <th>Comments</th>
               <td>Refer to <a href="#el-img">`img`</a> element.</td>
             </tr>
+            <tr tabindex="-1" id="att-validity">
+              <th>`validity [IDL]`</th>
+              <td class="elements">
+                HTML elements;
+                <a data-cite="html/#the-input-element:dom-cva-validity">`input`</a>
+              </td>
+              <td class="aria"><a class="core-mapping" href="#ariaInvalidTrue">`aria-invalid`</a> ="true"</td>
+              <td class="ia2">Use WAI-ARIA mapping</td>
+              <td class="uia">Use WAI-ARIA mapping</td>
+              <td class="atk">Use WAI-ARIA mapping</td>
+              <td class="ax">Use WAI-ARIA mapping</td>
+              <td class="comments">If the element has the `validity [IDL]` set and the `aria-invalid` attribute set, User Agents MUST expose only the`validity [IDL]` state.</td>
+            </tr>
           </tbody>
         </table>
         <h4 id="att-value-button">`value`</h4>


pkra:
Now pointing to main.

jnurthen:
@scottaohara is this ready for merge if we get a reviewer or is it waiting for implementations?

html-aam/index.html
```diff
@@ -15435,6 +15435,19 @@ <h4 id="att-usemap">`usemap`</h4>
...
+              <td class="elements">
+                HTML elements;
+                <a data-cite="html/#the-input-element:dom-cva-validity">`input`</a>
+              </td>
+              <td class="aria"><a class="core-mapping" href="#ariaInvalidTrue">`aria-invalid`</a> ="true"</td>
+              <td class="ia2">Use WAI-ARIA mapping</td>
+              <td class="uia">Use WAI-ARIA mapping</td>
+              <td class="atk">Use WAI-ARIA mapping</td>
+              <td class="ax">Use WAI-ARIA mapping</td>
+              <td class="comments">If the element has the `validity [IDL]` set and the `aria-invalid` attribute set, User Agents MUST expose only the`validity [IDL]` state.</td>
```
cookiecrook, line 15449: I think el.ariaInvalid would stilll need to be exposed… in part because I don't know how to remove it with a HTMLInputElement  ILD override, so we could leave it and expose as null. @rahimabdi what do you think?

annevk: It's not entirely clear to me what the requirement here is.

It's also not clear to me why tests couldn't be written for this.

Is the idea that if the form element is considered valid per HTML's requirements, `aria-invalid` is to be ignored? I think that should happen on the AT layer and attribute reflection shouldn't really be involved in that. (Thus if you set the `aria-invalid` attribute, the `ariaInvalid` getter will still reflect that, even if it's valid.)

rahimabdi:
The `validity` IDL override is already implemented in WebKit: https://bugs.webkit.org/show_bug.cgi?id=252549. I've updated the original issue comment as well.

Also noting that, per @cookiecrook's comment, it's not possible for user agents to only expose `validity` as `el.ariaInvalid` reflection is separate and should remain unaffected. To @annevk's point, the value of `aria-invalid` could be conveyed using `validity` as a downstream AT consideration, and it's likely more accurate to reference the source of the invalid state which is the [`:invalid` CSS pseudo-class](https://html.spec.whatwg.org/#selector-invalid) (since the `validity` IDL attribute reflects CSS `:invalid`).

 I would propose updating the html-aam spec to state something to the effect of:

> “Assistive technologies must expose the state of the [:invalid CSS pseudo-class](https://html.spec.whatwg.org/#selector-invalid) (i.e., [validity](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#the-constraint-validation-api:validitystate-3) IDL attribute).

annevk:
I did not mean to suggest to define it in terms of the `:invalid` pseudo-class. But rather, to define it in the same terms the `:invalid` pseudo-class is defined in (and perhaps abstract that set of terms into a shared abstraction). The `:invalid` pseudo-class, much like the `validity` IDL attribute, is public API, and is best not relied upon in algorithms.

html-aam/index.html
spectranaut: This comment is not about whether aria-invalid is reflected in javascript -- this is the html-aam, which means it is about how the user agent exposes that information to the accessibility API. The comment is about what to expose in the accessibility API when the IDL validity and aria-invalid conflict.

spectranaut: Also we can't write tests for that yet -- but I'm working on it :)

spectranaut:
I think this can land, but @cookiecrook @rahimabdi it could use more positive reviews.

html-aam/index.html
annevk: I suspected as much, but I continue to think this is very confusing. We shouldn't refer to public API, as public API can be overwritten in JavaScript.

Instead you probably want to have some kind of algorithm:

> Is element _element_ to be considered valid by AT:
> 1. If _element_ is a form-associated element:
>    1. return true if it satisfies its constraints (https://html.spec.whatwg.org/#concept-fv-valid); otherwise false.
> 2. If it has an aria-invalid attribute, then return false.
> 3. Return true.

spectranaut: I think I see your point about the algorithm, that is clear to me, except for the language `to be considered valid by AT` - this spec isn't about what ATs do, it is about what browsers expose over the accessibility APIs. So we have to use the language `User agents expose`, like in the original note.

Also I'm curious for my own understanding what you mean by "refer to public API"? What do you mean by public API?

annevk: Anything defined in IDL. We shouldn't be calling that directly.

rahimabdi:
@spectranaut I think this PR may need more work to address @annevk's comments above. I'll see if @scottaohara would like to take the algorithm approach suggested, i.e., (written in current html-aam style):

**`aria-invalid`**
- If a form control element [suffers from a validity constraint](https://html.spec.whatwg.org/#validity-states) and `aria-invalid` is also set, user agents MUST expose only the validity state.

PR #2224, opened 2024-05-24
[html-aam PR 359] Change figure & figcaption accName computation

### LINKED ISSUE #325: Should figcaption participate in name/desc computation?
I have seen a growing number of occurrences where people are properly using the figcaption as a child of figure, and have useful information in the caption, but are failing a checker because figcaption is not included in the computation for either name and description.

As per my issue opened with wcag https://github.com/w3c/wcag/issues/852 I think a case can be made that figcaption could be used for either a name or description, depending on circumstances. Has consideration been given to including it?

If you proceed with its incorporation, it is my intention to create a complementary technique for WCAG that helps authors appropriately use it.

### LINKED ISSUE #564: Stop automatically exposing details relation for figure -> figcaption
I thought we had filed this, so there may be a dupe somewhere.
We should stop automatically exposing the details relation between figure and figcaption. They are nearby in the DOM anyway. It was so noisy on some pages that it caused JAWS to stop reading details relation entirely. 

+ @scottaohara

pkra:
Moved from https://github.com/w3c/html-aam/pull/359/

closes: https://github.com/w3c/html-aam/issues/564
closes: https://github.com/w3c/html-aam/issues/325

Edit: wpt - https://github.com/web-platform-tests/wpt/pull/49647

* [ ] WPT tests: https://github.com/web-platform-tests/wpt/pull/49647
   * [x] WebKit: https://bugs.webkit.org/show_bug.cgi?id=292490
   * [ ] Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=2047263
   * [x] Blink: - https://issues.chromium.org/issues/434213625 (just for naming image from figcaption)
* [ ] Does this need AT implementations? No.
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR: Needs to be filed as MDN currently states figcaption is accname for figure: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/figure.


pkra:
Now pointing to main.

html-aam/index.html
```diff
@@ -2251,38 +2251,45 @@ <h4 id="el-figcaption">`figcaption`</h4>
...
+                <div class="role"><span class="type">Role:</span> Use WAI-ARIA mapping</div>
               </td>
             </tr>
             <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
             <tr>
               <th>Comments</th>
-              <td></td>
+              <td>
+                <p>
+                  Only the first child instance of a <a href="#el-figcaption">`figcaption`</a> element that is present in the accessibility tree (e.g., not 'hidden') provides details for its parent
```
rahimabdi, line 2289: @jcsteh's [migrated comment](https://github.com/w3c/html-aam/pull/359/files#r1429366035) provided below:

> For figure, we're explicit that aria-labelledby, aria-describedby or aria-details prevent exposure of the details relation. However, we're not similarly explicit here for the reverse relation. Common sense will hopefully prevail, but I don't love relying on common sense for a spec. Is there something I'm missing that would otherwise make this explicit?
> 
> This is a similar situation to https://github.com/w3c/html-aam/pull/481/files#r1424943049. It seems we tend to use ARIA to express roles in a cross-platform manner wherever possible now, but we don't do that for relations, I guess largely because reverse relations and the details-roles attribute complicate things.

rahimabdi:
Re-adding reviewers @cookiecrook @aleventhal @jcsteh.

html-aam/index.html
```diff
@@ -2251,38 +2251,45 @@ <h4 id="el-figcaption">`figcaption`</h4>
...
             </tr>
             <tr>
               <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
               <td>
-                <div class="role"><span class="type">Role:</span> `ATK_ROLE_CAPTION`</div>
+                <div class="role"><span class="type">Role:</span> Use WAI-ARIA mapping</div>
                 <div class="relations">
                   <span class="type">Relations:</span>
-                  `ATK_RELATION_LABEL_FOR` with parent <a href="#el-figure">`figure`</a> element
+                  If this is the first instance of a child `figcaption` to its parent <a href="#el-figure">`figure`</a> element, then expose `RELATION_DETAILS_FOR` referencing the parent `figure`
```
aleventhal, line 2273: @scottaohara @jcsteh 
JAWS turned off their details relation setting by default because it made some pages overly verbose. Is it really necessary when the figcaption is most likely just a sibling of the image being described anyway? Should we file a separate issue for this to remove the automatic assignment of the details relation for it? I don't think I understand why we need it.

scottaohara: i'm not opposed to removing the details relationship.  i think you're right here.  i think what's most important here is to make sure it's clear what the figure content is vs the caption.  the divide of where each start/end.  but the details relationship isn't needed for that.

scottaohara: if people are good with removing this, then i can pull all that out of this PR.  it'd definitely make it a lot simpler.

jcsteh: If we remove the details relationship, am I correct that the only thing this PR will do is remove the label association between figure and figcaption?

> i think what's most important here is to make sure it's clear what the figure content is vs the caption. the divide of where each start/end.

How will this PR achieve that? Or am I misunderstanding something here?

scottaohara: @jcsteh right now if we don't do anything else but remove the detail relationship from the PR, then yes, this PR would simply be about removing figcaption from contributing to the accName of the figure.

per my comment on making the distinction between the figure and its caption clear, i'm open to ideas if the existing caption role for the figcaption can't be the delineator that screen readers could use to identify this separation.

jcsteh: Personally, I think the role is a sufficient delineator.

html-aam/index.html
```diff
@@ -2251,38 +2251,31 @@ <h4 id="el-figcaption">`figcaption`</h4>
...
-                <div class="subrole"><span class="type">AXSubrole:</span> `(nil)`</div>
-                <div class="roledesc"><span class="type">AXRoleDescription:</span> `"group"`</div>
+                <div class="role"><span class="type">Role:</span> Use WAI-ARIA mapping</div>
               </td>
             </tr>
             <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
             <tr>
               <th>Comments</th>
-              <td></td>
+              <td> </td>
```
scottaohara:
```suggestion
              <td></td>
```

html-aam/index.html
```diff
@@ -2338,7 +2325,7 @@ <h4 id="el-figure">`figure`</h4>
...
```
scottaohara:
```suggestion
              <td></td>
```

scottaohara:
Details relationships have been removed from this PR.

So, what this PR does now is 
1. removes the mappings for figure receiving a name from the figure and figcaption elements from their mapping tables
2. removes figcaption from the name calculation steps in that section of the spec, and explicitly states in that section that figcaption does not contribute to name or description unless specified by the author (e.g., via aria)

With these changes, screen readers probably SHOULD, if they don't do so already, make sure that
1. the figure element has its role exposed (announced), whether it has a name or not(?), so that people can be aware of the start/end of a figure
2. the figcaption element SHOULD similarly have its caption role exposed (announced) so it can be made clear where the figure content vs caption content start/end.

Obviously, i'm not saying any screen reader 'must' do what i outlined, but it seemed the simplest enough way to indicate that some sort of changes to make sure people can effectively understand the figure's content vs caption can be made - since now the overly verbose/redundancy of figcaptions as names to figures is removed - and we will no longer be asking a details relationship to be exposed, since that had the unintended effect of trying to solve one verbosity issue with a different once.

aleventhal:
Landed the removal of the auto details relation for Chrome/Edge in https://chromium-review.googlesource.com/c/chromium/src/+/5891779

aleventhal (approved):

giacomo-petri (approved):

cookiecrook (approved):

scottaohara:
created web platform test for this PR (also added to OP) - https://github.com/web-platform-tests/wpt/pull/49647

html-aam/index.html
```diff
@@ -2309,24 +2302,18 @@ <h4 id="el-figure">`figure`</h4>
...
-                <div class="relations"><span class="type">Relations:</span> `IA2_RELATION_LABELLED_BY` with child <a href="#el-figcaption">`figcaption`</a> element</div>
               </td>
             </tr>
             <tr>
               <th><a href="https://msdn.microsoft.com/en-us/library/ms726297%28v=VS.85%29.aspx">UIA</a></th>
               <td>
                 <div class="role"><span class="type">Role:</span> Use WAI-ARIA mapping</div>
-                <div class="general">
-                  Accessible name derived from `figcaption` according to the <a href="#figure-element-accessible-name-computation">`figure` Element Accessible Name Computation</a>
-                </div>
```
cookiecrook, line 2321: Sorry I missed this... 

> removes figcaption from the name calculation steps in that section of the spec, and explicitly states in that section that figcaption does not contribute to name or description unless specified by the author (e.g., via aria)

What's the reason for this removal? WebKit and Safari implementations agree on using figcaption as the name if none other is provided via ARIA. Seems logical to me to keep this.

cookiecrook (dismissed):
flagging a section i may have missed earlier

html-aam/index.html
scottaohara: here is one of the examples that was referenced as rational for this change [specifically the link to the wicg thread where people are using figcaption to provide transcripts for a video within a figure](https://github.com/w3c/html-aam/pull/359#issuecomment-1017555985)

cookiecrook: yeah, that's a good example

cookiecrook: Scott and I chatted, and are working on a updated that will account for [this HTML text](https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element):

> 2. If the image is a descendant of a figure element that has a child figcaption element, and, ignoring the figcaption element and its descendants, the figure element has no [flow content](https://html.spec.whatwg.org/multipage/dom.html#flow-content-2) descendants other than [inter-element whitespace](https://html.spec.whatwg.org/multipage/dom.html#inter-element-whitespace) and the img element, then return the contents of the first such figcaption element.

scottaohara: new tests added to account for the quoted html spec text

scottaohara: updated the img element's name calculation to include figcaption per the above text from the HTML spec

rahimabdi (approved):

html-aam/index.html
```diff
@@ -16311,13 +16299,19 @@ <h4 id="el-img-name">`img` Element Accessible Name Computation</h4>
...
+              If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty: use `alt` attribute, even if its value is the empty string.
               <div class="note">
                 An `img` with an `alt` attribute whose value is the empty string is mapped to the
                 <a class="core-mapping" href="#role-map-presentation">`presentation`</a> role. It has no accessible name.
               </div>
             </li>
-            <li>Otherwise, if there is no `alt` attribute use the `title` attribute.</li>
+            <li>If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty and there is no `alt` attribute, use the `title` attribute.</li>
+            <li>
+              If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty and there are no `alt` or `title` attributes, and the `img` is a descendent of a `figure`
```
rahimabdi, line 16310:
```suggestion
              If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty and there are no `alt` or `title` attributes, and the `img` is a descendant of a `figure`
```

rahimabdi:
@scottaohara I've added an implementation tracker to this issue, and a link to the WebKit bug. I've submitted a WebKit PR (https://github.com/WebKit/WebKit/pull/45222#issuecomment-2870017030) and came across a question during implementation for `` accname:

> If the accessible name is still empty and there are no `alt` or `title` attributes, and the `img` is a descendant of a `figure` element with a child `figcaption` **but no other flow content or whitespace descendents**, then use the text equivalent computation of the `figcaption` element's subtree.

Should this be "but no other flow content or non-whitespace descendants"? I believe whitespace counts as flow content (and as text nodes), so curious if those should not factor here and should be stripped to ensure the `<figure>` only has an `` and `<figcaption>` as descendants.

html-aam/index.html
```diff
@@ -16311,13 +16299,19 @@ <h4 id="el-img-name">`img` Element Accessible Name Computation</h4>
...
                 An `img` with an `alt` attribute whose value is the empty string is mapped to the
                 <a class="core-mapping" href="#role-map-presentation">`presentation`</a> role. It has no accessible name.
               </div>
             </li>
-            <li>Otherwise, if there is no `alt` attribute use the `title` attribute.</li>
+            <li>If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty and there is no `alt` attribute, use the `title` attribute.</li>
+            <li>
+              If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty and there are no `alt` or `title` attributes, and the `img` is a descendant of a `figure`
+              element with a child `figcaption` but no other <a data-cite="html/dom.html#flow-content-2">flow content</a> or
+              <a data-cite="html/dom.html#inter-element-whitespace">whitespace</a> descendents, then use the <a data-cite="accname-1.2/#mapping_additional_nd_te">text equivalent computation</a> of the
```
scottaohara, line 16312:
```suggestion
              element with a child `figcaption` but no other <a data-cite="html/dom.html#flow-content-2">flow content</a> descendants, then use the <a data-cite="accname-1.2/#mapping_additional_nd_te">text equivalent computation</a> of the
```

scottaohara:
these updates are in chromium browsers and webkit, so we are merging this PR (finally)

scottaohara:
bug filed on chromium specifically for the figcaption not providing an image an accessible name - https://issues.chromium.org/issues/434213625

PR #2227, opened 2024-05-28
[Monorepo] Editors-draft-actions

daniel-montalvo:


spectranaut:
hey @daniel-montalvo, can you explain what part isn't working? Seems like everything is passing here. Is that because it is not landed? How have you been testing, btw?

@jnurthen or I can try to take a look once we have a bit more context

daniel-montalvo:
> hey @daniel-montalvo, can you explain what part isn't working? Seems like everything is passing here. Is that because it is not landed? How have you been testing, btw?

Thanks.

I think this is not even running on this PR because there's no core-aam directory, thus there's not been any changes to the core-aam/ directory.

I was getting stuck at the download artifact step, but I think something like  https://github.com/daniel-montalvo/aria/actions/runs/9299954322/ should work.

This is under my own forks that I set up earlier for testing.

@spectranaut @jnurthen @pkra  Does this sound like a path forward to replicate with the other specs?

daniel-montalvo:
@spectranaut @pkra @jnurthen Have you had a chance to review this?

If this sounds OK to you I can then replicate the action for the rest of the specs.

I think having an action per spec is best because we optimize the runs based on the paths, and avoid having unnecessary runs.

pkra:
@daniel-montalvo I'm not sure what kind of review comments make sense. Since it will only trigger on push to main, there's no test run to review, right? I don't know the artifact actions too well but it looks ok to me.

daniel-montalvo:
Thanks.

 

For now, I was only looking to see if that’s an approach that makes sense for you all. This is tricky to test, and I think we won’t really be able to test it until we merge.

 

Next steps for me are to replicate all the actions needed per spec and to add the steps for it to be able to publish on TR space.

--

Daniel Montalvo Charameli

Accessibility Specialist

W3C/WAI

.github/workflows/build-copy-and-publish-accname.yml
```diff
@@ -0,0 +1,64 @@
+name: Build, Copy, and Publish accname
+on:
+  push:
+    branches: [main]
+    paths: ['accname/**']
```
spectranaut, line 5: I think all of these paths should also include commons, correct?

.github/workflows/build-copy-and-publish-accname.yml
```diff
@@ -0,0 +1,64 @@
...
+          repository: w3c/accname
+          ref: gh-pages
+          path: accname
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Download artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: accname
+          path: accname
```
spectranaut, line 51: do we need `common` directory here as well?

spectranaut (changes requested):
thanks for this!! one question about the common directory -- see comments in line. otherwise I think this looks good, and maybe we need to land to test

.github/workflows/build-copy-and-publish-accname.yml
daniel-montalvo: What I am aiming for here is trust the spec-prod action to pick up all it needs from common. To see detailed steps on how this action runs, you can see https://github.com/w3c/accname/actions/runs/9424164070/job/25963922169#step:3:531 especially where it copies the relevant assets. It then details which assets it's copying and then it uploads them as an artifact together with the generated HTML snapshot.

.github/workflows/build-copy-and-publish-accname.yml
daniel-montalvo: Thanks for catching this. Yes, I do need to update these trigger events to include the common path.

PR #2237, opened 2024-06-07
Presentational Roles Conflict Resolution - global WAI-ARIA states or properties on inherited presentational roles

### LINKED ISSUE #2233: [Clarification] Presentational Roles Conflict Resolution - global WAI-ARIA states or properties on inherited presentational roles
In the chapter on [Presentational Roles Conflict Resolution](https://w3c.github.io/aria/#conflict_resolution_presentation_none), the 3rd bullet states:

> If an element has [global](https://w3c.github.io/aria/#dfn-global) WAI-ARIA states or properties, user agents MUST ignore the [none](https://w3c.github.io/aria/#none)/[presentation](https://w3c.github.io/aria/#presentation) role and instead expose the element's implicit role. However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.

Does this also encompass elements with an inherited presentational role? For instance:

```
<table role="presentation">
  <tbody>
    <tr>
      <td aria-description="Month">January</td>
      <td aria-description="Incoming">$100</td>
    </tr>
    <tr>
      <td aria-description="Month">February</td>
      <td aria-description="Incoming">$80</td>
    </tr>
  </tbody>
</table>
```
Is the third bullet relevant in this context?

I've observed that most browsers don't expose the td with a cell role (which seems reasonable), but perhaps we should clarify that the third bullet pertains solely to "explicit presentational roles."

giacomo-petri:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2237--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2237--wai-aria.netlify.app%2Findex.html)

Closes: #2233

- clarified global WAI-ARIA states or properties on inherited presentational roles
- reorganized the structure of the 3rd bullet to visually and semantically emphasize that the examples below strictly refer to this bullet. The final paragraph related to the author's responsibility is instead applicable to all points.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2237.html" title="Last updated on Dec 12, 2025, 11:07 AM UTC (74750ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2237/bf8e158...74750ac.html" title="Last updated on Dec 12, 2025, 11:07 AM UTC (74750ac)">Diff</a>


index.html
```diff
@@ -13524,16 +13524,18 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
-      <li>If an element has <a>global</a> WAI-ARIA states or properties, user agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and instead expose the element's implicit role. However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</li>
+			<li>
+				<p>If an element with an explicit presentational role has <a>global</a> WAI-ARIA states or properties, user agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and instead expose the element's implicit role. Conversely, if the presentational role is inherited and the element has <a>global</a> WAI-ARIA states or properties, user agents MUST NOT expose the element, maintaining its presentational role regardless of any <a>global</a> ARIA state or property. However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</p>
```
scottaohara, line 13528: >the element MUST NOT be exposed unless the

the element or the element's role?

giacomo-petri: Good point, Scott!
While looking through the ARIA specifications, I found similar wording and re-used it, assuming the intent was to not expose the element's role rather than the element itself (I can't recall the exact location right now). This interpretation seems supported by the following statement:

> Thus, the [none](https://w3c.github.io/aria/#none)/[presentation](https://w3c.github.io/aria/#presentation) role causes a given element to be treated as having no role or to be removed from the [accessibility tree](https://w3c.github.io/aria/#dfn-accessibility-tree), but does not cause the content contained within the element to be removed from the accessibility tree.

But I'm more than happy to use your wording to clarify that we do not expect the content to be ignored, only the element's role.

giacomo-petri: Oh, I found it! It was the original statement I didn't change:

> However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.

I'll update both

giacomo-petri, line 13528:
```suggestion
				<p>If an element with an explicit presentational role has <a>global</a> WAI-ARIA states or properties, user agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and instead expose the element's implicit role. Conversely, if the presentational role is inherited and the element has <a>global</a> WAI-ARIA states or properties, user agents MUST NOT expose the element's role, maintaining its presentational role regardless of any <a>global</a> ARIA state or property. However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element's role MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</p>
```

pkra:
@giacomo-petri I believe you need to rebase or cherry pick the commit from #2244

giacomo-petri:
@pkra, it seems there are other stray end tags...

cookiecrook:
Unassigning myself because I don't have the capacity to review a 30,000+ line diff.

jnurthen:
> Unassigning myself because I don't have the capacity to review a 30,000+ line diff. It won't render on GitHub because it's too large.

Hmm yeah - We could probably review this by just looking at the commits (other than the prettier one). But it may be easier to just cherry pick them into a new PR. @spectranaut @pkra do you have any suggestions?

pkra:
@jnurthen not sure what's going on here. The prettier action should ignore the aria  spec but it seems to have formatted it. 

We can try reverting the prettier commit. But if the actions are flaky, I'm wondering if it isn't time to bite the bullet and activate prettier (and update all PRs). Since I did that manually for the PRs from other specs, I should be able to do this for ARIA.

pkra:
I reversed the prettier commit, rebased on main, and force-pushed here. It looks like this solved the prettier problem.

@giacomo-petri, @jnurthen, @scottaohara, @cookiecrook could please take another look?

pkra:
@giacomo-petri I've rebased on main.

spectranaut:
@giacomo-petri does this need tests?

giacomo-petri:
- WPT tests created: https://github.com/web-platform-tests/wpt/pull/54126
- Results available here: https://wpt.fyi/results/wai-aria/role/role/role_none_conflict_resolution.tentative.html?label=pr_head&max-count=1&pr=54126


**Note:** 

When I originally wrote this PR, I believe browsers were not exposing the roles of children within elements assigned a presentational role. Now that all major browsers do expose the content of such children, should we update the phrasing in this PR from:

> Conversely, if the presentational role is inherited

to something like:

> Conversely, if the element has a [required accessibility parent role](https://w3c.github.io/aria/#scope) and that parent has a presentational role

or something similar?

index.html
```diff
@@ -16680,25 +16680,18 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
           <li>
-            If an element has <a>global</a> WAI-ARIA states or properties, user agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and instead expose the element's implicit role.
-            However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit
-            non-presentational role is applied.
+            <p>If an element with an explicit presentational role has <a>global</a> WAI-ARIA states or properties, user agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and instead expose the element's implicit role. Conversely, if the presentational role is inherited and the element has <a>global</a> WAI-ARIA states or properties, user agents MUST NOT expose the element's role, maintaining its presentational role regardless of any <a>global</a> ARIA state or property. However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element's role MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</p>
```
jnurthen:
```suggestion
              <p>If an element has <a>global</a> WAI-ARIA states or properties, the behavior depends on whether the presentational role is explicit or inherited:</p>
              <ul>
                <li>
                  <strong>Explicit presentational role:</strong> User agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and instead expose the element's implicit role.
                </li>
                <li>
                  <strong>Inherited presentational role:</strong> User agents MUST NOT expose the element's role, maintaining its presentational role regardless of any <a>global</a> ARIA state or property.
                </li>
              </ul>
              <p>However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element's role MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</p>

```

index.html
```diff
@@ -16680,25 +16680,18 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+            <p class="note">Some <a>global</a> WAI-ARIA states and properties are <a>prohibited</a> on certain roles. These states and properties are still considered global for the purposes of Presentational Role Conflict resolution.</p>
+            <p>For example, <pref>aria-describedby</pref> is a global attribute and would always be applied; <pref>aria-level</pref> is not a global attribute and would therefore only apply if the element was not in a presentational state.</p>
+            <pre class="example highlight">
+              <span class="comment">&lt;!-- 1. [role="none"] is ignored due to the global aria-describedby property. --&gt;</span>
+              &lt;h1 role="none" aria-describedby="comment-1"&gt; Sample Content &lt;/h1&gt;
+              <span class="comment">&lt;!-- 2. [role="none"] negates both the implicit 'heading' and the non-global aria-level. --&gt;</span>
+              &lt;h1 role="none" aria-level="2"&gt; Sample Content &lt;/h1&gt;
+            </pre>
+          </li>
```
jnurthen: remove duplicate closing list

jnurthen (changes requested):
I think this is clearer.

index.html
giacomo-petri: Thanks @jnurthen, great improvement.

index.html
```diff
@@ -16680,25 +16680,28 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
           </li>
```
giacomo-petri, line 16789:
```suggestion
```

index.html
```diff
@@ -16680,25 +16680,27 @@ <h3>Presentational Roles Conflict Resolution</h3>
             explicit role. If the action of exposing the explicit role causes the accessibility tree to be malformed, the expected results are undefined.
```
front-endian: Happy to move this to a separate issue to avoid bloating this PR, but does the following only apply to this second exception, or does this apply to all of them?
> If the action of exposing the explicit role causes the accessibility tree to be malformed, the expected results are undefined.

index.html
```diff
@@ -16680,25 +16680,27 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+              <p>If an element has <a>global</a> WAI-ARIA states or properties, the behavior depends on whether the presentational role is explicit or inherited:</p>
+              <ul>
+                <li>
+                  <strong>Explicit presentational role:</strong> User agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and instead expose the element's implicit role.
+                </li>
+                <li>
+                  <strong>Inherited presentational role:</strong> User agents MUST NOT expose the element's role, maintaining its presentational role regardless of any <a>global</a> ARIA state or property.
+                </li>
+              </ul>
+              <p>However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element's role MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</p>
```
front-endian: It is a bit hard to parse all this in context since there is some jumping back and forth between exception, non-exceptions and different ways of writing things.

Because all of this is preempted by,

> User agents MUST NOT expose elements having explicit or inherited presentational role in the accessibility tree, with these exceptions

... I wonder if this might be easier to parse if some of the parts explaining when elements _must not_ be exposed were left out? That way the reader doesn't have to jump back and forth between examples of "must not" and "must" and can focus in on the few exceptions.

Maybe something like:

> *[first two items in list of exceptions]*
> - If an element has global WAI-ARIA states or properties, and has an explicit presentational role, user agents MUST ignore the none/presentation role and instead expose the element's implicit role.
> - If an element has an inherited presentational role but an explicit non-presentational role is applied to it as well, user agents MUST ignore the inherited presentational role and expose the explicit role applied to the element.

Then maybe the, "however" isn't needed because the above exception clearly states it only applies when there are **global** states or properties.

index.html
```diff
@@ -16680,25 +16680,27 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+
+            <p class="note">Some <a>global</a> WAI-ARIA states and properties are <a>prohibited</a> on certain roles. These states and properties are still considered global for the purposes of Presentational Role Conflict resolution.</p>
+            <p>For example, <pref>aria-describedby</pref> is a global attribute and would always be applied; <pref>aria-level</pref> is not a global attribute and would therefore only apply if the element was not in a presentational state.</p>
```
front-endian: Small edit to keep tenses consistent:

```suggestion
            <p>For example, <pref>aria-describedby</pref> is a global attribute and would always be applied; <pref>aria-level</pref> is not a global attribute and would therefore only apply if the element is not in a presentational state.</p>
```

front-endian (changes requested):
Have some requested changes to simplify/clarify a few parts.

index.html
front-endian: I added the second suggested bullet point because it was effectively being stated with what's currently in the PR, but I would have no problems leaving it out as I believe other parts of the spec make it clear that an explicit role overrides an inherited role in cases like this.

index.html
giacomo-petri: I think you are right. Any of these author errors could result in a malformed accessibility tree. @spectranaut, should we open a separate issue for this, or would it be fine to handle it in the current PR by removing the sentence from the second bullet and adding it as a separate note at the end?

index.html
giacomo-petri: This seems to go against @jnurthen’s proposal here: https://github.com/w3c/aria/pull/2237#discussion_r2270813839

While I understand that "Inherited presentational role" isn’t within the scope of the first IF condition, I find the current version clearer due to the list structure and the upfront term definition.

@front-endian @jnurthen, would it be a good compromise to remove the nested list in favor of @front-endian’s suggestion:

> If an element has global WAI-ARIA states or properties, and has an explicit presentational role, user agents MUST ignore the none/presentation role and instead expose the element's implicit role.

and then add a note at the bottom explaining what happens when an element has both global WAI-ARIA states/properties and an inherited presentational role?

front-endian: Looking at this again after the WG meeting this week: I think there is a more minor edit that would address my concerns about the potential for misinterpretation, while also preserving the good suggestion @jnurthen had previously. I'll write up a simpler suggested edit shortly.

front-endian: I agree that the nested list is easier to read and keeping it would be better. The source of my confusion reading this is the "however" sentence for non-global states/properties, since the reader is flipping between "must not" and "must" a few times.

What about keeping the nested list and moving the "however" sentence about non-global states/properties into its own item? It could mirror the item for global states/properties and use a nested list since the structure of the rule is similar. Something like:

> _[no changes]_ User agents MUST NOT expose elements having explicit or inherited presentational role in the accessibility tree, with these exceptions:
> - _[...]_
> - _[...]_
> - _[...]_
> - If an element **only** has **non-global** WAI-ARIA states or properties, the behavior depends on the element's role:
>   - **Explicit non-presentational role:** User agents MUST expose the element's non-presentation role, regardless of whether the element has an inherited presentational role.
>   - **Explicit presentational role:** User agents MUST NOT expose the element's role, maintaining the element's presentational role.
>   - **Inherited presentational role without an explicit non-presentational role:** User agents MUST NOT expose the element's role, maintaining its presentational role.

Or alternatively, it could be simpler and only mention the exception:

> _[no changes]_ User agents MUST NOT expose elements having explicit or inherited presentational role in the accessibility tree, with these exceptions:
> - _[...]_
> - _[...]_
> - _[...]_
> - If an element with an implicit presentational role and explicit non-presentational role **only** has **non-global**, role specified WAI-ARIA states or properties, the element's role MUST be exposed.

front-endian (commented):
Suggesting a more minor edit to preserve the easier to follow nested list.

index.html
front-endian: Because other tasks are blocked on this, if neither of these edits seem better to you @giacomo-petri  I am okay with closing these suggested revisions, approving this PR, and I can open a separate issue after this PR is merged.

index.html
front-endian: Pulled this out into a separate issue so this PR can proceed faster: #2638

giacomo-petri:
@front-endian,

sorry, I wasn't able to work on this earlier. I've revisited this ticket after a couple of weeks (which actually helped), since I was able to review it with a fresh perspective and without the prior context.

I noticed that the second sub-bullet under "If an element has global WAI-ARIA states or properties…" is "out of scope" for that section. The section's description states:

> User agents MUST NOT expose elements having explicit or inherited presentational roles in the accessibility tree, with these exceptions.

However, the second sub-bullet says:

> User agents MUST NOT expose the element's role,

which contradicts the section's intent (it's not an exception).

I've updated the wording as follows:
- specified that if an element has global WAI-ARIA states or properties and the presentational role is explicit, user agents MUST ignore the none/presentation role and instead expose the element's implicit role.
- added a new bullet stating that if an element has only non-global, role-specific WAI-ARIA states or properties, the presentational role is inherited, and an explicit non-presentational role is applied, then user agents MUST ignore the inherited presentational role and instead expose the element's explicit role.
- Moved the "However…" part into a note, since it's not part of the exceptions.
- Moved the "Inherited presentational role: User agents MUST NOT expose the element's role…" sentence into a note as well, since this too isn't part of the exceptions.


Note:
I've kept the original proposal as Option 1 for reference. The new version is Option 2. If the new proposal addresses everyone's concerns, I'll remove Option 1.

Thanks!

index.html
front-endian: Resolving as the newest version (option 2) resolves all my concerns.

front-endian (approved):
@giacomo-petri , sorry for the delay re-reviewing this, but Option 2 looks great to me. Clears up all my concerns and is easy to follow. Thank you!

Approved assuming we go with something like Option 2.

giacomo-petri:
I recently realized that I also need to address the following case:

```
<ul role="none">
    <li role="none" aria-describedby="test" data-testname="presentational roles conflict - ul[role=none] &gt; li[aria-describedby]" class="ex-generic">a</li>
    <li aria-describedby="test">b</li>
    <li aria-describedby="test">c</li>
  </ul>
```

This works as expected across all UA, but the current behavior is not aligned with the specification (both the current one and the version I proposed). 

According to the current specifications, the first list item should technically be exposed with a listitem role. However, in practice, all major user agents correctly treat it as presentational, which is also the intended behavior.
So I'll need to apply one more small adjustment.

index.html
```diff
@@ -16670,6 +16670,7 @@ <h3>States and Properties</h3>
         <h3>Presentational Roles Conflict Resolution</h3>
         <p>There are a number of ways presentational role conflicts are resolved.</p>
         <p>User agents MUST NOT expose [=elements=] having explicit or inherited presentational role in the accessibility tree, with these exceptions:</p>
```
cookiecrook: Do we want to make it clear here that "MUST not expose elements" does not mean "MUST not expose any descendant elements either"? This was one of the biggest points of web author confusion with the name of the "presentation" role, and the reason we later added the synonym "none" role.

index.html
```diff
@@ -16670,6 +16670,7 @@ <h3>States and Properties</h3>
...
+		<h4>Option 1</h4>
         <ul>
           <li>
             If an element is focusable, user agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and expose the element with its implicit role, in order to ensure that the element
```
cookiecrook: What happens if you have an inherited none role on a focusable elements whose required accessibility parent role is explicitly none (and valid)?

```html
<ul role="none">
  <tr>
    <td tabindex="0"><!-- is this an orphaned `cell` or a minimum role… maybe `group`? -->
```

FYI @scottaohara

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
             explicit role. If the action of exposing the explicit role causes the accessibility tree to be malformed, the expected results are undefined.
           </li>
           <li>
-            If an element has <a>global</a> WAI-ARIA states or properties, user agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and instead expose the element's implicit role.
-            However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit
-            non-presentational role is applied.
+              <p>If an element has <a>global</a> WAI-ARIA states or properties, the behavior depends on whether the presentational role is explicit or inherited:</p>
```
```suggestion
              <p>If an element has <a>global</a> WAI-ARIA attributes, the behavior depends on whether the presentational role is explicit or inherited:</p>
```

cookiecrook: ARIA editors policy is to avoid the state vs prop distinction except in the very few cases where the difference is relevant.

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+              <ul>
+                <li>
+                  <strong>Explicit presentational role:</strong> User agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and instead expose the element's implicit role.
+                </li>
+                <li>
+                  <strong>Inherited presentational role:</strong> User agents MUST NOT expose the element's role, maintaining its presentational role regardless of any <a>global</a> ARIA state or property.
+                </li>
+              </ul>
+              <p>However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element's role MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.</p>
```
cookiecrook: passive voice.... "the element's role MUST NOT be exposed" by what?

Presumably you want:
```suggestion
              <p>However, if an element has only non-global, role-specific WAI-ARIA states or properties, User Agents MUST NOT expose the element's role unless the presentational role is inherited and an explicit non-presentational role is applied.</p>
```

…though I'm having trouble parsing the second half of the sentence. How can there be both an inherited presentational role and an explicit non-presentational role?

cookiecrook: FYI @front-endian since she raised a similar clarity concern in the resolved comment before this one.

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+
+            <p class="note">Some <a>global</a> WAI-ARIA states and properties are <a>prohibited</a> on certain roles. These states and properties are still considered global for the purposes of Presentational Role Conflict resolution.</p>
```
```suggestion
            <p class="note">Some <a>global</a> WAI-ARIA attributes are <a>prohibited</a> on certain roles. These states and properties are still considered global for the purposes of Presentational Role Conflict resolution.</p>
```

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+            <p>For example, <pref>aria-describedby</pref> is a global attribute and would always be applied; <pref>aria-level</pref> is not a global attribute and would therefore only apply if the element is not in a presentational state.</p>
```
cookiecrook: Would prefer to avoid the word "state" here given the other use of that term in the ARIA context.

Perhaps this?
```suggestion
            <p>For example, <pref>aria-describedby</pref> is a global attribute and would always be applied; <pref>aria-level</pref> is not a global attribute and would therefore only apply if the element is not presentational.</p>
```

cookiecrook: It might be more clear to avoid using the term "presentational" anywhere it can be avoided across the section.
```suggestion
            <p>For example, <pref>aria-describedby</pref> is a global attribute and would always be applied; <pref>aria-level</pref> is not a global attribute and would therefore only apply if the element does not have an explicit or inherited role of `none`.</p>
```

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
-<span class="comment">&lt;!-- 1. [role="none"] is ignored due to the global aria-describedby property. --&gt;</span>
-&lt;h1 role="none" aria-describedby="comment-1"&gt; Sample Content &lt;/h1&gt;
-<span class="comment">&lt;!-- 2. [role="none"] negates both the implicit 'heading' and the non-global aria-level. --&gt;</span>
-&lt;h1 role="none" aria-level="2"&gt; Sample Content &lt;/h1&gt;
-		</pre>
+        <p>Authors MUST NOT use <rref>presentation</rref> and <rref>none</rref> on elements where user agents will ignore that role because it conflicts with one of the above items.</p>
+		<h4>Option 2</h4>
+		<ul>
+          <li>
+            If an element is focusable, user agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and expose the element with its implicit role, in order to ensure that the element
```
cookiecrook: Thinking about this more, I don't think we can force engines to expose the implicit role on elements with a validly inherited role of none, like in my list `example` from above.

```html
<ul role="none">
  <tr>
    <td tabindex="0"><!-- is this an orphaned `cell` or a minimum role… maybe `group`? -->
```

Maybe the solution is (hand-wavy, half-considered draft text follows):

- If element is focusable or otherwise has ~some attributes], ignore any **_explicit_** role of none/presentation.
- If element is focusable or otherwise has ~some attributes, and the element element's required accessibility parent has a role of none, do not inherit the none role, and instead, expose the element's minimum role.

This may be difficult to implement, if you all settle on any specific language that can solve this, I'd want to run it by more engine accessibility engineers for each of the major browsers.

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+            is <a>operable</a>.
+          </li>
+          <li>
+            If an <a href="#mustContain">allowed child element</a> has an explicit non-presentational role, user agents MUST ignore an inherited presentational role and expose the element with its
+            explicit role. If the action of exposing the explicit role causes the accessibility tree to be malformed, the expected results are undefined.
+          </li>
```
cookiecrook: > If an allowed child element has an explicit non-presentational role, user agents MUST ignore an inherited presentational role

the first part seems unnecessary: if it has an explicit role, it doesn't have an inherited role, so this doesn't need to be stated.

> If the action of exposing the explicit role causes the accessibility tree to be malformed, the expected results are undefined.

This doesn't seem relevant to the "presentational role conflict" section we're in, since this could happen with or without the explicitly-overridden-so-therefore-not-inherited `none` role.

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+		  <li>
+			If an element with an explicit presentational role has <a>global</a> WAI-ARIA states or properties, user agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and instead expose the element’s implicit role.
```
```suggestion
			If an element with an explicit presentational role has <a>global</a> WAI-ARIA attributes, user agents MUST ignore the <rref>none</rref>/<rref>presentation</rref> role and instead expose the element’s implicit role.
```

Or minimum role?

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+		  	<p class="note">Some <a>global</a> WAI-ARIA states and properties are <a>prohibited</a> on certain roles. These states and properties are still considered global for the purposes of Presentational Role Conflict resolution.</p>
```
```suggestion
		  	<p class="note">Some <a>global</a> WAI-ARIA attributes are <a>prohibited</a> on certain roles. These attributes are still considered global for the purposes of Presentational Role Conflict resolution.</p>
```

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+            <p>For example, <pref>aria-describedby</pref> is a global attribute and would always be applied; <pref>aria-level</pref> is not a global attribute and would therefore only apply if the element is not in a presentational state.</p>
```
```suggestion
            <p>For example, <pref>aria-describedby</pref> is a global attribute and would always be applied; <pref>aria-level</pref> is not a global attribute and would therefore only apply if the element does not have an explicit or inherited role of <rref>none</rref> or <rref>presentation</rref>.</p>
```

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+            <pre class="example highlight">
+              <span class="comment">&lt;!-- 1. [role="none"] is ignored due to the global aria-describedby property. --&gt;</span>
```
```suggestion
              <span class="comment">&lt;!-- 1. [role="none"] is ignored due to the global aria-describedby attribute. --&gt;</span>
```

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+              &lt;h1 role="none" aria-describedby="comment-1"&gt; Sample Content &lt;/h1&gt;
+              <span class="comment">&lt;!-- 2. [role="none"] negates both the implicit 'heading' and the non-global aria-level. --&gt;</span>
```
cookiecrook: good examples

index.html
```diff
@@ -16680,25 +16681,53 @@ <h3>Presentational Roles Conflict Resolution</h3>
...
+              &lt;h1 role="none" aria-level="2"&gt; Sample Content &lt;/h1&gt;
+            </pre>
+		  </li>
+		  <li>
+			<p>If an element with an inherited presentational role but also an explicit non-presentational role, has only non-global, role-specific WAI-ARIA states or properties, then user agents MUST ignore the inherited presentational role and instead expose the element’s explicit role.</p>
```
cookiecrook: Again, I don't know what it means to have both of these. You can't have both an inherited presentational role and an explicit non-presentational role.

PS. When you modify this next, there's also another instance of "states or properties" that should be "attributes"

cookiecrook (changes requested):

index.html
scottaohara: agreed... in situations like this the thing that might make the most sense is to fall back to an element's minimum role or generic.  not ideal, but would make more sense than exposing a lone cell

index.html
front-endian: I second that if we mention what presentation roles do in this section it would be helpful to clarify that given how common/unintuitive/impactful that misunderstanding is.

On the other hand, do we even have to document any of this section since this is starting to duplicated the existing documentation of `role="none"`? Could it be simplified to something along the lines of:
> User agents follow the expected behavior for elements with a presentational role in the accessibility tree, with these exceptions:

index.html
front-endian: This looked like a good edit to me.

> How can there be both an inherited presentational role and an explicit non-presentational role?

I think the second half is trying to clarify that an explicit `role="blah"` overrides any situation where a presentational role would be implicitly applied? It does feel like over clarifying things and I'm not sure I understand how this situation would happen on a website.

index.html
front-endian: @cookiecrook, [this came up previously](https://github.com/w3c/aria/pull/2237#discussion_r2321257351) and there is an open issue to make this change: #2638 . It was moved to a separate issue since this bit is carried over from the current published spec, things were getting a bit scope creep-y with what this PR is intended to fix, and to help this PR make progress.

That being said, I agree that that edit should be made at some point. With #2638 I was waiting until this PR landed before pushing an edit, but I could also push an edit now since this PR is taking a while.

index.html
giacomo-petri: This is the original text (as written in [ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/#conflict_resolution_presentation_none)). My interpretation is that the intention is to clarify that we are dealing specifically with a presentational-role conflict: an implicit presentational role is creating the conflict, and that implicit role is overridden by an explicit non-presentational role (which is quite obvious and not needed to mention).

Given that, and considering your concerns, we can replace:
> However, if an element has only non-global, role-specific WAI-ARIA states or properties, the element MUST NOT be exposed unless the presentational role is inherited and an explicit non-presentational role is applied.

with:
> However, if an element with an explicit or implicit presentational role has only non-global, role-specific WAI-ARIA attributes, user agents MUST NOT expose the element's role.

The original "unless" clause simply states that the element actually has a non-presentational role, which puts it outside the scope of a presentational-role conflict.

If you agree, I'll move forward with this change.

index.html
giacomo-petri: I see some parallels with my [last comment](https://github.com/w3c/aria/pull/2237#issuecomment-3521271299) 

My initial hypothesis (though it needs a deeper review) is that this applies to any condition (focusable, global attribute, ...) where the element’s required accessibility parent is presentational.

If that turns out to be correct, I’ll add a condition for it.

index.html
giacomo-petri: I've just realised that we don't actually define "Presentational Roles Conflict Resolution" anywhere. We mention that there are several ways such conflicts are resolved, but we never state what it is.

I'll add a definition that explains the concept explicitly, also clarifying that "MUST not expose elements" does not mean "MUST not expose any descendant elements" and maybe including an example to illustrate this.

Edit:
While editing, including examples felt excessive since this section is about conflicts, not presentational roles. I've added a brief clarification in parentheses, but this will probably be better addressed in the role="none" section.

index.html
giacomo-petri: I'm running the following examples through multiple UA

```
<h2>ul[role"none"][tabindex="0"] > li[tabindex="0"]</h2>
    <ul role="none" tabindex="0">
      <li tabindex="0">a</li>
    </ul>

    <h2>ul[role="none"][tabindex="0"] > li</h2>
    <ul role="none" tabindex="0">
      <li>a</li>
    </ul>

    <h2>ul[role="none"] > li[tabindex="0"]</h2>
    <ul role="none">
      <li tabindex="0">a</li>
    </ul>

    <h2>table[role="none"][tabindex="0"] td[tabindex="0"]</h2>
    <table role="none" tabindex="0">
      <tr>
        <td tabindex="0">test</td>
      </tr>
    </table>

    <h2>table[role="none"][tabindex="0"] td</h2>
    <table role="none" tabindex="0">
      <tr>
        <td>test</td>
      </tr>
    </table>

    <h2>table[role="none"] td[tabindex="0"]</h2>
    <table role="none">
      <tr>
        <td tabindex="0">test</td>
      </tr>
    </table>
```

The results are very inconsistent (not just across different UA, but even within the same UA depending on the element involved).

Additionally, if we say that in: 
```
<ul role="none">
  <tr>
    <td tabindex="0">
```
the `<td>` should be exposed as a generic, then by the same logic a

`<section role="none" tabindex="0">`

should also be exposed as a generic. In the first example the presentational role is inherited (which might indicate author oversight), but in the `<section>` example the `role="none"` and the `tabindex` are on the same element, implying the author is making a more intentional choice (at least in theory).

Since we don't have a clearly defined definition for presentational conflicts, should we simply scope focus-related presentational conflicts to focusable elements that have an explicit presentational role but an implicit widget role?
In every other situation there is no presentational conflict, so the author's role="none" should be honored.

Based on that logic, in all of the examples at the beginning of my comment, both the required parent and the acc child should be exposed as generic/minimum role.

giacomo-petri:
TODO: Missing condition for non-focusable widget roles(?) nested inside a presentational child element. 

```
<div role="button">
    <button>
```
This scenario is already covered by the focusable bit, where we state that focusable elements must be exposed with their implicit role (need to tweak it based on @cookiecrook [example](https://github.com/w3c/aria/pull/2237#discussion_r2594338802))

----

```
<button>
    <div role="button">
```
However, this case is not covered, because the nested non-focusable button should instead be treated as presentational (i.e., its children should be presentational). Per the ARIA specs, it should be exposed as a generic element, but all user agents currently expose it as a button, as confirmed by the WPT reults in [WPT-50045](https://github.com/web-platform-tests/wpt/pull/50045))

index.html
cookiecrook: > However, if an element with an explicit or implicit presentational role has only non-global, role-specific WAI-ARIA attributes, user agents MUST NOT expose the element's role.

This part is now ambiguous too I think because the earlier part of the sentence referenced two different potential roles: "MUST NOT expose the element's role."

UAs MUST NOT expose the `none` role? Or UAs MUST NOT the other role that's associated with the role-specific attributes?

giacomo-petri:
I've updated the PR:
- I removed option 1 for clarity, since previous comments favored option 2.
- I believe the concerns highlighted in earlier comments are now addressed, with clearer definitions of what happens in scenarios such as:
  ```
  <ul role="none">
    <tr>
      <td tabindex="0">
  ```
    I also added a couple of additional examples to make this behavior clearer.
- While working on:
  ```
  <button>
    <div role="button">button</div>
  </button>
  ```
    I encountered the issue we discussed some time ago regarding "presentational children" and how UAs treat them, which makes the concept of "presentational children" somewhat theoretical in practice. However, by addressing this within the presentational roles conflict resolution, we can clarify that UAs may continue their current behavior for the benefit of end users, while still treating it as a conflict and therefore an authoring error.

index.html
```diff
@@ -16741,28 +16741,57 @@ <h3>States and Properties</h3>
...
+<span class="comment">&lt;!-- 3. &lt;td&gt; inherits [role="none"] from &lt;ul role=none&gt; and the inherited presentational role is not ignored even though the element has a global WAI-ARIA attribute, because the required ancestor has a presentational role --&gt;</span>
+&lt;table role="none"&gt;
+	&lt;tr&gt;
+		&lt;td aria-describedby="comment-1"&gt;Content&lt;/td&gt;	
+	&lt;/tr&gt;
+&lt;/table&gt;
+            </pre>
+		  </li>
+		  <li>
+			If an element has a presentational role as a result of <a href="#childrenArePresentational">children presentational</a>, user agents MAY ignore the inherited presentational role and expose the element with its role, to preserve the element's implicit role semantics.
```
spectranaut, line 16781: It seems to me that this is in agreement with the request to deprecate children presentational, as requested in this issue: https://github.com/w3c/aria/issues/2509

The WG didn't feel comfortable making that change, because there are scenarios where assistive technologies cannot handle that content.

spectranaut (changes requested):
If we are keeping children presentation (which somehow I've been thrown into question on whether we should) I think we should land this, but without the bullet I commented on, because as far as I can tell that is equivalent to removing the children presentation requirement entirely.

PR #2245, opened 2024-06-13
Add heading and separator roles to feed's allowed accessibility children

### LINKED ISSUE #2236: Consider adding headings to feed's allowed accessibility children

I've run into a use case where headings are used to break up sections of multiple articles within a feed. These are not article-specific headings (which should be within the article node); in my case, the headings have date/time text that group multiple articles below them.

My specific use case is a chat feed with timestamps that group several messages. Another more traditional use case could be a blog feed with headings per month or year.

I haven't run into any practical issues with this semantic structure -- keyboard focus doesn't land on the headings, and no one has complained about them in user testing yet.

smhigley:
Closes #2236

Based on the working group discussion of the linked issue, we decided to make a quick PR to add heading & separator to feed, and work on ARIA role content models separately. At some point, the content model approach may replace the heading/separator addition when it's ready.

* [x] Browser implementations (link to issue or commit): N/A
   * WebKit:
   * Gecko:
   * Blink:
* [x] Does this need AT implementations? no
* [x] MDN Issue/PR: N/A


shirsha:
In The feed role , "Authors  MAY  use [ heading ](https://w3c.github.io/aria/#heading) and [ separator ](https://w3c.github.io/aria/#separator) children in a  feed . For example, in a feed of news stories there could be headings separating sets of articles into sequential dates. If a  feed  does have [ headings ](https://w3c.github.io/aria/#headings) or [ separators ](https://w3c.github.io/aria/#separators) as children, they  SHOULD NOT  be focusable or contain information necessary for understanding or navigating the  feed  's [ articles ](https://w3c.github.io/aria/#article)."

Confused with last sentence "If a  feed  does have [ headings ](https://w3c.github.io/aria/#headings) or [ separators ](https://w3c.github.io/aria/#separators) as children, they  SHOULD NOT  be focusable or contain information necessary for understanding or navigating the  feed  's [ articles ](https://w3c.github.io/aria/#article)". May be provide examples like timestamp and name etc. Provide some explicit info

giacomo-petri:
Should we consider adding the `scrollbar` role as an allowed child element?

When dealing with a high but "static" number of articles, showing them on scroll for any reason (e.g. for performance reasons), this setup might be one of the rare cases where an author would want to create a custom scrollbar. 
By doing so, it would give the impression that all the content is already fully loaded, allowing users to see the entire scroll progress right from the initial page load.

I understand that the scrollbar can be positioned outside the feed element, but I don't see why the author shouldn't be allowed to place it inside, as the scrollbar is an integral part of the feed document.

Regarding the rest of the addition, I like it. The details about focusability and not including important text are crucial for maintaining a smooth experience for keyboard users.

smhigley:
> Should we consider adding the scrollbar role as an allowed child element?

@giacomo-petri this makes sense to me! Though I think scrollbar should generally be an allowed child of any composite widget, not just feed. Probably as a separate PR, or as part of the roles categorization that @scottaohara mentioned on the last ARIA call -- WDYT?

smhigley:
@shirsha the sentence you quoted contains an example in it -- "For example, in a feed of news stories there could be headings separating sets of articles into sequential dates." Does that address your comment?

There's nothing added to the heading and separator roles, since this change doesn't affect their usage outside of feed.

pkra:
@smhigley rebased on main.

index.html
```diff
@@ -4058,6 +4058,11 @@ <h2>Definition of Roles</h2>
               set on the <rref>article</rref> or one of its descendant elements. For example, in HTML, each <rref>article</rref> element should have a <code>tabindex</code> value of either
               <code>-1</code> or <code>0</code>.
             </p>
+            <p>
+              Authors MAY use <rref>heading</rref> and <rref>separator</rref> children in a <code>feed</code>. For example, in a feed of news stories there could be headings separating sets of
```
pkra, line 4045:
```suggestion
              Authors MAY use elements with role <rref>heading</rref> and <rref>separator</rref> as <a>accessibility children</a> of an element with role <code>feed</code>. For example, in a feed of news stories there could be headings separating sets of
```

index.html
```diff
@@ -4058,6 +4058,11 @@ <h2>Definition of Roles</h2>
...
+              articles into sequential dates. If a <code>feed</code> does have <rref>headings</rref> or <rref>separators</rref> as children, they SHOULD NOT be focusable or contain information
+              necessary for understanding or navigating the <code>feed</code>'s <rref data-lt="article">articles</rref>.
```
pkra, line 4047: I'm not sure about the two SHOULD NOTs. 

For the first part about focusability. Why should headings not be focusable? I'm not sure if the APG example is any good, but thinking about this change from that example, it seems to me a user may want to stop at headings when navigating through the feed.

The second part is about the contents. I feel that's out of scope for ARIA - it feels WCAG-ish to me. Without much deep diving, I couldn't find anything like it in the spec. Are we sure we want to start with this?

pkra: For comparison, https://github.com/w3c/aria/pull/2159 uses a note for similar author guidance.

PR #2261, opened 2024-06-27
Editorial: Added definition of "host language"

### LINKED ISSUE #2035: "host language attribute" is not defined
"host language attribute" is not defined in main ARIA spec (and related)

I heard today that a few ARIA veterans didn't know what this was. .... e.g. and HTML or SVG attr since those are "host languages" for ARIA

skeetergraphics:
Closes #2035 

Used existing verbiage to make a definition of "host language" while keeping the content as in tact as possible.


index.html
```diff
@@ -243,7 +243,12 @@ <h2>User Agent Support</h2>
 	</section>
 	<section id="co-evolution">
 		<h2>Co-Evolution of <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> and Host Languages</h2>
-		<p><abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> is intended to augment semantics in supporting languages like [[HTML]] and [[SVG2]], or to be used as an accessibility enhancement technology in other markup-based languages that do not explicitly include support for ARIA. It clarifies semantics to assistive technologies when authors create new types of objects, via style and script, that are not yet directly supported by the language of the page, because the invention of new types of objects is faster than standardized support for them appears in web languages.</p>
+		<p>A host language is a markup-based language that either:</p>
+		<ul>
+			<li>supports <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr>, using it to augment semantics; or</li>
+			<li>does not explicitly support <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr>, using it as an accessibility enhancement technology.</li>
```
spectranaut, line 249: I feel like this bullet is a little questionable when pulled out of the original paragraph. it's kind of question in the original paragraph to begin with.

Personally, I'd remove the bullets, and say something like what you wrote:
```
A host language is a markup-based language in which ARIA can be used as an accessibility enhancement technology. Examples include [[HTML]] and [[SVG2]], which both explicitly support the use of ARIA. ARIA roles and properties clarify semantics to assistive technologies...
```

spectranaut (changes requested):
One suggestion! :)
And @jnurthen can give advice about exporting definitions, because I'm less familiar with it.

index.html
skeetergraphics: I agree with @spectranaut’s direction of removing the bullets and feel we have two slightly different definitions or concepts which the two bullets represented.

If ARIA as an “accessibility enhancement technology” is well-understood and if explicitly supporting ARIA is the same as supporting ARIA, this statement may be sufficient:

A host language is a markup-based language that supports the use of ARIA―examples include [[HTML]] and [[SVG2]].

skeetergraphics: .

cookiecrook: +1 to @spectranaut's rewrite... I don't think the explicit vs implicit support distinction you're making in the original PR text is especially relevant.

pkra:
@skeetergraphics I'm afraid I cannot help rebasing this to the changes on main since it lives on a fork. I hope it won't be too painful.

MelSumner (approved):

pkra:
I've opened #2563 which aims to replace this PR.

PR #2264, opened 2024-06-27
HTML AAM: Clarification for popovertargetaction

scottaohara:
this line was meant to mean that whether someone explicitly defines the popovertargetaction=auto or if an author leaves off this attribute, making it an implicit popovertargetaction.

changing this to hopefully be clearer about this, and not have people think it only applied to if someone explicitly declared that attribute/value.


keithamus (approved):
Yep, LGTM!

adampage (approved):
Ditto! 🚀

PR #2269, opened 2024-07-08
Editorial: adds link that was previously empty

MelSumner:
Editorial. When reviewing the code, I found an empty link.

### DIFF
--- a/accname/index.html
+++ b/accname/index.html
@@ -295,7 +295,7 @@
         </dd>
       </dl>
       <p>
-        The <cite><a class="specref" href="#">Accessible Rich Internet Applications (WAI-ARIA) 1.2</a></cite> [[!WAI-ARIA]] specification provides lists of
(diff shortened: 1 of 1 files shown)


accdc (approved):

accname/index.html
```diff
@@ -295,7 +295,7 @@ <h2>Name and Description</h2>
         </dd>
       </dl>
       <p>
-        The <cite><a class="specref" href="#">Accessible Rich Internet Applications (WAI-ARIA) 1.2</a></cite> [[!WAI-ARIA]] specification provides lists of
+        The <cite><a class="specref" href="https://www.w3.org/TR/wai-aria-1.2/">Accessible Rich Internet Applications (WAI-ARIA) 1.2</a></cite> [[!WAI-ARIA]] specification provides lists of
```
pkra, line 298: This leads to two links next to each other - the first one (just fixed) is pointing to ARIA 1.2, the second one (generated by respec) link points to ARIA 1.1 (more specificcally, to the spec's bibliography which points to ARIA 1.1).

I'm not sure which one is desired here.

scottaohara: The build process seems to already auto populate the hyperlinks href as well, since the link works on the current published spec without this change. 

Agree with Peter, probably just need one link here

PR #2272, opened 2024-07-11
update UIA mapping for maxlength

scottaohara:
This PR supersedes https://github.com/w3c/html-aam/pull/521

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -12050,7 +12050,7 @@
             <tr>
               <th><a href="https://msdn.microsoft.com/en-us/library/ms726297%28v=VS.85%29.aspx">UIA</a></th>
               <td>
-                <div class="general">Not mapped</div>
+                <div class="general">`RangeValue.Maximum`</div>
               </td>
             </tr>
             <tr>


scottaohara:
status update: no work to implement this has started yet, so waiting on that to merge.

spectranaut (approved):
approval-stamp, as this was approved by Ben in the previous PR

spectranaut:
@benbeaudry is there an issue for this week in chrome? see the superseded PR for more information.

benbeaudry:
@spectranaut Not that I know of, and no planned work around it either. Feel free to file one! What's the timeline here, and what are the dependencies? Trying to see if I can justify prioritizing this over other work.

spectranaut:
Someone needs to open bugs on browsers -- I think both Chrome and Firefox?

HaTheo:
I've filed https://issues.chromium.org/issues/520003834 & https://bugzilla.mozilla.org/show_bug.cgi?id=2045180.

@benbeaudry / @scottaohara do we have confirmation that elements with both a RangeValuePattern and a ValuePatern will properly read, it seems like the RangeValuePattern is not typically on a text input control (usually applied to a Slider or Spinner Control) so I am just looking to confirm there are examples of this working as Text inputs in UIA does not typically report maxvalue.

jcsteh:
I'm rather confused by this mapping. According to the [UIA documentation for the Edit control type](https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-supporteditcontroltype#required-control-patterns):

> "All edit controls that take a numeric range must expose the RangeValue control pattern."
> Maximum: "This property must be the largest value to which the contents of the edit control can be set."

The maximum length of the control is *very* different to the maximum numeric value. How is an AT supposed to differentiate these cases?

This might be fine if there were no such thing as a numeric text input... but there is; e.g. `<input type="number">`. Although UIA maps that to spinner, it still exposes the Text pattern because the text is editable. I guess you could say an AT differentiates this purely based on the control type, but that seems very tenuous to me.

Furthermore, the [UIA documentation for the RangeValue control pattern](https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-implementingrangevalue#required-members-for-irangevalueprovider) says that Value, LargeChange, SmallChange, Maximum, Minimum, etc. are *required* members, yet there's no way these could ever be provided for this case. Do we just ignore this guidance?

HaTheo:
@jcsteh Agreed.  When I filed this bug I wrote the comment and question to clarify the approach because, while it seems like the intended direction, I have concerns since I can't find any existing Windows UI examples of this pattern. We used unconventional UIA mappings for Office UI before, which later became standard. If UIA/Microsoft are committed to taking this route here, it should be fine. We just need to ensure the UIA documentation is updated to reflect this in the long term and that ATs can handle it.

HaTheo:
@spectranaut, I am adding an agenda item so we can draw attention to this and ask @scottaohara during the weekly sync on Thursday.

HaTheo:
@smhigley to confirm with @douggeoffray , and look into possibility of changing UIA documentation to better indicate support.

jcsteh:
Changing the documentation doesn't address the concern of ambiguity between an editable, numeric input and a non-numeric editable input with a maximum character length, not to mention the possibility of both: a numeric input which also has a maximum character length.

Ultimately, as far as implementation in Gecko goes, sure, I can do this. But IMO, it just doesn't make any sense for a client. Clients will end up having to special case this, which seems like precisely the sort of thing UIA's design is supposed to eliminate.

smhigley:
I think @douggeoffray may also weigh in here, but I don't believe it should be ambiguous -- a number input will always use the RangeValue for the numerical constraint. The `maxlength` attribute doesn't apply to number inputs, only `text`, `tel`, `email`, and `password`: https://html.spec.whatwg.org/multipage/input.html#concept-input-apply (which makes sense to me, becuase the maximum value of a number input already constrains the character count).

I think there are some other instances of patterns being announced differently based on ControlType -- e.g. the TogglePattern in checkbox, radio, menuitemcheckbox/radio, switch, options, and toggle buttons. Those are maybe more spiritually similar, but IIRC there are some variations in announcement based on control type. I'm definitely open to other opinions on whether that's a bigger deal here though.

jcsteh:
The Toggle pattern is announced differently, but it effectively communicates the same thing: this thing is enabled or disabled. In contrast, IMO, maximum value and maximum length are fundamentally different things. This is not just a cosmetic difference; the UX has to be entirely different. For example, a screen reader might want to alert the user when the number of characters is approaching the maximum, but they won't be doing that for a numeric input. This kind of conflation is exactly the sort of thing that undermines the UIA promise of clean design and [IMO makes UIA insufficient for the web](https://www.jantrid.net/2025/03/19/why-uia-insufficient-web/).

douggeoffray:
Today, UIA clients already special-case patterns. Based on the control type, they typically look for only the patterns relevant to that control. I haven’t seen a screen reader generically expose every pattern for every control type.
This proposal would add another case for edit controls: support for the RangeValue pattern.

That would address scenarios where an edit control uses a max count, as well as scenarios where it uses both min and max counts.

For numeric input control, when you use input type=”number” this already gets converted to a UIA spin button and the RangeValue pattern is used.

Is there a real case where you would have a numeric input that isn’t using type=”number” where you have a range and a max count? In the case of a spin button you wouldn’t need a max count as you could just limit the max number.

The only other option would be to add two new custom UIA properties where browsers could expose the min and max counts. That would leave the RangeValue pattern to do its thing and still expose the min/max counts. The only advantage of this is it would allow your case of a numeric input not using type=”number” that has a number range and a max char count…is this really needed? The cons are it would be another set of properties to create, document, and consume.

Regarding UIA documentation, it is very much out of date. Many things have been added/changed since that documentation was last touched. Updating it to discuss using the RangeValue pattern for edit controls is a perfect example of something that needs to be done.

smhigley:
@douggeoffray I think the benefit would only be theoretical / about spec meaning in this case -- on the HTML side combining `maxlength` with a number input / `spinbutton` is already disallowed 👍

jcsteh:
> Today, UIA clients already special-case patterns. Based on the control type, they typically look for only the patterns relevant to that control.

There's a difference between the question of whether a pattern is used/queried vs having entirely different behaviour for a pattern depending on the control type. The toggle pattern causes different announcements depending on the control type, but that's purely cosmetic. As noted above, this conflation of the RangeValue pattern requires entirely different behaviour. It may as well be a different pattern, and that is ultimately how clients are going to have to treat it.

Regardless, I don't think I'm going to convince you, and ultimately, I don't have any authority where UIA is concerned, so I'm going to move on. I still think it undermines UIA's promise of clean design and I suspect it might result in ugliness/problems later, but it's far from the first time an accessibility API suffered from design issues, so I'll grant that's not a particularly strong argument. From a browser perspective, it's easy enough to implement this, which is ultimately the only say I have here.

Issue #212, opened 2016-01-21
Where things are the same across different areas, and where they should be treated differently


MichielBijl:
Taken from the README.

PR #2274, opened 2024-07-12
[accname] Update text node definition

rahimabdi:
Closes [accname #212](https://github.com/w3c/accname/issues/212)

Update text node definition using the [DOM text interface](https://dom.spec.whatwg.org/#interface-text) definition.

### DIFF
--- a/accname/index.html
+++ b/accname/index.html
@@ -285,7 +285,7 @@
         </dd>
         <dt>contents</dt>
         <dd>
-          name is generated from the Text [=nodes=] associated with the [=element=]. Although this may be allowed in addition to "author" in some <a class="termref">roles</a>, "content" is used only
+          name is generated from the [=Text=] node associated with the [=element=]. Although this may be allowed in addition to "author" in some <a class="termref">roles</a>, "content" is used only
           if higher priority "author" features are not provided. Priority is defined by the <a href="#mapping_additional_nd_te">text equivalent computation</a> algorithm.
         </dd>
         <dt>prohibited</dt>
@@ -706,7 +706,7 @@
                   </div>
                 </li>
                 <li id="comp_text_node">
-                  <span id="step2G"><!-- Don't link to this legacy numbered ID. --></span><em>Text Node:</em> Otherwise, if the <code>current node</code> is a Text [=Node=], return its textual
+                  <span id="step2G"><!-- Don't link to this legacy numbered ID. --></span><em>Text Node:</em> Otherwise, if the <code>current node</code> is a [=Text=] node, return its textual
                   contents.
                 </li>
                 <li id="comp_recursive_name_from_content">


accname/index.html
```diff
@@ -706,8 +706,8 @@ <h4>Computation steps</h4>
                   </div>
                 </li>
                 <li id="comp_text_node">
-                  <span id="step2G"><!-- Don't link to this legacy numbered ID. --></span><em>Text Node:</em> Otherwise, if the <code>current node</code> is a Text [=Node=], return its textual
-                  contents.
+                  <span id="step2G"><!-- Don't link to this legacy numbered ID. --></span><em>Text Node:</em> Otherwise, if the <code>current node</code> is a [=exclusive Text node|text node=], return
```
jnurthen, line 709:
```suggestion
                  <span id="step2G"><!-- Don't link to this legacy numbered ID. --></span><em>Text Node:</em> Otherwise, if the <code>current node</code> is an [=exclusive Text node|text node=], return
```

smockle: This would result in “Otherwise, if the current node is an [sic] text node, return”, right?

The [preview deployment](https://deploy-preview-2274--wai-aria.netlify.app/accname/#comp_text_node) has this:
> G. Text Node: Otherwise, if the current node is a [text node](https://dom.spec.whatwg.org/#exclusive-text-node), return its textual contents.

jnurthen: ah - sorry got it the wrong way round. My mistake.

accname/index.html
```diff
@@ -285,8 +285,8 @@ <h2>Name and Description</h2>
         </dd>
         <dt>contents</dt>
         <dd>
-          name is generated from the Text [=nodes=] associated with the [=element=]. Although this may be allowed in addition to "author" in some <a class="termref">roles</a>, "content" is used only
-          if higher priority "author" features are not provided. Priority is defined by the <a href="#mapping_additional_nd_te">text equivalent computation</a> algorithm.
+          name is generated from the [=exclusive Text node|text node=] associated with the [=element=]. Although this may be allowed in addition to "author" in some <a class="termref">roles</a>,
```
smockle, line 288: This changes from “the nodes” (plural) to “the text node” (singular)—is that intended? An accessible name can be generated from multiple text nodes.

For example, this URL—
```
data:text/html,<h1>Hello, <!--COMMENT-->world!</h1>
```

—plus this JavaScript—

```JS
getAccessibleName(document.querySelector("h1"))
// 'Hello, world!'

document.querySelector("h1").childNodes
// NodeList(3) [text, comment, text]
```

rahimabdi: Agreed "text nodes" should be plural, thank you @smockle. This does introduce inconsistency for the ReSpec linking; aliases are strongly discouraged (let alone differently aliased links) so curious if anyone disagrees with simply making it plural while in other cases we use "text node" as the link. 

Changed as part of https://github.com/w3c/aria/pull/2274/commits/bde4f90bfc312be58e09e04842d55136713a924a.

rahimabdi:
I think this PR is good to merge.

keithamus (approved):
LGTM :+1:. Nice work

smockle (approved):

scottaohara (approved):

PR #2275, opened 2024-07-12
[html-aam] - UIA input mappings cleanup

### LINKED ISSUE #173: HTML-AAM: should <legend> map to aria role=label?
feels label-y

scottaohara:
closes [#173](https://github.com/w3c/html-aam/issues/173)

the localized control types for the UIA mappings for various input elements is not how these controls are actually reflected... nor, per the issue filed, how they likely should be.

being 'edit' fields with accNames that will more than likely convey their purpose is likely enough, and the localized control types don't need to reflect the type attribute value.

This is not a normative change, since this updates the spec to reflect the current implementation.

Additional inputs may need to be reviewed / changed, but those can be updated in a separate PR.

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -3203,7 +3203,8 @@
               <td>
                 <a data-cite="html">`input`</a>
                 <span class="el-context"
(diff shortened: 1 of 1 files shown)


benbeaudry (approved):
I double checked that Chromium wasn't setting any of those localized control type mapping. LGTM.

PR #2278, opened 2024-07-15
[html-aam] fix ax iframe mapping

scottaohara:
closes [html aam 130](https://github.com/w3c/html-aam/issues/130)

the iframe mapping for the ax api listed `AXWebArea` as the platform role, but this is incorrect / is the role of the child web area, not the iframe itself.

speaking with @minorninth (who opened the original issue), 'not mapped' seems most appropriate at this time.


minorninth:
Thanks!

cookiecrook (approved):

rahimabdi (approved):

PR #2284, opened 2024-07-17
Remove Core-AAM UA requirement to disallow orphaned role context.

cookiecrook:
Remove Core-AAM UA requirement to disallow orphaned role context.

Closes https://github.com/w3c/aria/issues/2166

### DIFF
--- a/core-aam/index.html
+++ b/core-aam/index.html
@@ -548,18 +548,13 @@
 &lt;a href="#" role="button"&gt; &lt;!-- computedrole returns "button" --&gt;</pre
             >
           </aside>
-          <p>
-            When an element has a role but is not contained in the required context (for example, an orphaned `listitem` without the required accessible parent of role `list`), User Agents MUST ignore
-            the role token, and return the `computedrole` as if the ignored role token had not been included.
+          <p class="note">
+            When an element has a role but is not contained in the required context (for example, an orphaned `listitem` without the required accessible parent of role `list`), this is an authoring error, but the user agent behavior is not specified as a single rule. For most roles, user agents can either recover the error by ignoring the role, or respect the author's intended role in scenarios deemed by the implementation to be harmless. This unspecified permissiveness in how engines treat author role usage errors is occasionally overridden for a specific role in other specs such as [[HTML-AAM]].
           </p>
           <aside class="example">
             <!-- ReSpec needs these examples to be unindented. -->
             <pre>
-&lt;div role="listitem"&gt; &lt;!-- Author error: orphaned listitem. computedrole returns "generic" --&gt;
-
-&lt;div role="list"&gt; &lt;!-- computedrole returns "list" --&gt;
-  &lt;div role="listitem"&gt; &lt;!-- computedrole returns "listitem" in the required context. --&gt;</pre
-            >
+&lt;div role="listitem"&gt; &lt;!-- Author error: orphaned listitem. computedrole is unspecified. --&gt;</pre>
           </aside>
           <p>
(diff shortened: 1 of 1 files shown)


core-aam/index.html
```diff
@@ -548,18 +548,15 @@ <h2>Computed Role</h2>
...
             >
           </aside>
-          <p>
-            When an element has a role but is not contained in the required context (for example, an orphaned `listitem` without the required accessible parent of role `list`), User Agents MUST ignore
-            the role token, and return the `computedrole` as if the ignored role token had not been included.
+          <p class="note">
+            When an element has a role but is not contained in the required context (for example, an orphaned `listitem` without the required accessible parent of role `list`), this is an authoring
+            error, but the user agent behavior is not specified as a single rule. For most roles, user agents can either recover the error by ignoring the role, or respect the author's intended role
+            in scenarios deemed by the implementation to be harmless. This unspecified permissiveness in how engines treat author role usage errors is occasionally overridden for a specific role in
+            other specs such as [[HTML-AAM]].
```
spectranaut, line 555: Hmmm I wish there was a way to word that last line that is a little less vague and vaguely ominious. Maybe we should say something a bit more specific, like:
"This unspecified permissiveness in how engines treat author role usage errors is occasionally overridden for specific HTML tags without ARIA role overrides, as defined in [[HTML-AAM]]"...? Because I think only HTML does this anyway?

spectranaut (changes requested):
I'd like to understand the exceptions a bit better!

core-aam/index.html
```diff
@@ -548,18 +548,15 @@ <h2>Computed Role</h2>
...
           </p>
           <aside class="example">
             <!-- ReSpec needs these examples to be unindented. -->
-            <pre>
-&lt;div role="listitem"&gt; &lt;!-- Author error: orphaned listitem. computedrole returns "generic" --&gt;
-
-&lt;div role="list"&gt; &lt;!-- computedrole returns "list" --&gt;
-  &lt;div role="listitem"&gt; &lt;!-- computedrole returns "listitem" in the required context. --&gt;</pre
-            >
+            <pre>&lt;div role="listitem"&gt; &lt;!-- Author error: orphaned listitem. computedrole is unspecified. --&gt;</pre>
```
cookiecrook, line 559: Actually, I think I should keep the second example.
```suggestion
            <pre>
&lt;div role="listitem"&gt; &lt;!-- Author error: orphaned listitem. computedrole is unspecified. --&gt;

&lt;div role="list"&gt; &lt;!-- computedrole returns "list" --&gt;
  &lt;div role="listitem"&gt; &lt;!-- computedrole returns "listitem" in the appropriate context. --&gt;</pre>
```

core-aam/index.html
cookiecrook, line 555:
```suggestion
            in scenarios deemed by the implementation to be harmless. Please note that this permissiveness in how engines treat author role errors might be overridden in a language-specific mapping document such as [[HTML-AAM]].
```

cookiecrook: @spectranaut The **Prettier** bot broke the thread (because of the line breaks it needlessly added) but this is in response to [your feedback suggestion](https://github.com/w3c/aria/pull/2284/files#r1683523037).

cookiecrook: [Thread continues here](https://github.com/w3c/aria/pull/2284#discussion_r1684768065)... (Prettier bot's fault, not mine.)

spectranaut: Ok, this seems better to me!

spectranaut (approved):
Approval with the change to list html-aam specifically

scottaohara (approved):
lgtm when the suggested changes are merged in

spectranaut:
@cookiecrook if you include your proposed changes I will land!

PR #2289, opened 2024-07-19
Editorial: Delete gitattributes, gitignore, pr-preview, readme, and other unwanted files for this monorepo

daniel-montalvo:


spectranaut:
Actually, I think we left most of these files on purpose. The various specs might want to keep their READMEs, maybe we should update them to be more useful. And I assume the editors had the git-ignores for a reason, and these can be file specific. The pr-previews should go though, and probably the .gitattributes -- I'm curious what other editors think though.

pkra:
My initial reaction was "let's keep the readme's" - but I really don't have strong feelings about this.

spectranaut (changes requested):
How about keeping the READMEs, @daniel-montalvo, and deleting everything else?

daniel-montalvo:
Will update accordingly, @spectranaut

daniel-montalvo:
I am closing this without merging. This is not causing any issues now, and could potentially cause problems for some editors according to [Val's comment above](https://github.com/w3c/aria/pull/2289#issuecomment-2239717261)

If there's updates needed for these files we could open new PRs for them as we need them.

PR #2290, opened 2024-07-20
Relax `aria-valuenow` requirement on `slider` role

### LINKED ISSUE #2260: Should aria-valuenow be required if aria-valuetext is used?
Happened across this issue in axe-core today - https://github.com/dequelabs/axe-core/issues/4515, and I think it raises a good question.  Is there a good reason (for a user) for an author to need to maintain both an aria-valuenow and an aria-valuetext?

take for instance the youtube video scrubber:
```
<div class="ytp-progress-bar" tabindex="0" role="slider" 
  aria-label="Seek slider" aria-valuemin="0" aria-valuemax="30" aria-valuenow="13" 
  aria-valuetext="0 Minutes 13 Seconds of 0 Minutes 30 Seconds" style="">
```

if one were to remove the aria-valuenow, from my quick testing with Narrator, VoiceOver (macos/ios), NVDA and JAWS, there was no negative user impact to understanding the slider since the aria-valuetext was still being conveyed.

the visual styling of the indicator for sliders, progressbars, etc., are often not even based on the value of aria-valuenow 

so largely this seems like it might just be 'extra work' for developers if they're using aria-valuetext, but still being required to declare an aria-valuenow which doesn't get exposed to users, and is not often the source of the visual styling for the slider/progressbar/etc. indicator.

But, maybe i'm missing something and this attribute still can be accessed and has value to users I'm not thinking about???

I'm less concerned about if authors might have use for it - cause if they do, great.  Keep using it.  This issue is just focused on determining why it needs to be an author requirement to specify the attribute if an author doesn't need it / it isn't being exposed to users anyway?

adampage:
Closes #2260

Relaxes a requirement for the `slider` role by permitting authors to omit `aria-valuenow` in favor of `aria-valuetext` alone.

I purposefully left the adjacent “_browsers MAY implement the repair techniques..._” statement unchanged, as well as the `slider`-specific repair technique that it cross-references. Both still seemed compatible with this change.

My introduction of “**_or_**” in the `slider`’s Required States and Properties cell [on line 7738](https://github.com/w3c/aria/compare/main...adampage:aria:relax-aria-valuenow?expand=1#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R7738) looks unprecedented in the spec in terms of its format. I’d be grateful for the editors to weigh in on that?

* [ ] "author MUST" tests:
* [X] Browser implementations (link to issue or commit): n/a
   * WebKit:
   * Gecko:
   * Blink:
* [X] Does this need AT implementations? n/a
* [ ] ACT review?
* [X] MDN Issue/PR: n/a

***
<a href="https://pr-preview.s3.amazonaws.com/adampage/aria/pull/2290.html" title="Last updated on Jul 17, 2025, 9:44 PM UTC (53b0cab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2290/ff9dd6b...adampage:53b0cab.html" title="Last updated on Jul 17, 2025, 9:44 PM UTC (53b0cab)">Diff</a>


adampage:
Hi @daniel-montalvo, could you advise me on [this prettier error](https://github.com/w3c/aria/actions/runs/10017230520/job/27691148879?pr=2290#step:2:735)?

daniel-montalvo:
Thanks @adampage  I approve this since it is coming from a fork.

@pkra  This is complaining because it can't find the branch relax-aria-valuenow, which I guess makes sense because this is coming from a fork. Is this expected behavior?

In any case, weren't we supposed to disable prettier for ARIA for now?

index.html
```diff
@@ -7658,8 +7658,38 @@ <h5>Presentational Role Inheritance</h5>
...
 				<p>Elements with the role <code>slider</code> have an implicit <pref>aria-orientation</pref> value of <code>horizontal</code>.</p>
+				<div class="note">
+					<p>It is usually advisable for a slider to express its current value as a parsable <a href="#valuetype_number">number</a> using <pref>aria-valuenow</pref>. However, there are use cases where <pref>aria-valuetext</pref> alone provides a superior user experience.</p>
+					<p>For example, pain can be measured along a range of values in which each value is most helpfully expressed by descriptive text rather than superficially expressed by an unintuitive number.</p>
+				</div>
+				<pre class="example highlight">
+&lt;label&gt;
+  Pain rating:
+  &lt;input id="pain" type="range" min="0" max="10" value="0" aria-valuetext="No pain" /&gt;
```
scottaohara, line 7670: should probably use a fully custom ARIA example.  with the input element, the value is going to still be specified by default.

spectranaut:
> Thanks @adampage I approve this since it is coming from a fork.
> 
> @pkra This is complaining because it can't find the branch relax-aria-valuenow, which I guess makes sense because this is coming from a fork. Is this expected behavior?
> 
> In any case, weren't we supposed to disable prettier for ARIA for now?

I opened an issue for the prettier issue: https://github.com/w3c/aria/issues/2294

index.html
adampage: Gah, of course. On it.

scottaohara (changes requested):
added some suggested text to clarify the aria-valuenow repair should only occur if both aria-valuetext and aria-valuenow are missing.

I'd also like to see user agent or AT SHOULD NOT expose an aria-valuenow fallback value if the aria-valuetext is specified but aria-valuenow is not.  For instance, in my recent testing i recall screen readers (though don't remember exactly which ones) would expose a fallback value when aria-valuenow wasn't specified (usually a 50%)

adampage:
> I'd also like to see user agent or AT SHOULD NOT expose an aria-valuenow fallback value if the aria-valuetext is specified but aria-valuenow is not.

Here’s [a stab at that](https://pr-preview.s3.amazonaws.com/adampage/aria/pull/2290.html#:~:text=However%2C%20both%20user%20agents%20and%20assistive%20technologies%20SHOULD%20NOT%20expose%20a%20fallback%20value%20for%20aria%2Dvaluenow%20if%20the%20author%20has%20only%20specified%20the%20aria%2Dvaluetext%20attribute.), @scottaohara.

adampage:
Heya @jnurthen, quick editors question: I just noticed that the spec seems to prefer “_user agents_” over “_browsers_” for normative statements? I found only 2 matches for `\bBrowser(s)?\s+(MAY|SHOULD|MUST)\b` versus 77 for `user agent(s)`.

One of those 2 “_browser_” matches was in this PR’s `slider` section, so I swapped it out in [this latest commit](https://github.com/w3c/aria/pull/2290/commits/957880b825d219f51bb1996cc11480cf5f691d3e#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R7661). Would it be scope creep for me to just take care of [the last remaining one (in the `scrollbar` section)](https://www.w3.org/TR/wai-aria-1.3/#:~:text=current%20thumb%20position.-,If%20aria%2Dvaluenow%20is%20missing%20or%20has%20an%20unexpected%20value%2C%20browsers%20MAY%20implement%20the%20repair%20techniques%20specified%20in%20the%20section%20describing%20handling%20author%20errors%20in%20states%20and%20properties%2C%20which%20are%20equivalent%20to%20the%20repair%20techniques%20for%20%3Cinput%20type%3D%22range%22%3E%20in%20HTML.,-Elements%20with%20the).

index.html
```diff
@@ -7658,8 +7658,34 @@ <h5>Presentational Role Inheritance</h5>
...
+    &lt;li&gt;Hardly noticeable pain&lt;/li&gt;
+    &lt;li&gt;Somewhat distracting pain&lt;/li&gt;
+    &lt;li&gt;Pain interrupts some activities&lt;/li&gt;
+    &lt;li&gt;Pain prevents daily activities&lt;/li&gt;
+    &lt;li&gt;Unbearable pain, cannot do anything&lt;/li&gt;
+    &lt;li&gt;As bad as pain can be, nothing else matters&lt;/li&gt;
+  &lt;/ol&gt;
+&lt;/div&gt;
+&lt;script&gt;
+  document.querySelectorAll('#ratings li').forEach(function(rating, index) {
```
smhigley, line 7682: Even with the comment about the example being incomplete, I think I'd lean towards leaving off the script entirely. Maybe adding a comment above the `<ol>` clarifying that those are the visible marks & labels along the slider.

The reason being that even with all the warnings in the world, people still go to the ARIA spec and say "oh this is how I should do things", then go tell other people "this is what we should be doing, this is the spec" 😅. And this script example doesn't include keyboard, drag, etc., and with `aria-hidden` + click events, mobile SRs and voice control won't be able to interact, etc. etc. (that's not a critique of your example code, I know showing all that is impossible in a small example, but just a reason to not demo any of the script since anything we do will be super inaccessible in an aria slider scenario)

adampage: Oof, yes. 🤦🏻  I agonized a bit over how far to take this.

In support of this spec change, I was hoping that the code example would at least illustrate the essential aspect of _only_ `aria-valuetext` being updated and `aria-valuenow` being absent altogether. I’ll take another run at this.

index.html
```diff
@@ -7658,8 +7658,34 @@ <h5>Presentational Role Inheritance</h5>
...
+    rating.addEventListener('click', function() {
+      document.getElementById('pain').setAttribute('aria-valuetext', this.textContent);
+    });
+  });
```
adampage, line 7686:
```suggestion
  // Not shown: Custom Javascript to implement a fully conformant ARIA slider pattern.
  // Also, the `aria-valuetext` attribute of `div#pain` will be dynamically updated with
  // the user’s chosen pain value corresponding to the visible mark and label presented
  // by `ol#ratings` (e.g., "Pain interrupts some activities").
```

How about this, @smhigley?

smhigley: I like it!

smhigley (approved):
With that additional change/comment, LGTM!

index.html
adampage: Thank you!

scottaohara (approved):

cookiecrook:
@adampage will you update the checklist in the "Test, Documentation and Implementation tracking" section?

index.html
```diff
@@ -7705,7 +7730,7 @@ <h5>Presentational Role Inheritance</h5>
 						<th class="role-required-properties-head">Required States and Properties:</th>
 						<td class="role-required-properties">
 							<ul>
-								<li><pref>aria-valuenow</pref></li>
+								<li><pref>aria-valuenow</pref> or <pref>aria-valuetext</pref></li>
```
jnurthen, line 7733: I don't think we normally do ors here. The requirement is already specified in prose as a MUST so I think we should just remove the required property here

jnurthen (changes requested):
other than removing the required property LGTM

jnurthen:
> Would it be scope creep for me to just take care of [the last remaining one (in the `scrollbar` section)](https://www.w3.org/TR/wai-aria-1.3/#:~:text=current%20thumb%20position.-,If%20aria%2Dvaluenow%20is%20missing%20or%20has%20an%20unexpected%20value%2C%20browsers%20MAY%20implement%20the%20repair%20techniques%20specified%20in%20the%20section%20describing%20handling%20author%20errors%20in%20states%20and%20properties%2C%20which%20are%20equivalent%20to%20the%20repair%20techniques%20for%20%3Cinput%20type%3D%22range%22%3E%20in%20HTML.,-Elements%20with%20the).

I know its a pain but i'd prefer a separate PR

pkra:
@adampage sorry, I can't help rebasing this since it's on a fork. I hope it won't be too painful :pensive:

index.html
adampage: Thanks @jnurthen — done, in [53b0cab](https://github.com/w3c/aria/pull/2290/commits/53b0cab8f029508bab31cf54ced86d1d5024a94c).

adampage:
@pkra No trouble at all, thanks for the heads-up. I’ve rebased and resolved the merge conflicts.

adampage:
> I know its a pain but i'd prefer a separate PR

@jnurthen, no pain at all — I just opened #2472 for this. 👍🏻

jnurthen (approved):

pkra:
@jnurthen @spectranaut should we seek ACT review for this?

spectranaut:
Since we haven't heard anything from ACT, I think we need to open an issue on validators, and merge this.

HaTheo:
@adampage, @scottaohara, and @patrickhlauke.

I’ve found that in all cases(across tested screen readers, browsers, platforms, etc.) screen readers announce **_only_** the human-readable value text when it is provided over the valuenow. The only issue I found was with Orca (on all Linux browsers). It seems that Orca is looking for a change only when valuenow is present. I have confirmed that the value is being updated, and I do see property changed events using so I think it is likely an Orca bug.

I quickly made this example https://codepen.io/HaTheo/pen/RNKKwxw (Note: I did use AI and then modified/validated it). 

| | Jaws    | NVDA         | Narrator     | VoiceOver                 | Orca         | 
|---------|--------------|--------------|---------------------------|--------------|--------------------|
| Firefox | ✓ Text/Text  | ✓  Text/Text | X No change / X No change | ✓ Text/Text  | X No change / Text |
| Chrome  | ✓  Text/Text | ✓ Text/Text  | ✓ Text/Text               | ✓ Text/Text  | X No change / Text |
| Edge    | ✓ Text/Text  | ✓ Text/Text  | ✓  Text/Text              | ✓ Text/Text  | X No change / Text |
| Safari  |       -       |       -       |            -               | ✓  Text/Text |X No change / Text |

Does the fact that this isn't working on Orca change our opinion about checking this in? I am happy to file a bug on Orca, and it actually might be easier to get leverage with this filed.

patrickhlauke:
> Does the fact that this isn't working on Orca change our opinion about checking this in? I am happy to file a bug on Orca, and it actually might be easier to get leverage with this filed.

Late reply, but for my part, I'd go with it and file an Orca bug

HaTheo:
bug filed here:
https://gitlab.gnome.org/GNOME/orca/-/work_items/721

HaTheo:
Orca v51 includes the fix, with no plans to backport it. Because UX remains unaffected(screen readers universally read out the `valuetext` when provided, even when `valuenow` is added), I propose updating the change note to encourage authors to provide both aria-valuenow and aria-valuetext for legacy screen reader compatibility. 

https://gitlab.gnome.org/GNOME/orca/-/work_items/721#note_2792724
____
### UPDATE(correction):
____

@adampage, @scottaohara, @patrickhlauke: per the conversation and @joanmarie's callout, we need to rethink the default and repair techniques if we take this approach, as ATs are getting 50% or (min+max)/2 from user agents.

One option is to add a comment like the one I suggested in the discussion on the issue, introducing a normative requirement that "User agents MUST NOT apply repair techniques to calculate a `aria-valuenow` if `aria-valuetext` is provided." But I don't love that..

There are a few other alternatives/options:
* We could deprecate the `valuenow` repair technique entirely.
* We could change this downstream in Core-AAM instead of the ARIA spec.
* We could leave it as is... and keep the author MUSTs for valuenow, or at least say authors SHOULD provide it.

joanmarie:
TL;DR: IF you're going to do this, consider at least exposing `STATE_INDETERMINATE` on the object so Orca can identify repair-techniqued values and not present bogus relative slider positions in its "Where Am I?" command.

-----

Since I've been (at) summoned.... :grin: 

I personally don't object to this change, but as a non-author I don't understand it. In my mind, a slider represents a bounded numeric range with a known minimum, a known maximum, AND a known current value. As a sighted user, I can see that "Pain interrupts some activities" is the mid point and that "Unbearable pain, cannot do anything" -- which certainly sounds to me like the maximum -- actually isn't the maximum. That little red dot can be nudged one more time to the right.

That said, given the interest in relaxing the authoring requirement, here's the technical issue: Orca has a "Where Am I?" command which, in the case of ranges, includes the relative position in the range as a percentage. In other words, this command conveys what I can visually see about the position of that little red dot.

* With `aria-valuenow`: "As bad as pain can be, nothing else matters. 100 percent."
* With just `aria-valuetext`: "As bad as pain can be, nothing else matters. 50 percent."

The latter being due to the reasons @HaTheo cited, namely assistive technologies are given values from the user agents. They don't know where those values come from or if they are valid. They have to trust that they are valid if there's no evidence to the contrary. With this in mind:

I'm afraid -- at least in the case of `AtspiValue`, and possibly the equivalent API for the other platforms -- you cannot simply tell user agents that they MUST NOT implement repair techniques. [`atspi_value_get_current_value()`](https://docs.gtk.org/atspi2/method.Value.get_current_value.html) returns a `gdouble`. What should the value of that `gdouble` be? Same question applies to the min and max getters. Maybe user agents should return -1 instead of [what's in the repair table](https://w3c.github.io/aria/#authorErrorDefaultValuesTable). But they have to return *something*.  

If you want to remove `aria-valuenow` and also make Orca stop saying 50% in response to its "Where Am I?" command (e.g. the Core-AAM fix): if `atspi_value_get_current_value()` returns -1 AND the accessible object's state set includes `STATE_INDETERMINATE`, Orca will not present the percent. I'd be willing to relax Orca's requirement and just check for the presence of `STATE_INDETERMINATE` as the indicator that the values are untrustworthy. But you will still need the user agent implementations for that change.

And as long as I'm here... Looking at the merge request, I question this statement:

```html
<p>It is usually advisable for a slider to express its current value as a parsable <a href="#valuetype_number">number</a> using <pref>aria-valuenow</pref>. However, there are use cases where <pref>aria-valuetext</pref> alone provides a superior user experience.</p>
```

Based on @HaTheo's report, it appears that all screen readers -- including Orca -- already present the helpful descriptive text and not the unintuitive number, even though that number is present/exposed to ATs. So how does the removal of `aria-valuenow` provide this superior user experience?

Also if this proposed change lands, it might make sense to update the text related to [`aria-valuetext`](https://w3c.github.io/aria/#aria-valuetext) which currently states:

> If the [aria-valuetext](https://w3c.github.io/aria/#aria-valuetext) attribute is set, authors SHOULD also set the [aria-valuenow](https://w3c.github.io/aria/#aria-valuenow) attribute, **unless that value is unknown** (for example, on an indeterminate [progressbar](https://w3c.github.io/aria/#progressbar)).

(emphasis added)

Similarly, it looks like the [`aria-valuenow`](https://w3c.github.io/aria/#aria-valuenow) guidance needs some slight tweaking regardless of whether or not this change lands. Because it currently states:

> For elements with role [slider](https://w3c.github.io/aria/#slider) and [spinbutton](https://w3c.github.io/aria/#spinbutton), assistive technologies SHOULD render the actual value to users.

Based on @HaTheo's report, no AT is rendering the actual value of sliders to users in response to value changes if `aria-valuetext` is present. And Orca, through it's "Where Am I?" command, currently represents it as a percent (which I personally think is more meaningful than the actual value).

Anyway, I've said (more than) enough. :innocent: Back to my downstream cave.

adampage:
Hi @HaTheo and @joanmarie, thanks very much for your review and thoughtful feedback.

On the merits of allowing authors to omit `aria-valuenow`, I find #2260 persuasive, as well as the video scrubber example in the cited dequelabs/axe-core#4515 issue.

In [the pain slider example](https://pr-preview.s3.amazonaws.com/adampage/aria/pull/2290.html#example-19), I omitted not just `aria-valuenow`, but also `aria-valuemin` and `aria-valuemax`. My goal was to show how a slider can be a useful control for presenting a range of options that conceptually increase or decrease, but whose values or bounds might be unintuitive for an author to express as decimal numbers.

> One option is to add a comment like the one I suggested in the discussion on the issue, introducing a normative requirement that "User agents MUST NOT apply repair techniques to calculate a aria-valuenow if aria-valuetext is provided." But I don't love that..

FWIW, I intended to discourage user agents from performing the `aria-valuenow` repair in [this portion of the PR](https://github.com/w3c/aria/pull/2290/changes#:~:text=However%2C%20both%20%3Ca%3Euser%20agents%3C/a%3E%20and%20%3Ca%3Eassistive%20technologies%3C/a%3E%20SHOULD%20NOT%20expose%20a%20fallback%20value%20for%20%3Ccode%3Earia%2Dvaluenow%3C/code%3E%20if%20the%20author%20has%20only%20specified%20the%20%3Ccode%3Earia%2Dvaluetext%3C/code%3E%20attribute.%3C/p%3E). But thanks to @joanmarie’s recommendation:

> TL;DR: IF you're going to do this, consider at least exposing STATE_INDETERMINATE on the object so Orca can identify repair-techniqued values and not present bogus relative slider positions in its "Where Am I?" command.

...I now understand that Core AAM should also address this.

@spectranaut, could I ask for your guidance here? If I’m on the right track, then I think [Core AAM’s `aria-valuenow` property mapping table](https://w3c.github.io/core-aam/#ariaValueNow) should conditionally map to an indeterminate value when only `aria-valuetext` is present? It looks like [`aria-setsize` is doing something similar-ish](https://w3c.github.io/core-aam/#ariaSetsize:~:text=If%20the%20author,also%3A%20Group%20Position)? I’d be happy to make that change myself; just looking for a sanity check.

@scottaohara, I’m also now thinking I should have updated [the `slider` row in the ARIA spec’s fallback value table](https://w3c.github.io/aria/#document-handling_author-errors_states-properties:~:text=aria%2Dvaluemax.-,slider,-aria%2Dvaluenow) to further discourage the fallback when `aria-valuetext` is present. I’ll let this review thread play out a bit longer, but I’ll plan to take a stab at that if you agree.

> Also if this proposed change lands, it might make sense to update the text related to [aria-valuetext](https://w3c.github.io/aria/#aria-valuetext)...

@joanmarie, thanks for catching this. I’ve just pushed up a tweak for that in 595a311. What do you think?

> Similarly, it looks like the [aria-valuenow](https://w3c.github.io/aria/#aria-valuenow) guidance needs some slight tweaking...

Oof, yes, thank you. 🤦🏻  I’ve added a tweak in 6acf0b0. It does still ask the reader to hang tight on fully interpreting “**_actual_** value” (emphasis mine) until they read through the following paragraph and note. Do you think that’s sufficient?

adampage:
@pkra or @daniel-montalvo, is there any chance you could help me make sense of this [“roleInfo PR check” failure](https://github.com/w3c/aria/actions/runs/28145267913/job/83351036236?pr=2290)?

I found [this bit in the repo’s readme](https://github.com/w3c/aria#:~:text=The%20ariaChild.js,common/script%20directory.) but am a little lost. 😅

index.html
```diff
@@ -9309,12 +9309,33 @@ <h5>Presentational Role Inheritance</h5>
...
             <p>Elements with the role <code>slider</code> have an implicit <pref>aria-orientation</pref> value of <code>horizontal</code>.</p>
+            <div class="note">
+              <p>It is usually advisable for a slider to express its current value as a parsable <a href="#valuetype_number">number</a> using <pref>aria-valuenow</pref>. However, there are use cases where <pref>aria-valuetext</pref> alone provides a superior user experience.</p>
```
joanmarie: I would change this prose. As best as I can tell, it is factually inaccurate. My understanding is that screen readers are presenting the same thing* regardless of whether or not `aria-valuenow` is present. If the user experience is unchanged, it is not "superior."

If there are cases in which the absence of `aria-valuenow` actually does make the user experience superior, explaining how it does this and providing a concrete example is worth doing. If there are no such cases, this potentially benefits the author who no longer has to maintain/expose a value.

*With the exception of Orca's "Where Am I?" command in which the presence of `aria-valuenow` makes it possible to indicate the relative position of that little red dot.

joanmarie:
> > Similarly, it looks like the [aria-valuenow](https://w3c.github.io/aria/#aria-valuenow) guidance needs some slight tweaking...
> 
> Oof, yes, thank you. 🤦🏻 I’ve added a tweak in [6acf0b0](https://github.com/w3c/aria/commit/6acf0b031057f656139f0a7a9ba63dfbdbaa164f). It does still ask the reader to hang tight on fully interpreting “**_actual_** value” (emphasis mine) until they read through the following paragraph and note. Do you think that’s sufficient?

Apologies for not being more clear. I was specifically referring to the following spec text:

> For elements with role [slider](https://w3c.github.io/aria/#slider) and [spinbutton](https://w3c.github.io/aria/#spinbutton), assistive technologies SHOULD render the actual value to users.

This is the text in the paragraph before the text you changed.

For elements with role `slider`, assistive technologies SHOULD render the actual value to users **if there is no value for `aria-valuetext`.**

pkra:
@adampage 

> @pkra or @daniel-montalvo, is there any chance you could help me make sense of this [“roleInfo PR check” failure](https://github.com/w3c/aria/actions/runs/28145267913/job/83351036236?pr=2290)?
> 
> I found [this bit in the repo’s readme](https://github.com/w3c/aria#:~:text=The%20ariaChild.js,common/script%20directory.) but am a little lost. 😅

Oh! thanks for pointing to the readme - it needs an update.

Anyway. For this PR, the check should fail.

This PR is changing the characteristics of a role, so the abstract information gathered in roleInfo.js will (upon merge) be updated.

The roleInfo check generates a diff https://github.com/w3c/aria/actions/runs/28145267913/job/83351036236?pr=2290 t

We should document that either the PR author or a reviewer must check that the diff corresponds to the changes.

adampage:
> We should document that either the PR author or a reviewer must check that the diff corresponds to the changes.

Ah, got it — thanks, @pkra. I’ve reviewed and [applied the diff](https://github.com/w3c/aria/pull/2290/commits/ce199e8327e9b4ca655087f52b046d6407ee472f).

index.html
adampage:
```suggestion
              <p>It is usually advisable for a slider to express its current value as a parsable <a href="#valuetype_number">number</a>. However, there are use cases where a descriptive text value provides a superior user experience.</p>
```

Thanks, @joanmarie, I see. How about ☝🏻 this tweak, focusing on the desired user outcome rather than the implementation specifics?

adampage:
> For elements with role slider, assistive technologies SHOULD render the actual value to users if there is no value for aria-valuetext.

@joanmarie — terrific, thanks for the content suggestion. I’ve just pushed that in 38ad64f.

index.html
joanmarie: For the pain slider, the descriptive text provided through `aria-valuetext` does indeed provide a superior user experience. So I'm good with this change. Thanks!

HaTheo:
Cool, I like this change. I think it addresses the issue, and I will file bugs in the browser, as no browser is following the "if a value is not provided for either valuetext or valuenow, do not repair" change. I will wait to talk about it in an hour, though.

pkra:
@adampage 
> Ah, got it — thanks, @pkra. I’ve reviewed and [applied the diff](https://github.com/w3c/aria/pull/2290/commits/ce199e8327e9b4ca655087f52b046d6407ee472f).

That's not necessary. It will also be applied automatically after merging.

joanmarie:
> Cool, I like this change. I think it addresses the issue, and I will file bugs in the browser, as no browser is following the "if a value is not provided for either valuetext or valuenow, do not repair" change. I will wait to talk about it in an hour, though.

Just wanted to triple check that you saw what I said earlier:

> I'm afraid -- at least in the case of `AtspiValue`, and possibly the equivalent API for the other platforms -- you cannot simply tell user agents that they MUST NOT implement repair techniques. [`atspi_value_get_current_value()`](https://docs.gtk.org/atspi2/method.Value.get_current_value.html) returns a `gdouble`. What should the value of that `gdouble` be? Same question applies to the min and max getters. Maybe user agents should return -1 instead of [what's in the repair table](https://w3c.github.io/aria/#authorErrorDefaultValuesTable). But they have to return _something_.

Before you file those bugs, it might be worth defining *exactly* what "do not repair" means. Because if it means "return nothing", that might not be implementable. On some platforms, the implementation is not just some property that can be absent/empty/null.

adampage:
> That's not necessary. It will also be applied automatically after merging.

@pkra Oh, I see. So we do want that check to fail and persist through the merge? Would it be better for me to back out my manual commit of that diff? Sorry for my bumbling on this. 🤦🏻‍♂️

pkra:
@adampage 
> @pkra Oh, I see. So we do want that check to fail and persist through the merge? Would it be better for me to back out my manual commit of that diff? Sorry for my bumbling on this. 🤦🏻‍♂️

no worries. Just FYI. It's like the RFC keyword check - it can be ignored. But it doesn't matter if you do it.

PR #2292, opened 2024-07-24
[html-aam] - address title attribute and name prohibited elements

scottaohara:
This PR does a few things:

which i will list in a bit...

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -16267,7 +16267,7 @@
           </ol>
         </section>
         <section>
-          <h4>`a` Element Accessible Name Computation</h4>
+          <h4>`a` (with `href`) Element Accessible Name Computation</h4>
           <ol>
             <li>
               If the `a` element has an <a data-cite="wai-aria-1.2/#aria-label">`aria-label`</a> or an <a data-cite="wai-aria-1.2/#aria-labelledby">`aria-labelledby`</a> attribute the
@@ -16280,7 +16280,7 @@
           </ol>
         </section>
         <section>
-          <h4>`area` Element Accessible Name Computation</h4>
+          <h4>`area` (with `href`) Element Accessible Name Computation</h4>
           <ol>
             <li>
               If the `area` element has an <a data-cite="wai-aria-1.2/#aria-label">`aria-label`</a> or an <a data-cite="wai-aria-1.2/#aria-labelledby">`aria-labelledby`</a> attribute the
@@ -16309,7 +16309,7 @@
           </p>
         </section>
         <section>
-          <h4>Section and Grouping Element Accessible Name Computation</h4>
+          <h4>Sectioning Content Elements Accessible Name Computation</h4>
           <ol>
             <li>
               If the element has an <a data-cite="wai-aria-1.2/#aria-label">`aria-label`</a> or an <a data-cite="wai-aria-1.2/#aria-labelledby">`aria-labelledby`</a> attribute the
@@ -16321,21 +16321,33 @@
           </ol>
         </section>
         <section>
-          <h4>Text-level Element Accessible Name Computation</h4>
-          <p>
-            <a>`abbr`</a>, <a>`b`</a>, <a>`bdi`</a>, <a>`bdo`</a>, <a>`br`</a>, <a>`cite`</a>, <a>`code`</a>, <a>`dfn`</a>, <a>`em`</a>, <a>`i`</a>, <a>`kbd`</a>, <a>`mark`</a>, <a>`q`</a>,
-            <a>`rp`</a>, <a>`rt`</a>, <a>`ruby`</a>, <a>`s`</a>, <a>`samp`</a>, <a>`small`</a>, <a>`strong`</a>,
-            <a data-cite="HTML/text-level-semantics.html#the-sub-and-sup-elements">`sub` and `sup`</a>, <a>`time`</a>, <a>`u`</a>, <a>`var`</a>, <a>`wbr`</a>
-          </p>
+          <h4>Name Prohibited Elements</h4>
+            <p>The following elements are prohibited from being named by authors.</p>
+            <p>
+              <a>`a`</a> (no `href` attribute), `abbr`</a>, <a>`area`</a> <span class="el-context">(no `href` attribute)</span>, <a data-cite="HTML">autonomous custom element (if implicit `generic` role)</a>, 
+              <a>`b`</a>, <a>`bdi`</a>, <a>`bdo`</a>, <a>`br`</a>, <a>`body`</a>, <a>`caption`</a>, <a>`cite`</a>, <a>`code`</a>, <a>`data`</a>, <a>`del`</a>, <a>`dfn`</a>, <a>`div`</a>, <a>`em`</a>, 
+              <a>`figcaption`</a>, <a data-cite="html/custom-elements.html#custom-elements-face-example">form-associated custom element (if implicit `generic` role)</a>, <a>`html`</a>, <a>`i`</a>, <a>`ins`</a>, 
+              <a>`kbd`</a>, <a>`label`</a>, <a>`legend`</a>, <a>`li`</a>, <a>`mark`</a>, <a>`p`</a>, <a>`pre`</a>, <a>`picture`</a>, <a>`q`</a>, <a>`rp`</a>, <a>`rt`</a>, <a>`s`</a>, <a>`samp`</a>, <a>`small`</a>, 
+              <a>`span`</a>, <a>`strong`</a>, <a data-cite="HTML/text-level-semantics.html#the-sub-and-sup-elements">`sub` and `sup`</a>, <a>`time`</a>, <a>`u`</a>, <a>`var`</a>.
+            </p>
           <ol>
             <li>
               If the element has an <a data-cite="wai-aria-1.2/#aria-label">`aria-label`</a> or an <a data-cite="wai-aria-1.2/#aria-labelledby">`aria-labelledby`</a> attribute the
               <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is to be calculated using the algorithm defined in
               <a href="" class="accname">Accessible Name and Description: Computation and API Mappings</a>.
             </li>
-            <li>Otherwise use the `title` attribute.</li>
-            <li>If none of the above yield a usable text string there is no <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.</li>
+            <li>Otherwise there is no <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.</li>
           </ol>
+          <p class="note">
+            Name Prohibited elements do not use the `title` attribute in their name computation. The `title` attribute contributes to the accessible description of the element.
+          </p>
+        </section>
+        <section>
+          <h4>Elements which cannot be named</h4>
+          <p>
+            <a>`base`</a>, <a>`br`</a>, <a>`col`</a>, <a>`colgroup`</a>, <a>`head`</a>, <a>`input type=hidden`</a>, <a>`link`</a>, <a>`map`</a>, <a>`meta`</a>, <a>`noscript`</a>, <a>`param`</a>, 
+            <a>`script`</a>, <a>`slot`</a>, <a>`source`</a>, <a>`style`</a>, <a>`template`</a>, <a>`title`</a>, <a>`track`</a>, <a>`wbr`</a>.
+          </p>
         </section>
       </section>
       <section id="accdesc-computation">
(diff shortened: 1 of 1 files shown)


html-aam/index.html
```diff
@@ -16321,29 +16321,41 @@ <h4>Section and Grouping Element Accessible Name Computation</h4>
...
-          <h4>Text-level Element Accessible Name Computation</h4>
-          <p>
-            <a>`abbr`</a>, <a>`b`</a>, <a>`bdi`</a>, <a>`bdo`</a>, <a>`br`</a>, <a>`cite`</a>, <a>`code`</a>, <a>`dfn`</a>, <a>`em`</a>, <a>`i`</a>, <a>`kbd`</a>, <a>`mark`</a>, <a>`q`</a>,
-            <a>`rp`</a>, <a>`rt`</a>, <a>`ruby`</a>, <a>`s`</a>, <a>`samp`</a>, <a>`small`</a>, <a>`strong`</a>,
-            <a data-cite="HTML/text-level-semantics.html#the-sub-and-sup-elements">`sub` and `sup`</a>, <a>`time`</a>, <a>`u`</a>, <a>`var`</a>, <a>`wbr`</a>
-          </p>
+          <h4>Name Prohibited Elements</h4>
+            <p>The following elements are prohibited from being named by authors.</p>
+            <p>
+              <a>`a`</a> (no `href` attribute), `abbr`</a>, <a>`area`</a> <span class="el-context">(no `href` attribute)</span>, <a data-cite="HTML">autonomous custom element (if implicit `generic` role)</a>,
```
scottaohara, line 16327:
```suggestion
              <a>`a`</a> (no `href` attribute), <a>`abbr`</a>, <a>`area`</a> <span class="el-context">(no `href` attribute)</span>, <a data-cite="HTML">autonomous custom element (if implicit `generic` role)</a>, 
```

jcsteh:
> identify elements which are name prohibited, reducing these items accName algorithm to only accept names from aria-label/aria-labelledby.

So does "name prohibited" mean that aria-label and aria-labelledby are still allowed? Isn't that effectively "name from author"? And doesn't this contradict ARIA/AccName, where "name prohibited" disallows aria-label/aria-labelledby, or am I misconstruing something?

scottaohara:
@jcsteh ARIA's name prohibited is an author must not.  

The rational behind the [name prohibited step added to accName is here](https://github.com/w3c/accname/commit/45dbaa2)

so the element itself (e.g., with a generic role) is now not to be given a name, but the aria-label/ledby on the element can still be used in the calculation of another element's name.  (i honestly still don't understand this use case and why the button couldn't have been given an aria-label, rather than the span within it... this just seems confusing).

While there's a bit of edits to recategorize things, effectively the primary change of this PR (for the html aam spec) is only removing title from these element's accName steps.

jcsteh:
Ah. I get it. I was misinterpreting the "Name Prohibited Elements" section as being the instructions for computing the name of a name prohibited element itself; i.e. when such an element is the root element. But if I understand you correctly, that section is also meant to be used when the name is being computed during a descendant traversal (and similar for the other elements under "Accessible Name Computations By HTML Element"). Probably just my misreading, but I wonder if anyone else might make the same mistake.

aleventhal:
Presumably name prohibited elements are still a pass through when computing the name of ancestor.

rahimabdi (changes requested):

html-aam/index.html
```diff
@@ -16309,7 +16309,7 @@ <h4>`iframe` Element Accessible Name Computation</h4>
           </p>
         </section>
         <section>
-          <h4>Section and Grouping Element Accessible Name Computation</h4>
+          <h4>Sectioning Content Elements Accessible Name Computation</h4>
```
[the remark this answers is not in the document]
scottaohara: Yes. But this section needs to be redone, so I’d rather not do this just to undo it later

html-aam/index.html
```diff
@@ -16321,30 +16321,41 @@ <h4>Section and Grouping Element Accessible Name Computation</h4>
...
         <section>
-          <h4>Text-level Element Accessible Name Computation</h4>
+          <h4>Name Prohibited Elements</h4>
+          <p>The following elements are prohibited from being named by authors.</p>
           <p>
-            <a>`abbr`</a>, <a>`b`</a>, <a>`bdi`</a>, <a>`bdo`</a>, <a>`br`</a>, <a>`cite`</a>, <a>`code`</a>, <a>`dfn`</a>, <a>`em`</a>, <a>`i`</a>, <a>`kbd`</a>, <a>`mark`</a>, <a>`q`</a>,
-            <a>`rp`</a>, <a>`rt`</a>, <a>`ruby`</a>, <a>`s`</a>, <a>`samp`</a>, <a>`small`</a>, <a>`strong`</a>,
-            <a data-cite="HTML/text-level-semantics.html#the-sub-and-sup-elements">`sub` and `sup`</a>, <a>`time`</a>, <a>`u`</a>, <a>`var`</a>, <a>`wbr`</a>
+            <a>`a`</a> (no `href` attribute), <a>`abbr`</a>, <a>`area`</a> <span class="el-context">(no `href` attribute)</span>,
+            <a data-cite="HTML">autonomous custom element (if implicit `generic` role)</a>, <a>`b`</a>, <a>`bdi`</a>, <a>`bdo`</a>, <a>`br`</a>, <a>`body`</a>, <a>`caption`</a>, <a>`cite`</a>,
```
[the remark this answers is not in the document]
scottaohara: that's my bad for putting the "(if ...)" into the `<a>`.  fixing that now.

pkra:
@scottaohara rebased on main.

html-aam/index.html
```diff
@@ -16404,30 +16404,41 @@ <h4>Section and Grouping Element Accessible Name Computation</h4>
...
         <section>
-          <h4>Text-level Element Accessible Name Computation</h4>
+          <h4>Name Prohibited Elements</h4>
+          <p>The following elements are prohibited from being named by authors.</p>
           <p>
-            <a>`abbr`</a>, <a>`b`</a>, <a>`bdi`</a>, <a>`bdo`</a>, <a>`br`</a>, <a>`cite`</a>, <a>`code`</a>, <a>`dfn`</a>, <a>`em`</a>, <a>`i`</a>, <a>`kbd`</a>, <a>`mark`</a>, <a>`q`</a>,
-            <a>`rp`</a>, <a>`rt`</a>, <a>`ruby`</a>, <a>`s`</a>, <a>`samp`</a>, <a>`small`</a>, <a>`strong`</a>,
-            <a data-cite="HTML/text-level-semantics.html#the-sub-and-sup-elements">`sub` and `sup`</a>, <a>`time`</a>, <a>`u`</a>, <a>`var`</a>, <a>`wbr`</a>
+            <a>`a`</a> <span class="el-context">(no `href` attribute)</span>, <a>`abbr`</a>, <a>`area`</a> <span class="el-context">(no `href` attribute)</span>,
+            <a data-cite="HTML">autonomous custom element</a> <span class="el-context">(if implicit `generic` role)</span>, <a>`b`</a>, <a>`bdi`</a>, <a>`bdo`</a>, <a>`br`</a>, <a>`body`</a>,
```
scottaohara, line 16411: removed br.  it was here by mistake and is already listed in the elements which cannot be named section.

```suggestion
            <a data-cite="HTML">autonomous custom element</a> <span class="el-context">(if implicit `generic` role)</span>, <a>`b`</a>, <a>`bdi`</a>, <a>`bdo`</a>,  <a>`body`</a>,
```

html-aam/index.html
```diff
@@ -16404,30 +16404,41 @@ <h4>Section and Grouping Element Accessible Name Computation</h4>
...
             <li>
               If the element has an <a data-cite="wai-aria-1.2/#aria-label">`aria-label`</a> or an <a data-cite="wai-aria-1.2/#aria-labelledby">`aria-labelledby`</a> attribute the
               <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is to be calculated using the algorithm defined in
               <a href="" class="accname">Accessible Name and Description: Computation and API Mappings</a>.
             </li>
-            <li>Otherwise use the `title` attribute.</li>
-            <li>If none of the above yield a usable text string there is no <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.</li>
+            <li>Otherwise there is no <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.</li>
           </ol>
+          <p class="note">Name Prohibited elements do not use the `title` attribute in their name computation. The `title` attribute contributes to the accessible description of the element.</p>
```
scottaohara: @rahimabdi per our chat today, how does this update look?

```suggestion
          <p class="note">Name Prohibited elements do not use the `title` attribute in their name computation. The `title` attribute generally contributes to the accessible description of an element, with exceptions for elements where the attribute has other specified behaviors or mappings.</p>
```

rahimabdi: > where the attribute has other specified behaviors or mappings.

LGTM, thanks! Since `title` is part of the accname for several elements, should we add another exception clause such as "or otherwise specified in an element's accessible name"? This may already be covered by "other specified behaviors".

> of an element, with exceptions

Is there an extra space before the word "with" here? It's difficult to tell but does appears so.

html-aam/index.html
```diff
@@ -16835,30 +16835,41 @@ <h4>Section and Grouping Element Accessible Name Computation</h4>
           </ol>
         </section>
         <section>
-          <h4>Text-level Element Accessible Name Computation</h4>
+          <h4>Name Prohibited Elements</h4>
```
rahimabdi, line 16851: Some additional elements for your consideration:

- br (no mappings)
- wbr (unsure about this, but does have mappings)
- embed (?)
- output
- meter
- slot
- template

rahimabdi: Looks like most of these are rightfully covered under "elements which cannot be named" 👍🏾

scottaohara: meter is covered under "other form elements"  (though i'd like to revise that section to overtly name the elements covered in another PR)

output has its own name calc section.

br, wbr, slot and template are in cannot be named.  

embed is not overtly mentioned, but should be similarly to the computation steps for iframe - i'll make a separate issue for that

html-aam/index.html
```diff
@@ -16848,30 +16848,41 @@ <h4>Section and Grouping Element Accessible Name Computation</h4>
...
       </section>
       <section id="accdesc-computation">
         <h3>Accessible Description Computation</h3>
         <p>
           Authors MAY provide an <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a> for any HTML element that is a valid child of the `body` element. The following list
           represents the order of precedence for <a class="termref">user agents</a> to compute the <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a> of an element. As
-          defined by <a data-cite="accname-1.2/#mapping_additional_nd_description">Accessible Name and Description Computation: Description Computation </a>, <a class="termref">user agents</a> MUST
-          use the first applicable description source, even if its use results in an empty description.
+          defined by <a data-cite="accname-1.2/#mapping_additional_nd_description">Accessible Name and Description Computation: Description Computation</a>, <a class="termref">user agents</a> MUST use
+          the first applicable description source, even if its use results in an empty description.
```
rahimabdi: Minor nit: would the word "available" fit better here rather than "applicable"?

html-aam/index.html
```diff
@@ -16896,7 +16907,10 @@ <h3>Accessible Description Computation</h3>
               </li>
             </ul>
           </li>
-          <li>Otherwise, use the flat string of the `title` attribute if it was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> for the element.</li>
+          <li>
+            Otherwise, use the flat string of the `title` attribute if it was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> for the element, and if the flat
```
rahimabdi: Could you clarify the wording here for "flat string"; is this to account for scenarios where `title` contains non-alphanumeric characters such as emojis?

html-aam/index.html
scottaohara:
```suggestion
          the first available description source, even if its use results in an empty description.
```

html-aam/index.html
scottaohara:
```suggestion
            Otherwise, use the text string value of the `title` attribute if it was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> for the element, and if the flat
```

scottaohara: thanks for catching that - not really applicable here since the title attribute can only have a text string value - and there isn't really anything to flatten, like with other elements where the accName is a calculation of the flattened text / text-equivalent of an element's contents.

PR #2293, opened 2024-07-25
Editorial: Remove important terms from CORE-AAM

spectranaut:
There were only two important terms listed in CORE-AAM, and one of them had no references. The other was only used in one table, so I moved the definition to above the table.


core-aam/index.html
```diff
@@ -10655,7 +10628,11 @@ <h3>Changes to document content or node visibility</h3>
               </tr>
             </tbody>
           </table>
-          <p>Fire these events for node changes where the node in question is an element and has an <a class="termref">accessible object</a>:</p>
+          <p>
+            Fire the following events for node changes where the node in question is an element and has an <a class="termref">accessible object</a>. The accessibility subtree of a node is its
+            [=accessible object=] in the <a>accessibility tree</a> and its descendants in that tree. It does not include objects which have relationships other than parent-child in that tree. For
```
rahimabdi, line 10633: I see that "in that tree" is used more than once; is this redundant?

rahimabdi (changes requested):
Since one of the two definitions ("activation behavior") was removed because it was never referenced, I'm curious if it would be valuable to summarize/include it elsewhere such as [section 4.8 Events](https://w3c.github.io/core-aam/#mapping_events).

Or alternatively, since accessible object is now linked to the ARIA spec, should the ARIA [activation behavior](https://www.w3.org/TR/wai-aria-1.2/#dfn-activation-behavior) definition be referenced rather than entirely removing it?

core-aam/index.html
spectranaut: I just spent a moment thinking about removing it, but ultimately I think it's more clear with it in. In fact, to be even more clear, I think we should write: 
```suggestion
            [=accessible object=] in the <a>accessibility tree</a> and its descendants in that tree. It does not include accessible objects which have relationships other than parent-child in that tree. For
```

spectranaut:
Thanks Rahim, for the thoughts/thoroughness! But there is no reference to anything like activation behavior in the events section, so from my perspective, there is no reason to include it. You got me curious about what that term was originally used for -- so I looked up the commit that added it: https://github.com/w3c/core-aam/pull/92/files

@jnurthen -- you added the term `activation behavior`.. but even when you added it, there was no reference to it in the spec. Do you know why? Is it ok to remove?

jnurthen (approved):

spectranaut:
@rahimabdi can you see my comment and re-review this PR?

curtbellew (approved):

Issue #2296, opened 2024-07-29
Do HTML `<form>` and `<dialog>` elements require an accessible name?


gsharpsharp:
# Describe your concern

In the WAI-ARIA spec, the [`form`](https://w3c.github.io/aria/#form) and [`dialog`](https://w3c.github.io/aria/#dialog) roles are labeled "Accessible Name Required." Does this mean that HTML authors are required to specify an accessible name for `<form>` and `<dialog>` elements (e.g. with an `aria-label` or `aria-labelledby` attribute) in order to comply with the WAI-ARIA spec?

The question could theoretically be asked about the `<img>` and `<table>` elements as well, but it looks like they have other native mechanisms for specifying accessible names as described in HTML Accessibility API Mappings, [section 4.1.10](https://w3c.github.io/html-aam/#img-element-accessible-name-computation) and [section 4.1.11](https://w3c.github.io/html-aam/#table-element-accessible-name-computation) respectively.

# Link to the version of the specification or documentation you were looking at at.

Link to documentation: https://w3c.github.io/aria/

Does the issue exists in [the editors draft](https://w3c.github.io/aria/) (the editors draft is the most recent draft of the specification)? Yes

stevefaulkner:
> In the WAI-ARIA spec, the [form](https://w3c.github.io/aria/#form) and [dialog](https://w3c.github.io/aria/#dialog) roles are labeled "Accessible Name Required." Does this mean that HTML authors are required to specify an accessible name for <form> and <dialog> elements (e.g. with an aria-label or aria-labelledby attribute) in order to comply with the WAI-ARIA spec?

no. What it means is if you use an **explicit** role. 
`<div role=“dialog”>` for example, you must provide an accessible name. The ARIA spec cannot define conformance requirements for native HTML elements.

scottaohara:
to further @stevefaulkner's answer, here's the issue i filed to help correct this common misunderstanding - https://github.com/w3c/aria/issues/2180

Issue #1764, opened 2022-07-08
Further clarifications to form role


scottaohara:
Re: https://github.com/w3c/core-aam/issues/11#issuecomment-1179418567 and https://github.com/w3c/html-aria/issues/423 - requiring a form be named seems like it should be scoped to only if someone wants to expose the form as a landmark.

curtbellew:
This makes sense to me.  Because my ACC checker does check for duplicate landmarks, I would add that I think it's also best to name a form when there is more than one form on the page.  Similar to the nav tag which can often appear more than once on a page - it's best in that case to name the nav regions ie 'Main' vs 'Some other'.  That practice seems to apply here.

scottaohara:
to help clarify this further:

The ARIA spec presently says that a `role=form` is required to be named.

But a form's role may still be announced even if it doesn't have a name, as the Core AAM issue/PR has resulted in.  

Previously, a form (HTML element or if role=form was used) was only consistently exposed as a form if it had a name. Now, naming the form will promote it to a landmark... but it may still be exposed as a form per the linked change.

So the clarity here should be around "why" is a name for a form required.  And that likely could be written up to touch on two topics:
1. to promote it to a landmark
2. to help distinguish different forms from each other that are present on the same web page.

Issue #427, opened 2016-08-29
Radio Group should not have accessible name required: true


jnurthen:
While most radio groups do need an  accessible name, there are some instances where the radio buttons themselves sufficiently describe their operation without requiring an accessible name on the radio group itself.
HTML native radio groups do not require an associated fieldset/legend surrounding them so why should aria ones have this additional restriction.

cookiecrook:
If ARIA 1.2 is still limited to HTML Role Parity, this issue is not in scope.

mcking65:
This is another change that is nearly editorial except for the fact that it has authoring impacts. Changing this would not slow 1.2.

jnurthen:
I'm having a hard time thinking of a good example where a radio group is better without a name. If someone comes up with one please add it and we will progress this bug - but until that time I'm going to park this with needsinfo.

scottaohara:
```
<div role=radiogroup>
<label><input type=radio name=f> I think radiogroups should always have a name</label>
<label><input type=radio name=f> I do not think radiogroups should always have a name</label>
</div>
```

pkra:
Discussed on today's editors meeting https://www.w3.org/2024/04/15-aria-editors-minutes.html#t01

pkra:
Closing in favor of https://github.com/w3c/aria/issues/2180

PR #2297, opened 2024-07-29
Remove name required from alertdialog, dialog, form, grid, radiogroup, table

### LINKED ISSUE #2180: Consider loosening 'requirements' for naming certain container elements
Follow on to #427, radiogroup, table, grid, dialog, etc. - these are currently 'required' to be named from author, but this requirement is commonly confused as being a wcag requirement (these roles being unnamed won't actually always be a wcag failure) - and there can be instances where naming these containers can simply cause information redundancy.  

This issue is being filed to talk with the larger working group about changing the author musts to author shoulds for these sorts of roles.

scottaohara:
closes #2180
related to recent question: #2296
related to https://github.com/w3c/aria/issues/1764

This PR removes requiring names for alertdialog, dialog, form, grid, radiogroup and table roles and replaces the author MUST with an author SHOULD. (for grid/table there was no author MUST naming paragraph, so the author SHOULD paragraph was added to these roles).

With this change there will likely be necessary updates to related APG/MDN docs.

* [x] Related APG Issue/PR: https://github.com/w3c/aria-practices/issues/3244
* [x] MDN Issue/PR: https://github.com/mdn/content/issues/38262
* [ ] IBM Equal Access Toolkit issue

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2297.html" title="Last updated on Feb 21, 2025, 9:03 AM UTC (c5b1911)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2297/077621f...c5b1911.html" title="Last updated on Feb 21, 2025, 9:03 AM UTC (c5b1911)">Diff</a>


pkra (approved):

giacomo-petri:
Hi group,

I prefer to keep things simple, and I really like this change. However, I strongly believe there’s a significant difference between:
- alertdialog
- dialog (with aria-modal="true")

and
- form
- grid
- radiogroup
- table 

I understand that repeating the main heading of a modal in its accessible name can sometimes feel redundant. However, if focus moves to the modal and the screen reader only announces it as a "modal dialog," users are forced to navigate inside just to understand its purpose, which isn’t efficient either. In this case, I’d prefer some redundancy over making users enter a modal they may not even need.

That said, I can live with either approach, but I want to note that I do think expectations for the roles above differ.

smockle:
> …if focus moves to the modal and the screen reader only announces it as a "modal dialog", users are forced to navigate inside just to understand its purpose, which isn't efficient…

How common is it to focus the modal itself? Is that something web authors should be doing?

At GitHub, we typically move focus to an interactive element _inside_ a modal dialog (e.g. a close button). ARIA APG’s [Modal Dialog Example](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/) and [Date Picker Dialog Example](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/) do this too—in the former example, the “Street” input is focused; in the latter, a `<td>` with `tabindex=0` representing the current date.

❧

To speed things along, should we split this PR? One PR could remove ‘name required’ from `form`, `grid`, `radiogroup`, and `table`; the other could remove ‘name required’ from `alertdialog` and `dialog` (with `aria-modal="true"`). That’d let us address the confusion in https://github.com/w3c/aria/issues/2431 right away, while giving us time to discuss the dialog case further.

giacomo-petri:
If an alert dialog requires a choice (e.g., "Accept" or "Decline") and cannot be dismissed without selection, immediately shifting focus to one of the options without providing context, I believe it creates a poor user experience. 

Similarly, opening a dialog and moving focus directly to the close button is also ineffective. Ideally, the topic or purpose of the dialog should be announced first before presenting the option to close it (IMO).

scottaohara:
fwiw, @giacomo-petri and i spoke today and we agree that i many cases it would be beneficial for dialogs to have accessible names.  but, the point of this update is not to say that dialogs wouldn't benefit form accnames, but to acknowledge the fact that not every dialog needs an accname - and thus why an author MUST is too strict, and it out of alignment with HTML _not_ requiring dialogs to be named.

all that said, if there is any further editorial one would like to see added to this PR to further note that accNames are recommended _but not required_ for dialogs, please provide a suggestion of such text, or we should make a separate PR for that.  But i would very much like to not regress on the previous decision and split this PR apart.

smockle (approved):

smockle:
Thanks @scottaohara; that makes sense. 👍

Incidentally, reviewing the [2024-05-23 minutes](https://www.w3.org/2024/05/23-aria-minutes.html#t06), you said almost the same thing then:
> **scott:** …I'm not arguing that there aren't cases where things do need names, the issue is that this is a pass/fail boolean right now, and we're saying that these always need a name and _that's not true_
> **scott:** just want to be absolutely clear that I'm not saying dialogs don't require names, that would be a misunderstanding. I still agree that in most cases, it would be good, and people should provide names to these things. I think it's an overreach to say they're always required

Wildebrew:
I hate to introduce a complication, but I think radiogroup is a very different role from the rest.
Radiogroup represents a group of form controls. Radio buttons trigger screen readers to switch into forms/application mode, which makes reading static content complicated (you have to switch in and out of browse mode), not to mention that usres expect to be able to navigate forms with the tab key.

Given the tendency of using standard labels for radio buttons and relying on the group label to differentiate them, it would significantly complicate the screen reader user experience not to have an accessible name for each group.


Imagine a standard survey with 8 or 9 questions on the same page.
The questions will touch on various aspects of a recent user experience.
The radio buttons in each group have the same labels, e.g. "yes" and "no", or "agree" "somewhat agree" and "disagree".
Without the group label you have no idea, as a screen reader user, what you are responding to.
Yes, you could force the screen reader out of application/forms mode and then navigate up past the first radio button to find the static content that represents the question, hopefully the question is coded as a heading to make this process easier. But switching interaction modes for a screen reader is not exactly a super simple trick, especially with auto forms mode on (which is the default behavior of most screen readers).

In my experience, not having a group label on such a page is likely to increase the time required to answer a set of 5 questions with standard group labels by a factor of 4 or more. I admit that I do not have official study statistics to back this up but I've used a screen reader for over 30 years myself, trained dozens of people on how to use a screen reader, and even translated NVDA (into Icelandic). 

To sum up, given radiogroup's strong association with forms I recommend to remove it from this list and continuing to require an accessible name for it.

In fact I would go so far as to argue that the group role should be upgraded to require an accessible name as well, since it is inevitably used for grouping form controls, including radio buttons or text inputs that trigger screen readers to switch to application/forms mode and users navigating those with the tab key instead of carefully perusing the content.

Keep in mind that the `<fieldset>` element maps to the group role, and if you are going to group radio buttons with HTML, that is the element you use.

stevefaulkner:
yeah i agree with @Wildebrew if a `radiogroup` role is used to identify a set of radio buttons it must have an accname, otherwise the group label of the set of controls will not understood.

please can someone provide the use case for radiogroup with no label? it serves no one

smockle:
Here’s an example @scottaohara provided in https://github.com/w3c/aria/issues/427#issuecomment-2057234198:

```HTML
<div role=radiogroup>
  <label><input type=radio name=f> I think radiogroups should always have a name</label>
  <label><input type=radio name=f> I do not think radiogroups should always have a name</label>
</div>
```

> per the current spec, it would be required to add a name. but the names of the radio groups make it pretty clear what the intent of the group is asking one to decide.

Wildebrew:
That's not how the vast majority of radio buttons are coded.
The whole idea of grouping controls is to remove repeated text and make the
experience easier to understand.



If I saw that example in the wild I would make an accessibility usability
recommendation to change it to

```
<div role="radiogroup" aria-labelledby="q2">
<h2 id="q2">Do you think radiogroup should have an accessible name?</h2>
<label> <input type="radio" name="q22"> <span>yes</span></label>
<label> <input type="radio" name="q22"> <span>No</span></label>
</div>
```

Advantages:
* You can skim the questions by navigating by headings before you start
* It's much easier to understand with a screen reader
* For mobile devices, you save screen space by having shorter labels

(verbose labels are tiresome and confusing for a screen reader user,
especially long labels with repeated text)

stevefaulkner:
@smockle @scottaohara What purpose does the `role="radiogroup"` role serve in this instance? 

Who does its semantics serve? 

What does it impart that its absence wouldn't?

`radiogroup` becomes `group` in the acctree (on windows). Without an accesssible name both the `group` and `radiogroup` role are ignored by screen readers as I recall.

```html
<div role=radiogroup>
  
<label><input type=radio name=f> I think radiogroups should always have a name</label>
  
<label><input type=radio name=f> I do not think radiogroups should always have a name</label>

</div>
```

giacomo-petri:
This mirrors the discussion I had with Scott, where we agreed that while users often benefit from the accessible name in these roles, perhaps even the majority of the time, there are cases where it's unnecessary or redundant, making a strict "MUST" requirement too rigid.

For example, I've seen many quizzes where each page contains a single question as an h1, followed by answer choices wrapped in a `<fieldset>` tag. In such cases, assigning an accessible name to the radio group isn't required, as it would simply duplicate the `<h1>`. Here we're not debating whether it's a must, a best practice, a nice to have, or not needed at all; we're just pointing out that it's not always a strict requirement.

Regarding the question, 

> Who does its semantics serve?

we're not saying an accessible name shouldn't be set, just that it is not always necessary. In HTML, a `<fieldset>` can exist without a `<legend>`, right? We're aiming to clarify that while setting an accessible name is generally recommended, it's not always mandatory.

giacomo-petri (approved):

stevefaulkner:
@giacomo-petri 

What I am saying is there is no scenario, that has been offered so far, that shows when `<div role="radiogroup">` without an accessible name has any benefit. Would appreciate use cases. 

In regards to `<fieldset>` it is similar, but not the same as it has a default visual semantic (border), there is no wilful addition of a role annotation specifically for accessibility, to convey a semantic meaning.

accdc:
I agree, there is no reason that I can think of where having a radiogroup with no accessible name will be of any benefit to accessibility for anyone.

scottaohara:
>That's not how the vast majority of radio buttons are coded.

You're right, the vast majority are.  But then there are some that aren't.  The simple example provided of having radio buttons with labels where their purpose is clear, but the radiogroup has no accessible name is one of the reasons conformance checkers like [axe](https://github.com/dequelabs/axe-core/issues/3658), the html validator and arc toolkit _do not_ file an issue for radiogroups that don't have names.  (as a counter - IBM's equal access checker _does_ flag when a radiogroup lacks a name). There's potential for false positives.  Per the example cited, where a radiogroup is used but the choices / purpose of the group are clear from the labels - i would find it very suspect if someone raised that in an accessibility audit, unless it was cited as a best practice (which i'd agree with and advocate for - and seems we agree on that one @wildebrew, per your mention you'd also recommend it be changed as a usability improvement).

@stevefaulkner regarding:
> What purpose does the `role="radiogroup"` role serve in this instance?  

This issue isn't about saying using a radiogroup without a name is a good idea.  As stated multiple times now, authors absolutely SHOULD provide all these roles names whenever appropriate.  Rather, this issue/PR is to align with the fact that there are cases where this role / the others mentioned can still have content that is understandable in cases when names aren't provided.  An author MUST leaves no wiggle room for use cases where the name isn't needed.  If the radiogroup role were to be exposed even without a name, this would be more useful for the no-name scenarios. Again, i'm not arguing that.  It's to mitigate against the fact that not all checkers are even implementing the author must rule, since it has potential as a false positive when reviewing for WCAG.  What is required by ARIA is not necessarily what is required by HTML or WCAG.  If other specs had requirements that weren't being implemented, then they'd be downgraded/removed.  This isn't a PR about one's opinions, if it was, then I'd have left this role alone...

With all that said, if people strongly want to keep radiogroup as a role that must be named by authors, then I *personally* am fine with that.   But I will point out again that doing so is advocating for a rule that doesn't necessarily align with WCAG, and isn't a check respected by all a11y checkers.  So, keeping this an author must will not actually match reality unless more checkers get on board with surfacing this as an error.

stevefaulkner:
thanks @scottaohara this is not a hill i would die on, but I am still interested in real world examples of cases where a radiogroup sans accname is useful, [Eric Eggert provided an interesting example](https://yatil.social/@yatil/114018388961326972)

wittjeff:
I'm generally onboard with @scottaohara's thinking here. Two notes:
1) I find the labeling of clickthrough consent contracts to be surprisingly non-optimal a lot of the time, and I think the reason for that (overly brief labels that required reading of prior text, such as "I agree") bumps into this question re naming of fieldsets.  But what if you have a situation where the author has written nicely explicit consent checkboxes in a group, such as "I agree to the Terms of Service" and "I agree to the Privacy Policy".  Are we then going to force the author to make up a label for the group of them when the scope is quite succinct already?  I can imagine a couple of grouping labels that in some situations might be appropriate or slightly better than no group label, but I'm not over the line enough to say it should always be required here.

2) IMO requiring meaningful labels for iframes should be softened for essentially the same reason.  Some sources note that the title attribute is really only needed if the iframe is visually distict (which I think is still arguable), but I think most automated testing tools consider this a hard requirement.  If a sighted person can't tell the difference between a div and an iframe, I don't see why a blind user always needs more semantic information in that context.

spectranaut:
Discussed in today's ARIA working group meeting: https://www.w3.org/2025/02/20-aria-minutes#eafc

Thank you for the feedback, everyone. The conclusion is that, even if the scenarios are very rare, so long as authors can create accessible mark up using these roles and no accessible name, then the MUST needs to be a SHOULD.

pkra:
I've filed an MDN issue.

scottaohara:
yes @mcking65 - i know the way the set / position in set is calculated a bit differently in chromium based browsers.  but webkit and firefox do seem to use radiogroup (named or not) to indicate the set size / position in set for radios.

as a comparison, just placing radios in a generic container or even a container with role=group did not result in the radio buttons getting their set size or position in set being calculated / exposed.

as mentioned, chromium is a bit different in the way they are calculating radio buttons x of y.  but, where radiogroup and group are helpful here is that help chrome identify these are separate sets of radio buttons (with the radiogroup / group not needing a name to be acknowledged).  

For instance:

```
<div role=radio>1</div>
<div role=radio>2</div>

<div role=radiogroup>
  <div role=radio>1b</div>
  <div role=radio>2b</div>
</div>

<div role=radio>3</div>
```

with chrome, radio buttons 1, 2 and 3 are considered one set. The radiogroup role (as well as other non-generic/ignored roles) is seen in the a11y tree, and those radio buttons are treated as their own group, separate from the ones previously mentioned.  It doesn't matter if the radiogroup is named or not - the browser still sees it and accounts for it in how it communicates the a11y tree, even if screen readers don't explicitly announce the radiogroup due to lack of name.

so yeah, thanks for bringing this up, since just helps explain further that the role is not without use if it doesn't have a name...

spectranaut:
I think we decided we can merge this now @pkra ?

PR #2299, opened 2024-07-31
Editorial: Update computation step "comp_embedded_control" (formerly step 2c) text for clarity

### LINKED ISSUE #244: Computation steps, section 2C could be more clear
@aleventhal and I were reading [the computation steps, section 2C](https://w3c.github.io/accname/#comp_embedded_control):

> Otherwise, if the current node is a control embedded within the label (e.g. any element directly referenced by aria-labelledby) for another [widget](https://www.w3.org/TR/wai-aria-1.2/#dfn-widget), where the user can adjust the embedded control's value, then return the embedded control as part of the text alternative in the following manner:

We're not 100% sure what this means and we thought others might also find it confusing. Would it be possible to revise this step for improved clarity? Including additional examples could be helpful as well. Thanks!

MelSumner:
Closes [accname issue 244](https://github.com/w3c/accname/issues/244)

If merged, this PR updates step "comp_embedded_control" (formerly step 2c) to be the agreed upon text per this issue: https://github.com/w3c/accname/issues/244

### DIFF
--- a/accname/index.html
+++ b/accname/index.html
@@ -554,7 +554,8 @@
                   </div>
                 </li>
                 <li id="comp_embedded_control">
-                  <span id="step2C"><!-- Don't link to this legacy numbered ID. --></span><em>Embedded Control:</em> Otherwise, if the <code>current node</code> is a user-adjustable control embedded within the label, and the control is not the element being named or described, then return the embedded control's value as part of the text alternative in the following manner:
+                  <span id="step2C"><!-- Don't link to this legacy numbered ID. --></span><em>Embedded Control:</em> Otherwise, if the <code>current node</code> is a user-adjustable control embedded
+                  within the label, and the control is not the element being named or described, then return the embedded control's value as part of the text alternative in the following manner:
                   <ol>
                     <li id="comp_embedded_control_textbox"><em>Textbox:</em> If the embedded control has role <a class="role-reference" href="#textbox">textbox</a>, return its value.</li>
                     <li id="comp_embedded_control_combobox_or_listbox">
--- a/accname/index.html
+++ b/accname/index.html
@@ -554,7 +554,7 @@
                   </div>
                 </li>
                 <li id="comp_embedded_control">
-                  <span id="step2C"><!-- Don't link to this legacy numbered ID. --></span><em>Embedded Control:</em> Otherwise, if the <code>current node</code> is a user-adjustable control embedded
+                  <span id="step2C"><!-- Don't link to this legacy numbered ID. --></span><em>Embedded Control:</em> Otherwise, if the <code>current node</code> is a user-perceivable control embedded
                   within the label, and the control is not the element being named or described, then return the embedded control's value as part of the text alternative in the following manner:
                   <ol>
                     <li id="comp_embedded_control_textbox"><em>Textbox:</em> If the embedded control has role <a class="role-reference" href="#textbox">textbox</a>, return its value.</li>
--- a/accname/index.html
+++ b/accname/index.html
@@ -554,9 +554,7 @@
                   </div>
                 </li>
                 <li id="comp_embedded_control">
-                  <span id="step2C"><!-- Don't link to this legacy numbered ID. --></span><em>Embedded Control:</em> Otherwise, if the <code>current node</code> is a control embedded within the label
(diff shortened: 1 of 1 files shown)


aleventhal (approved):

accdc (approved):

accname/index.html
```diff
@@ -554,9 +554,8 @@ <h4>Computation steps</h4>
                   </div>
                 </li>
                 <li id="comp_embedded_control">
-                  <span id="step2C"><!-- Don't link to this legacy numbered ID. --></span><em>Embedded Control:</em> Otherwise, if the <code>current node</code> is a control embedded within the label
-                  (e.g. any element directly referenced by <code>aria-labelledby</code>) for another <a class="termref">widget</a>, where the user can adjust the embedded control's value, then return
-                  the embedded control as part of the text alternative in the following manner:
+                  <span id="step2C"><!-- Don't link to this legacy numbered ID. --></span><em>Embedded Control:</em> Otherwise, if the <code>current node</code> is a user-adjustable control embedded
```
cookiecrook, line 557: "user-adjustable" doesn't account for an embedded `readonly` field. 

Do we care about this?
```html
<label for=""cb>
  <input type="cb" name="cb" id="id">
  Flash the screen <input readonly value="3"> times.
</label>
```

Will the new language, the (admittedly contrived) label above would be "  Flash the screen times."

cookiecrook: Actually I guess the old text doesn't account for that either. "user-perceivable" perhaps? Or just remove that adjective?

cookiecrook: "user-perceivable" perhaps? 

```suggestion
                  <span id="step2C"><!-- Don't link to this legacy numbered ID. --></span><em>Embedded Control:</em> Otherwise, if the <code>current node</code> is a user-perceivable control embedded
```

Or just remove that adjective?

accname/index.html
```diff
@@ -554,9 +554,8 @@ <h4>Computation steps</h4>
...
+                  within the label, and the control is not the element being named or described, then return the embedded control's value as part of the text alternative in the following manner:
```
cookiecrook, line 558: Nit: take or leave. "as part of the text alternative" makes it sound (a little bit) like you're overriding how this returned value would be included in the over text alternative computation... I think  it stands without that clause, but feel free to keep it and resolve this comment if you disagree.

```suggestion
                  within the label, and the control is not the element being named or described, then return the embedded control's value in the following manner:
```

accname/index.html
cookiecrook: Note: This includes disabled fields too, but I think that's good. What do you all think about this (also contrived) example?

```html
<label for=""cb>
  <input type="cb" name="cb" id="id">
  Flash the screen <input disabled value="3"> times.
</label>
```

accdc:
Simply referring to control sounds good to me if it's clear to everyone else.

scottaohara:
the argument about just saying "control" is that a display none control's value should not participate.  and i'd also argue that one should not anticipate a `input type=checkbox value=nope` would have its value participate either.

a non-hidden control with a user-perceivable value (e.g., text input value or chosen option from a select, but not checkbox or radio button value) seems more like what we would actually want covered here.

cookiecrook:
> a non-hidden control with a user-perceivable value (e.g., text input value or chosen option from a select, but not checkbox or radio button value) seems more like what we would actually want covered here.

"user-perceivable value" for a checkbox (true or false, checked or unchecked) would also be perceivable. If that's the case, the value for each control type would need to be specified. Yuck.

Should it be limited this to "user-perceivable control with a _string_ value"?

scottaohara:
when i was referring to value i meant the string value.  e.g., `input type=checkbox value=foo`  the state (checked or not) shouldn't be considered as part of the calculation.  agreed that'd be yucky.

accname/index.html
MelSumner: it might but I think it was part of the feedback to add it in the first place.

MelSumner:
@MelSumner note for myself to rebase and also address comments that don't have suggestions.

spectranaut:
@MelSumner this has been open for a bit want to take a look and try to get it over the finish line? :)

PR #2307, opened 2024-08-08
Editorial: Handle PRs from forks

### LINKED ISSUE #2294: Prettier throws errors on PRs opened from forks
See: https://github.com/w3c/aria/pull/2290

daniel-montalvo:
Fixes #2294

### DIFF
--- a/.github/workflows/prettier.yaml
+++ b/.github/workflows/prettier.yaml
@@ -34,4 +34,4 @@ jobs:
 
       - name: Handle commit failure
         if: ${{ steps.prettify.outcome == 'failure' && steps.check_fork.outputs.is_fork == 'true' }}
-        run: echo "Commit failed due to lack of permissions. Please update permissions."
+        run: echo "Please run prettier locally"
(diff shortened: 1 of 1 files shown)


pkra:
I was hoping we could run a prettier check before trying to write. 

That way, we can ask external PR authors to run prettier locally to pass the check, avoiding the write access problem.

spectranaut (changes requested):
Thanks for this, Daniel, but I agree with @pkra ! It would be better if we did a prettier dry run on forks and if it fails, ask the person who made the PR to run prettier locally.

From a quick look it looks like a possibility: https://github.com/marketplace/actions/prettier-action

spectranaut (approved):
For now we need this, and we just need to request that people run prettier if they come from a fork.

daniel-montalvo:
@pkra @spectranaut 

I think this now matches your suggestions.

spectranaut (approved):
awesome thanks!!

MelSumner (approved):

PR #2309, opened 2024-08-08
[html-aam] Update ATK URL

smockle:
Initially filed as https://github.com/w3c/html-aam/pull/560

https://gnome.pages.gitlab.gnome.org/atk/ (the URL currently referenced throughout [html-aam](https://w3c.github.io/html-aam)) 404s, even after logging in to the GNOME GitLab instance[^1].

https://docs.gtk.org/atk/ (this PR’s replacement URL) seems to be the new location. [GTK’s “API Docs” page](https://www.gtk.org/docs/apis/) links to this URL.

[^1]: That said, [GNOME’s “Technologies” page](https://www.gnome.org/technologies/) still links to this URL. 🤷

❧

<details>
<summary>Test, Documentation and Implementation tracking</summary>

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:
</details>

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -309,7 +309,7 @@
               </td>
             </tr>
             <tr>
-              <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
+              <th><a href="https://docs.gtk.org/atk/">ATK</a></th>
               <td>
                 <div class="general">Use WAI-ARIA mapping</div>
               </td>
@@ -359,7 +359,7 @@
               </td>
             </tr>
             <tr>
-              <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
+              <th><a href="https://docs.gtk.org/atk/">ATK</a></th>
               <td>
                 <div class="general">Use WAI-ARIA mapping</div>
               </td>
@@ -412,7 +412,7 @@
               </td>
             </tr>
             <tr>
-              <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
+              <th><a href="https://docs.gtk.org/atk/">ATK</a></th>
               <td>
                 <div class="role"><span class="type">Role:</span> `ATK_ROLE_STATIC`</div>
                 <div class="objattrs">
@@ -467,7 +467,7 @@
               </td>
             </tr>
             <tr>
-              <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
+              <th><a href="https://docs.gtk.org/atk/">ATK</a></th>
               <td>
                 <div class="general">Use WAI-ARIA mapping</div>
               </td>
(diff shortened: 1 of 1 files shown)


scottaohara (approved):

scottaohara:
so.... initially i was thinking this would be good - replace all the links with the new link.  but this doesn't actually fix the link in the referenced terms.  that still goes to the old documentation page, and that's an issue in core aam and ARIA (all the specs i'd assume) as well.  

but also, core aam doesn't link to each of the platform pages within each feature table.... and that got me wondering why this spec does/should continue to do so?  

i'm thinking maybe the fix here is to fix the link in the informative references section, and instead change the atk links to use the reference link `[[ATK]]`, as well as the platform references?

scottaohara (changes requested):
see prior comment

smockle:
Thanks @scottaohara; that makes sense. 👍

I’ll update the PR to do what you’re suggesting:
> …fix the link in the informative references section, and instead change the atk links to use the reference link `[[ATK]]`, as well as the platform references…

❧

One question: I found https://gnome.pages.gitlab.gnome.org/at-spi2-core/atk/ today. Between that and https://docs.gtk.org/atk/, which should we link to?

scottaohara:
@spectranaut do you know which we should be referring to, per @smockle's question above?  seems the api version is the same in both, but the library version is different.

smockle:
Quick updates:

- I discovered someone had already corrected the bibliographic entry: https://github.com/w3c/aria/blob/6723ad2bf812395c1b32661a2431e61c02fe6e05/html-aam/biblio.js#L2-L7 However, that correction is not being loaded, so I pushed a commit to start loading it: https://github.com/w3c/aria/pull/2309/commits/71891a45d70f69f138730bd51be66626eea667f6.

  Now, https://gitlab.gnome.org/GNOME/at-spi2-core/ is the URL used in this PR. We could keep that, since it _doesn’t_ 404, or we could replace it with https://gnome.pages.gitlab.gnome.org/at-spi2-core/atk/ or https://docs.gtk.org/atk/.

- I replaced the hardcoded links throughout the document with `[[ATK]]`: https://github.com/w3c/aria/pull/2309/commits/4122754f9a200214b58a12b9e1b7185eff3be540. I wrapped these in `class="informative"`, to keep the ATK reference from moving into “B.1 Normative references”.

scottaohara:
looks good to me.  i am thinking we leave the link that's in the biblio.js already...

we can tackle changing the other platform api links to instead link to their reference section link in a follow on PR, if we dont' want to just do those here as well.

spectranaut:
Sorry for the late reply, I think https://gnome.pages.gitlab.gnome.org/at-spi2-core/atk/ is the right choice!

Edit: also this change looks good. seems fine to link from every table, even if core-aam doesn't.

scottaohara:
@smockle can you look at resolving the merge conflicts.  then we can merge

daniel-montalvo:
@spectranaut @scottaohara @smockle 

I resolved conflicts, if I am not mistaken this should now be ready to merge.

Issue #2318, opened 2024-08-22
Automate PR labeling


daniel-montalvo:
Create gh automation to add labels based on either files changed  or title
https://github.com/actions/labeler

daniel-montalvo:
This now succeeds, for example in #2370 

Closing, feel free to reopen if there's anything else I am missing here.

PR #2321, opened 2024-08-23
Editorial: Add ./github/labeler

daniel-montalvo:


pkra:
Looks intriguing! Is there some documentation on how this works?

daniel-montalvo:
@pkra 
See https://github.com/actions/labeler

My understanding is that something like this should work

pkra:
Thanks @daniel-montalvo. 

IIUC their documentation correctly, the file needs to be placed in `.github/labeler.yml`  (not `.github/ISSUE_TEMPLATE/labeler.yml`).  We'd also need an accompanying workflow file, e.g., `.github/workflows/labeler.yml` described at https://github.com/actions/labeler?tab=readme-ov-file#create-workflow

pkra (approved):
From what I can tell (without experience with that action), this looks good.

daniel-montalvo:
Closes #2318

Issue #2242, opened 2024-06-13
Netlify to replace pr-review


spectranaut:
We also want to look to add the follow features to our netlify publications:
- [ ] links from the netlify publication back to the PR it was generated from
- [ ] successful links between specs generated by the PR
- [ ] publish generated respec document (instead of running respec client side)

daniel-montalvo:
We now have statically-generated PR previews
w3c/aria#2265

spectranaut:
About `Eventually, we will need successful links between specs generated by the PR instead of links to the editor's drafts` -- If, for core-aam, we add the ED links like this: 
```
  ariaSpecURLs: {
    ED: "../index.html",
    CR: "https://w3c.github.io/aria/",
    CRD: "https://w3c.github.io/aria/",
    PR: "https://www.w3.org/TR/wai-aria/",
    REC: "https://www.w3.org/TR/wai-aria/",
  },
  accNameURLs: {
    ED: "../accname/index.html",
    CR: "https://w3c.github.io/accname/",
    CRD: "https://w3c.github.io/accname/",
    PR: "https://www.w3.org/TR/accname-1.2/",
    REC: "https://www.w3.org/TR/accname-1.2/",
  },
```
It works in PR preview, which you could see before I realized w3c.github.io/core-aam editor draft would be broken: https://github.com/w3c/aria/pull/2298

So maybe we really do need to set up all the w3c.github.io/* to directed back to this repository. Then if we do this for all the specs, the links will be correct in PR preview... right...?

daniel-montalvo:
@spectranaut @pkra @jnurthen and all.

I have followed an approach similar to @spectranaut comments above in https://github.com/daniel-montalvo/aria

The build process now has a bash script that replaces specStatus with "unofficial" and then includes the unofficial keys for the resolveReferences function to generate relative links in the PRs.

For a prototype of how this could work, see
https://staging-aria.netlify.app
and add /<shortName> to get to the other specs as desired.

I am using the Netlify production context for now because this is just a test. If we are happy with this approach, I will adjust the main ARIA netlify configuration file accordingly so that this only runs in PRs.

pkra:
@daniel-montalvo I thought #2325 and #2386 had fixed cross references to link within PRs but on #2416 in https://deploy-preview-2416--wai-aria.netlify.app I find ~5000 links starting with `https://w3c.github.io/aria/#`. Should I open a new issue?

daniel-montalvo:
[...]

PR #2325, opened 2024-08-28
Editorial: Update netlify build to get relative links to specs in PRs

daniel-montalvo:
- Update our Netlify.toml config file
- add a new script that
  - changes specStatus to unofficial
  - creates new URL mappings for unofficial status to allow for relative links to specs

Part of #2242

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,53 +1,43 @@
 spec:aria:
-  - any:
-    - changed-files:
-        - any-glob-to-any-file: 'index.html'
+  - changed-files:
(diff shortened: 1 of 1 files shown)


pkra (approved):
I spot checked a few links in a few specs. It's slightly confusing at first that the regular bibliography links remain the same. 

I can still see it being helpful for, e.g., html-aam entries linking the roles to the current PR draft.

Perhaps we should ask someone who requested this if this is helpful.

pkra:
@cookiecrook would you mind taking a look if this is the direction you had in mind?

PR #2341, opened 2024-09-25
Update: fix inconsistency around aria-setsize=-1

### LINKED ISSUE #1759: Revisit aria-setsize="-1" authoring requirement
The current text of `aria-setsize` has the following requirement for when the size is unknown:

> When specifying aria-setsize, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of aria-setsize to -1.

The issue is that while there is an author requirement to set the value to `-1`, there is no corresponding UA requirement for how to handle `-1`. There is also no apparent support that I could find for `aria-setsize="-1"` indicating an unknown number of items (though I did not test on Linux).

In core AAM the only mapping to special-case `-1` is ATK/AT-SPI, which has the following:
> State: STATE_INDETERMINATE if the author-provided value is -1

In tests, I found a mix of directly announcing the set size as -1, announcing the set size as 0, and ignoring the attribute entirely in favor of calculating the set size based on DOM. The result is weird or confusing UX, such as landing on "row 1 of 0" (in the case of a grid with a header row and loading body rows).

I think the author "SHOULD" should be removed in favor of "MAY", or potentially the entire section on "-1" should be taken out until there is at least minimal support for it.

related: #1323

scottaohara:
closes #1759

currently drafting...

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:


core-aam/index.html
```diff
@@ -9852,9 +9852,12 @@ <h3>Group Position</h3>
             <a class="property-reference" href="#aria-level"><code>aria-level</code></a
             >, <a class="property-reference" href="#aria-posinset"><code>aria-posinset</code></a
             >, and <a class="property-reference" href="#aria-setsize"><code>aria-setsize</code></a> are all 1-based. When the [=ARIA/property=] is not present or is &quot;0&quot;, it indicates the
-            property is not computed or not supported. If any of these properties are specified by the author as either &quot;0&quot; or a negative number, [=user agents=] SHOULD use &quot;1&quot;
-            instead.
+            property is not computed or not supported.</p>
+	  <p>If the <a class="property-reference" href="#aria-level"><code>aria-level</code></a> or <a class="property-reference" href="#aria-posinset"><code>aria-posinset</code> properties are specified by the author as either &quot;0&quot; or a negative number, [=user agents=] SHOULD use &quot;1&quot; instead.
```
MelSumner, line 9856:
```suggestion
	  <p>If the <a class="property-reference" href="#aria-level"><code>aria-level</code></a> or <a class="property-reference" href="#aria-posinset"><code>aria-posinset</code></a> properties are specified by the author as either &quot;0&quot; or a negative number, [=user agents=] SHOULD use &quot;1&quot; instead.
```

scottaohara: Thanks

index.html
```diff
@@ -12904,10 +12904,10 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p>If all items up to the current item in a set are present in the document structure, it is not necessary to set this <a>attribute</a>, as the <a>user agent</a> can automatically calculate the position for these items. However, if all previous items in the set are not present in the document structure at a given moment, the author MUST set this <a>attribute</a>to provide an explicit indication of an element's position.</p>
-				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.</p>
+				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown (indeterminate), authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.</p>
```
aleventhal, line 12910: How about "if the total number of items is intedeterminate, and the DOM is currently incomplete such that the user agent would not be able to compute the correct number, authors SHOULD ..."

MelSumner (approved):
This seems to resolve the confusion, thank you!

index.html
scottaohara, line 12910:
```suggestion
				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown (indeterminate), and the DOM is currently incomplete such that user agents would not be able to compute the correct number of items in a set, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.</p>
```

index.html
```diff
@@ -12904,10 +12904,10 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p>If all items up to the current item in a set are present in the document structure, it is not necessary to set this <a>attribute</a>, as the <a>user agent</a> can automatically calculate the position for these items. However, if all previous items in the set are not present in the document structure at a given moment, the author MUST set this <a>attribute</a>to provide an explicit indication of an element's position.</p>
-				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.</p>
+				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown (indeterminate), and the DOM is currently incomplete such that user agents would not be able to compute the correct number of items in a set, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.</p>
```
scottaohara, line 12910:
```suggestion
				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown (indeterminate), and the DOM is currently incomplete such that user agents would not be able to compute the correct number of items in a set, authors SHOULD set the value of <code>aria-setsize</code> to a negative number (e.g., <code>-1</code>) or <code>0</code>.</p>
```

scottaohara: completely align the core aam change and the ARIA spec.   So, not just -1 but _any_ negative number OR zero

scottaohara:
talking with @aleventhal about this some more, we should also add in what to do about if someone uses the empty string, or puts in literal "garbage" text strings. 

So this can probably be 
positive full numbers works as presently spec'd 
0 and any negative number == indeterminate
anything else == undefined and thus ignore.

aleventhal:
> So this can probably be positive full numbers works as presently spec'd 0 and any negative number == indeterminate anything else == undefined and thus ignore.

I know this is going to happen in a follow-up, but it's probably good if it's not too strict, e.g. spaces before and after the number are ok. I was going to do the following:
1. Empty string or not present = undefined and thus ignore
2. Pass it to a non-strict int conversion function, and comes back < 1, treat as follows:
    - aria-posinset or aria-level -> 1
    - aria-setsize -> -1 (indeterminate)

index.html
```diff
@@ -12904,10 +12904,10 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 				<p>If all items up to the current item in a set are present in the document structure, it is not necessary to set this <a>attribute</a>, as the <a>user agent</a> can automatically calculate the position for these items. However, if all previous items in the set are not present in the document structure at a given moment, the author MUST set this <a>attribute</a>to provide an explicit indication of an element's position.</p>
-				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of <code>aria-setsize</code> to <code>-1</code>.</p>
+				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown (indeterminate), and the DOM is currently incomplete such that user agents would not be able to compute the correct number of items in a set, authors SHOULD set the value of <code>aria-setsize</code> to a negative number (e.g., <code>-1</code>) or <code>0</code>.</p>
```
aleventhal, line 12910: We probably should just keep recommending -1 because some implementations may not update soon, and -1 was previously the way.

That doesn't mean the AAM wouldn't handle 0, but there's no reason to give authors extra options here.

scottaohara, line 12910:
```suggestion
				<p>When specifying <code>aria-setsize</code>, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown (indeterminate), and the DOM is currently incomplete such that user agents would not be able to compute the correct number of items in a set, authors SHOULD set the value of <code>aria-setsize</code> to a <code>-1</code>.</p>
```

core-aam/index.html
```diff
@@ -9852,8 +9852,15 @@ <h3>Group Position</h3>
...
-            property is not computed or not supported. If any of these properties are specified by the author as either &quot;0&quot; or a negative number, [=user agents=] SHOULD use &quot;1&quot;
-            instead.
+            property is not computed or not supported.
+          </p>
+          <p>
+            If the <a class="property-reference" href="#aria-level"><code>aria-level</code></a> or <a class="property-reference" href="#aria-posinset"><code>aria-posinset</code></a> properties are
+            specified by the author as either &quot;0&quot; or a negative number, [=user agents=] SHOULD use &quot;1&quot; instead.
+          </p>
+          <p>
+            If the <a class="property-reference" href="#aria-setsize"><code>aria-setsize</code></a> property is specified by the author as any string value other than a positive whole number, the set
```
aleventhal, line 9862: actually, "" is not indeterminate either.

aleventhal (changes requested):

core-aam/index.html
scottaohara: yeah, that's covered in another part of the spec.  guess this should have said "non-empty string value..."  i'll change it to that.

scottaohara, line 9862:
```suggestion
            If the <a class="property-reference" href="#aria-setsize"><code>aria-setsize</code></a> property is specified by the author as any non-empty string value other than a positive whole number, the set
```

aleventhal (approved):

cookiecrook (approved):

rahimabdi:
@aleventhal @scottaohara Would this be important enough to flag a console message?

aleventhal:
setsize=-1 is really rare in my experience. I've never actually seen an example with it.
I'm not sure what/when the console message would be, but my quick answer based on prevalence is that it's not worth any code in the browser engine.

spectranaut:
@adampage to investigate the status of browsers w.r.t. this change and file issues if necessary

spectranaut:
Discussed in today's meeting: https://www.w3.org/2025/05/15-aria-minutes.html#ffff

adampage:
Hey all, I’ve finished investigating this and have <a href="https://cdpn.io/pen/debug/pvvMpvy/4fbdc2bf82d22b5c94151314d76c782e">published some results</a>.

Some fun trivia:

- Webkit reported a set size of `184346744073798551615` for [some `aria-setsize="-1` tests specimens](https://cdpn.io/pen/debug/pvvMpvy/4fbdc2bf82d22b5c94151314d76c782e#:~:text=%E2%9D%8C-,Tomato%2C%20Tomato%2C%20(5%20of%2018%2C446%2C744%2C073%2C709%2C551%2C615),-%E2%9D%8CBullet%20Tomato%2C%20selected).

- `article` was the only role that exhibited no issues, seemingly on account of ignoring `aria-setsize` or `aria-posinset` altogether. Windows + Firefox + Narrator was [the one exception](https://cdpn.io/pen/debug/pvvMpvy/4fbdc2bf82d22b5c94151314d76c782e#articles-table-heading:~:text=Windows%2011%20%2B%20Narrator%20%2B%20Firefox,-OS).

- The main browser repair techniques for calculating set size seem to be 1) reporting the actual number of DOM elements within the group, or 2) accepting the highest value reported by the author via `aria-posinset`.

- Aside from repairing or misreporting the _set size_, many browser + AT combos also repaired or misreported each item’s stated position within the set.

The tl;dr is that every browser exhibits issues with both `aria-setsize="0"` and `aria-setsize="-1"` in relation to this PR. [`menuitem`](https://cdpn.io/pen/debug/pvvMpvy/4fbdc2bf82d22b5c94151314d76c782e#menu-items-table-heading), [`option`](https://cdpn.io/pen/debug/pvvMpvy/4fbdc2bf82d22b5c94151314d76c782e#listbox-options-table-heading), and [`radio`](https://cdpn.io/pen/debug/pvvMpvy/4fbdc2bf82d22b5c94151314d76c782e#radio-buttons-table-heading) had the most.

cc: @scottaohara @spectranaut

adampage:
The PR checklist now has [issue links for Webkit, Gecko, and Blink](https://github.com/w3c/aria/pull/2341#:~:text=Browser%20implementations%20%28link,org/issues/420965168) cc: @rahimabdi, @aleventhal, @jcsteh

rahimabdi:
An incredibly thorough investigation, thank you @adampage!

PR #2350, opened 2024-10-08
Clarifications for HTML's autocomplete and aria-autocomplete

### LINKED ISSUE #2337: Clarify `aria-autocomplete` and `autocomplete` semantic equivalency
# Describe your concern

I've been reading through the specs regarding these two attributes and can't clearly identify if `aria-autocomplete` is semantically equivalent to HTML's `autocomplete`. I see a significant number of implementations of ComboBoxes using `autocomplete="off"` to instruct the browser to not use it's native autofill functionality (despite sometimes not being respected [like in chrome](https://issues.chromium.org/issues/40093420)). They additionally use `aria-autocomplete="list"` on the same element to convey the usage/presence of a custom listbox.

Is it an acceptable practice to use both `aria-autocomplete` and `autocomplete` on a ComboBox and other elements/widgets, or should only one of these be used?

* https://www.w3.org/TR/wai-aria-1.2/#host_general_conflict
* https://www.w3.org/TR/wai-aria-1.2/#implicit_semantics
* https://www.w3.org/TR/core-aam-1.2/#ariaAutocompleteInlineListBoth

Yes

Related: https://github.com/w3c/aria-practices/issues/639

scottaohara:
Closes #2337

- updates `autocomplete` mapping for the `form` element (it doesn't map directly to the form element, but rather owned form controls will expose a supports autocompletion mapping, or not). 
- Calls out that aria-autocomplete is not supported on the form element or elements with role=form
- identify that it can be ok to use `aria-autocomplete=" [ valid value ] " ` and `autocomplete=off` together, but that for `autocomplete=" [ valid value ] "` and `aria-autocomplete="[ valid value ] "`, UA behavior should win out.

not sure how much more detail needs to be provided, as attempting to do so felt like treading into repeating what is already defined in each spec for these attributes (html / aria).

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:


html-aam/index.html
```diff
@@ -8599,39 +8599,56 @@ <h4 id="att-autocomplete">`autocomplete`</h4>
             <tr>
               <th>[[WAI-ARIA-1.2]]</th>
               <td>
-                <p><a class="core-mapping" href="#ariaAutocompleteInlineListBoth">`aria-autocomplete`</a></p>
-                <div class="note"><strong>Note:</strong> the ARIA attribute and the HTML attribute have disparate features.</div>
+                <div class="general">
+                  <p>If specivied `autocomplete=off` or the control is owned by a `form` with `autocomplete=off` - <a class="core-mapping" href="#ariaAutocompleteNone">`aria-autocomplete=none`</a></p>
```
MelSumner, line 8603:
```suggestion
                  <p>If specified `autocomplete=off` or the control is owned by a `form` with `autocomplete=off` - <a class="core-mapping" href="#ariaAutocompleteNone">`aria-autocomplete=none`</a></p>
```

MelSumner (approved):
Other than a suggested change for a spelling error, this looks okay to ship IMO.

cookiecrook (approved):

adampage (approved):
LGTM! 🚀  These clarifications are helpful.

spectranaut:
@scottaohara is this non-normative? Are you just updating to represent reality, or do browser needs implementations?

spectranaut:
@scottaohara @rahimabdi I think this can land, unless issues need to be opened on browsers?

PR #2354, opened 2024-10-11
Add command & commandfor attribute related mappings

keithamus:
This adds the mapping for `command` and `commandfor` attributes as per the [Invokers Explainer](https://open-ui.org/components/invokers.explainer/).

Additions:

 - [3.6.25 command (in the Toggle, Show, and Hide popover states)](https://deploy-preview-2354--wai-aria.netlify.app/html-aam/#att-command-popovers)
 - [3.6.26 command (in the Close and Show Modal states)](https://deploy-preview-2354--wai-aria.netlify.app/html-aam/#att-command-dialogs)
 - [3.6.27 commandfor](https://deploy-preview-2354--wai-aria.netlify.app/html-aam/#att-commandfor)

* [ ] "author MUST" tests:
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=269672
   * Gecko: 
   * Blink: https://chromium-review.googlesource.com/c/chromium/src/+/5942728
* [x] Does this need AT implementations? Yes
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

WebKit internal tests: https://searchfox.org/wubkat/source/LayoutTests/accessibility/mac/expanded-notification-commandfor-popover.html

Chrome internal tests: https://source.chromium.org/chromium/chromium/src/+/main:content/test/data/accessibility/html/commandfor-api-popover.html;l=1?q=commandfor.*api&sq= & https://source.chromium.org/chromium/chromium/src/+/main:content/test/data/accessibility/html/commandfor-api-dialog.html

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2354.html" title="Last updated on Mar 10, 2025, 10:29 AM UTC (5cf0bfe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2354/6d5a569...5cf0bfe.html" title="Last updated on Mar 10, 2025, 10:29 AM UTC (5cf0bfe)">Diff</a>


html-aam/index.html
```diff
@@ -9365,6 +9365,132 @@ <h4 id="att-controls">`controls`</h4>
...
+        <h4 id="att-command">`command`</h4>
+        <table aria-labelledby="att-command">
+          <tbody>
+            <tr>
+              <th>HTML Specification</th>
+              <td>`command`</td>
+            </tr>
+            <tr>
+              <th>Element(s)</th>
+              <td><a data-cite="html/popover.html#attr-command">`button`</a>;</td>
```
scottaohara, line 9377:
```suggestion
              <td><a data-cite="html/form-elements.html#attr-button-command">`button`</a>;</td>
```

html-aam/index.html
```diff
@@ -9365,6 +9365,132 @@ <h4 id="att-controls">`controls`</h4>
...
+        <h4 id="att-commandfor">`commandfor`</h4>
+        <table aria-labelledby="att-commandfor">
+          <tbody>
+            <tr>
+              <th>HTML Specification</th>
+              <td>`commandfor`</td>
+            </tr>
+            <tr>
+              <th>Element(s)</th>
+              <td><a data-cite="html/form-elements.html#attr-commandfor">`button`</a>;</td>
```
scottaohara, line 9426: dupe comment: 
was the support for this removed from "input" buttons ?

the html PR still mentions command / commandfor being allowed on input and button in the indices.html

html-aam/index.html
scottaohara, line 9377: was the support for this removed from "input" buttons ?

the html PR still mentions command / commandfor being allowed on input and button in the indices.html

html-aam/index.html
```diff
@@ -9365,6 +9365,132 @@ <h4 id="att-controls">`controls`</h4>
...
+            </tr>
+            <tr>
+              <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
+              <td>
+                <div class="general">Not mapped</div>
+              </td>
+            </tr>
+            <tr>
+              <th>Comments</th>
+              <td>The value of this attribute can impact the mappings of its related <a href="#att-commandfor">`commandfor`</a> attribute.</td>
```
scottaohara, line 9413: so the rational (flawed or not) popovertargetaction was listed as not having mappings, and instead adjusted the popovertarget attribute's mappings is because popovertarget was the constant for the button that invoked the popover.

with command/commandfor though, it doesn't seem like there would be a default action that would occur without a command value specified, and in such a case this attribute would have no additional mappings for the button?

where i'm going with this, is that if this were just going to ever have commands for popover/dialogs, then it'd probably make sense to stick with how this is spec'd now.

but since this has the potential to expand into different commands, where each of those commands might then need to update the mappings further, maybe it would end up being easier to spec if we treated this similar to how `input type` elements are spec'd, where each value (or collection of values?) for `type` is given its own mapping table.

what lead me to think about this is that the current commandfor mappings don't mention the "show-modal" value, and how all the mappings that are there for the popover related values need to go away for that one.


would this change in how the spec is written have an impact on how this has been implemented?   if so, then maybe there's just a different formatting to the spec that could be done to make this easier to maintain (and i can handle that in another PR down the line)

html-aam/index.html
```diff
@@ -9365,6 +9365,132 @@ <h4 id="att-controls">`controls`</h4>
...
+                  <p>
+                    If the element with the `commandfor` attribute has a `command` attribute with value of "`show-popover`", or "`toggle-popover`" and the associated element is hidden, or not present
+                    in the DOM:
+                    <a class="core-mapping" href="#ariaExpandedFalse">`aria-expanded=false`</a>
+                  </p>
+                  <p>
+                    If the element with the `commandfor` attribute has a `command` attribute with value of "`show-popover`", or "`toggle-popover`", and the associated element is an accessibility
+                    ancestor of the element with the `commandfor` attribute:
+                    <a class="core-mapping" href="#ariaExpandedUndefined">`aria-expanded=undefined`</a>
+                  </p>
```
scottaohara, line 9445: per mention in previous comment, missing expectations for commandfor + command='show-modal'

html-aam/index.html
```diff
@@ -9365,6 +9365,132 @@ <h4 id="att-controls">`controls`</h4>
...
+                  User Agents MUST expose an <a class="core-mapping" href="#ariaDetails">`aria-details`</a> relation with the associated element <strong>except</strong> under the following conditions:
+                </p>
+                <ul>
+                  <li>The element's `command` attribute value is a value other than "`show-popover`" or "`toggle-popover`",</li>
+                  <li>The associated popover element is the next immediate accessibility sibling to the invoking element,</li>
+                  <li>The element's `command` attribute is in the "`toggle-popover`" state and the element is a descendant of the `popover` it is associated with.</li>
+                  <!-- accessibility sibling will be defined in the ARIA specification -->
+                </ul>
+                <p class="note">A button that a [=button/form owner=] without `type` in the "`button`" state cannot invoke a command.</p>
+              </td>
```
scottaohara, line 9490: just another instance of where indicating expectations for show-modal would be needed (or not if they had their own tables, per my prior comment)

scottaohara (changes requested):
feel free to reach out if you have questions / want to talk through what i was suggesting

html-aam/index.html
keithamus: Yes, that sounds like an oversight in the spec. There was a decision to only support the attributes on the button element.

keithamus:
Okay @scottaohara I've expanded it as you described, I think it makes a lot more sense. Now each state has its own table which prescibes how that state should map. I've included all currently specified states, toggle/show/hide popover, close, and show modal.

scottaohara:
ok i think this can work.

we still need attribute tables for command and command for.  but as far as i'm concerned, those attribute tables could be simplified to just list the attribute, and point to the instances you made in the element section.

e.g., like how the [`type`](https://deploy-preview-2354--wai-aria.netlify.app/html-aam/#att-type-input) attribute for inputs is listed - but simpler... i could see that just removing all the individual mapping rows and in the comments, just link to the first of the button (command in the x state) tables you made.

(mental note to reduce redundancy and unnecessary page link by squashing all "not mapped" or "see ..." rows into a single comment row for other attributes too...)

html-aam/index.html
```diff
@@ -1257,6 +1257,328 @@ <h4 id="el-button">`button`</h4>
             </tr>
           </tbody>
         </table>
+        <h4 id="el-button-command-togglepopover">`button` <span class="el-context">(`command` attribute in the Toggle popover state)</span></h4>
```
aleventhal, line 1260: Is there a difference in what we expose for toggle vs show? Can we the combine mapping entries for those into one?

html-aam/index.html
```diff
@@ -1257,6 +1257,328 @@ <h4 id="el-button">`button`</h4>
...
+              <td>
+                <p>
+                  User Agents MUST expose an <a class="core-mapping" href="#ariaDetails">`aria-details`</a> relation with the associated element <strong>except</strong> under the following conditions:
+                </p>
+                <ul>
+                  <li>The associated popover element is the next immediate accessibility sibling to the invoking element,</li>
+                  <li>The element is a descendant of the `popover` it is associated with.</li>
+                  <!-- accessibility sibling will be defined in the ARIA specification -->
+                </ul>
+                <p class="note">A button that has a [=button/form owner=] without `type` in the "`button`" state cannot invoke a command.</p>
```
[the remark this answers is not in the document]
keithamus: Discussing further in https://github.com/whatwg/html/pull/9841 I think I misunderstood past decisions, but a form owner button with an implicit type should revert to button when it has the `command`/`commandfor` attributes so this is actually wrong, I'll fix.

html-aam/index.html
```diff
@@ -1257,6 +1257,328 @@ <h4 id="el-button">`button`</h4>
...
+                <a data-cite="html">`button`</a>
+                <span class="el-context">(<a data-cite="html/button.html#attr-input-command">`command`</a> attribute in the <a data-cite="html/input.html#attr-button-command-show-modal-state">Show modal</a> state)</span>
+              </td>
+            </tr>
+            <tr>
+              <th>[[WAI-ARIA-1.2]]</th>
+              <td><a class="core-mapping" href="#role-map-button">`button`</a> role.</td>
+            </tr>
+            <tr>
+              <th>
```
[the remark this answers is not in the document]
scottaohara: being a button that opens a modal dialog and where focus will automatically be put into the dialog once it's opened, there's no need for any additional mappings.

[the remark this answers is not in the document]
aleventhal: I probably need to apply the patch and view the HTML to make sure this fact stands out enough.

html-aam/index.html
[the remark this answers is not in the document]
keithamus: Okay I've fixed this by adjusting the note, and the HTML spec now specifically carves out some exceptions around when a button is a "submit button" which allows us to easily share the same note with popovers/command.

annevk:
https://github.com/whatwg/html/pull/9841 is close to ready. Is this expected to be merged soon @scottaohara & @aleventhal? Ideally we can merge both around the same time.

html-aam/index.html
```diff
@@ -9184,6 +9184,204 @@ <h4 id="att-colspan">`colspan`</h4>
...
+              <td>
+                <div class="general">See comments</div>
+              </td>
+            </tr>
+            <tr>
+              <th>Comments</th>
+              <td>
+                <p>The `commandfor` attribute identifies the associated element for the `button` element.</p>
+                <p>
+                  The specified `command` will indicate if a mapping needs to be exposed between the `button` and
```
scottaohara, line 9378:
```suggestion
                  The specified `command` state will determine if a relationship mapping needs to be exposed between the `button` and
```

scottaohara:
@keithamus i have taken some of aaron's feedback and reconfigured this a bit.

The related states are now their own mapping tables popover (toggle,show,hide) and dialog (close, show-modal).  These have been moved to the attribute section, rather than in the element section.  This was more just a partial re-consolidation and moving around, where the only major change I made was removed the mappings for aria-expanded for the close command.  Exposing that state should have been similar to a close button within a popover, where the expanded state isnt' necessary to communicate in that context. 

i added the missing `commandfor` table - but it's largely there to indicate that it serves as a way to associate the button and the element it commands - but again the mappings under the `command` states are where any programmatic associations that need to be exposed to users are defined.

cc @aleventhal @jcsteh @cookiecrook / @rahimabdi  for reviews for your platforms - but not much should be different here since these mappings were already similarly specified for `popover`/`popovertarget`.

I'm particularly curious if Webkit is looking to implement any of the aria-details relationships that chrome/firefox have implemented, and if anything more specific needs to be added to the ax row of the command=show,hide,toggle (or popovertarget for that matter)

scottaohara:
@annevk pending reviews, i think @keithamus did a good job getting this setup, and i spent parts of today finishing it up.
pinged reviewers in my previous comment.

cookiecrook (commented):
@muan and @nt1m may have some comments on this in the context of:
- https://github.com/whatwg/html/issues/10499

lukewarlow:
If https://github.com/whatwg/html/pull/11045 is merged this PR will need updating to handle the request-close command alongside the close command.

html-aam/index.html
```diff
@@ -9272,6 +9272,204 @@ <h4 id="att-colspan">`colspan`</h4>
...
+            </tr>
+            <tr>
+              <th>[[WAI-ARIA-1.2]]</th>
+              <td>
+                <p>
+                  If the associated element is displayed as a popover:
+                  <a class="core-mapping" href="#ariaExpandedTrue">`aria-expanded=true`</a>
+                </p>
+                <p>
+                  If the the associated element is hidden, or not present in the DOM:
```
lukewarlow, line 9304:
```suggestion
                  If the associated element is hidden, or not present in the DOM:
```

html-aam/index.html
```diff
@@ -9272,6 +9272,204 @@ <h4 id="att-colspan">`colspan`</h4>
...
+                  <a class="core-mapping" href="#ariaExpandedFalse">`aria-expanded=false`</a>
+                </p>
+                <p>
+                  If the the associated element is an accessibility ancestor of the element with the `command` attribute:
```
lukewarlow, line 9308:
```suggestion
                  If the associated element is an accessibility ancestor of the element with the `command` attribute:
```

html-aam/index.html
```diff
@@ -9272,6 +9272,204 @@ <h4 id="att-colspan">`colspan`</h4>
...
+              <th>Comments</th>
+              <td>
+                <p>
+                  User Agents MUST expose an <a class="core-mapping" href="#ariaDetails">`aria-details`</a> relationship with the associated element (identified via the specified `commandfor` attribute) <strong>except</strong> under the following conditions:
+                </p>
+                <ul>
+                  <li>The associated popover element is the next immediate accessibility sibling to the invoking element,</li>
+                  <li>The element is a descendant of the `popover` it is associated with.</li>
+                </ul>
+                <p class="note">A button that represents a <a data-cite="html/forms.html#concept-submit-button">submit button</a> with a [=button/form owner=] cannot invoke a command.</p>
```
lukewarlow, line 9351: The same is also true of reset buttons

html-aam/index.html
```diff
@@ -9272,6 +9272,204 @@ <h4 id="att-colspan">`colspan`</h4>
...
+                <span class="el-context">(`command` in the 
+                <a data-cite="html/input.html#attr-button-command-close-state">Close state</a>) and
+                <a data-cite="html/input.html#attr-button-command-show-modal-state">Show Modal state</a>)</span>
+              </td>
+            </tr>
+            <tr>
+              <th>[[WAI-ARIA-1.2]]</th>
+              <td>
+                <div class="general">
+                  A `command` attribute in the `close` and `show-modal` states provide no additional accessibility mappings to the `button` element.
```
lukewarlow, line 9378: cc @scottaohara is this deliberately omitting the case of a close button pointing to a non-modal dialog (where it *can* be outside of the dialog)?

html-aam/index.html
```diff
@@ -9270,6 +9270,197 @@ <h4 id="att-colspan">`colspan`</h4>
...
+            </tr>
+            <tr>
+              <th>[[WAI-ARIA-1.2]]</th>
+              <td>
+                <p>
+                  If the associated element is displayed as a popover:
+                  <a class="core-mapping" href="#ariaExpandedTrue">`aria-expanded=true`</a>
+                </p>
+                <p>
+                  If the associated element is hidden, or not present in the DOM:
```
lukewarlow, line 9300: To clarify what counts as "hidden" here. If it's pointed at a div that's not a popover, is that hidden?

I take this to mean the below but it would be good to clarify


commandfor="valid-open-popover" -> aria-expanded: true
commandfor="valid-closed-popover" -> aria-expanded: false
commandfor="not-in-dom" -> aria-expanded: false
commandfor="other-element" -> aria-expanded: false ?? Should this be undefined?

scottaohara: >To clarify what counts as "hidden" here. If it's pointed at a div that's not a popover, is that hidden?

[hidden](https://w3c.github.io/aria/#dfn-hidden) as informative defined by ARIA - which is why i'm not linking to it, because then i'd have normative text link to an informative definition and respec yells at me if i do that...

so for popovertarget or commandfor pointing to an element that doesn't have a popover attribute - i'd expect it to either not expose an expanded state at all, or undefined.  

actually, looking at current behavior for popovertarget referencing an element that isn't in the DOM, it appears that no browsers expose that in the false state... which thinking about it, is probably the right decision.  so this needs to be updated in the spec as saying it should be false doesn't match any implementation.

html-aam/index.html
scottaohara: this was deliberate.   if people can provide use cases where this is needed for real ui, then we can always reconsider.  but for now, the general idea would be that close buttons would commonly be within the dialog (modal or not), and if there was a control to close the non-modal dialog, outside of the dialog, it is more likely it'll be a toggle than a singular close button - or the close button might disappear from the DOM - or need to be marked as disabled (since you can't close a closed element), so exposing an expanded state here wouldn't necessarily be useful, either.

html-aam/index.html
```diff
@@ -9184,6 +9184,204 @@ <h4 id="att-colspan">`colspan`</h4>
...
+            <tr>
+              <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
+              <td>
+                <div class="general">Not mapped</div>
+              </td>
+            </tr>
+            <tr>
+              <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
+              <td>
+                <div class="general">Not mapped</div>
```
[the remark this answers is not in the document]
scottaohara: changed to tbd

lukewarlow:
(authoritative non-a11y person) LGTM

html-aam/index.html
```diff
@@ -9270,6 +9270,201 @@ <h4 id="att-colspan">`colspan`</h4>
             </tr>
           </tbody>
         </table>
+        <h4 id="att-command-popovers">`command` <span class="att-context">(in the Toggle, Show, and Hide popover states)</span></h4>
+        <table aria-labelledby="att-command-popovers">
```
pkra, line 9274:
```suggestion
        <table class="data" aria-labelledby="att-command-popovers">
```

html-aam/index.html
```diff
@@ -9270,6 +9270,201 @@ <h4 id="att-colspan">`colspan`</h4>
...
+                <p class="note">
+                  A button that represents a <a data-cite="html/forms.html#concept-submit-button">submit button</a> or is in the
+                  <a data-cite="html/input.html#reset-button-state-(type=reset)">reset state</a> with a [=button/form owner=] cannot invoke a command.
+                </p>
+              </td>
+            </tr>
+          </tbody>
+        </table>
+        <h4 id="att-command-dialogs">`command` <span class="att-context">(in the Close and Show Modal states)</span></h4>
+        <table aria-labelledby="att-command-dialogs">
```
pkra, line 9358:
```suggestion
        <table class="data" aria-labelledby="att-command-dialogs">
```

scottaohara (approved):

cookiecrook (approved):

pkra (approved):

accname/index.html
```diff
@@ -628,7 +628,8 @@ <h4>Computation steps</h4>
                 <li id="comp_host_language_label">
                   <span id="step2E"><!-- Don't link to this legacy numbered ID. --></span><em>Host Language Label:</em> Otherwise, if the <code>current node</code>'s native markup provides an
                   [=attribute=] (e.g. <code>alt</code>) or [=element=] (e.g. HTML <code>label</code> or SVG <code>title</code>) that defines a text alternative, return that alternative in the form of
-                  a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or <code>role="none"</code>).
+                  a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or
+                  <code>role="none"</code>).
```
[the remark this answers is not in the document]
scottaohara: simpler is better.  thanks

html-aam/index.html
```diff
@@ -9270,6 +9270,201 @@ <h4 id="att-colspan">`colspan`</h4>
...
+            <tr>
+              <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
+              <td>
+                <div class="general">Not mapped</div>
+              </td>
+            </tr>
+            <tr>
+              <th>Comments</th>
+              <td>
+                <p class="note">A button that represents a <a data-cite="html/forms.html#concept-submit-button">submit button</a> with a [=button/form owner=] cannot invoke a command.</p>
```
lukewarlow, line 9409: This bit is missing the additional reset button mention that was added for popover section above.

accname/index.html
```diff
@@ -639,8 +639,7 @@ <h4>Computation steps</h4>
                 <li id="comp_host_language_label">
                   <span id="step2E"><!-- Don't link to this legacy numbered ID. --></span><em>Host Language Label:</em> Otherwise, if the <code>current node</code>'s native markup provides an
                   [=attribute=] (e.g. <code>alt</code>) or [=element=] (e.g. HTML <code>label</code> or SVG <code>title</code>) that defines a text alternative, return that alternative in the form of
-                  a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or
-                  <code>role="none"</code>).
+                  a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (e.g., <code>role="none"</code>).
```
scottaohara, line 642: undoing this change as I'm thinking this might have been an accidental addition - since this update shouldn't have anything to do with the accname spec.
```suggestion
                 a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or
                  <code>role="none"</code>).
```

accname/index.html
```diff
@@ -639,7 +639,7 @@ <h4>Computation steps</h4>
                 <li id="comp_host_language_label">
                   <span id="step2E"><!-- Don't link to this legacy numbered ID. --></span><em>Host Language Label:</em> Otherwise, if the <code>current node</code>'s native markup provides an
                   [=attribute=] (e.g. <code>alt</code>) or [=element=] (e.g. HTML <code>label</code> or SVG <code>title</code>) that defines a text alternative, return that alternative in the form of
-                  a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or
+                 a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or
```
scottaohara, line 642:
```suggestion
                  a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or
```

scottaohara:
As this has landed in the html spec, is shipped in chrome/edge and is being implemented by other browsers, i'm merging this PR.

Issue #2361, opened 2024-10-24
Reconsider whether ARIA should ship new features that have incomplete API mappings and missing AT support.


cookiecrook:
Reconsider whether ARIA should ship new features that have incomplete API mappings and missing AT support.

This problem came to my attention due to some Interop 2024 failures in [the aria-attribute-reflection test](https://wpt.fyi/results/html/dom/aria-attribute-reflection.html?label=master&label=experimental&aligned&view=interop&q=label%3Ainterop-2024-accessibility) for aria-colindextext and aria-rowindextext, which are new attributes in ARIA 1.3

WebKit engineers chose not to implement these two because there is no API mapping for it on Mac, and no near-term expectation for AT support. As such, the other browsers on Mac are reflecting the DOM properties, but do nothing with them other than pass the WPT test. I have an [open WPT PR to remove those tests from Interop](https://github.com/web-platform-tests/interop-accessibility/issues/151), since they were added mid-year.

It's also not clear from the Core-AAM listings whether there is AT support, or whether the Windows API mappings are speculative for Windows screen readers. (If so, I have concern with that approach overall.) I'll file a separate issue for Core-AAM to do a better job of indicating whether there is known downstream AT support. That's a larger feature request for the spec that we should discuss as a group.

As a related datapoint, Apple has proposed several ARIA features that have support in Apple AT, that the group has been reluctant to ship since Windows ATs are missing related features. The following are a few examples of those ARIA proposals that have not shipped because no Windows AT has similar feature support yet:

- #1215
- #744

The group should discuss whether the same should apply to these.... Especially if it turns out there is no Windows or Linux AT support and these are just speculative API mappings.

spectranaut:
@joanmarie does Orca support these attributes, aria-colindextext and aria-rowindextext?

joanmarie:
> @joanmarie does Orca support these attributes, aria-colindextext and aria-rowindextext?

Yup, it has for quite some time. It also does a fallback check for `aria-coltext` and `aria-rowtext` because (last time I checked) Google Sheets still exposes that instead of the (imho) more-clearly-named versions in the spec.

As a related aside, LibreOffice Calc followed the ARIA lead and exposes the row and column labels/indices via IA2 and ATK using the `colindextext` and `rowindextext` object attributes (this is how the ARIA properties are expected to be exposed when used in web content).

stes-acc:
The practical thing about rowindextext is that you can use it to declare lead object names for a row.

[...]

PR #2362, opened 2024-10-25
[core-aam] add new AX API mappings for colindextext/rowindextext

cookiecrook:
Closes https://github.com/w3c/core-aam/issues/53

* [x] "author MUST" tests: n/a
* [x] "user agent MUST" tests: 
   * [x] not testable in WPT, so n/a
   * [ ] @spectranaut may wish to add platform mapping tests
   * [ ] WebKit: [#257142](https://webkit.org/b/257142)
   * [ ] Gecko: [#1931415](https://bugzilla.mozilla.org/show_bug.cgi?id=1931415)
   * [x] Blink: [#378698863](https://issues.chromium.org/issues/378698863)
* Does this Mac-specific change need AT implementation? 
   * VoiceOver change is not prioritized b/c there is not a platform equivalent
* [x] Related APG Issue/PR: n/a
* [x] MDN Issue/PR: n/a

### DIFF
--- a/core-aam/index.html
+++ b/core-aam/index.html
@@ -6372,7 +6372,7 @@
               <tr>
                 <th><abbr title="macOS Accessibility Protocol">AX API</abbr></th>
                 <td>
-                  <span class="property">Property: TBD</span>
+                  <span class="property">Property: <code>AXARIAColumnIndexText</code>: <code>&lt;value&gt;</code></span>
                 </td>
               </tr>
             </tbody>
@@ -9209,7 +9209,7 @@
               <tr>
                 <th><abbr title="macOS Accessibility Protocol">AX API</abbr></th>
                 <td>
-                  <span class="property">Property: TBD</span>
+                  <span class="property">Property: <code>AXARIARowIndexText</code>: <code>&lt;value&gt;</code></span>
                 </td>
               </tr>
             </tbody>
--- a/core-aam/index.html
+++ b/core-aam/index.html
@@ -6372,7 +6372,7 @@
               <tr>
                 <th><abbr title="macOS Accessibility Protocol">AX API</abbr></th>
                 <td>
-                  <span class="property">Property: <code>AXARIAColumnIndexText</code>: <code>&lt;value&gt;</code></span>
+                  <span class="property">Property: <code>AXColumnIndexDescription</code>: <code>&lt;value&gt;</code></span>
                 </td>
               </tr>
             </tbody>
@@ -9209,7 +9209,7 @@
               <tr>
                 <th><abbr title="macOS Accessibility Protocol">AX API</abbr></th>
                 <td>
-                  <span class="property">Property: <code>AXARIARowIndexText</code>: <code>&lt;value&gt;</code></span>
+                  <span class="property">Property: <code>AXRowIndexDescription</code>: <code>&lt;value&gt;</code></span>
                 </td>
               </tr>
             </tbody>


cookiecrook:
@twilco @minorninth FYI

core-aam/index.html
```diff
@@ -6372,7 +6372,7 @@ <h4 id="ariaColIndexText"><code>aria-colindextext</code></h4>
               <tr>
                 <th><abbr title="macOS Accessibility Protocol">AX API</abbr></th>
                 <td>
-                  <span class="property">Property: TBD</span>
+                  <span class="property">Property: <code>AXARIAColumnIndexText</code>: <code>&lt;value&gt;</code></span>
```
cookiecrook, line 6375: @twilco et al discussing whether we should double-down on having "ARIA" in the property name just because that was added to the related `AXARIAColumnIndex` property.... We may deprecate the ARIA-specific names in the future anyway.
```suggestion
                  <span class="property">Property: <code>AXColumnIndexText</code>: <code>&lt;value&gt;</code></span>
```

core-aam/index.html
```diff
@@ -9209,7 +9209,7 @@ <h4 id="ariaRowIndexText"><code>aria-rowindextext</code></h4>
               <tr>
                 <th><abbr title="macOS Accessibility Protocol">AX API</abbr></th>
                 <td>
-                  <span class="property">Property: TBD</span>
+                  <span class="property">Property: <code>AXARIARowIndexText</code>: <code>&lt;value&gt;</code></span>
```
cookiecrook, line 9212: ditto.
```suggestion
                  <span class="property">Property: <code>AXRowIndexText</code>: <code>&lt;value&gt;</code></span>
```

jnurthen:
@aleventhal @jcsteh @spectranaut We'd like to merge this before the CR snapshot - can we get this reviewed as soon as possible please. Thanks.

aleventhal (approved):

spectranaut (commented):
@cookiecrook it seems like this new mapping is not implemented in any other the browsers, from a quick code search?

I added the implementation tracking, because it seems like we should be doing implementation tracking for this CL.  Before it lands we should have issues on the browsers to implement.

Because this still needs implementation, should it really block the CORE-AAM snapshot? Perhaps if you really want it in, we can at least make issues on the browsers before merging.

aleventhal:
Funny you mention that, I'm working on a CL for it -- doesn't work yet:
https://chromium-review.googlesource.com/c/chromium/src/+/6012688

cookiecrook:
Yes, it's new. So that ARIA WG doesn't have to publish a "TBD" in the spec.

cookiecrook:
> should it really block the CORE-AAM snapshot? 

Otherwise I would expect it to be pulled or marked as At-Risk because this is a new feature without AT support on one platform, and without preceding vendor commitment. I outlined the concerns at:

- https://github.com/w3c/aria/issues/2361

cookiecrook:
> Perhaps if you really want it in, we can at least make issues on the browsers before merging.

I will link the WebKit tracker. Sounds like Aaron has one for Blink already. I hesitate to file one for Gecko without acknowledgement from @jcsteh or @nmlapre that they want to do the mapping work before there is AT commitment.

spectranaut:
> I have linked the WebKit tracker and the one Aaron filed for Blink. I hesitate to file one for Gecko without acknowledgement from @jcsteh or @nmlapre that they want to do the mapping work before there is AT commitment.

Sometimes I open issues on gecko and link back to the ARIA PR when I want to get review from firefox :) Personally I think it's fine to open an issue in this case, I don't think this change is controversial and besides it is coming from the API maintainers of AX API.

spectranaut:
Discussed in today's ARIA working group meeting: https://www.w3.org/2024/11/14-aria-minutes.html#8070

We will add this to the AAM, and open issues on browsers, as eventually, probably, VoiceOver will do something with this value.

spectranaut (approved):
Per working group discussion, merging.

PR #2369, opened 2024-11-01
Update: add mapping tables for customizable select

scottaohara:
Initial draft of select / button as child of a select mappings.

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -1257,6 +1257,79 @@
             </tr>
           </tbody>
         </table>
+
+        <h4 id="el-button-select-child">`button` as a child of a `select` element</h4>
+        <table aria-labelledby="el-button-select-child">
+          <tbody>
+            <tr>
+              <th>HTML Specification</th>
+              <td>
+                <a data-cite="HTML">`button`</a>
+              </td>
+            </tr>
+            <tr>
+              <th>[[wai-aria-1.2]]</th>
+              <td>
+                <div class="general"><a href="#el-button-select-child-comments">See comments</a>.</div>
+                <!-- a button that is a child of a select is not a typical "button" element.
+                     it does not serve as a submit, reset, or generic 'button' type. Rather,
+                     its purpose is directly tied to the rendering of a customized select
+                     serving as the visible indicator/styleable node to serve as the 
+                     invoking element for the select's listbox popup.
+                -->
+              </td>
+            </tr>
+            <tr>
+              <th><a data-cite="core-aam-1.2/#roleMappingComputedRole">Computed Role</a></th>
+              <td class="role-computed">
+                <div class="general"><a href="#el-button-select-child-comments">See comments</a>.</div>
+              </td>
+            </tr>
+            <tr>
+              <th>
+                <a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a>
+              </th>
+              <td>
+sdfsddsaf
+              </td>
+            </tr>
+            <tr>
+              <th><a href="https://msdn.microsoft.com/en-us/library/ms726297%28v=VS.85%29.aspx">UIA</a></th>
+              <td>
+                <div class="general">Use WAI-ARIA mapping</div>
+              </td>
+            </tr>
+            <tr>
+              <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
+              <td>
+                <div class="general">Use WAI-ARIA mapping</div>
+              </td>
+            </tr>
+            <tr>
+              <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
+              <td>
+                <div class="general">Use WAI-ARIA mapping</div>
+              </td>
+            </tr>
+            <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
+            <tr>
+              <th id="el-button-select-child-comments">Comments</th>
+              <td>
+                <p>User agents MUST treat a `button` that is a child element of a `select` element as one item  in the accessibility tree.</p>
+                <p>
+                  User agents MUST use the <a data-cite="accname-1.2/#mapping_additional_nd_te">text equivalent computation</a> of the subtree of the
+                  `button` element as the exposed value of the `select` element. If the `button` contains no content, then there is no value to expose 
+                  for the collapsed state of the `select`.
+                </p>
+                <p class=note>
+                  TBD: indicate that global html/aria attributes and html/aria attributes specific to buttons are not to be respected on a button that is a child of a select?  Since the button/select are meant to be treated as one entity in the a11y tree, and the button's content is meant to be treated as the value for the select, and not the name of the select, it doesn't make much sense to treat this button as if it were actually a button... 
+                </p>
+                <p>See also <a href="#el-select-combobox">`select`<span class="el-context"> (with NO `multiple` attribute and NO `size` attribute having value greater than `1`)</span></a></p>
+              </td>
+            </tr>
+          </tbody>
+        </table>
+        
         <h4 id="el-canvas">`canvas`</h4>
         <table aria-labelledby="el-canvas">
           <tbody>
@@ -6339,7 +6412,10 @@
             <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
             <tr>
               <th>Comments</th>
-              <td></td>
+              <td>
+                <p>User agents MUST treat a `button` that is a child element of a `select` element as one item  in the accessibility tree.</p>
+                <p>See also <a href="#el-button-select-child">`button` element as a child of a `select` element.</p>
+              </td>
             </tr>
           </tbody>
         </table>


html-aam/index.html
```diff
@@ -1257,6 +1257,79 @@ <h4 id="el-button">`button`</h4>
...
+              <th id="el-button-select-child-comments">Comments</th>
+              <td>
+                <p>User agents MUST treat a `button` that is a child element of a `select` element as one item  in the accessibility tree.</p>
+                <p>
+                  User agents MUST use the <a data-cite="accname-1.2/#mapping_additional_nd_te">text equivalent computation</a> of the subtree of the
+                  `button` element as the exposed value of the `select` element. If the `button` contains no content, then there is no value to expose 
+                  for the collapsed state of the `select`.
+                </p>
+                <p class=note>
+                  TBD: indicate that global html/aria attributes and html/aria attributes specific to buttons are not to be respected on a button that is a child of a select?  Since the button/select are meant to be treated as one entity in the a11y tree, and the button's content is meant to be treated as the value for the select, and not the name of the select, it doesn't make much sense to treat this button as if it were actually a button...
```
scottaohara, line 1325:
```suggestion
                  TBD: indicate that global html/aria attributes and html/aria attributes specific to buttons are not to be respected on a button that is a child of a select?  Since the button/select are meant to be treated as one entity in the a11y tree, and the button's content is meant to be treated as the value for the select, and not the name of the select, it doesn't make much sense to treat this button as if it were actually a button... 
            <br>
            <br>
            suggest ignoring accesskey, autofocus, contenteditable, draggable, inert, popover, tabindex, title
```

html-aam/index.html
```diff
@@ -6339,7 +6416,31 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
...
+                  <li>If the `select` element does not have an `aria-valuetext` attribute and does not contain a `button` part, then the accessible value is calculated from the accessible name of the selected `option` element.</li>
+                </ol>
+                <!-- provide a link to the ARIA spec when the https://github.com/w3c/aria/pull/2417 PR lands --> 
+
+                <p class=note>The following text is still being refined</p>
+                <p>
+                  The calculated role of the customizable `select` element's popup picker is a `listbox`, by default.</p>
+                <p>
+                  If authors render invalid elements, in regard to the select element's content model, as descendants of the select element, and those invalid elements result in accessible objects being present as siblings in the accessibility tree to the valid descendent elements of the select, then user agents SHOULD expose the popup picker with the role of a modeless dialog.</p>
```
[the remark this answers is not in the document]
scottaohara: thanks @mfreed7 - role of dialog in the modeless state is really what i should have said there.

for the first comment, i said "render" because i would rather flags not be raised / repairs not be made if someone includes an invalid element, but it is set to display: none.  but i'm fine with your suggested wording change as well.

html-aam/index.html
```diff
@@ -1291,6 +1291,83 @@ <h4 id="el-button">`button`</h4>
...
+              <td>
+                <div class="general"><a href="#el-button-select-child-comments">See comments</a></div>
+              </td>
+            </tr>
+            <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
+            <tr>
+              <th id="el-button-select-child-comments">Comments</th>
+              <td>
+                <p>
+                  User agents MUST treat the child `button` element of a `select` element as inert. For instance, the UA stylesheet applying <span data-x="css-interactivity">'interactivity: inert'</span> to the button part.
```
scottaohara, line 1353: betting this xref was causing the build fail?

```suggestion
                  User agents MUST treat the child `button` element of a `select` element as inert. For instance, the UA stylesheet applying 'interactivity: inert' to the button part.
```

html-aam/index.html
```diff
@@ -6427,7 +6504,47 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
...
+
+                  <pre><code>
+&lt;select>
+  <button><selectedcontent></selectedcontent></button>
+  <article></article>
+  <article>
+    <option>item 1
+    <option>item 2
+  </article>
+  <article> </article>
```
zcorpan, line 6539: Escape `<`

html-aam/index.html
```diff
@@ -1291,6 +1291,83 @@ <h4 id="el-button">`button`</h4>
...
+              <td>
+                <div class="general"><a href="#el-button-select-child-comments">See comments</a></div>
+              </td>
+            </tr>
+            <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
+            <tr>
+              <th id="el-button-select-child-comments">Comments</th>
+              <td>
+                <p>
+                  User agents MUST treat the child `button` element of a `select` element as inert. For instance, the UA stylesheet applying 'interactivity: inert' to the button part.
```
zcorpan, line 1353: I think this shouldn't be a requirement, but a statement of fact, referencing HTML's rendering section. (If authors override the UA style, is the UA violating _this_ requirement?)

zcorpan (changes requested):

html-aam/index.html
scottaohara, line 6539:
```suggestion
  &lt;button>&lt;selectedcontent>&lt;/selectedcontent>&lt;/button>
  &lt;article></article>
  &lt;article>
    &lt;option>item 1
    &lt;option>item 2
  &lt;/article>
  &lt;article> &lt;/article>
```

html-aam/index.html
scottaohara, line 1353:
```suggestion
                  User agents treat the child `button` element of a `select` element as inert. 
```

html-aam/index.html
```diff
@@ -6427,7 +6504,47 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
...
+              <td>
+                <p>To calculate the accessible value of a `select` element:</p>
+                <ol>
+                  <li>If `aria-valuetext` is specified on the `select` element, use the attribute's value, including the empty string.</li>
+                  <li>Else if there is a `button` part, calcluate the accessible value from the <a data-cite="accname-1.2/#mapping_additional_nd_te">text equivalent computation</a> of the subtree of that element, even if it returns the empty string.</li>
+                  <li>If the `select` element does not have an `aria-valuetext` attribute and does not contain a `button` part, then the accessible value is calculated from the accessible name of the selected `option` element.</li>
+                </ol>
+                <!-- provide a link to the ARIA spec when the https://github.com/w3c/aria/pull/2417 PR lands --> 
+
+                <p class=note>The following text is still being refined</p>
```
scottaohara, line 6516: TODO

html-aam/index.html
```diff
@@ -6427,7 +6504,47 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
...
+                  <li>If the `select` element does not have an `aria-valuetext` attribute and has not calculated a usable value from its button part, then the accessible value is calculated from the accessible name of the selected `option` element, even if the accessible name of the selected `option` is the empty string.</li>
+                </ol>
+                <!-- provide a link to the ARIA spec when the https://github.com/w3c/aria/pull/2417 PR lands --> 
+
+                <p class=note>The following text is still being refined</p>
+                <p>
+                  The calculated role of the customizable `select` element's popup picker is a `listbox`, by default.</p>
+                <p>
+                  If authors include invalid elements (according to the the `select` element's content model) as descendants of the `select` element, and those invalid elements result in accessible objects being present as siblings in the accessibility tree to the valid descendent elements of the select, then user agents SHOULD expose the popup picker with the role dialog in the modeless state.</p>
```
scottaohara, line 6520: provide direct link to the select element in HTML, here

html-aam/index.html
```diff
@@ -6427,7 +6504,47 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
...
+                  In the event that invalid elements are present, but are empty, include only whitespace, or are only used as wrapping elements for the valid descendent element's of a select, then user agents SHOULD ignore these elements in the accessibility tree.
+                </p>
+
+                <div class=example>
+                  <p>In the following example the declared `article` elements are invalid per the `select` element's content model. But as each `article` element is empty, contains only whitespace, or contains only valid `option` elements, the `article` element's role can be ignored in the user agent's accessibility tree:</p>
+
+                  <pre><code>
+&lt;select>
+  &lt;button>&lt;selectedcontent>&lt;/selectedcontent>&lt;/button>
+  &lt;article></article>
```
scottaohara, line 6534:
```suggestion
  &lt;article></article> <!-- empty so can be ignored -->
```

html-aam/index.html
```diff
@@ -6427,7 +6504,47 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
...
+  &lt;article>
```
scottaohara, line 6535:
```suggestion
  &lt;article> <!-- only contains option elements, so article role can be ignored -->
```

html-aam/index.html
```diff
@@ -6427,7 +6504,47 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
...
+                  In the event that invalid elements are present, but are empty, include only whitespace, or are only used as wrapping elements for the valid descendent element's of a select, then user agents SHOULD ignore these elements in the accessibility tree.
+                </p>
+
+                <div class=example>
+                  <p>In the following example the declared `article` elements are invalid per the `select` element's content model. But as each `article` element is empty, contains only whitespace, or contains only valid `option` elements, the `article` element's role can be ignored in the user agent's accessibility tree:</p>
+
+                  <pre><code>
+&lt;select>
+  &lt;button>&lt;selectedcontent>&lt;/selectedcontent>&lt;/button>
+  &lt;article> &lt;/article> <!-- empty so can be ignored -->
```
scottaohara, line 6534:
```suggestion
  &lt;article> &lt;/article> <!-- empty / containing only white space so can be ignored -->
```

html-aam/index.html
```diff
@@ -6427,7 +6504,47 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
...
+              <td>
+                <p>To calculate the accessible value of a `select` element:</p>
+                <ol>
+                  <li>If `aria-valuetext` is specified on the `select` element, use the attribute's value, including the empty string.</li>
+                  <li>Else if there is a `button` part, calculate the accessible value from a non-empty string <a data-cite="accname-1.2/#mapping_additional_nd_te">text equivalent computation</a> of the subtree of that element.</li>
+                  <li>If the `select` element does not have an `aria-valuetext` attribute and has not calculated a usable value from its button part, then the accessible value is calculated from the accessible name of the selected `option` element, even if the accessible name of the selected `option` is the empty string.</li>
+                </ol>
+                <!-- provide a link to the ARIA spec when the https://github.com/w3c/aria/pull/2417 PR lands --> 
+
+                <p class=note>The following text is still being refined</p>
```
scottaohara, line 6516:
```suggestion
```

html-aam/index.html
```diff
@@ -6428,7 +6505,46 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
...
+    &lt;option>item 2
+  &lt;/article>
+  &lt;article hidden> foo &lt;/article> <!-- article is in the hidden state, so is ignored -->
+&lt;/select>
+                  </code></pre> 
+                </div>
+
+                <p class=note>Treating the picker as a `dialog` can help inform users that content beyond the expected elements of a `listbox` popup are present, as they can then inspect the content of the dialog with their assistive technology to discover such content.</p>
+
+                <p>See also <a href="#el-button-select-child">`button` element as a child of a `select` element.</p>
```
giacomo-petri, line 6538: For better readability, should we move (or duplicate) the link to the 'button as a child of a select element' section closer to the second bullet point in the accessible name computation? It might help readers understand, for example, why an aria-label set on the button doesn’t contribute to the accessible name computation.

html-aam/index.html
```diff
@@ -6428,7 +6505,46 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
...
+                  <li>If the `select` element does not have an `aria-valuetext` attribute and has not calculated a usable value from its button part, then the accessible value is calculated from the accessible name of the selected `option` element, even if the accessible name of the selected `option` is the empty string.</li>
+                </ol>
+                <!-- provide a link to the ARIA spec when the https://github.com/w3c/aria/pull/2417 PR lands --> 
+
+                <p>
+                  The calculated role of the customizable `select` element's popup picker is a `listbox`, by default.</p>
+                <p>
+                  If authors include invalid elements (according to the the `select` element's content model) as descendants of the `select` element, and those invalid elements result in accessible objects being present as siblings in the accessibility tree to the valid descendent elements of the select, then user agents SHOULD expose the popup picker with the role of `dialog` in the modeless state.</p>
```
zcorpan, line 6512: Implementation requirements that depend on content model are unusual. Browsers typically don't include conformance checkers, so can't easily tell if the content model is being violated or not. What is the purpose here?

scottaohara: thanks for pointing this out @zcorpan 

the rational here was talked about a lot in open ui, but the high level gist is that since the parsing for select was changed to allow for anything, the role of the popup needs to let users know whether they're interacting with a version that follows the content model (and thus can be exposed as a listbox) or has elements in it which are important to make AT users aware of / doesn't follow the content model, and thus was decided to be exposed as a dialog to help surface this to users.

with that said though, i talked with @mfreed7 about this and we were thinking that to mitigate this explicit reference to the content model, we this can be updated to more overtly call out what HTML elements / allowed implicit roles will allow for the default role=listbox, and how the inclusion of others would result in the popup needing to be role=dialog.

html-aam/index.html
```diff
@@ -6420,7 +6497,46 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
             <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
             <tr>
               <th>Comments</th>
-              <td></td>
+              <td>
+                <p>To calculate the accessible value of a `select` element:</p>
+                <ol>
+                  <li>If `aria-valuetext` is specified on the `select` element, use the attribute's value, including the empty string.</li>
+                  <li>Else if there is a `button` part, calculate the accessible value from a non-empty string <a data-cite="accname-1.2/#mapping_additional_nd_te">text equivalent computation</a> of the subtree of that element.</li>
```
jcsteh, line 6504: Is there anything in the HTML spec that prevents the author from putting something other than the value in the `button` part? For example, what if the author puts a label in the `button` part? It would be very weird to expose a label as the value. Obviously, we can't stop authors from doing this, but I wonder if they should at least be warned against it, though that probably belongs in the HTML spec itself.

html-aam/index.html
jcsteh: While I understand why it's being proposed, this is the part of customisable select accessibility that worries me the most.

Mutations are a big concern here. What if the `<select>` started out entirely valid when it was first rendered, but then an author added more elements such that it became invalid? Or on the flip side, what if it were invalid and then became valid? Since we're effectively talking about UA repair for invalid authoring, it's not super important *how* we fail, but the failure mode needs to be clearly defined. This can't be undefined (unspecified) behaviour if we want interoperable implementations that don't later cause web compat pain.

Perhaps we could state that the role is determined based on the initial rendering of the `<select>`. If it starts out as valid and an author later makes it invalid, that's their loss. We'll only correct for authoring error to a point. I think this would be the most pragmatic solution and would be the least likely to cause implementation pain and performance problems.

If we do want to handle mutations from valid to invalid and vice versa, there are some difficult issues we need to address:

1. Most accessibility APIs don't have role change events (or at least not role change events that clients heed). Should the role just mutate silently, risking confusing the client, or should the subtree be recreated whenever validity changes? Recreating subtrees is expensive all the way through the accessibility stack, particularly if there are a lot of nodes. Even mutating the role silently has performance costs; see below.
2. Checking whether validity has changed is not trivial. Unless I'm missing something, we'd need to re-evaluate this every time a descendant is added to a select. At the very least, this probably means walking the ancestry for every insertion or removal, since caching a select ancestor on every node would waste a lot of memory. Anything that adds even the tiniest bit of work for all insertions into the tree, even if it's just an ancestor walk with a simple check, is a potential performance cliff. This possibly also means walking all the non-generic children of the select as well, though we could maybe get around this caching the validity state on the select. In general, any specified behaviour here needs to take performance concerns into account.

CC @eeejay for awareness and input, since he will likely be doing the Gecko implementation for this and has also done some recent work regarding role calculation. CC also @lucasradaelli from Chrome accessibility, since aleventhal is no longer working on that.

html-aam/index.html
```diff
@@ -6420,7 +6497,46 @@ <h4 id="el-select-combobox">`select`<span class="el-context"> (with NO `multiple
...
+
+                <p>
+                  The calculated role of the customizable `select` element's popup picker is a `listbox`, by default.</p>
+                <p>
+                  If authors include invalid elements (according to the the `select` element's content model) as descendants of the `select` element, and those invalid elements result in accessible objects being present as siblings in the accessibility tree to the valid descendent elements of the select, then user agents SHOULD expose the popup picker with the role of `dialog` in the modeless state.</p>
+
+                <p>
+                  In the event invalid elements are present, but are not rendered, then no user agent role repair is necessary.</p>
+                <p>
+                  In the event that invalid elements are present, but are empty, include only whitespace, or are only used as wrapping elements for the valid descendent element's of a select, then user agents SHOULD ignore these elements in the accessibility tree.
```
jcsteh, line 6517: This is even more concerning. It's probably feasible to do this for empty elements in the initially rendered subtree. It's slightly trickier (but probably still feasible) to do this when new empty elements are inserted into an otherwise valid select. But what if someone inserts an empty `<article>` into a `<select>` (so it's ignored), but later inserts a child that makes the `<article>` invalid? Now we're not just making decisions based on the type of the element. The decisions also have to take into account the contents of that element. That means we're doing even more walk on every insertion into the tree.

Also, does "include only whitespace" explicitly refer to direct children or is this suggesting that we need to evaluate the contents overall. For example, does `<article><span> <span> </span></span></article>` get ignored or not? What about `<article>&nbsp</article>`? I'm guessing this is only intended to cover cases like `<article> </article>`, but this probably needs to be clarified. Even just saying "includes only a single whitespace child" is probably sufficient clarification.

html-aam/index.html
eeejay: I agree that this is a pain point.

eeejay: In the hypothetical here that the select has a mix of invalid elements and `option` elements, the role would be a dialog. But I am unclear of how a platform API or AT would deal with `option` accessibles that are children of `dialog` and have no `listbox` ancestor.

eeejay: My humble proposal:
* If the `select` initially has a valid content model, and then an invalid child is added, it should remain a `listbox` with unspecced behavior, just like how a non-`group`/`option` child would affect an explicit aria `listbox`.
* If the `select` initially has an invalid content model its role would be a dialog. All `select` children would get generic roles. The `dialog` role is immutable and won't change to a `listbox` if the invalid children are removed.

html-aam/index.html
eeejay: I'm not sure why empty invalid tags are an exception here. What is the use-case?

In my proposal above, I could see an empty invalid tag as a method of insuring the picker gets a role of dialog even if there is no initial content, eg:
```html
<select>
  <button>
    <selectedcontent>Select a pet:</selectedcontent>
  </button>
  <article></article>
</select>
```

PR #2235, opened 2024-05-31
Empty or non-matching values for ID reference and ID reference list - Clarification on author responsibilities

giacomo-petri:
Closes: #2225

Clarification on author responsibilities concerning empty or invalid references for properties accepting ID reference (or ID reference list) as values. 

The intention is, considering the dynamic nature of contemporary websites, to allow authors to set empty or "invalid" values, acknowledging the DOM can be populated accordingly when necessary.

PR #2373, opened 2024-11-08
Update: clarify / expand value attribute mappings

scottaohara:
replacement for https://github.com/w3c/html-aam/pull/396

closes https://github.com/w3c/html-aam/issues/314

This should largely be reflecting what is currently happening in browsers.

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -7951,6 +7951,9 @@
             </div>
           </li>
           <li>All elements having an accessible object in IAccessible2 mapping are supposed to implement IAccessible, IAccessible2 and IAccessible2_2 interfaces.</li>
+          <li>
+            User interaction with certain HTML elements will set an element's <a data-cite="html/form-control-infrastructure.html#concept-fe-dirty">dirty value flag</a>. For these elements which are identified in [HTML], if specified such attributes represent only the default value or state for the element. 
+          </li>
         </ul>
         <h4 id="att-abbr">`abbr`</h4>
         <table aria-labelledby="att-abbr">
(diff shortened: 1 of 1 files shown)


MelSumner (approved):
:shipit:

html-aam/index.html
```diff
@@ -15770,46 +15773,232 @@ <h4 id="att-value-input">`value`</h4>
...
+              <td><div class="general">See comments</div></td>
+            </tr>
+            <tr>
+              <th>Comments</th>
               <td>
-                <div class="name">
-                  Associates the accessible value for entry type input elements and <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> for button type input elements
-                </div>
+                If specified, the value of the attribute will contribute to the
+                <a href="#input-type-button-input-type-submit-and-input-type-reset-accessible-name-computation">accessible name computations</a> for these `input` elements in the `button`, `reset` and `submit` states, rendering as the button `input` element's text label.
```
cookiecrook: I think "contribute to" may be vague since it is not defined elsewhere in the doc?

```suggestion
                If specified, the value of the attribute will be the <a data-cite="accname-1.2/#comp_host_language_label">host language label</a> used in the <a href="#input-type-button-input-type-submit-and-input-type-reset-accessible-name-computation">accessible name computation</a> for `input` elements in the `button`, `reset` and `submit` states.
```

html-aam/index.html
```diff
@@ -15770,46 +15773,232 @@ <h4 id="att-value-input">`value`</h4>
...
+            </tr>
+            <tr>
+              <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
+              <td><div class="general">See comments</div></td>
+            </tr>
+            <tr>
+              <th>Comments</th>
+              <td>
+                If specified, and the `input` in the `image` state has no `alt` attribute specified, then the value of the attribute will contribute to the
+                <a href="#input-type-button-input-type-submit-and-input-type-reset-accessible-name-computation">accessible name computations</a>, and will render as text if the image source is broken. Otherwise, the attribute is ignored.
```
cookiecrook: ditto
```suggestion
                If specified, and the `input` in the `image` state has no `alt` attribute specified, then the value of the attribute will be the <a data-cite="accname-1.2/#comp_host_language_label">host language label</a> used in the <a href="#input-type-button-input-type-submit-and-input-type-reset-accessible-name-computation">accessible name computation</a>, and will render as text if the image source is broken. Otherwise, the attribute is ignored.
```

html-aam/index.html
```diff
@@ -7951,6 +7951,9 @@ <h3>HTML Attribute State and Property Mappings</h3>
             </div>
           </li>
           <li>All elements having an accessible object in IAccessible2 mapping are supposed to implement IAccessible, IAccessible2 and IAccessible2_2 interfaces.</li>
+          <li>
+            User interaction with certain HTML elements will set an element's <a data-cite="html/form-control-infrastructure.html#concept-fe-dirty">dirty value flag</a>. For these elements which are identified in [HTML], if specified such attributes represent only the default value or state for the element.
```
spectranaut: I think I don't understand the second sentence of this.. specifically:

> if specified such attributes represent only the default value or state of the element.

Is it me, or could it be made more clear?

spectranaut: after a discussion with scott, here is  much wordier option:

> User interaction with certain HTML elements will set an element's <a data-cite="html/form-control-infrastructure.html#concept-fe-dirty">dirty value flag</a>. These elements, as identified in [HTML], could have attributes that represent the default value or state of the element. When the dirty flag is set, that default value or state does not necessarily determine the value or state of the element or corresponding accessible node.

adampage (approved):

spectranaut:
We need to fix this branch before landing, something crazy is happening.

pkra:
@spectranaut this looks similar to the problem in https://github.com/w3c/aria/pull/2235. @daniel-montalvo do you have any insights?

daniel-montalvo:
> @spectranaut this looks similar to the problem in #2235. @daniel-montalvo do you have any insights?

@pkra @spectranaut 
I merged main and pushed the changes (didn't rebase though). Now this shows only one changed file `html-aam/index.html` and 
we have a more comprehensible diff. Is this any better?

html-aam/index.html
spectranaut: I committed the change

spectranaut:
I think this is ready to merge @scottaohara !

daniel-montalvo:
@spectranaut @scottaohara

I solved one conflict originated from merging j #2309 so as Val said I think this is ready to merge.

scottaohara:
@daniel-montalvo @spectranaut looks like something didn't work right in publishing after merge.

spectranaut:
> @daniel-montalvo @spectranaut looks like something didn't work right in publishing after merge.

I think because of this: https://github.com/w3c/aria/pull/2616

Can you review/merge?

PR #2375, opened 2024-11-15
Update: minor popover clarification

scottaohara:
to acknowledge there are different popover states (manual, auto/empty-string, and the proposed hint) this minor update clarifies that the minimum role of group applies to all popover states.

Is it worth also adding this note, or something like it?

There are no unique mappings for different popover states. Any additional accessibility mapping changes would be the responsibility of the author. e.g, using different base HTML elements, attributes, or ARIA attributes to make such changes.

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -13368,7 +13368,7 @@
                   <li>when the element has a `popovertarget=hide` attribute value,</li>
                   <li>or when the element is a descendant of the `popover` and its `popovertarget` is the "`auto`" state.</li>
                 </ul>
-                <p>If specified on an element with an implicit role of `generic`, then the element's role instead maps to <a class="core-mapping" href="#role-map-group">`group`</a>.</p>
+                <p>If specified on an element with an implicit role of `generic`, then the element's role instead maps to <a class="core-mapping" href="#role-map-group">`group`</a> for all <a data-cite="html/popover.html#the-popover-attribute">`popover` states</a>.</p>
               </td>
             </tr>
           </tbody>
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -13365,7 +13365,7 @@
                 <p>User agents MUST NOT expose a details relation between a `popover` and its invoking element under the following conditions:</p>
                 <ul>
                   <li>when the `popover` is the next immediate accessibility sibling to the invoking element,</li>
-                  <li>when the element has a `popovertarget=hide` attribute value,</li>
+                  <li>when the element has a `popovertargetaction=hide` attribute value,</li>
                   <li>or when the element is a descendant of the `popover` and its `popovertarget` is the "`auto`" state.</li>
                 </ul>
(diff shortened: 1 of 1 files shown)


html-aam/index.html
```diff
@@ -13368,7 +13368,7 @@ <h4 id="att-popover">`popover`</h4>
                   <li>when the element has a `popovertarget=hide` attribute value,</li>
```
adampage, line 13368:
```suggestion
                  <li>when the element has a `popovertargetaction=hide` attribute value,</li>
```

I know this isn’t related to this PR, but should this be `popovertargetaction`?

adampage (approved):
LGTM. 🚀 

> Is it worth also adding this note, or something like it?
> 
> There are no unique mappings for different popover states...

I think that’d be helpful. 👍🏻

html-aam/index.html
scottaohara: yes!  thank you!

html-aam/index.html
```diff
@@ -13365,10 +13365,10 @@ <h4 id="att-popover">`popover`</h4>
                 <p>User agents MUST NOT expose a details relation between a `popover` and its invoking element under the following conditions:</p>
                 <ul>
                   <li>when the `popover` is the next immediate accessibility sibling to the invoking element,</li>
-                  <li>when the element has a `popovertarget=hide` attribute value,</li>
+                  <li>when the element has a `popovertargetaction=hide` attribute value,</li>
                   <li>or when the element is a descendant of the `popover` and its `popovertarget` is the "`auto`" state.</li>
                 </ul>
-                <p>If specified on an element with an implicit role of `generic`, then the element's role instead maps to <a class="core-mapping" href="#role-map-group">`group`</a>.</p>
+                <p>If specified on an element with an implicit role of `generic`, then the element's role instead maps to <a class="core-mapping" href="#role-map-group">`group`</a> for all <a data-cite="html/popover.html#the-popover-attribute">`popover` states</a>.</p>
```
scottaohara, line 13371:
```suggestion
                <p>If specified on an element with an implicit role of `generic`, then the element's role instead maps to <a class="core-mapping" href="#role-map-group">`group`</a> for all <a data-cite="html/popover.html#the-popover-attribute">`popover` states</a>.</p>
                <p class="note">There are no unique mappings for the different `popover` states. Any accessibility mapping changes for the popover element would be the responsibility of the author. e.g., using different base HTML elements, attributes, or ARIA attributes to make such changes.</p>
```

smhigley (approved):
Seems worth adding to me too 👍

scottaohara:
@rahimabdi once you give this a look over, and if you have nothing to add/change, then we can merge.

rahimabdi:
LGTM @scottaohara , thank you!

rahimabdi (approved):

scottaohara:
@spectranaut @jnurthen - while i merged this, i think there may be something off with publishing?

looking at the failure - https://github.com/w3c/aria/actions/runs/12718859000/job/35458034879
... well, that's all i can do is look at it.  these failure logs do very little to tell me anything... i'm sure others might understand them, but they're esoteric nonsense to me.

PR #2377, opened 2024-11-18
[accname] Do “AriaLabel” substep before “Embedded Control” substep

sideshowbarker:
This change reorders the _“AriaLabel”_ substep in the set of steps in the accname Computation algorithm such that it occurs before the _“Embedded Control”_ substep, rather than after that step.

Otherwise, without this change, the spec as written doesn’t match the behavior as implemented in existing UAs, nor the corresponding requirements at https://w3c.github.io/html-aam/#accname-computation in the HTML-AAM spec:

> If the [embedded-control] element has an `aria-label` or an `aria-labelledby` attribute the accessible name is to be calculated using the algorithm defined in Accessible Name and Description: Computation and API Mappings.

(Which is the first step in accname computation for each control)

Also, without this change, the spec doesn’t match the expectations in http://wpt.fyi/results/accname/name/comp_label.html for the _“input with label for association is superceded by aria-label”_ subtest.

* [x] "user agent MUST" tests: https://wpt.fyi/results/accname/name/comp_label.html
* [X] Browser implementations: WebKit, Gecko, and Blink all implement the AriaLabel-computation-step-before-Embedded-Control-computation-step behavior in this patch.
* [ ] MDN Issue/PR: N/A

### DIFF
--- a/accname/index.html
+++ b/accname/index.html
@@ -546,6 +546,35 @@
                     </aside>
                   </div>
                 </li>
+                <li id="comp_label">
+                  <span id="step2D"><!-- Don't link to this legacy numbered ID. --></span><em>AriaLabel:</em> Otherwise, if the <code>current node</code> has an <code>aria-label</code> [=attribute=]
+                  whose value is not undefined, not the empty string, nor, when trimmed of [=ascii whitespace|whitespace=], is not the empty string:
+                  <ol>
+                    <li>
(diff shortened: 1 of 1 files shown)


accdc:
Hi, this seems like a significant spec change, which would break the following if the order was changed like this where aria-label was processed before the embedded control.

```
<div id="desc">
Destroy all my data in
<input aria-label="Number of days" type="number" min="0" value="2">
days.
</div>
<button aria-describedby="desc">Are you sure?</button>
```

Issue #1746, opened 2022-05-24
Consider adding a 'decorative image' role


scottaohara:
# Description of bug or feature request

Creating this issue to start up a conversation relating to one of the proposed solutions to https://github.com/w3c/html-aam/issues/404.  

Presently the importance of images is treated as an odd boolean where authors either need to decide (or are told) that certain images are important to convey to AT, while others are decorative and can be treated as if they were hidden. This need to markup an image as 'informative or decorative' essentially means that authors are telling users of AT what is important for them to know about, and doesn't take into account that some users may want to know what these decorative images represent.  

However, if there were a role (or other `aria-*` attribute... whichever tbh) that could be used to indicate that an image with alternative text is available to AT, but has also been explicitly marked as decorative, so AT can allow users to decide if they want to interact with these images or not.

As mentioned, this could then be a way to resolve the above mentioned HTML AAM issue, where an author has specified an image as having a null `alt` - the long standing indicator that an image is decorative - but also provided a `title`, indicating that there is some level of information available, but is either an author error in how the image should have been marked up, or is communicating potentially nice to have (or nonsense) information that users should be able to decide if they want to make it discoverable, or not.

## Will this require a change to [CORE-AAM](https://w3c.github.io/core-aam/)?

It will if this proposal gets any traction.

## Will this require a change to the [ARIA authoring guide](https://w3c.github.io/aria-practices/)?

It will if this proposal gets any traction.

# PR tracking (for feature tracking, leave as is when submitting)

* ARIA PR:
   * Editor's Draft commit:
   * Working Draft commit:
* Core AAM PR:
* AccName PR:
* APG PR:

# Implementation tracking (for feature tracking, leave as is when submitting)

* WPT test:
* Implementations (link to issue or when done, link to commit):
   * WebKit: [ISSUE]()
   * Gecko: [ISSUE]()
   * Blink: [ISSUE]()

jnurthen:
Discussed at https://www.w3.org/2022/06/02-aria-minutes.html#t05

scottaohara:
potential alternative name "contextual-image"

giacomo-petri:
Hi everyone, 

I’ve discussed this topic internally with our company teams.
I'll use "decorative but with meaning" images to define this kind of images.
We've identified pros and cons:

### PROS
- allow users using assistive technologies to understand the visual presentation of an image
[...]

PR #2378, opened 2024-11-18
[html-aam] Make HTML img accname use “title” text if alt is empty

sideshowbarker:
This change updates the accessible-name computation steps for the `img` element to require that in the case of an `alt=""` (empty `alt`) `img` element which has a non-empty `title` attribute, the value of the `title` attribute is included in computation of the element’s accessible name.

Otherwise, without this change, the spec as written doesn’t match the behavior as implemented in existing UAs, nor the corresponding requirements in https://w3c.github.io/accname/#comp_host_language_label in the AccName spec — specifically, the requirement to skip past the _“Host Language Label”_ step in the `alt=""` case, and to move on to subsequent steps, culminating in the _“Tooltip”_ step, which for HTML elements requires using the value of any `title` attribute, if present.

Also, without this change, the spec doesn’t match the expectations in https://wpt.fyi/results/accname/name/comp_tooltip.html for the _“img with tooltip label with empty alt”_ subtest.

* [X] "user agent MUST" tests: https://wpt.fyi/results/accname/name/comp_tooltip.html
* [X] Browser implementations (link to issue or commit): WebKit, Gecko, and* Blink all implement the if-alt-is-empty-use-title behavior in this patch.
* [ ] MDN Issue/PR: N/A

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -16218,13 +16218,13 @@
               <a href="" class="accname">Accessible Name and Description: Computation and API Mappings</a>.
             </li>
             <li>
-              Otherwise use `alt` attribute, even if its value is the empty string.
+              Otherwise use the `alt` attribute, unless its value is the empty string.
               <div class="note">
                 An `img` with an `alt` attribute whose value is the empty string is mapped to the
-                <a class="core-mapping" href="#role-map-presentation">`presentation`</a> role. It has no accessible name.
+                <a class="core-mapping" href="#role-map-presentation">`presentation`</a> role. And for any `role="presentation"` element, the algorithm defined in <a href="" class="accname">Accessible Name and Description: Computation and API Mappings</a> requires skipping past the “Host Language Label” step (which would otherwise result in using the value of the `alt` attribute) to subsequent steps — culminating in the “Tooltip” step, which for HTML elements requires using the value of the `title` attribute, if present.
               </div>
             </li>
-            <li>Otherwise, if there is no `alt` attribute use the `title` attribute.</li>
+            <li>Otherwise, use the `title` attribute.</li>
             <li>Otherwise there is no <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.</li>
           </ol>
         </section>


scottaohara:
well, this is the opposite outcome that had been decided about 2.5 years ago - https://github.com/w3c/html-aam/pull/322.

cookiecrook:
Unfortunately I missed this addition in review... https://github.com/web-platform-tests/wpt/pull/42093

cookiecrook:
Since the browsers aligned on this variant that conflicts with the HTML-AAM spec, I think we should get UA implementor consensus to remove the WPT test and roll back this change. We'll do it in a way that does not affect the current Interop 2024 Focus Area.

We need comments from:
- @nmlapre and/or @jcsteh
- @aleventhal and/or @tranjocelyn  (cc @benbeaudry)
- @twilco

scottaohara:
> ~Unfortunately I missed this addition in review...~ [web-platform-tests/wpt#42093](https://github.com/web-platform-tests/wpt/pull/42093)
> 
> Actually, despite the typo... It seems I suggested this new test in review.
> 
> * [Update `/accname/name/comp_tooltip.html` web-platform-tests/wpt#42093 (comment)](https://github.com/web-platform-tests/wpt/pull/42093#discussion_r1336445287)
> 
> Sorry.

no reason to apologize - looking at that wpt PR, you gave me plenty of pings to come review, and i missed every one of them.
and it does seem that there were parallel efforts to work on this overarching issue, but both of the specs / the related issues were not being considered together.   shrug.   things happen, and opinions can change, but I just want to make sure that what was resolved by the working group and reflected in HTML AAM was considered, since it seems maybe that was missed when the test was created.

nmlapre:
> I think we should get UA implementor consensus to remove the WPT test and roll back this change

Are we talking about removing the "img with tooltip label with empty alt" test? What change are we considering rolling back? From my perspective, what the browsers are doing now is desirable behavior, which seems to make the `` test valid (albeit not by the current wording of the spec).

scottaohara:
@nmlapre for mozilla specifically, there would be no rollback, per se, as Firefox was the browser that originally treated a `title=foo` as the name for an image that had `alt=""`.  Whereas back when the HTML AAM issue/PR began, Chromium and Webkit did not algin with Firefox's behavior and did not use `title` as a means to re-expose an image that had `alt=""`.  (or if memory serves, webkit did still 'name' the image with the title attribute, but _also_ marked the image as ignored in the a11y tree due to the existence of the `alt=""`)

The linked HTML AAM issue has a lot of back and forth on whether or not `title` should be a naming mechanism for this contradictory markup.  I'm personally not interested in redebating that - so everything here after is not me trying to change one's opinion, but providing some summary/rational that maybe everyone didn't read the first time around - hence this discrepency.

I think both points of view are valid - and can be beneficial to users - depending on the use cases.  (e.g., `` is important to expose.  but `` is unfortunately not uncommon, and not generally useful.

As an FYI, accessibility checkers such as IBM's a11y assessment tool, ARC Toolkit, the Wave Evaluation tool and the ssa.gov ANDI accessibility checker identify instances of `img alt="" title=foo"` as "decorative" images and raise alerts/errors for using these attributes together in this way.  That leads me to believe that none of these tools are aware of the changes that browsers made here - and i mention this not as an argument to say what browsers have done here is _wrong_, but more to point out it doesn't match how these tools / developers have been led to believe these attributes are meant to be used together - particularly since there's long standing guidance, predating ARIA, that `alt=""` means the image is meant to be decorative.

So, if the decision is to keep this new unified behavior - and the spec needs to be adjusted to reflect that, then I'd submit there's also a need to make this update more widely known so that guidance can be updated and so that checkers aren't sending people warnings/alerts about what have become false positives (or change their messaging to point out that authors have provided conflicting information and that _something_ should be done to address that).

sideshowbarker:
> I think we should get UA implementor consensus to remove the WPT test

For that I opened https://github.com/web-platform-tests/wpt/pull/49214

sideshowbarker:
> Since the browsers aligned on this variant that conflicts with the HTML-AAM spec, I think we should get UA implementor consensus to remove the WPT test

You mean the test at https://wpt.fyi/results/accname/name/comp_tooltip.html that all the UAs are currently passing? If so, I guess I’m confused. Why do we want to remove a test that everybody’s passing?

> and roll back this change

You mean roll back the https://github.com/w3c/html-aam/pull/322/files change? If so, that’s what the https://github.com/w3c/aria/pull/2378/files patch in this new PR is doing, right?

scottaohara:
@sideshowbarker 
> If so, I guess I’m confused. Why do we want to remove a test that everybody’s passing?

I’ll try explaining this again…
The  test was written with a passing expectation that was the opposite of what the spec was expecting. Specificity the test alt=“” title=foo, asserting  “foo” is the “passing” result is not in alignment with what the spec said the result should have been.   UAs are passing an inaccurate test. 

As mentioned, at the time of the spec update, WebKit and chromium were in alignment with html aam.  Now they pass a test and not the spec. 

The change to roll back would be the change made in browsers (specifically WebKit and chromium) as they did match what html aam expected, and now they don’t. 


and to be clear again, while I have my preference, I am not trying to advocate for one way or the other.  It is up to those @cookiecrook mentioned to weigh in as to what should happen. I’m more trying to point out that there was a miss here (partially my fault for not catching the errant test result when asked to review last year) and that browsers updated in a way that moved away from what the spec / aria wg had resolved on.

sideshowbarker:
@scottaohara I see now. Thanks much for the explanation, and thanks much for the details. It's all extremely helpful to me. I'm not really yet socialized into the AAM and accname — but it's my intention to get it all interoperably implemented in Ladybird. I mean, implementing myself, and maintaining it — unless/until somebody else who has more accessibility implementation experience than me eventually shows up.

And to be clear, I have no strong opinions personally about any of it. I don't have the experience nor insights into the domain that'd take to be able to form opinions. Someday maybe, after I've done a lot more implementation work.

Regardless, I just want to have Ladybird working in the best way it can for AT users. And I want to make all the tests turn green.

So that’s the context for this PR — and for any further PRs or issues I may end up opening here later

cookiecrook:
Post Thanksgiving catch-up. Don't we need a test removed in comp_tooltip.html and a new one added to comp_tooltip.tentative.html

scottaohara:
related from the last time this was talked about - https://github.com/w3c/aria/issues/1746

jnurthen:
@adampage to create WPT PR to resolve the WPT issue

spectranaut:
Discussed in today's meeting: https://www.w3.org/2024/12/05-aria-minutes#8ad9

sideshowbarker:
Closing this, given that https://github.com/web-platform-tests/wpt/pull/49552 updated the WPT test expectations to match the existing HTML-AAM spec requirements.

cookiecrook:
@sideshowbarker are there implementation trackers for getting this change back into the browsers? The test file is `.tentative.` so browser engineers may not make that change without a tracking issue.

cookiecrook:
Got it... Normally that's what this section of the New ARIA Issue template is for: 
template showing unchecked items for linking implementation bug tracker links

cookiecrook:
Maybe we need to clone this or track as a new issue, since the spec and WPT is now as-intended, but the 3 major engines don't pass it...

cookiecrook:
WebKit: https://webkit.org/b/289290
Chromium: https://issues.chromium.org/issues/401294444
Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1952401

sideshowbarker:
> Got it... Normally that's what this section of the New ARIA Issue template is for: template showing unchecked items for linking implementation bug tracker links

Right — but this PR as written was about changing the spec to match the behavior of existing engines. So the spec patch I wrote didn’t require any browser bugs, nor any changes to WPT tests.

As far as I can see, https://github.com/w3c/html-aam/pull/322 is instead the PR which should have followed the template and included links to spec bugs. But regardless, wherever it should have happened, it’s not this PR.

sideshowbarker:
See also my comment at https://github.com/web-platform-tests/wpt/pull/49552#issuecomment-2559495334
> Chromium, WebKit, and Mozilla had all been passing this test previously, and with this change they’re now failing. So it seems like bugs should be rightly be filed in their bug trackers — to alert them to need to update their implementations.

PR #2383, opened 2024-11-22
Update: change default role for custom elements

### LINKED ISSUE #2303: 9.3 Presentational Roles Conflict Resolution does not consider custom element use cases
# Describe your concern

It's natural to provide ARIA properties on a custom element and then have the custom element implementation copy that property to a descendant. For example, aria-label, aria-description or aria-placeholder could be provided by the author on the custom element. The custom element would then copy the ARIA properties to the actual input element in its subtree, and it could then place role="none" on the custom element to avoid 1) Having an extra element in the a11y tree with duplicate properties and 2) Using a name on the custom element which likely has a generic role -- name on generic is prohibited.

Currently the role="none" (or role="presentation") on the custom element is ignored because we are afraid of losing the semantics. However, it seems like a reasonable use case to me. I don't think custom elements should get their role="none" markup ignored when used with aria-foo properties. It should still ignore role="none" if it has a tabindex.

https://w3c.github.io/aria

Link to documentation:

https://w3c.github.io/aria/#conflict_resolution_presentation_none

Yes

scottaohara:
closes #2303

This updates a custom element's default role to none to allow attribute reflection from the custom element parent to its internals.

This change provides additional clarification about how a custom element can be provided a role by authors, and what caveats would change a custom element's default role of none, to another implicit minimum role.

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -863,7 +863,7 @@
                   <li>If the author assigned a conforming role using the `role` attribute, or by the custom element's internals: map to the specified role.</li>
                   <li>else if the author assigned HTML attributes that result in a <a>minimum role</a>:  map to the minimum role.</li>
                   <li>else if the custom element is focusable: map to the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
-                  <li>else if an author specifies a global ARIA attribute on the custom element that creates a relation with another element: map to the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
+                  <li>else if an author specifies a global ARIA attribute on the custom element that creates a relation with another element: map to the <a class="core-mapping" href="#role-map-group">`group`</a> role</li>
                   <li>else if the custom element has no attached shadow root: map to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
                   <li>else if the custom element has an `aria-live` attribute: map to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
                   <li>else if element internals are used to set a global ARIA property: map to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role.</li>
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -862,7 +862,7 @@
                 <ul>
                   <li>If the author assigned a conforming role using the `role` attribute, or by the custom element's internals: map to the specified role.</li>
                   <li>else if the author assigned HTML attributes that result in a <a>minimum role</a>:  map to the minimum role.</li>
-                  <li>else if the custom element is focusable: map to the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
+                  <li>else if the custom element is focusable: map to the <a class="core-mapping" href="#role-map-group">`group`</a> role</li>
                   <li>else if an author specifies a global ARIA attribute on the custom element that creates a relation with another element: map to the <a class="core-mapping" href="#role-map-group">`group`</a> role</li>
                   <li>else if the custom element has no attached shadow root: map to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
                   <li>else if the custom element has an `aria-live` attribute: map to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -901,7 +901,7 @@
               <th id=custom-element-comments>Comments</th>
               <td>
                 <p>
-                  Along with the attributes defined in this specification that contribute to modifying an element's <a>minimum role</a>,
+                  Along with the attributes defined in this specification that participate in modifying an element's <a>minimum role</a>,
                   the following conditions will also alter a custom element's computed role, if an author has not otherwise specified an
                   explicit role for the element:
                 </p>
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -259,7 +259,7 @@
             specified which would require a more specific <a>minimum role</a> to be exposed.
           </li>
           <li>
-            Where an element is indicated as having &quot;No corresponding (WAI-ARIA) role&quot;, or is mapped to the <a class="core-mapping" href="#role-map-generic">`generic`</a>
+            Where an element is mapped to the <a class="core-mapping" href="#role-map-generic">`generic`</a>
             or <a class="core-mapping" href="#role-map-none">`none`</a> roles, user agents
             MUST NOT expose the <a class="core-mapping" href="#ariaRoleDescription">`aria-roledescription`</a> property value in the <a class="termref">accessibility tree</a> unless the element has an
             explicit, conforming `role` attribute value which [[WAI-ARIA-1.2]] does not prohibit the use of `aria-roledescription`.
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -841,29 +841,33 @@
             </tr>
           </tbody>
         </table>
-        <h4 id="el-autonomous-custom-element">autonomous custom element</h4>
-        <table class="data" aria-labelledby="el-autonomous-custom-element">
+        <h4 id="el-autonomous-custom-element">
+          autonomous custom elements
+          and
+          <span id="el-form-associated-custom-element">form-associated custom elements</span>
+        </h4>
+        <table aria-labelledby="el-autonomous-custom-element">
           <tbody>
             <tr>
               <th>HTML Specification</th>
               <td>
                 <a data-cite="HTML">autonomous custom element</a>
+                and
+                <a data-cite="html/custom-elements.html#custom-elements-face-example">form-associated custom element</a>
               </td>
             </tr>
             <tr>
               <th>[[wai-aria-1.2]]</th>
               <td>
-                <p>
-                  If the author assigned a conforming ARIA role using the `role` attribute, 
-                  or by the custom element's internals, map to the specified role.
-                </p>
-                <p>
-                  If the author assigned attributes that result in a <a>minimum role</a>, then map
-                  to the minimum role (see <a href="#custom-element-comments">comments</a>).
-                </p>
-                <p>
-                  Otherwise, the element maps to the <a class="core-mapping" href="#role-map-none">`none`</a> role.
-                </p>
+                <ul>
+                  <li>If the author assigned a conforming ARIA role using the `role` attribute, or by the custom element's internals, map to the specified role.</li>
+                  <li>else if the author assigned HTML attributes that result in a <a>minimum role</a>, then map to the minimum role.</li>
+                  <li>else if the custom element is focusable: expose the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
+                  <li>else if an author specifies an ARIA attribute on the custom element that creates a relation with another element: expose the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
+                  <li>else if the custom element has an `aria-live` attribute: expose the <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
+                  <li>else if element internals are used to set a global ARIA property: expose the <a class="core-mapping" href="#role-map-generic">`generic`</a> role.</li>
+                  <li>Otherwise, the element maps to the <a class="core-mapping" href="#role-map-none">`none`</a> role.</li>
+                </ul>
               </td>
             </tr>
             <tr>
@@ -898,20 +902,8 @@
             </tr>
             <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
             <tr>
-              <th id=custom-element-comments>Comments</th>
-              <td>
-                <p>
-                  Along with the attributes defined in this specification that participate in modifying an element's <a>minimum role</a>,
-                  the following conditions will also alter a custom element's computed role, if an author has not otherwise specified an
-                  explicit role for the element:
-                </p>
-                <ul>
-                  <li>if the custom element is focusable: expose the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
-                  <li>else if an author specifies an ARIA attribute on the custom element that creates a relation with another element: expose the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
-                  <li>else if the custom element has an `aria-live` attribute: expose the <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
-                  <li>else if element internals are used to set a global ARIA property: expose the <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
-                </ul>
-              </td>
+              <th>Comments</th>
+              <td></td>
             </tr>
           </tbody>
         </table>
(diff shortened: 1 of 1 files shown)


scottaohara:
marking as draft as the form associated custom element mapping table also needs updating - but i want to make sure i got this going in the right direction before i edit that table too

html-aam/index.html
```diff
@@ -849,8 +862,30 @@ <h4 id="el-autonomous-custom-element">autonomous custom element</h4>
             </tr>
             <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
             <tr>
-              <th>Comments</th>
-              <td></td>
+              <th id=custom-element-comments>Comments</th>
+              <td>
+                <p>
+                  Along with the attributes defined in this specification that contribute to modifying an element's <a>minimum role</a>,
```
scottaohara, line 868:
```suggestion
                  Along with the attributes defined in this specification that participate in modifying an element's <a>minimum role</a>,
```

html-aam/index.html
```diff
@@ -849,8 +862,30 @@ <h4 id="el-autonomous-custom-element">autonomous custom element</h4>
...
+                  explicit role for the element:
+                </p>
+                <ul>
+                  <li>if the custom element is focusable: expose as <a class="core-mapping" href="#role-map-group">`group`</a> role</li>
+                  <li>if the custom element has an `aria-live` or `aria-owns` attribute: expose as <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
+                  <li>if element internals are used to set a global ARIA property: expose as <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
+                  <li>if an author specifies an ARIA relation via an element attribute: expose <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
+                </ul>
+                <div class="example">
+                  <p>The following is an exmaple of an author specifiying an ARIA relation on a custom element, with another element in the DOM.</p>
```
aleventhal, line 879: ex-maple, not ex-oak?

aleventhal: Seriously, is this example for the previous bullet? Because aria-details won't change it from 'none', unless you do myCustomElement.ariaDetailsElements = [ el1, ... ];

html-aam/index.html
```diff
@@ -8680,7 +8715,7 @@ <h4 id="att-autofocus">`autofocus`</h4>
             </tr>
             <tr>
               <th>Comments</th>
-              <td>Provides a <a href="termref">minimum role</a> of <a class="core-mapping" href="#role-map-group">`group`</a>.</td>
+              <td>Provides a <a>minimum role</a> of <a class="core-mapping" href="#role-map-group">`group`</a>.</td>
```
aleventhal, line 8718: Did you intend to remove href="termref" here and in the other place?

aleventhal (changes requested):

html-aam/index.html
```diff
@@ -849,8 +862,30 @@ <h4 id="el-autonomous-custom-element">autonomous custom element</h4>
...
+                <ul>
+                  <li>if the custom element is focusable: expose as <a class="core-mapping" href="#role-map-group">`group`</a> role</li>
+                  <li>if the custom element has an `aria-live` or `aria-owns` attribute: expose as <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
+                  <li>if element internals are used to set a global ARIA property: expose as <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
+                  <li>if an author specifies an ARIA relation via an element attribute: expose <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
+                </ul>
+                <div class="example">
+                  <p>The following is an exmaple of an author specifiying an ARIA relation on a custom element, with another element in the DOM.</p>
+                  <pre>
+                    <code>
```
scottaohara, line 881:
```suggestion
                  <pre><code>
```

html-aam/index.html
```diff
@@ -849,8 +862,30 @@ <h4 id="el-autonomous-custom-element">autonomous custom element</h4>
...
+                      &lt;my-element aria-details=ex>...&lt;my-element>
+                      ...
+                      &lt;div id=ex> &lt;!-- details for my-element --> &lt;/div>
+                    </code>
+                  </pre>
```
scottaohara, line 886:
```suggestion
                  </code></pre>
```

html-aam/index.html
scottaohara: yes.  it was broken, this fixed the link

html-aam/index.html
scottaohara: ah ok.  thanks.  I didn't realize you were saying that this would only happen if it was set via IDL.

my thought was that this was another use case that didn't make sense to reflect, since the ID that was being referenced would likely be in the light DOM, so it wouldn't work if reflected into the custom element's shadow dom parts

html-aam/index.html
```diff
@@ -849,8 +862,38 @@ <h4 id="el-autonomous-custom-element">autonomous custom element</h4>
...
+    &lt;div id=ex> &lt;!-- details for my-element --> &lt;/div>
+    ...
+    &lt;ex-ample aria-owns=u>...&lt;/ex-ample>
+      ...
+    &lt;div id=u> &lt;!-- becomes the last a11y child of ex-ample --> &lt;/div>
+
+    &lt;id-lement> ... &lt;/id-lement>
+    &lt;script>
+      ...
+      document.querySelector('id-lement').ariaDescribedby = anotherEl;
```
scottaohara, line 892:
```suggestion
      document.querySelector('id-lement').ariaDescribedbyByElements = [ descElement ];
```

html-aam/index.html
```diff
@@ -849,8 +862,38 @@ <h4 id="el-autonomous-custom-element">autonomous custom element</h4>
...
+                <p>
+                  Along with the attributes defined in this specification that participate in modifying an element's <a>minimum role</a>,
+                  the following conditions will also alter a custom element's computed role, if an author has not otherwise specified an
+                  explicit role for the element:
+                </p>
+                <ul>
+                  <li>if the custom element is focusable: expose as <a class="core-mapping" href="#role-map-group">`group`</a> role</li>
+                  <li>if the custom element has an `aria-live` attribute: expose as <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
+                  <li>if element internals are used to set a global ARIA property: expose as <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
+                  <li>if an author specifies an ARIA attribute that creates a relation with another element: expose <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
```
scottaohara, line 876:
```suggestion
                  <li>if an author specifies an ARIA attribute on the custom element that creates a relation with another element: expose <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
```

aleventhal (approved):
Awesome, after many iterations

cookiecrook (changes requested):

html-aam/index.html
```diff
@@ -804,18 +805,35 @@ <h4 id="el-audio">`audio`</h4>
...
+              <td>
+                <ul>
+                  <li>If the author assigned a conforming ARIA role using the `role` attribute, or by the custom element's internals: map to the specified role.</li>
+                  <li>else if the author assigned HTML attributes that result in a <a>minimum role</a>:  map to the minimum role.</li>
+                  <li>else if the custom element is focusable: map to the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
+                  <li>else if an author specifies a global ARIA attribute on the custom element that creates a relation with another element: map to the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
+                  <li>else if the custom element has no attached shadow root: map to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
+                  <li>else if the custom element has an `aria-live` attribute: map to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
```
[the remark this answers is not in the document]
scottaohara: i'm not sure i follow this comment - since custom elements are presently generic anything that says to map this to generic is essentially keeping the status quo.

html-aam/index.html
```diff
@@ -226,7 +226,8 @@ <h3>HTML Element Role Mappings</h3>
             specified which would require a more specific <a>minimum role</a> to be exposed.
           </li>
           <li>
-            Where an element is indicated as having &quot;No corresponding (WAI-ARIA) role&quot;, or is mapped to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role, user agents
+            Where an element is indicated as having &quot;No corresponding (WAI-ARIA) role&quot;, or is mapped to the <a class="core-mapping" href="#role-map-generic">`generic`</a>
```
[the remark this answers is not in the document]
scottaohara: this text existed prior to the creation of the html-* platform roles.  i don't see a problem with revising this to just generic and none.  i think for some platform roles one's milage will vary if anything meaningful actually happens if one were to try and change the role description - but i can definitely see why maybe someone might want to use it on a video element, for instance.

html-aam/index.html
```diff
@@ -804,18 +805,35 @@ <h4 id="el-audio">`audio`</h4>
...
               </td>
             </tr>
             <tr>
               <th>[[wai-aria-1.2]]</th>
-              <td>If the author assigned a conforming ARIA role using the `role` attribute, map to that role. Otherwise, the <a class="core-mapping" href="#role-map-generic">`generic`</a> role.</td>
+              <td>
+                <ul>
+                  <li>If the author assigned a conforming role using the `role` attribute, or by the custom element's internals: map to the specified role.</li>
+                  <li>else if the author assigned HTML attributes that result in a <a>minimum role</a>:  map to the minimum role.</li>
+                  <li>else if the custom element is focusable: map to the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
```
scottaohara, line 829:
```suggestion
                  <li>else if the custom element is focusable: map to the <a class="core-mapping" href="#role-map-group">`group`</a> role</li>
```

html-aam/index.html
```diff
@@ -804,18 +805,35 @@ <h4 id="el-audio">`audio`</h4>
...
             <tr>
               <th>[[wai-aria-1.2]]</th>
-              <td>If the author assigned a conforming ARIA role using the `role` attribute, map to that role. Otherwise, the <a class="core-mapping" href="#role-map-generic">`generic`</a> role.</td>
+              <td>
+                <ul>
+                  <li>If the author assigned a conforming ARIA role using the `role` attribute, or by the custom element's internals: map to the specified role.</li>
+                  <li>else if the author assigned HTML attributes that result in a <a>minimum role</a>:  map to the minimum role.</li>
+                  <li>else if the custom element is focusable: map to the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
+                  <li>else if an author specifies a global ARIA attribute on the custom element that creates a relation with another element: map to the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
+                  <li>else if the custom element has no attached shadow root: map to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role</li>
```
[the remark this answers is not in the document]
scottaohara: custom elements are currently generic, so even without this spec change, this use case doesn't work now.

html-aam/index.html
```diff
@@ -804,18 +805,35 @@ <h4 id="el-audio">`audio`</h4>
...
               </td>
             </tr>
             <tr>
               <th>[[wai-aria-1.2]]</th>
-              <td>If the author assigned a conforming ARIA role using the `role` attribute, map to that role. Otherwise, the <a class="core-mapping" href="#role-map-generic">`generic`</a> role.</td>
+              <td>
+                <ul>
+                  <li>If the author assigned a conforming ARIA role using the `role` attribute, or by the custom element's internals: map to the specified role.</li>
+                  <li>else if the author assigned HTML attributes that result in a <a>minimum role</a>:  map to the minimum role.</li>
+                  <li>else if the custom element is focusable: map to the <a class="core-mapping" href="#role-map-generic">`group`</a> role</li>
```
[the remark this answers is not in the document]
scottaohara: fixed the incorrect link.  i never thought to use rref in html aam - wasn't sure that'd work.  i don't see that used in core aam either.

maybe it'd work now with the specs all in one repo?

re: about mapping to group or something else - i'm open to ideas.  cc @aleventhal if you have thoughts?

spectranaut:
@cookiecrook can you respond to @scottaohara comments to help move this forward?

cookiecrook (approved):

html-aam/index.html
```diff
@@ -804,18 +805,35 @@ <h4 id="el-audio">`audio`</h4>
...
               </td>
             </tr>
             <tr>
               <th>[[wai-aria-1.2]]</th>
-              <td>If the author assigned a conforming ARIA role using the `role` attribute, map to that role. Otherwise, the <a class="core-mapping" href="#role-map-generic">`generic`</a> role.</td>
+              <td>
+                <ul>
+                  <li>If the author assigned a conforming role using the `role` attribute, or by the custom element's internals: map to the specified role.</li>
+                  <li>else if the author assigned HTML attributes that result in a <a>minimum role</a>:  map to the minimum role.</li>
+                  <li>else if the custom element is focusable: map to the <a class="core-mapping" href="#role-map-group">`group`</a> role</li>
```
jcsteh, line 865: Should this be done even if a custom element's ShadowRoot has delegatesFocus: true? It's a tricky one. On one hand, it's a pretty important principle to expose focusable elements to accessibility. On the other hand, when delegatesFocus is true, the element isn't in the tab order; it is effectively transparent as far as focus is concerned.

I think it's probably safer to still use role="group" even for delegatesFocus: true. For example, browser native date/time inputs expose the container and allow it to be focused, even though it isn't part of the tab order. Nevertheless, I thought I'd clarify because I haven't seen any discussion of delegatesFocus either here or in #2303.

jcsteh (commented):
Sorry for the ridiculously slow turnaround here. Just a couple of questions, but overall this looks good to me.

> This updates a custom element's default role to none to allow attribute reflection from the custom element parent to its internals.

Just to be extra clear, when you say "allow attribute reflection" here, you mean that the component could copy attributes itself (e.g. aria-label), not that these would be implicitly reflected somehow? Nothing in the proposed spec text suggests implicit reflectoin; I just wanted to double check my understanding.

spectranaut:
@scottaohara other than Jamie's question this might be ready for implementation bugs?

html-aam/index.html
jnurthen: Good question which we should answer. As delegatesFocus doesn't actually make the shadow host itself "focusable" per the HTML spec's definition (sequentially/click focusable, based on tabindex or native semantics), as worded now, a delegating host with no tabindex of its own would fall through the focusable rule entirely and land on the generic/none rules below so not group.

If we want the group outcome for consistency with native composite controls (date/time inputs, etc.), I think we need to say so explicitly rather than relying on "focusable" to cover it — something like:

"...or is the host of a shadow root whose delegatesFocus is true and which contains a focusable area"

It might also be worth linking "focusable" to [the HTML spec's definition](https://html.spec.whatwg.org/multipage/interaction.html#focusable-area) while we're at it

PR #2385, opened 2024-11-25
[accname] Explicitly state UAs must ignore “aria-label” for slots

sideshowbarker:
This change updates the **AriaLabel** substep of the “Computation steps” section of the accname spec to explicitly state that the step must not be performed if the node being visited is a `slot` element.

See the discussion at https://github.com/w3c/accname/issues/173. It seems a requirement to ignore the `aria-label` attribute in this case — or to ignore _any_ ARIA attributes present on `slot` elements — may follow from normative requirements stated in some other spec.

Nevertheless, a specific requirement should be stated at point of use in the text for this step in accname spec itself.

* [X] "user agent MUST" tests: https://wpt.fyi/results/accname/name/shadowdom/slot.html
* [X] Browser implementations (link to issue or commit): WebKit, Gecko, and Blink already implement this.
* [X] WPT Tests: [accname/name/shadowdom/slot.html](https://wpt.fyi/results/accname/name/shadowdom/slot.html?label=master&label=experimental&aligned&q=label%3Aaccessibility)
* [ ] MDN Issue/PR: N/A

### DIFF
--- a/accname/index.html
+++ b/accname/index.html
@@ -578,7 +578,7 @@
                   </aside>
                 </li>
                 <li id="comp_label">
-                  <span id="step2D"><!-- Don't link to this legacy numbered ID. --></span><em>AriaLabel:</em> Otherwise, if the <code>current node</code> has an <code>aria-label</code> [=attribute=]
+                  <span id="step2D"><!-- Don't link to this legacy numbered ID. --></span><em>AriaLabel:</em> Otherwise, if the <code>current node</code> is not a [=slot=] element and has an <code>aria-label</code> [=attribute=]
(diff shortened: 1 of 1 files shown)


keithamus (approved):
Apologies, this one slipped off my radar. This does indeed LGTM :+1:

MelSumner (approved):
Agree with this addition, it's the correct framing.

cookiecrook (approved):

accdc (approved):

cookiecrook:
I added a link above to the existing WPT tests for accname on slots... If more tests are needed, please file a new issue on [interop-accessibility](https://github.com/web-platform-tests/interop-accessibility/). Thanks Mike!

Issue #2230, opened 2024-05-28
[monorepo] move aria-common issues & PRs to ARIA repo


pkra:
Identify what we want to keep before archiving aria-common.

pkra:
I think we should try to keep the history after all. I'll give that a try.

pkra:
Coming back to this, I don't think the approach we used for the other repos works out for aria-common since we already have a folder with a history. However, well, we have a fold with a history.

Moving issues is probably not that useful since aria-common has so few of them (and future ones can be filed here).

This leaves https://github.com/w3c/aria-common/pull/106 which I will re-create. I don't see a way to maintain its 94 commits but perhaps that's for the best. Ultimately, this is a refactoring so we primarily need an unchanged HTML output.

pkra:
Oh, there's one other thing: the eslint setup from aria-common should probably be moved here.

pkra:
As per editors' this can be closed. The aria-common issue tracker will be cleaned up in good time.

PR #2407, opened 2024-12-28
[aria] Drop “deprecated on this role” attributes from per-role tables

sideshowbarker:
This change causes the Inherited States and Properties rows of the per-role Characteristics tables in the ARIA spec to stop including all the attributes that has been marked “deprecated on this role in ARIA 1.2”.

If the intent is to clearly to signal to authors that those attributes shouldn’t be use with those roles, then it makes sense to just stop including those attributes in the tables for those roles — rather than indefinitely continuing to include them.

* [ ] "user agent MUST" tests: N/A
* [ ] Browser implementations (link to issue or commit): N/A
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR: none
* [ ] MDN Issue/PR: none

### DIFF
--- a/common/script/aria.js
+++ b/common/script/aria.js
@@ -479,12 +479,10 @@ function ariaAttributeReferences() {
                     var property = sortedList[k];
                     var req = "";
                     var dep = "";
+                    if (property.deprecated)
+                        continue;
                     if (property.required) {
                         req = " <strong>(required)</strong>";
-                    }
-                    if (property.deprecated) {
-                        dep =
-                            " <strong>(deprecated on this role in ARIA 1.2)</strong>";
                     }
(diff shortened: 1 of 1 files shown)


pkra:
Whatever the group decides, this must wait for https://github.com/w3c/aria/issues/2230

spectranaut:
Discussed in ARIA meeting, to be discussed in an editor's meeting: https://www.w3.org/2025/01/30-aria-minutes#7f1b

pkra:
As per today's editors call, we will close this issue. The WG did not want to go in this direction. Instead the WG tasked the editors with finding a different approach to present deprecation, possibly following the HTML spec's example. 

A new issue will track that work.

PR #2408, opened 2025-01-03
editorial: refactor aria.js

pkra:
Part of #2230

Partially moved from https://github.com/w3c/aria-common/pull/106 at [caf8e9e](https://github.com/w3c/aria-common/pull/106/commits/caf8e9edef64a279e0aa2406fefe8cac624aad63)

From the original PR:

Much work to be done.

* [x] add a very simple test
  * `test.sh` runs respec on ARIA before/after applying aria.js changes and diffs the two results
* rough goals 
  * clarify (e.g., switch to const/let, replace reused variable names with unique ones)
  * simplify (e.g., loops over querySelectorAll, use sets instead of filtering)
  * revise HTML string creation
  * extract loops
* steps
  * [x] pdef, sdef loop
  * [x] generating HTML for indices of (global) states&properties as well as role=roletype
  * [x] rdef loop
  * [x] roleInfo loop
  * [x] propList loop
  * [x] generate HTML for role-related indices, sub-Roles
  * [x] clean-up
* continuation
  * [x] rewrite subRoles code

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2408.html" title="Last updated on Mar 28, 2025, 8:51 AM UTC (d2bbc1e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2408/e5f0305...d2bbc1e.html" title="Last updated on Mar 28, 2025, 8:51 AM UTC (d2bbc1e)">Diff</a>


common/script/aria.js
```diff
@@ -7,746 +20,395 @@
...
-            }
+/**
+ *
+ * @param {HTMLElement} container - parent of sdef or pdef or rdef
+ */
+const rewriteDefContainer = (container) => {
+    // if we are in a div, convert that div to a section
+    // TODO:
+    // a) seems to be always the case.
+    // b) Why don't we author the spec this way?
```
pkra, line 84: Copied from https://github.com/w3c/aria-common/pull/106#discussion_r1462542533

> Seems like we should. Anything that is not generated should just be in the static html. It's odd because this container has `id=<attributename>` on it for the url fragment, which probably should be generated, since it is the same as the pdef/sdef

common/script/aria.js
```diff
@@ -7,746 +20,395 @@
...
+ * Replaces sdef/pdef with desired HTML
+ * @param {Object} propList -
+ * @param {HTMLElement} item - sdef or pdef, from nodeList.forEach
+ */
+const generateHTMLStatesAndProperties = function (propList, item) {
+    const title = item.getAttribute("title") || item.innerHTML;
+    const itemEntry = propList[title];
+    const dRef = item.nextElementSibling;
+    dRef.id = "desc-" + title; // TODO: too much of a side-effect?
+    dRef.setAttribute("role", "definition"); // TODO: ditto?
```
pkra, line 127: From https://github.com/w3c/aria-common/pull/106#discussion_r1462553999

> this kind of stuff is a little weird to be adding here, it's not generated, it should just be in the html, it seems.

common/script/aria.js
```diff
@@ -764,16 +425,229 @@ function ariaAttributeReferences() {
...
+    // process the document before anything else is done
+    // first get the properties
+    const pdefsAndsdefs = document.querySelectorAll("pdef, sdef");
+    const pdefsAndsdefsContainer = [...pdefsAndsdefs].map(
+        (node) => node.parentNode
+    );
+
+    pdefsAndsdefs.forEach(populatePropList.bind(null, propList));
+    pdefsAndsdefs.forEach(populateGlobalSP.bind(null, propList, globalSP));
+    pdefsAndsdefs.forEach(generateHTMLStatesAndProperties.bind(null, propList));
```
pkra, line 600: From https://github.com/w3c/aria-common/pull/106#discussion_r1462567065

> Each of these new function names could maybe follow a bit better of conventions... my top preference is at at least to rename any function that changes or produces html to have the prefix `render...`, and to describe specifically what it is creating -- like, `renderStatesAndPropertiesHeadings` instead of `generateHTMLStatesAndProperties`.
> 
> Also, related, I guess the "populate" is fine, but I might use a word like `build...` for anything that is just producing a data structure. Like `buildPropList` and `buildGlobalStatesAndPropertiesList` and wherever else data is just being mined from the html and then rearranged.
> 
> Also.... I know you are reusing variable names, but personally I would use no short hands -- so, "buildStateAndPropertyList" instead of "buildPropList".  I guess "pdefs" and "sdefs" etc are ok, considering they are using in our spec.
> 
> but right now I still don't really understand why there is a `globalSP` and `propList` and the difference between these two things.. could they somehow be names a bit more clearly? Or are you trying to reuse these original names to make it easier to compare to the original script?

common/script/aria.js
```diff
@@ -7,746 +20,395 @@
...
+
+    const applicabilityText = container.querySelector(
+        "." + itemEntry.is + "-applicability"
+    ).innerText;
+    const isDefault = applicabilityText === "All elements of the base markup";
+    const isProhibited =
+        applicabilityText ===
+        "All elements of the base markup except for some roles or elements that prohibit its use";
+    const isDeprecated =
+        applicabilityText === "Use as a global deprecated in ARIA 1.2";
```
pkra, line 64: From https://github.com/w3c/aria-common/pull/106#discussion_r1462542533

> I see, this is awful. Definitely, it makes more sense to start with a JSON object rather than looking for this text..

pkra:
After the individual comments above, here's the closing comment from @spectranaut's review at https://github.com/w3c/aria-common/pull/106#pullrequestreview-1837579645

> Ok Peter I took another crack at going over this -- I wonder how much you would be interested in renaming functions or variables in this PR? :)

pkra:
>  I wonder how much you would be interested in renaming functions or variables in this PR? :)

I'm interested and I will try to get to this before the editors' call next week. (fingers crossed)

spectranaut:
whoo!! I'll take a look tomorrow!

spectranaut:
Ok @pkra -- what do you think should be the goal here? How close of a review do you want at this stage?

Should we just merge this, as an intermediate step, then decide how to clean up more in the future? I'm fine with merging it as is without a close review, as it's just clean up and the spec is the same.

pkra:
> Should we just merge this, as an intermediate step, then decide how to clean up more in the future? 

That would be my suggestion. That way, we can start fixing the bugs we already know about (cf. the editors' board's "blocked" column). And of course we should continue to refactor it (incl. changing the spec to simplify the post-processing).

> I'm fine with merging it as is without a close review, as it's just clean up and the spec is the same.

I did my best to test that, diffing local respec builds before/after. The local respec builds would always fail to find the funders.html include so I checked against the gh-pages branch just now, too.

pkra:
I've moved the TODOs from the top card to #2501

Issue #2394, opened 2024-12-09
Text-to-background contrast in tables under ARIA's Definition of Roles too low


stiiin:
The headings of the tables under the [Definition of Roles](https://w3c.github.io/aria/#role_definitions) sections are unreadable under dark mode. This is because the `th` elements have `color: var(--text)` (through `base.css`), which computes to `color: #ddd` (when `dark.css` is applied), and `background-color: #eee` (through [this repository's `common.css`](https://github.com/w3c/aria/blob/5973a12f7d225be85be774fc4c2093c8e5572843/common/css/common.css#L93)).

pkra:
> A solution for this either lies in, or is blocking on, #2226.

Could you leave a comment on the other issue? Then we can close this to avoid duplicate issues.

stiiin:
Commented as requested.

Issue #2229, opened 2024-05-28
Move all AAM specs to use w3c's "data tables"


spectranaut:
We are trying to rely on less custom CSS in all of our specs.

One fix will be to remove the CSS related to the AAM tables.  We can use the W3C data tables instead: [respec CSS for def tables](https://www.w3.org/StyleSheets/TR/2021/README#def)

I made an experimental PR to do this on CORE-AAM: https://github.com/w3c/core-aam/pull/193/files

pkra:
Oh, this is the issue! I'll close #2435

pkra:
@spectranaut do you mind if I do this for all AAM specs in #2410?

spectranaut:
that would be awesome, thanks! :)

pkra:
> that would be awesome, thanks! :)

Already done, waiting for review 😉

PR #2410, opened 2025-01-09
editorial: revise custom stylesheets

### LINKED ISSUE #2226: remove spec specific css or inline it

pkra:
Resolves #2226

- [x] remove common.css
  - [x] reduce css as per https://github.com/w3c/aria/issues/2226#issuecomment-2136035094
  - [x] .image => remove after switching markup to figure
  - [x] remove all other styles
  - [x] remove stylesheet and switch to respec options
    - [x] add custom styles to accname to make up for nested OL styles
    - [x] adjust tables
       - [x] aria: replace with .def and .data
       - [x] accname: 1 table => .data
       - [x] core-aam: tables => .data (follow-up: #2462)
       - [x] dpub-aam: tables => .data (follow-up: #2462)
       - [x] dpub-aria: replace with .def
       - [x] graphics-aam: tables => .data (follow-up: #2462)
       - [x] graphics-aria: replace with .data
       - [x] html-aam: tables => .data (follow-up: #2462)
       - [x] mathml-aam: tables => .data (follow-up: #2462)
       - [x] pdf-aam: tables => .data (follow-up: #2462)
       - [x] svg-aam: tables => .data (follow-up: #2462)
       - [x] also update graphics2.html and aria-requirements.html
     - [x] remove common.css
     - [x] remove mapping-tables.css
     - [x] accname: add custom style block for list styles (follow-up: #2462)
  - [x] review color contrast issues
    - => https://github.com/speced/respec/issues/4871
    - [x] dark mode: `th{   background-color:#eee;  }` (cf. #2394 + duplicates)
      - will be resolved by dropping common.css (cf. above)
    - [x] syntax highlighting (light mode: IDL - dark mode: almost every code example) => #2433
      - [x] track: switch to more accessible highlight.js palette => #2433
      - [x] track incorrect language detection such as `#example-48` => https://github.com/speced/respec/wiki/ReSpec-Editor's-Guide#examples => #2433 
- [x] revise core-aam.css
  - [x] reduce css as per https://github.com/w3c/aria/issues/2226#issuecomment-2136038060
  - [x] remove core-aam.css as they are covered via common.css
- [x]  pdf-aam: remove link to (now removed) core-aam.css 
- [x] accname styles
  - [x] check https://github.com/w3c/aria/issues/2226#issuecomment-2136045884 => less than reduced common.css for aria [the stylesheet has already been dropped]
- [x] mapping-tables.css
  - [x] replace with `.data` as per #2435 ,  #2229 [cf. notes above]
- [x] editors.css - remove
- [x] slices.css - remove
- [x] html-aam.css - remove

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2410.html" title="Last updated on Mar 1, 2025, 11:49 AM UTC (b4f33bb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2410/786f112...b4f33bb.html" title="Last updated on Mar 1, 2025, 11:49 AM UTC (b4f33bb)">Diff</a>


pkra:
notes to self:

- the reduction of common.css in https://github.com/w3c/aria/issues/2226#issuecomment-2136035094  includes more declarations than the reductions in https://github.com/w3c/aria/issues/2226#issuecomment-2136038060 (core-aam) and https://github.com/w3c/aria/issues/2226#issuecomment-2136045884 (accname). So we can start there.
- core-aam loads both core-aam.css and common.css so we can reduce core-aam.css further

pkra:
It turns out, core-aam only sets styling that are already set in common.css.

pkra:
@jnurthen @spectranaut @daniel-montalvo I think I actually want to suggest to remove all of common.css. 

Almost all of it is minor stylistic differences to what respec does (i.e., everything after the table styles).

For tables, I would suggest to switch to respec's `table.def` styles. Very different but that's what they're for and it seems better to stick with respec for maintenance.

The only other bit are the nested lists styles. However, only accname actually needs these (and I feel like respec or W3C TR styles should be doing a better job here - but they probably don't want to risk it but perhaps be open to opt-in classes). 

[For completeness: there are 2 nested lists in aria in 4.3.2 - all other nested lists in specs are from respec TOCs where the styles are set by respec.]

I'll prep something for the next editors meeting so we can go through them together.

pkra:
Discussed in editors' call https://www.w3.org/2025/01/22-aria-editors-minutes.html#1b79

tl;dr I will continue this PR to effectively remove common.css

pkra:
@jnurthen looking through https://w3c.github.io/tr-design/src/README I see 1-2 other table styles: `.data` and `.index` (which is like `.data` but more cramped).  They look more like our current tables, but have noticeably more whitespace.

Are there others you had in mind?

jnurthen:
> Are there others you had in mind?

I think I was thinking about the default style without data for some of the tables at the end of the doc

pkra:
As per today's call, I will switch the other specs to using respec table classes. 

Then the editors can review each spec.

@MelSumner could you take a look at the nested list classes I've added to accname 42082b0? @spectranaut suggested it might be better to design the lists in a way so that new nested lists can be added without having to remember the right class. It seems prudent to ask a spec editor to do that :smile:

pkra:
accname has just 1 table. `.data` seems a good fit.

pkra:
For the AAM's, I'd have to adjust mapping-tables.css when dropping common.css. 

Instead I'll tackle https://github.com/w3c/aria/issues/2229 and remove mapping-table.css while adding respec's `.data` class to tables.

pkra:
This is ready for review.

While rebasing I fixed a typo in html-aam (extraneous closing parenthesis in the img alt section).

accname/index.html
```diff
@@ -143,6 +143,11 @@
         xref: ["dom", "core-aam-1.2", "wai-aria-1.2", "infra"],
       };
     </script>
+    <style>
+      .ol--upperAlpha{ list-style:upper-alpha; } /* replaces: ol ol */
+      .ol--lowerRoman { list-style:lower-roman; } /* replaces: ol ol ol */
+      .ol--lowerAlpha { list-style:lower-alpha; } /* replaces: ol ol ol ol */
```
spectranaut, line 148: Can't we do something like putting the class "computation_algorithm" [on the rool `ol`](https://github.com/w3c/aria/pull/2410/commits/f9522dc93db95b251cc6f100b2cfd17ee1274b4a#diff-c6c1825243b3e3b856bcd53c55febe98bf24dd51e27debb8ec71d71fa98a9a23L433) and then do:

```
.computation_algorithm ol ol { list-style:upper-alpha;  }
.computation_algorithm ol ol ol { list-style:lower-roman; }
.computation_algorithm ol ol ol ol {l ist-style:lower-alpha; }
```

And remove those custom classes on the nest `ol`s, so no one has to worry about adding then when they make a nested list in a computation?

spectranaut (changes requested):
I have one request about the CSS for nested lists, I'd like @MelSumner to weigh in as she is the visual editor of that spec :)

Also -- I think we should have, perhaps a follow up issue, to make all the data tables in the aams use `class=def` to get the blue definition tables: https://www.w3.org/StyleSheets/TR/2021/README#def

pkra:
> Also -- I think we should have, perhaps a follow up issue, to make all the data tables in the aams use `class=def` to get the blue definition tables: https://www.w3.org/StyleSheets/TR/2021/README#def

@spectranaut ugh, sorry, that's my fault. I misread #2229 and thought you wanted the `.data` style. 

I'll update the PR tomorrow.

accname/index.html
pkra: Two minor notes: 

- if `.computation_algorithm` goes on an OL, this needs to drop one ol from each line.
- then these styles would still affect the two lists in the Note and Example

As I said on the call, I have no strong feelings either way. But we can also work this out later. (I'm a little worried about merge conflicts as I'm touching all the specs.)

scottaohara:
i'm not really a fan of the blue tables for the aam specs.   i don't mind it in the ARIA spec - because there's also so much more to that spec than just the tables.  but for the AAMs, I'm personally not of the opinion this is a good design.   i'm actually much more of a fan of the `.data` class.  if that could be done and the first column could have a standardized width (around 200px looks like it'd work), the specs look at lot closer to how they do now - and everything isn't blue with blue links on them (which i know pass contrast because i checked - but i still find them not the easiest to read)

that said - if others really want the blue tables, then we will go with the majority.  i'm just not a fan.

pkra:
Thanks, @scottaohara. I had used `.data` at first but @spectranaut had pointed out in the review comment that  #2229 specifically asked for `.def`, so I updated it again. FWIW, I don't care either way.

accname/index.html
spectranaut: oh right. I forgot about the Notes and Example lists. Ok fine to land and change our minds later, if need be.

spectranaut:
@pkra @scottaohara I don't actually have a strong opinion on blue tables or not. The data tables look fine to me -- Peter do you want to drop the commit that changes it from data to def? Sorry for the extra work :)

pkra:
As per discussion on the editors list, I will 

- revert the accname change (and add the old list styles)
- file follow up issues to allow follow-ups for AAM tables and accname lists

Then merge.

pkra:
Oop, didn't see https://github.com/w3c/aria/pull/2410#issuecomment-2691356723 -- I'll do that then.

pkra:
The changes now match reviews from @spectranaut, @scottaohara, @rahimabdi 

Follow up issue: https://github.com/w3c/aria/issues/2462

PR #2413, opened 2025-01-24
ARIA IDL updates: convert eligible attributes to enumerated, new ARIA IDL guidance and examples

### LINKED ISSUE #2279: Remove old note in section 6.3.4 ARIA nullable DOMString Attribute
# Describe the change

During the [June 27 2024 "undefined" deep dive discussion](https://github.com/w3c/aria/issues/2177#issuecomment-2226727206), the ARIA WG largely agreed that the following note in [section 6.3.4 ARIA nullable DOMString Attribute](https://w3c.github.io/aria/#enumerated-attribute-values-html) should be removed because it is no longer relevant:

> Note: As of ARIA 1.2, all ARIA attributes exposed via IDL are defined as nullable [DOMString](https://webidl.spec.whatwg.org/#idl-DOMString)s. This matches the current implementation of all major rendering engines. This specification change should result in no implementation changes; it will merely represent the current reality of web engines. However, in a future draft, the ARIA Working Group intends to change several ARIA attributes to non-nullable DOMStrings, and seek implementations. The proposed change will bring ARIA into alignment with the HTML’s usage of [enumerated attributes](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute).

Since the ARIA IDL shipped in ARIA v1.2 with nullable `DOMString`s, this note may be outdated and unplanned as a future update.

https://w3c.github.io/aria/

Link to documentation:
https://w3c.github.io/aria/#enumerated-attribute-values-html

Yes.

### LINKED ISSUE #2281: Add missing/invalid value defaults to attribute characteristics tables
# Describe the change

During the [June 27 2024 "undefined" deep dive discussion](https://github.com/w3c/aria/issues/2177#issuecomment-2226727206), @cookiecrook proposed (ostensibly with WG agreement) expanding every ARIA attribute's characteristics table to include a value for both [missing value default](https://html.spec.whatwg.org/#missing-value-default) and [invalid value default](https://html.spec.whatwg.org/#invalid-value-default). Some states/properties specify an "Implicit Value for Role" which is the missing value default; note that the missing/invalid value defaults don't reflect to IDL as specified in the characteristics tables (see [6.3.2 IDL reflection of ARIA attributes](https://w3c.github.io/aria/#idl-reflection-attribute-values)).

Per @jnurthen, the table in [section 9. Handling Author errors](https://w3c.github.io/aria/#document-handling_author-errors) should not be changed/modified because validators rely on them to surface author errors.

Link to documentation: https://w3c.github.io/aria

Yes.

rahimabdi:
Closes #2281 
Closes #2279 

* [ ]  add enumerated attribute definitions for the following attributes including missing value default, invalid value default, keywords/states and canonical keyword:
  * aria-atomic
  * aria-autocomplete
  * aria-busy
  * aria-checked
  * aria-current
  * aria-disabled
  * aria-expanded
  * aria-haspopup
  * aria-hidden
  * aria-invalid
  * aria-live
  * aria-modal
  * aria-multiline
  * aria-multiselectable
  * aria-orientation
  * aria-pressed
  * aria-readonly
  * aria-required
  * aria-selected
  * aria-sort
 * [ ] Revise "A. Mapping WAI-ARIA Value types to languages" to include enumerated attributes
 * [ ] Revise "6.2.4 Value" for permissible values (remove all other types except "enumerated"?)
 * [ ] Update "6.3 ARIA Attributes"
   * Remove obsolete note in "6.3.4 ARIA nullable DOMString Attributes" about ARIA transitioning to non-nullable DOMString?
   * Remote spec guidance around getting/setting to only apply to non-enumerated attributes
   * Provide updated examples for "6.3.4.1 Example Attribute Usage"

* [ ] "author MUST" tests:
* [ ] "user agent MUST" tests: https://github.com/web-platform-tests/interop-accessibility/issues/177
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:
  - https://developer.mozilla.org/en-US/docs/Glossary/Enumerated

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2413.html" title="Last updated on Mar 6, 2025, 6:50 PM UTC (509091f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2413/a2d34fe...509091f.html" title="Last updated on Mar 6, 2025, 6:50 PM UTC (509091f)">Diff</a>

### DIFF
--- a/common/css/common.css
+++ b/common/css/common.css
@@ -99,10 +99,10 @@ caption{
 	margin:1em 0 0.1em;
 	padding:0 0 0 0.3em;
 }
-table.role-features th, table.role-features td, table.state-features th, table.state-features td, table.property-features th, table.property-features td, table.value-descriptions th, table.value-descriptions td {
+table.role-features th, table.role-features td, table.state-features th, table.state-features td, table.property-features th, table.property-features td, table.value-descriptions th, table.value-descriptions td, table.state-descriptions th, table.state-descriptions td {
 	min-width:20em;
 }
-table.role-features tbody th, table.state-features tbody th, table.property-features tbody th, table.value-descriptions tbody th {
+table.role-features tbody th, table.state-features tbody th, table.property-features tbody th, table.value-descriptions tbody th, table.state-descriptions tbody th {
     text-align: left !important;
 }
 th+th, td+td{
--- a/index.html
+++ b/index.html
@@ -12335,7 +12335,7 @@ button.ariaPressed; // null</pre>
 					</tr>
 					<tr>
 						<th class="property-value-head" scope="row">Value:</th>
-						<td class="property-value"><a href="#valuetype_true-false">true/false</a></td>
+						<td class="property-value"><a href="#valuetype_enumerated">Enumerated</a></td>
 					</tr>
 				</tbody>
 			</table>
@@ -12343,21 +12343,26 @@ button.ariaPressed; // null</pre>
 				<caption>Values:</caption>
 				<thead>
 					<tr>
-						<th scope="col">Value</th>
+						<th scope="col">Value (keyword)</th>
+                        <th scope="col">State</th>
 						<th scope="col">Description</th>
 					</tr>
 				</thead>
 				<tbody>
 					<tr>
-						<th class="value-name" scope="row"><strong class="default">false (default)</strong></th>
+						<th class="value-name" scope="row"><strong class="default">false, empty string (<code>&quot;&quot;</code>)</strong></th>
+						<th class="state-description" scope="row">False</th>
 						<td class="value-description">Only one item can be selected.</td>
 					</tr>
 					<tr>
 						<th class="value-name" scope="row">true</th>
+						<th class="state-description" scope="row">True</th>
 						<td class="value-description">More than one item in the widget can be selected at a time.</td>
 					</tr>
 				</tbody>
 			</table>
+			<p>The attribute's <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the False state.</p>
+			<p>For the purposes of <a data-cite="html/common-dom-interfaces.html#reflect">reflection</a>, the canonical keyword for the False state is the "false" keyword.</p>
 		</div>
 		<div class="property" id="aria-orientation">
 			<pdef>aria-orientation</pdef>
--- a/index.html
+++ b/index.html
@@ -11913,7 +11913,7 @@ button.ariaPressed; // null</pre>
 					</tr>
 					<tr>
 						<th class="state-value-head" scope="row">Value:</th>
-						<td class="state-value"><a href="#valuetype_token">token</a></td>
+						<td class="property-value"><a href="#valuetype_enumerated">Enumerated</a></td>
 					</tr>
 				</tbody>
 			</table>
@@ -11921,29 +11921,36 @@ button.ariaPressed; // null</pre>
 				<caption>Values:</caption>
 				<thead>
 					<tr>
-						<th scope="col">Value</th>
+						<th scope="col">Value (keyword)</th>
+                        <th scope="col">State</th>
 						<th scope="col">Description</th>
 					</tr>
 				</thead>
 				<tbody>
 					<tr>
 						<th class="value-name" scope="row">grammar</th>
+						<th class="state-description" scope="row">Grammar</th>
 						<td class="value-description">A grammatical error was detected.</td>
 					</tr>
 					<tr>
-						<th class="value-name" scope="row"><strong class="default">false (default)</strong></th>
+						<th class="value-name" scope="row"><strong class="default">false, empty string (<code>&quot;&quot;</code>)</strong></th>
+						<th class="state-description" scope="row">False</th>
 						<td class="value-description">There are no detected errors in the value.</td>
 					</tr>
 					<tr>
 						<th class="value-name" scope="row">spelling</th>
+						<th class="state-description" scope="row">Spelling</th>
 						<td class="value-description">A spelling error was detected.</td>
 					</tr>
 					<tr>
 						<th class="value-name" scope="row">true</th>
+						<th class="state-description" scope="row">True</th>
 						<td class="value-description">The value entered by the user has failed validation.</td>
 					</tr>
 				</tbody>
 			</table>
+			<p>The attribute's <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> is the False state, and its <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> is the True state.</p>
+			<p>For the purposes of <a data-cite="html/common-dom-interfaces.html#reflect">reflection</a>, the canonical keyword for the False state is the "false" keyword.</p>
 		</div>
 		<div class="property" id="aria-keyshortcuts">
 			<pdef>aria-keyshortcuts</pdef>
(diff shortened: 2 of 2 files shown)


rahimabdi:
Adding this PR for discussion at the Feb 27 2025 ARIA WG meeting.

common/css/common.css
```diff
@@ -99,10 +99,10 @@ caption{
 	margin:1em 0 0.1em;
 	padding:0 0 0 0.3em;
 }
-table.role-features th, table.role-features td, table.state-features th, table.state-features td, table.property-features th, table.property-features td, table.value-descriptions th, table.value-descriptions td {
+table.role-features th, table.role-features td, table.state-features th, table.state-features td, table.property-features th, table.property-features td, table.value-descriptions th, table.value-descriptions td, table.state-descriptions th, table.state-descriptions td {
```
pkra, line 102: This change will cause a conflict with #2410. Could you use a respec class name instead?

pkra (changes requested):

index.html
```diff
@@ -10492,29 +10496,33 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 						<td class="value-description">Assistive technologies will present only the changed node or nodes.</td>
 					</tr>
 					<tr>
 						<th class="value-name" scope="row">true</th>
+						<th class="state-description" scope="row">True</th>
 						<td class="value-description">Assistive technologies will present the entire changed region as a whole, including the author-defined label if one exists.</td>
 					</tr>
 				</tbody>
 			</table>
+			<p>The attribute's <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the Missing state.</p>
```
keithamus, line 10525: HTML has a concept of states that don't map to a keyword, but each state must have a definition, the definition allows mapping back to a keyword, or null. As an example the `popover` attribute has a missing value default of [no-popover state](https://html.spec.whatwg.org/multipage/popover.html#attr-popover-none-state) which let's the spec say things like "if the popover attribute is in the no-popover state".

Additionally, traditional enumerated attributes will map an invalid state back to a particular keyword; for example `popover`'s invalid attribute state is the `manual` state, which maps to the keyword `"manual"`. It's important to know what the invalid state maps back to because it determines what value is returned in the IDL for those values.

Note: not all missing values return null (e.g. the `kind` attribute on `<track>` elements has a missing value default of the subtitles state, so a track's `.kind` property returns `"subtitles"` even if the attribute is not present, another example is the `type` attribute on `<input>` which defaults to `"text"`).


With that said it's possible to be in "no state". In fact, an enumerated attribute without a missing value default will be in "no state", as will invalid. No state, AIUI, will always map to null.

So here, the simplest thing for us to do here, rather than the implied missing state is to remove this, and lean on the _lack of_ a missing/invalid state returning the no state. IOW this line could simply be removed:

```suggestion
```

If you think missing the definitions create more ambiguity, then we can instead do something like:

```suggestion
			<p>The attribute has no <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> or <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a>.</p>
```

index.html
```diff
@@ -10826,6 +10849,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 					</tr>
 				</tbody>
 			</table>
+			<p>The attribute's <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the Undefined state.</p>
```
keithamus, line 10852: The `<tr>` for the `"undefined"` keyword is missing a `<th>` to name its state. In other words there should be a line above L10848 with `<th class="state-description" scope="row">Undefined</th>`. (I cannot comment on L10848 sadly).

I don't think we want the missing value default here to return the Undefined state, that will mean the `.ariaChecked` property will return `"undefined"` when `aria-checked` isn't on the element. I presume we want to return `null` instead? Browsers currently return null, so missing should be "no state", i.e. removed:

```suggestion
			<p>The attribute has no <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and its <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the Undefined state.</p>
```

index.html
```diff
@@ -11386,29 +11420,34 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 					</tr>
 				</tbody>
 			</table>
 			<table class="value-descriptions">
 				<caption>Values:</caption>
 				<thead>
 					<tr>
-						<th scope="col">Value</th>
+						<th scope="col">Value (keyword)</th>
+                        <th scope="col">State</th>
```
keithamus, line 11432: Looks like a mix of tabs/spaces here.

index.html
```diff
@@ -11744,30 +11788,37 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
 						<td class="value-description">Indicates the popup is a <rref>grid</rref>.</td>
 					</tr>
 					<tr>
 						<th class="value-name" scope="row">dialog</th>
+						<th class="state-description" scope="row">Dialog</th>
 						<td class="value-description">Indicates the popup is a <rref>dialog</rref>.</td>
 					</tr>
 				</tbody>
 			</table>
+			<p>The attribute's <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> is the Undefined state, and its <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> is the Menu state.</p>
```
keithamus, line 11821: Again, here, the missing value default being the `Undefined` state means `.ariaHasPopup` will return `"undefined"` which I don't think is desirable. Instead:

```suggestion
			<p>The attribute has no <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a>, and its <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> is the Menu state.</p>
```

keithamus (changes requested):
@rahimabdi I've left a couple of review notes which I think apply in a lot of places, but rather than inundating you with review comments hopefully you can take the couple here and pattern match across the rest of the spec.

spectranaut:
Discussed briefly in today's ARIA meeting: https://www.w3.org/2025/03/06-aria-minutes#cd2e

pkra:
@rahimabdi I'm unable to help since this is on your fork. This looks like it will be a bit messy - very sorry!

rahimabdi:
@pkra Would it be easier to create a fresh branch from current main? I can re-do all of the updates and copy over the comments.

pkra:
> @pkra Would it be easier to create a fresh branch from current main? I can re-do all of the updates and copy over the comments.

@rahimabdi I suspect the answer is yes. Sorry again, also to @keithamus for creating additional review work.

rahimabdi:
@pkra No worries, thanks!

common/css/common.css
rahimabdi: Got it. The new PR (https://github.com/w3c/aria/pull/2484) does not modify common.css.

index.html
rahimabdi: Fixed in new PR: https://github.com/w3c/aria/pull/2484/files.

index.html
rahimabdi: Fixed in new PR: https://github.com/w3c/aria/pull/2484.

rahimabdi:
As this PR precedes substantial spec style sheet work (https://github.com/w3c/aria/pull/2410), I've opened a new PR branched from current main: https://github.com/w3c/aria/pull/2484.

Closing.

PR #2417, opened 2025-01-29
Allow aria-valuetext for use with combobox role

### LINKED ISSUE #711: aria-valuetext would be useful on many more roles

### LINKED ISSUE #2382: Consider allowing aria-valuetext for combobox role
In attempting to decide upon a way for the customizable select element to calculate the value to expose to users, use cases arose where allowing an author to set `aria-valuetext` on the `select` element (role=comobox) instead of having to fiddle with the returned markup from the select element's chosen option, seemed like a good idea.

Consider the following:

```
<!-- this option was specifically marked up so that the value would be "smile icon 1" rather than something like "smile smile icon 1" -->
<option>
  <span role=img aria-label="smile icon 1">🙂</span>
  <span class=description aria-hidden=true>smile icon 1</span>
</option>
```
In the collapsed state, the description is set to display none so that only the emoji is displayed. 

The following is a lot easier for a developer to do to get the option to announce in the intended way - but the returned content to the selectedoption element will only be the emoji, not the aria-label
```
<option aria-label="smile icon 1">🙂</option>
```

so, if someone could do
```
<select aria-valuetext="smile icon 1">
  <button><selectedcontent>🙂</selectedcontent></button>
  ...
</select>
```
that'd be a lot nicer than having to fiddle with very specific markup patterns, ensuring that every bit was marked up properly to convey the intended value when interacting with the collapsed select and when reviewing individual options.

cc @smhigley @aleventhal

scottaohara:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2417--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2417--wai-aria.netlify.app%2Findex.html)

closes #2382
and is related to #2344 / #2369 

This PR adds aria-valuetext as a supported property for the combobox role.

the `aria-valuetext` definition is slightly updated to indicate that the attribute can be used on other supported roles.  I'm not sure if we want to do any further updates to that section in this PR - or if that really should be handled in a larger PR to resolve https://github.com/w3c/aria/issues/711

* [ ] "author MUST" tests: n/a
* [X] "user agent MUST" tests: we can't currently write tests
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=313141
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=2034547
   * Blink: https://issues.chromium.org/issues/392649975
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2417.html" title="Last updated on Mar 7, 2025, 5:00 PM UTC (4c77ae3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2417/40f1062...4c77ae3.html" title="Last updated on Mar 7, 2025, 5:00 PM UTC (4c77ae3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -16014,10 +16014,7 @@ button.ariaPressed; // null</pre
           <div class="property-description">
             <p><a>Defines</a> the human readable text alternative of <pref>aria-valuenow</pref> for a range <a>widget</a>, or a text alternative for the value of custom <abbr>ARIA</abbr> or native widgets that support the attribute.</p>
             <p>This property is used, for example, on a range widget such as a slider or progress bar.</p>
-            <p>
-              If the <pref>aria-valuetext</pref> attribute is set, authors SHOULD also set the <pref>aria-valuenow</pref> attribute, unless that value is unknown (for example, on an indeterminate
-              <rref>progressbar</rref>).
-            </p>
+            <p>If the <pref>aria-valuetext</pref> attribute is set on a range <a>widget</a>, authors SHOULD also set the <pref>aria-valuenow</pref> attribute, unless that value is unknown (for example, on an indeterminate <rref>progressbar</rref>).</p>
             <p>
               Authors SHOULD only set the <pref>aria-valuetext</pref> attribute when the rendered value cannot be meaningfully represented as a number. For example, a slider might have rendered values
               of <code>small</code>, <code>medium</code>, and <code>large</code>. In this case, the values of <pref>aria-valuenow</pref> could range from 1 through 3, which indicate the position of
--- a/index.html
+++ b/index.html
@@ -2843,6 +2843,7 @@
               following:
             </p>
             <ul>
+              <li>If the <code>aria-valuetext</code> attribute is specified on the element with the <code>combobox</code> role, the value of the attribute is the value of the <code>combobox</code> element.</li>
               <li>
                 If the <code>combobox</code> element is a host language element that provides a value, such as an HTML <code>input</code> element, the value of the combobox is the value of that
                 element.
@@ -2949,6 +2950,7 @@
                     <li><sref>aria-invalid</sref></li>
                     <li><pref>aria-readonly</pref></li>
                     <li><pref>aria-required</pref></li>
+                    <li><pref>aria-valuetext</pref></li>
                   </ul>
                 </td>
               </tr>
@@ -16012,7 +16014,7 @@ button.ariaPressed; // null</pre
         <div class="property" id="aria-valuetext">
           <pdef>aria-valuetext</pdef>
           <div class="property-description">
-            <p><a>Defines</a> the human readable text alternative of <pref>aria-valuenow</pref> for a range <a>widget</a>.</p>
+            <p><a>Defines</a> the human readable text alternative of <pref>aria-valuenow</pref> for a range <a>widget</a>, or other widget roles which support the attribute.</p>
             <p>This property is used, for example, on a range widget such as a slider or progress bar.</p>
             <p>
               If the <pref>aria-valuetext</pref> attribute is set, authors SHOULD also set the <pref>aria-valuenow</pref> attribute, unless that value is unknown (for example, on an indeterminate
--- a/index.html
+++ b/index.html
@@ -16012,7 +16012,7 @@ button.ariaPressed; // null</pre
         <div class="property" id="aria-valuetext">
           <pdef>aria-valuetext</pdef>
           <div class="property-description">
-            <p><a>Defines</a> the human readable text alternative of <pref>aria-valuenow</pref> for a range <a>widget</a>, or a text alternative for the value of supported custom or native widgets that support the attribute.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -13300,7 +13302,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
 		<div class="property" id="aria-valuetext">
 			<pdef>aria-valuetext</pdef>
 			<div class="property-description">
-				<p><a>Defines</a> the human readable text alternative of <pref>aria-valuenow</pref> for a range <a>widget</a>.</p>
```
aleventhal, line 13303: I suppose this kind of makes it sound like it creates the human-readable version of aria-valuenow for combobox, but aria-valuenow doesn't work in a combobox.

scottaohara: yeah.  you're right...  going to need to change this.  thanks

aleventhal (approved):

index.html
```diff
@@ -13300,7 +13302,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+				<p><a>Defines</a> the human readable text alternative of <pref>aria-valuenow</pref> for a range <a>widget</a>, or a text alternative for the value of custom <abbr>ARIA</abbr> or native widgets that support the attribute.</p>
 				<p>This property is used, for example, on a range widget such as a slider or progress bar.</p>
 				<p>If the <pref>aria-valuetext</pref> attribute is set, authors SHOULD also set the <pref>aria-valuenow</pref> attribute, unless that value is unknown (for example, on an indeterminate <rref>progressbar</rref>).</p>
```
spectranaut, line 13307:
```suggestion
				<p>If the <pref>aria-valuetext</pref> attribute is set on a range <a>widget</a>, authors SHOULD also set the <pref>aria-valuenow</pref> attribute, unless that value is unknown (for example, on an indeterminate <rref>progressbar</rref>).</p>
```

spectranaut (approved):
Looks good to me, one suggestion to clarify a SHOULD in the aria-valuetext description.

scottaohara:
> Looks good to me, one suggestion to clarify a SHOULD in the aria-valuetext description.

thank you!

adampage (approved):
Looks great. 🚀

pkra:
@scottaohara could you update the top card on what remains to do before merging this?

pkra:
@scottaohara I've tried to be very careful while rebasing this. I hope everything came out ok.

spectranaut:
@scottaohara this doesn't actually add aria-valuetext to the list of supported states and properties on combobox, it just updates the prose to not be range-specific.

EDIT: It was accidentally dropped in the prettier update :sweat_smile: I've now added it back!

lolaodelola:
Hi folks, I've updated the OP with the links to both the Gecko and Webkit bugs.

jcsteh:
Sorry for the late feedback, but I only learned about this today.

I'm concerned about what would happen if aria-valuetext was used on an editable combobox. When a control is editable, its value and editable text really should be the same. As I understand it, most screen readers will read the editable text for an editable combobox, not the value. If the aria-valuetext is different, that's going to lead to confusion for authors, who will expect that aria-valuetext will be reported.

I think aria-valuetext should be disallowed on editable comboboxes. Alternatively, there could be a rule that it must be the same as the editable text content, but that seems redundant and error prone.

spectranaut:
@mcking65 Hi Matt, when this was last discussed in the ARIA working group, you said you would open a PR against this PR. There was also a request to limit this to select only comoboxes. Can you make that update?

mcking65:
While working on a review and proposed changes, I developed serious questions about whether this PR is necessary and whether it would be a desirable direction. I shared my concerns and questions in [this comment on issue 2382](https://github.com/w3c/aria/issues/2382#issuecomment-5449799335).

front-endian:
As [discussed in today's meeting](https://github.com/w3c/aria/pull/2417#issuecomment-5529972132) I'm going to help Scott by taking a look at this in the coming week, adding any guardrails we agreed on to this PR, add notes if needed for authors about appropriate/dangerous usage, and will talk to @mcking65 about what changes are needed for acc-name too.

PR #2419, opened 2025-01-30
Add "Name from Content" accname step for ::marker pseudo-element

rahimabdi:
Closes https://github.com/w3c/accname/issues/203

Describe Change Here!

The accname algorithm should account for `::marker` CSS-generated content. This PR for provides spec guidance on ensuring `::marker` content (that is visible and where the current node supports it) is included in the accname computation.  

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:


accname/index.html
```diff
@@ -635,8 +635,12 @@ <h4>Computation steps</h4>
                       <span id="step2F.ii"><!-- Don't link to this legacy numbered ID. --></span><em>Name From Generated Content:</em> Check for
                       <abbr title="Cascading Style Sheets">CSS</abbr> generated textual content associated with the <code>current node</code> and include it in the <code>accumulated text</code>. The
                       <abbr title="Cascading Style Sheets">CSS</abbr> <a href="https://www.w3.org/TR/CSS2/generate.html#before-after-content"><code>::before</code> and <code>::after</code></a> pseudo
-                      elements [[!CSS2]] can provide textual content for [=element|elements=] that have a content model.
+                      elements [[!CSS2]] can provide textual content for [=element|elements=] that have a content model, in addition to [=element|elements=] that support the <a href="https://www.w3.org/TR/css-pseudo-4/#selectordef-marker"><code>::marker</code> pseudo element if it is visually rendered.
```
scottaohara, line 638: Not opposed to the current wording, but this is how my mind had to translate what was said.  just throwing it out as a suggestion as alternate wording.  

```suggestion
                      elements [[!CSS2]] can provide textual content for [=element|elements=] that have a content model. The CSS <a href="https://www.w3.org/TR/css-pseudo-4/#selectordef-marker"><code>::marker</code> pseudo element</a> can also provide textual content for [=element|elements=] that support the pseudo element.
```

accname/index.html
```diff
@@ -635,8 +635,12 @@ <h4>Computation steps</h4>
...
                       <ol>
+                        <li id="comp_name_from_content_pseudo_element_marker">
+                          For <code>::marker</code> pseudo elements, <a class="termref">User agents</a> MUST append <abbr title="Cascading Style Sheets">CSS</abbr> textual content, without a space,
+                          to the textual content of the <code>current node</code> if the <code>::marker</code> is visually rendered and the <code>current node</code> supports <code>::marker</code>.
```
scottaohara, line 642: just not sure about "visually rendered" here, per my previous comment.  but otherwise think this is fine.

MelSumner (approved):
I think the visually rendered part is fine, especially since this is related to CSS (which is all about what's visually rendered).

scottaohara:
> I think the visually rendered part is fine, especially since this is related to CSS (which is all about what's visually rendered.

but as i've already mentioned, there are properties in CSS that do _not_ visually render content (in that they are rendered, but not visible), but they are still included in the accName.

one can't even make the argument that "well they still take up space in the layout, so that counts as their visual rendering" because `visibility: hidden` content also still takes up space in the visual rendering of the layout - but THAT content is similar to `display: none` and not exposed to the a11y tree by default.

accname/index.html
rahimabdi: @scottaohara  Agreed. My only concern would be for lists, where `::marker` is automatically generated depending on `list-style-type` so I'm curious about the scenario where `list-style-type: none` visually suppresses the marker (and perhaps list semantics). For `::marker` only (and not other CSS properties like `opacity`, `scale`), should it be conveyed if it's not visually present?

rahimabdi:
> > I think the visually rendered part is fine, especially since this is related to CSS (which is all about what's visually rendered.
> 
> but as i've already mentioned, there are properties in CSS that do _not_ visually render content (in that they are rendered, but not visible), but they are still included in the accName.
> 
> one can't even make the argument that "well they still take up space in the layout, so that counts as their visual rendering" because `visibility: hidden` content also still takes up space in the visual rendering of the layout - but THAT content is similar to `display: none` and not exposed to the a11y tree by default.

Thank you @scottaohara @MelSumner. I'm happy to remove the "visually rendered" part. Scott brings up a good point that "visually rendered" isn't existing accname spec language so my only outstanding concern is with treatment of `list-style-type: none` and if `::marker` should participate in accname for that case. 

Also, since `list-style` supports unicode characters, a string value and even images (`list-style-image`), I'm curious if you think it would beneficial to call these out as well.

spectranaut:
@accdc ping to review :)

accdc (approved):
Thank you!

accname/index.html
scottaohara: sorry it took so long for me to get back to this.

imo, i wouldn't expect the marker/list-style-type to participate if it wasn't rendered, because that's how it works now.  if someone were to set an `ol` for instance to `list-style-type: none` then if the list is still exposed as such, the list items are just exposed as that, with no announcement of the markers that were removed.  i think that should be maintained.

rahimabdi:
Thank you @scottaohara, I've removed the "if visually rendered" part and simplified the step language.

Requesting a final round of reviews from @accdc @MelSumner @scottaohara.

scottaohara (approved):

accdc (approved):

MelSumner (approved):
:shipit:

PR #2425, opened 2025-02-03
editorial: doc-pagelist section has been moved in the correct alphabetical-order

giacomo-petri:
Closes: https://github.com/w3c/dpub-aria/issues/55

Very minor editorial change.

From the comparison, it may not be immediately clear, but the change simply repositions doc-pagelist to its correct alphabetical order. In [the W3C DPUB-ARIA 1.1 role definitions](https://www.w3.org/TR/dpub-aria-1.1/#role_definitions), the previous ordering was incorrect.

### DIFF
--- a/dpub-aria/index.html
+++ b/dpub-aria/index.html
@@ -3282,105 +3282,6 @@
         </div>
 
         <div class="role">
-          <rdef>doc-pagelist</rdef>
-
-          <div class="role-description">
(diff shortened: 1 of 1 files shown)


pkra (approved):
git is being very annoying with this change.

spectranaut (approved):

spectranaut:
This is editorial, has two reviews, so is good to merge

PR #2466, opened 2025-03-06
PDF-AAM initial HTML tables import

zakkinsey:
Closes PDF-AAM #7

This PR consists of the initial import of the HTML tables for the PDF-AAM mappings. My understanding is with this PR we will officially have our first public working draft of the PDF-AAM. 

**I am unsure what, if any, tests need to be run for this PR.** I expect at this stage most, or all, of these tests do not need to be run, please correct me if that assumption is incorrect. 

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

### DIFF
--- a/pdf-aam/index.html
+++ b/pdf-aam/index.html
@@ -53,7 +53,7 @@
           },
           {
             name: "Zak Kinsey",
-            company: "Target Stream",
+            company: "TargetStream",
             companyURL: "https://www.targetstream.com",
             w3cid: "160220",
           },
--- a/pdf-aam/index.html
+++ b/pdf-aam/index.html
@@ -4,9 +4,15 @@
     <title>PDF Accessibility API Mappings 1.0</title>
     <meta content="text/html; charset=utf-8" http-equiv="Content-Type" />
     <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
-    <script src="../common/script/resolveReferences.js" class="remove"></script>
-    <script src="../common/script/utility.js" class="remove"></script>
-    <script src="../common/biblio.js" class="remove" defer></script>
+    <script src="common/script/resolveReferences.js" class="remove"></script>
+    <script src="common/script/utility.js" class="remove"></script>
+    <script src="common/biblio.js" class="remove" defer></script>
+
+    <link href="common/css/mapping-tables.css" rel="stylesheet" type="text/css" />
+    <link href="common/css/common.css" rel="stylesheet" type="text/css" />
+
+    <!-- note: may need to move the core-aam style sheet to common at some point -->
+    <link href="../core-aam/css/core-aam.css" rel="stylesheet" type="text/css" />
 
     <script class="remove">
       var respecConfig = {
@@ -53,7 +59,7 @@
           },
           {
             name: "Zak Kinsey",
-            company: "TargetStream",
+            company: "Target Stream",
             companyURL: "https://www.targetstream.com",
             w3cid: "160220",
           },
@@ -74,7 +80,7 @@
           },
           {
             name: "Zak Kinsey",
-            company: "Target Stream",
+            company: "TargetStream",
             companyURL: "https://www.targetstream.com",
             /* todo: add w3cid */
           },
@@ -199,7 +205,7 @@
         <section id="mapping_role_table">
           <h3>Role Mapping Tables</h3>
           <h4 id="role-map-example"><code>example</code></h4>
-          <table class="data" aria-labelledby="role-map-example">
+          <table aria-labelledby="role-map-example">
             <tbody>
               <tr>
                 <th>ARIA Specification</th>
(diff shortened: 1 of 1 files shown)


spectranaut (changes requested):
Hi Zak!
A few things to get this landed:
1) Remove 3 references to CSS files, these files no longer exist (a change that just landed recently)
2) We are now using the respec `class="data"` on all the mapping tables, so you will want to add that to all of the pdf-aam mapping tables.
3) When there is a wai-aria analygous role, and the mapping should be the same, rather than repeat the mapping I think it would be best to do as HTML-AAM does, and say "See WAI-ARIA mappings", for example: https://w3c.github.io/html-aam/#el-a
4) As for the sections in the beginning, rather than duplicate I think it is best to reference when relevant.

pdf-aam/index.html
```diff
@@ -238,6 +246,12070 @@ <h4 id="role-map-example"><code>example</code></h4>
...
+                      </td>
+                  </tr>
+                  <tr>
+                      <th>
+                          <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+                      </th>
+                      <td>
+                          <span class="property">AXRole: <code>TODO</code></span><br>
+                          <span class="property">AXSubrole: <code>TODO</code></span><br>
+                      </td>
```
cookiecrook: Rather than duplicating the platform mappings here, I'd expect we'd want to point to the place (or as a ReSpec include) where generic is defined in the ARIA spec.

pdf-aam/index.html
```diff
@@ -238,6 +246,12070 @@ <h4 id="role-map-example"><code>example</code></h4>
...
+                  </tr>
+                  <tr>
+                      <th><abbr title="Notes">Notes:</abbr></th>
+                      <td>
+                          <span class="property">Note from PDFa: <code>We may need to wordsmith the description</code></span><br>
```
cookiecrook: it's not clear to me what description this is referencing.

pdf-aam/index.html
```diff
@@ -238,6 +246,12070 @@ <h4 id="role-map-example"><code>example</code></h4>
...
+          </table>
+          
+          
+          <h4 id="role-map-example"><code>Note Placement = Inline</code></h4>
+          <table aria-labelledby="role-map-example">
+              <tbody>
+                  <tr>
+                      <th>ARIA Specification</th>
+                      <td>
+                          <a class="role-reference" href="https://www.w3.org/TR/core-aam-1
```
cookiecrook: Consider adding these as smaller PRs... 5 or 10 roles at a time perhaps... Or a reasonable set of related role... Rather than a 10k-line diff.

cookiecrook (changes requested):

zakkinsey:
> Hi Zak! A few things to get this landed:
> 
> 1. Remove 3 references to CSS files, these files no longer exist (a change that just landed recently)
> 2. We are now using the respec `class="data"` on all the mapping tables, so you will want to add that to all of the pdf-aam mapping tables.
> 3. When there is a wai-aria analygous role, and the mapping should be the same, rather than repeat the mapping I think it would be best to do as HTML-AAM does, and say "See WAI-ARIA mappings", for example: https://w3c.github.io/html-aam/#el-a
> 4. As for the sections in the beginning, rather than duplicate I think it is best to reference when relevant.

Discussed in PDF-AAM Task Force meeting today:
Relating to your numbered items above: 
1: Daniel has helped remove those. Thanks Daniel!
2: I will work on getting my export script updated to add 'class="data"' to all the tables

Then in general as discussed on the call, this import's main intent was to build the template/framework for all of the tables. Fill out the structure to start opening issues against in the PDF-AAM repo and future PRs will be much more concise. As we are also trying to move from our spreadsheet into doing this work here on GitHub, having the base structure built out enables us to more effectively decouple from the old worksheet. 

In general, the TODOs were left to be flags so we could in fact reference "See WAI-ARIA mappings" either HTML-AAM or Core-AAM as the situation dictates. I am still unclear of how to properly do that reference, or in general best ways to reference re-spec tasks. But I expect that to be done in its own PR. 

4: Agreed.

pdf-aam/index.html
zakkinsey: As per our discussion below, initial import will be large so work can be moved to GH repo from spreadsheet, then issues will be opened for other tasks for smaller future PRs.

pdf-aam/index.html
zakkinsey: Agreed. Will do in future PRs after initial import.

pdf-aam/index.html
```diff
@@ -191,33 +141,40 @@ <h2>General rules</h2>
 
         <section id="mapping_role_table">
           <h3>Role Mapping Tables</h3>
-          <h4 id="role-map-example"><code>example</code></h4>
+          <h4 id="role-map-example"><code>Document Logical document within a parent document</code></h4>
           <table class="data" aria-labelledby="role-map-example">
```
daniel-montalvo, line 145:
```suggestion
          <h4 id="role-document-child"><code>Document (Child of a document)</code></h4>
```

pdf-aam/index.html
```diff
@@ -191,33 +141,40 @@ <h2>General rules</h2>
...
             <tbody>
```
daniel-montalvo, line 146: And then the aria-labelledby reference here should be updated to the id  of the h4.

pdf-aam/index.html
```diff
@@ -191,33 +141,40 @@ <h2>General rules</h2>
 
         <section id="mapping_role_table">
           <h3>Role Mapping Tables</h3>
-          <h4 id="role-map-example"><code>example</code></h4>
+          <h4 id="role-map-example"><code>Document Logical document within a parent document</code></h4>
```
cookiecrook, line 144:
```suggestion
          <h4 id="role-map-example"><code>document</code> Logical document within a parent document</h4>
```

Beyond the formatting diff, are there multiple types of documents you're trying to represent?

cookiecrook: For the most part, you'll want the output to use a similar source/structure as these tables in HTML-AAM or DPUB-AAM, which will get you a similar final ReSpec rendering to this:

ReSpec rending of the HTML a:link mapping, which cross-references ARIA's link role

cookiecrook: And then only in situations where there is no other mapping to through to would you define the system-specific API properties... Here's HTML-AAM's `abbr` as an example:

ReSpec rendering of the HTML-AAM abbr table

cookiecrook: And in that last case, the computed role would be something like `pdf-foo`

zakkinsey: Yes, at least one other entry, "Child of Structure Tree root", that we have identified as of now.

cookiecrook: There are probably places you'll want to point to the DPUB spec, too.

pdf-aam/index.html
```diff
@@ -191,33 +141,40 @@ <h2>General rules</h2>
...
-                  <span class="property">Role: <code>ROLE_SYSTEM_EXAMPLE</code></span
+                  <span class="property">Role: <code>Use WAI-ARIA mapping</code></span
                   ><br />
                 </td>
               </tr>
               <tr>
                 <th><abbr title="User Interface Automation">UIA</abbr></th>
                 <td>
-                  <span class="property">Control Type: <code>Example</code></span
+                  <span class="property">Control Type: <code>Use WAI-ARIA mapping</code></span
```
cookiecrook, line 170:
```suggestion
<div class="general">Use WAI-ARIA mapping</div>
```
You don't need the soft line breaks or `<br />` either, FWIW

pdf-aam/index.html
```diff
@@ -191,33 +141,40 @@ <h2>General rules</h2>
...
                   ><br />
                 </td>
               </tr>
               <tr>
                 <th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
                 <td>
-                  <span class="property">Role: <code>ROLE_EXAMPLE</code></span
+                  <span class="property">Role: <code>Use WAI-ARIA mapping</code></span
```
cookiecrook, line 177:
```suggestion
<div class="general">Use WAI-ARIA mapping</div>
```

pdf-aam/index.html
```diff
@@ -226,81 +183,549 @@ <h4 id="role-map-example"><code>example</code></h4>
                   <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
                 </th>
                 <td>
-                  <span class="property">AXRole: <code>AXxample</code></span
+                  <span class="property">AXRole: <code>Use WAI-ARIA mapping</code></span
+                  ><br />
+                  <span class="property">AXSubrole: <code>Use WAI-ARIA mapping</code></span
+                  ><br />
```
cookiecrook, line 189:
```suggestion
                  <div class="general">Use WAI-ARIA mapping</div>
```

pdf-aam/index.html
```diff
@@ -226,81 +183,549 @@ <h4 id="role-map-example"><code>example</code></h4>
...
+                </td>
+              </tr>
+              <tr>
+                <th><abbr title="Notes">Notes:</abbr></th>
+                <td>
+                  <span class="property"
+                    >Note from PDFa:
+                    <code
+                      >Added note - question for the group... A screen reader shouldn't "announce" when it hits the first Document tag but should it "announce" when it hits "nested" (subsequent)
+                      Document tags, to tell someone they are in a "second" document? AND... the same question applies to Document Fragment, too. If "Yes," how do we indicate that in the AAM?</code
```
cookiecrook, line 199: Normative AT requirements are out of scope for the AAM docs, but you can put in informative hints. For example, AT might treat this like a separate document, such as they do with an iframe.

pdf-aam/index.html
```diff
@@ -226,81 +183,549 @@ <h4 id="role-map-example"><code>example</code></h4>
...
+              <tr>
+                <th>Analogous HTML Element</th>
+                <td>
+                  <span class="property" href="https://www.w3.org/TR/html-aam-1.0/#el-section ">Role: <code>section </code></span><br />
+                </td>
+              </tr>
+              <tr>
+                <th>MSAA + IAccessible2</th>
+                <td>
+                  <span class="property">Role: <code>Use WAI-ARIA mapping</code></span
```
cookiecrook, line 227: ditto for this sectino... you don't need to list "Use another mapping" at the API property level... (Role, Control Type, AXRole, etc.) Just add it as text:

```html
<div class="general">Use WAI-ARIA mapping</div>
```

zakkinsey:
@cookiecrook & @daniel-montalvo My most recent commit attempts to apply the advice you all listed here. You will notice that we likely still need to figure out better id names for the link structure elements. Also, the last entry, "attribute" definitely needs to be changed/updated, I'm just not sure of the best way to do this. I will discuss it today with @PaulRayius in our PDFa meeting.

pdf-aam/index.html
zakkinsey: I think we will need to apply something along these lines for attribute entries, specifically the last entry in our initial import. @PaulRayius

pdf-aam/index.html
```diff
@@ -191,121 +141,567 @@ <h2>General rules</h2>
...
-          </div>
-        </section>
-      </section>
+          <h4 id="role-document"><code> Document (Child of Structure tree root)</code></h4>
+			<table class="data" aria-labelledby="role-document">
+				<tbody>
+					<tr>
+						<th>ARIA Specification</th>
+						<td>
+							<a class="role-reference" href="https://www.w3.org/TR/core-aam-1.2/#role-map-generic"><code>generic </code></a>
```
daniel-montalvo: If we want to link to core-aam here, we can use the fixCrosssReferences script.

More info on how to use these links is at https://github.com/w3c/aria/#special-elements-and-classes
```suggestion
							<a class="core-mapping" href="#role-map-generic"><code>generic </code></a>
```

pdf-aam/index.html
```diff
@@ -93,19 +79,20 @@
 
         // Spec URLs
         ariaSpecURLs: {
+          ED: "https://w3c.github.io/aria/",
```
daniel-montalvo, line 82: This correct the specURLs for cross-reference link.
```suggestion
          ED: "https://w3c.github.io/aria/",
          FPWD: "https://www.w3.org/TR/wai-aria-1.3",
          WD: "https://www.w3.org/TR/wai-aria-1.3",
```

pdf-aam/index.html
```diff
@@ -93,19 +79,20 @@
...
           CR: "https://w3c.github.io/aria/",
           CRD: "https://w3c.github.io/aria/",
           PR: "https://www.w3.org/TR/wai-aria/",
           REC: "https://www.w3.org/TR/wai-aria/",
         },
         accNameURLs: {
-          CR: "https://w3c.github.io/accname/",
+          ED: "https://w3c.github.io/accname/",
```
daniel-montalvo, line 91:
```suggestion
          ED: "https://w3c.github.io/accname/",
          FPWD: "https://www.w3.org/TR/accname-1.2/",
          WD: "https://www.w3.org/TR/accname-1.2/",
```

pdf-aam/index.html
```diff
@@ -93,19 +79,20 @@
...
           CRD: "https://w3c.github.io/accname/",
```
daniel-montalvo:
```suggestion
          CR: "https://www.w3.org/TR/accname-1.2/",
          CRD: "https://www.w3.org/TR/accname-1.2/",
```

pdf-aam/index.html
```diff
@@ -93,19 +79,20 @@
...
           PR: "https://www.w3.org/TR/accname-1.2/",
           REC: "https://www.w3.org/TR/accname-1.2/",
         },
```
daniel-montalvo, line 99:
```suggestion

		      coreMappingURLs: {
          ED: "https://w3c.github.io/core-aam/",
          FPWD: "https://www.w3.org/TR/core-aam-1.2/",
          WD: "https://www.w3.org/TR/core-aam-1.2/",
          CR: "https://www.w3.org/TR/core-aam-1.2/",
          CRD: "https://www.w3.org/TR/core-aam-1.2/",
          PR: "https://www.w3.org/TR/core-aam-1.2/",
          REC: "https://www.w3.org/TR/core-aam-1.2/",
		      },
        dpubModURLs: {
          ED: "https://w3c.github.io/dpub-aria/",
          FPWD: "https://www.w3.org/TR/dpub-aria-1.1/",
          WD: "https://www.w3.org/TR/dpub-aria-1.1/",
          REC: "https://www.w3.org/TR/dpub-aria-1.1/",
        graphicsMappingModURLs: {
          ED: "https://w3c.github.io/graphics-aam/",
          FPWD: "https://www.w3.org/TR/graphics-aam-1.0/",
          WD: "https://www.w3.org/TR/graphics-aam-1.0/",
          CR: "https://www.w3.org/TR/graphics-aam-1.0/",
          PR: "https://www.w3.org/TR/graphics-aam-1.0/",
          REC: "https://www.w3.org/TR/graphics-aam-1.0/",
        graphicsModURLs: {
          ED: "https://w3c.github.io/graphics-aria/",
          FPWD: "https://www.w3.org/TR/graphics-aria/",
          WD: "https://www.w3.org/TR/graphics-aria/",
          REC: "https://www.w3.org/TR/graphics-aria/",
        htmlMappingURLs: {
          ED: "https://w3c.github.io/html-aam/",
          FPWD: "https://www.w3.org/TR/html-aam-1.0/",
          WD: "https://www.w3.org/TR/html-aam/",
          REC: "https://www.w3.org/TR/html-aam/",
        practicesURLs: {
          ED: "https://www.w3.org/WAI/ARIA/apg/",
          WD: "https://www.w3.org/WAI/ARIA/apg/",
          FPWD: "https://www.w3.org/WAI/ARIA/apg/",
          CR: "https://www.w3.org/WAI/ARIA/apg/",
          REC: "https://www.w3.org/WAI/ARIA/apg/",
        },
```

pdf-aam/index.html
```diff
@@ -191,121 +141,567 @@ <h2>General rules</h2>
...
+					<tr>
+						<th>Analogous HTML Element</th>
+						<td>
+							<span class="property" href="https://www.w3.org/TR/html-aam-1.0/#el-body">Role: <code>body</code></span>
+						</td>
+					</tr>
+					<tr>
+						<th>MSAA + IAccessible2</th>
+						<td>
+							<div class="general">Use WAI-ARIA mapping</div>
```
daniel-montalvo: I'd prefer to avoid use of divs here. This may conflict with the aria-common scripts that we are using.

```suggestion
Use WAI-ARIA mapping
```

pdf-aam/index.html
```diff
@@ -191,121 +141,567 @@ <h2>General rules</h2>
...
+						</td>
+					</tr>
+					<tr>
+						<th><abbr title="User Interface Automation">UIA</abbr></th>
+						<td>
+						   <div class="general">Use WAI-ARIA mapping</div>
```
daniel-montalvo: Same as above

```suggestion
Use WAI-ARIA mapping
```

pdf-aam/index.html
```diff
@@ -191,121 +141,567 @@ <h2>General rules</h2>
...
+						</td>
+					</tr>
+					<tr>
+						<th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
+						<td>
+							<div class="general">Use WAI-ARIA mapping</div>
```
daniel-montalvo:
```suggestion
Use WAI-ARIA mapping
```

pdf-aam/index.html
```diff
@@ -191,121 +141,567 @@ <h2>General rules</h2>
...
+						</td>
+					</tr>
+					<tr>
+						<th>
+							<abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+						</th>
+						<td>
+							<div class="general">Use WAI-ARIA mapping</div>
```
daniel-montalvo:
```suggestion
Use WAI-ARIA mapping
```

pdf-aam/index.html
```diff
@@ -191,121 +141,567 @@ <h2>General rules</h2>
...
-          <div class="note" id="ftn.note1">
-            <p>
-              <sup>[Note 1]</sup>
-              Footnote content
-            </p>
-          </div>
-        </section>
-      </section>
+          <h4 id="role-document"><code> Document (Child of Structure tree root)</code></h4>
+			<table class="data" aria-labelledby="role-document">
```
daniel-montalvo, line 303: This table seems to be malformed. There are `tr` and `td`elements missing. An example of how this table could look like is below:

```
			<table class="data" aria-labelledby="role-document">
				<tbody>
					<tr>
						<th>ARIA Specification</th>
						<td>
							<a class="core-mapping" href="#role-map-generic"><code>generic </code></a>
						</td>
					</tr>
						<th>Analogous HTML Element</th>
							<td><a class="html-mapping" href="#el-body">Role: <code>body</code></a></td></tr>
						<tr><th>MSAA + IAccessible2</th>
							<td><div class="general">Use WAI-ARIA mapping</div></td></tr>
						<tr><th><abbr title="User Interface Automation">UIA</abbr></th>
						   <td>Use WAI-ARIA mapping</td></tr>
						<tr><th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
							<td>Use WAI-ARIA mapping</td></tr>
						<tr><th>
							<abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
						</th>
						<td>
							Use WAI-ARIA mapping
						</td></tr>
						<tr><th><abbr title="Notes">Notes:</abbr></th>
							<td><span class="property">Note from PDFa: <code>only processor instructions</code></span></td></tr>
				</tbody>
			</table>
```

daniel-montalvo (dismissed):
Overall looking much better. Just some more fine-tuning needed :-)

These changes should be applied across all tables.

pdf-aam/index.html
zakkinsey: Completed in commit: e847f1be072f48a48e37746045a6feef77f21859

pdf-aam/index.html
zakkinsey: Committing your suggestion as is.

pdf-aam/index.html
zakkinsey: Completed

pdf-aam/index.html
zakkinsey: This should be correct now. I did not find any missing structure elements.

pdf-aam/index.html
```diff
@@ -191,121 +188,467 @@ <h2>General rules</h2>
...
+						<th>ARIA Specification</th>
+						<td>
+						<!-- Guidance needed on how to handle this entry.-->
+							<a class="core-mapping" href="https://www.w3.org/TR/core-aam-1.2/#role-map-region   https://www.w3.org/TR/core-aam-1.2/#role-map-generic"><code>region IF there is an accessible name. Otherwise use generic</code></a>
+						</td>
+					</tr>
+					<tr>
+						<th>Analogous HTML Element</th>
+						<td>
+							<span class="property" href="https://www.w3.org/TR/html-aam-1.0/#el-section ">Role: <code>section </code></span>
```
daniel-montalvo, line 399:
```suggestion
							<a class="html-mapping" href="#el-body">Role: <code>body</code></a>
```
@zakkinsey This is how the link should look like if we want to link to the body role of the html aam. I think it'd be safer to work on the import script so that this is robust before landing this to avoid future issues.

spectranaut (changes requested):
This is starting to look really good!

My main feedback is about the TBD sections. I think these should be fixed before landing. 
For a lot of them, you can take inspiration from HTML-AAM, which refers to Core-AAM in the same way that PDF-AAM does/should.

1. I think you can remove the "Important Terms" section, and add it back if you need it later. 

2. For [PDF-AAM's](https://deploy-preview-2466--wai-aria.netlify.app/pdf-aam/#mapping) `2. Mapping WAI-ARIA to Accessibility APIs` can mostly copy [HTML-AAM's prose](https://w3c.github.io/html-aam/#mapping_general).

3. I think you can remove [Special Processing Requiring Additional Computation](https://w3c.github.io/core-aam/#mapping_additional), and add it back if PDF-AAM has "special processing require additional computation" for attributes or roles that don't just map to Core-AAM. For now, when you refer to Core-AAM, this Core-AAM section applies.

4. For the Actions section, if the Actions described in Core-AAM are applicable to a PDF, then it seems to me that you should actually refer to the Core-AAM here... or perhaps just copy it, and replace "DOM Element" with whatever the PDF concept of an element is?

5. For Privacy considerations, I don't know. Can a PDF with accessibility mark up lead to fingerprinting in a browser/over a network? If so then you need to include this section - and copy it with updated prose explaining that this can be done.

daniel-montalvo:
Thanks @spectranaut for your comments. @zakkinsey  is the prose something some in your team can have a look at or you require some help with that?

zakkinsey:
Yes, thank you @spectranaut! @daniel-montalvo we will take a first pass at this and if we run into anything we don't think we can handle we will reach out. I suspect we will lean towards removing sections we are unsure of for now and adding them back in later as our focus shifts to less on the AAM table mapping entries to more on these sections. Again, thanks for all the help and guidance through this process for us!

pdf-aam/index.html
cookiecrook: I think this thread is generally resolved now...

pdf-aam/index.html
cookiecrook: Re-opened as:
- https://github.com/w3c/pdf-aam/issues/18

Resolving this thread.

spectranaut (approved):
Approve landing this PR as draft to make future changes smaller and easier to review!

cookiecrook:
Remaining feedback raised as [new issues on PDF-AAM](https://github.com/w3c/pdf-aam/issues).

PR #2473, opened 2025-03-10
Editorial: Relative links for PR and branch netlify previews

daniel-montalvo:
Partially addresses w3c/aria#2242

This only picks up references to roles, states, and properties. Some definitions are still pointing to the editors draft or TR documents because they are not processed by the script due to the way they are written.

For example, `<a>assistive technologies</a>` would rely on [Xref](https://respec.org/xref) for processing.
We should decide if we are happy with landing this or we want another approach, like waiting for respec to process the files and then just search and replace all absolute links (no matter where they link to) with relative links.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -217,3 +217,6 @@ pip-log.txt
 
 #Mr Developer
 .mr.developer.cfg
+
+# Local Netlify folder
+.netlify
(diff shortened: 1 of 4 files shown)


pkra (approved):
Looks good!

PR #2474, opened 2025-03-11
Editorial: Point to "element reference attribute" definitions in HTML for IDREF attributes

dandclark:
https://github.com/whatwg/html/pull/10995 adds definitions for "single element reference" and "set of element references" attributes, formalizing the concept of attributes that reference other elements via ID.

In this change, update the "Mapping WAI-ARIA Value types to languages" to use these new definitions for "ID Reference" and "ID Reference List" attributes.

This whole section is non-normative.

***
<a href="https://pr-preview.s3.amazonaws.com/dandclark/aria/pull/2474.html" title="Last updated on Mar 17, 2025, 10:19 PM UTC (72adb11)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2474/5b667ad...dandclark:72adb11.html" title="Last updated on Mar 17, 2025, 10:19 PM UTC (72adb11)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -17168,14 +17168,13 @@ el.getAttribute("aria-label"); // Returns "Publish"</pre
         </tr>
         <tr>
           <td>ID reference</td>
-          <td>The value of a defined <a data-cite="html/dom.html#the-id-attribute">id attribute</a> on another element</td>
+          <td><a data-cite="html/common-microsyntaxes.html#single-element-reference">Single element reference attributes</a></td>
           <td><a data-cite="xmlschema11-2/#IDREF">IDREF</a></td>
         </tr>
         <tr>
           <td>ID reference list</td>
           <td>
-            The value of one or more defined <a data-cite="html/dom.html#the-id-attribute">id attributes</a> on other element(s), represented as
-            <a data-cite="html/common-microsyntaxes.html#space-separated-tokens">Space-separated tokens</a>
+            <a data-cite="html/common-microsyntaxes.html#set-of-element-references">Set of element references attributes</a>
           </td>
           <td><a data-cite="xmlschema11-2/#IDREFS">IDREFS</a></td>
         </tr>


rahimabdi:
Thank you @dandclark! In addition to the changes you proposed in alignment with HTML, I'm curious if there needs to be more extensive changes where the ARIA spec references "ID reference" (now "single element reference") and "ID reference list" (now "set of element references").

Other places where id reference and id reference list are mentioned include the "Values" table for ARIA attributes, the "Identifies" relationship definition in [Important Terms](https://w3c.github.io/aria/#terms), section 6.2.4 Value defining [Characteristics of States and Properties](https://w3c.github.io/aria/#state_prop_att) and 8.6.1 ID Reference Error Processing in [State and Property Attribute Processing](https://w3c.github.io/aria/#state_property_processing).

Questions for the ARIA WG:

- Should the [ARIA Values schema](https://w3c.github.io/aria/#typemapping) change the WAI-ARIA type for id reference/id reference list to align with HTML? This would also mean changing all instances of "id reference" to "single element reference", and "id reference list" would be changed to "set of element references" including attribute Values tables
- Are there any other changes that should be made to WAI-ARIA value types? For example, `aria-relevant` is the only attribute that is "token list". As an FYI, part of the IDL work in https://github.com/w3c/aria/issues/2281 entails removing the following WAI-ARIA types and making them "enumerated": true/false, tristate, true/false/undefined, token.

spectranaut:
> * Should the [ARIA Values schema](https://w3c.github.io/aria/#typemapping) change the WAI-ARIA type for id reference/id reference list to align with HTML? This would also mean changing all instances of "id reference" to "single element reference", and "id reference list" would be changed to "set of element references" including attribute Values tables

I think this is unrelated to this particular PR, but I also think it's a good question. The AAMs use "IDREF" language ([example](https://w3c.github.io/core-aam/#ariaControls)) which is inconsistent with ARIA's use of "id reference"/"id reference list". Since ARIA is host language agnostic, we don't need to use the HTML terms... and possibly we shouldn't, and we should just stick to "id reference" and "single element reference", but I'd like to hear why if you think we should change, Rahim!

alice:
> * Should the [ARIA Values schema](https://w3c.github.io/aria/#typemapping) change the WAI-ARIA type for id reference/id reference list to align with HTML?

I think it would make sense to at least name the WAI-ARIA type in a way that doesn't limit it to ID association, yeah. This was already an issue once the [IDL attributes](https://w3c.github.io/aria/#ARIAMixin) which take an `Element` or `FrozenArray<Element>` were added.

I think it might also make sense to add a section discussing (whatever name you decide) attributes under the [ARIA Attributes](https://w3c.github.io/aria/#aria-attributes) with more information about how these attributes are interpreted, possibly defining a concept like [refers to](https://whatpr.org/html/10995/common-microsyntaxes.html#single-element-reference-refers-to) or [_attr_-associated element(s)](https://whatpr.org/html/10995/common-microsyntaxes.html#get-the-attr-associated-element), which could be referred to from the AAMs.

Currently, for example, the [Core-AAM guidance for `aria-controls`](https://www.w3.org/TR/core-aam-1.2/#ariaControls) says

> Relation: IA2_RELATION_CONTROLLER_FOR points to accessible nodes matching IDREFs

I think we might want something like 

> Relation: IA2_RELATION_CONTROLLER_FOR points to accessible nodes matching elements referred to by `aria-controls`

OR

> Relation: IA2_RELATION_CONTROLLER_FOR points to accessible nodes matching `aria-controls`-associated elements

but, again, not necessarily using the HTML terminology; either way, where the term is defined in the ARIA spec and mapped out to host languages much like the attribute types are.

rahimabdi:
> I think this is unrelated to this particular PR, but I also think it's a good question. The AAMs use "IDREF" language ([example](https://w3c.github.io/core-aam/#ariaControls)) which is inconsistent with ARIA's use of "id reference"/"id reference list". Since ARIA is host language agnostic, we don't need to use the HTML terms... and possibly we shouldn't, and we should just stick to "id reference" and "single element reference", but I'd like to hear why if you think we should change, Rahim!

@spectranaut Agreed, except in the case when references point to a `Element` or `FrozenArray<>` as @alice noted.  Perhaps this could be "id or element reference" and "id reference or element list"?

> I think it would make sense to at least name the WAI-ARIA type in a way that doesn't limit it to ID association, yeah. This was already an issue once the [IDL attributes](https://w3c.github.io/aria/#ARIAMixin) which take an Element or FrozenArray<Element> were added.

@Alice I've raised a new issue to investigate this so as to unblock this PR. Thanks!
- https://github.com/w3c/aria/issues/2487

> I think it might also make sense to add a section discussing (whatever name you decide) attributes under the [ARIA Attributes](https://w3c.github.io/aria/#aria-attributes) with more information about how these attributes are interpreted, possibly defining a concept like [refers to](https://whatpr.org/html/10995/common-microsyntaxes.html#single-element-reference-refers-to) or [attr-associated element(s)](https://whatpr.org/html/10995/common-microsyntaxes.html#get-the-attr-associated-element), which could be referred to from the AAMs.

Added to an upcoming ARIA WG agenda for discussion.

rahimabdi (approved):
Thank you @dandclark.

Issue #2016, opened 2023-08-29
add term "accessibility ancestor"


pkra:
Extracted from https://github.com/w3c/aria/pull/2010#discussion_r1308602573

Tooltip currently references owning element receiving focus. #2010 will update that to accessibility parent. It seems like "accessiblity ancestor" could be more fitting - but we lack that term.

pkra:
Currently proposed in https://github.com/w3c/aria/pull/2483

PR #2094, opened 2023-12-17
Remove group as allowed child of tree

mcking65:
Closes #2014

Removes group as an allowed accessibility child of tree to fix inconsistency with definition of treeitem. Direct children of tree can only be parent nodes or end nodes. Leaf nodes must be a child of a group that is a child of a parent node treitem. Even in a virtualized tree, if the elements that represent the ancestry of rendered nodes were not present in the DOM, the tree would be inoperable.

# PR tracking
Check these when the relevant issue or PR has been made, OR after you have confirmed the
related change is not necessary (add N/A). Leave unchecked if you are unsure. Read the
[Process Document](https://github.com/w3c/aria/wiki/ARIA-WG-Process-Document/_edit) or
[Test Overview](https://github.com/w3c/aria/wiki/ARIA-Test-Overview) for more information.

* [x] Related Core AAM Issue/PR: not required
* [x] Related AccName Issue/PR: not required
* [x] Related APG Issue/PR: not required

# Implementation tracking

* [x] "author MUST" tests: https://github.com/act-rules/act-rules.github.io/issues/2309
* [x] "user agent MUST" tests: not required
* [x] Browser implementations: not required
* [x] Does this need AT implementations? no

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2094.html" title="Last updated on Feb 14, 2025, 8:49 AM UTC (694277d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2094/b5a3d7f...694277d.html" title="Last updated on Feb 14, 2025, 8:49 AM UTC (694277d)">Diff</a>

PR #2483, opened 2025-03-21
Clarify Accessibility Parent-Child Relationships for menu, group, and menuitem

### LINKED ISSUE #2438: Group as allowed acc child of role menu if acc child of menuitem
# Describe your concern

```
<div role="menu">
	<div role="group">
		<span role="menuitem">Item 1</span>
		<div role="group">
			<span role="menuitem">Item 2</span>
			<span role="menuitem">Item 3</span>
		</div>
	</div>
</div>
```

Per [ARIA specs](https://www.w3.org/TR/wai-aria-1.3/#menu), the role="menu" accepts group as acc child only if one of the following is true:
- [group](https://www.w3.org/TR/wai-aria-1.3/#group) with [accessibility child](https://www.w3.org/TR/wai-aria-1.3/#dfn-accessibility-child) [menuitem](https://www.w3.org/TR/wai-aria-1.3/#menuitem)
- [group](https://www.w3.org/TR/wai-aria-1.3/#group) with [accessibility child](https://www.w3.org/TR/wai-aria-1.3/#dfn-accessibility-child) [menuitemradio](https://www.w3.org/TR/wai-aria-1.3/#menuitemradio)
- [group](https://www.w3.org/TR/wai-aria-1.3/#group) with [accessibility child](https://www.w3.org/TR/wai-aria-1.3/#dfn-accessibility-child) [menuitemcheckbox](https://www.w3.org/TR/wai-aria-1.3/#menuitemcheckbox)

but it's unclear whether a group containing another group (group > (or with generic intervening) group) is acceptable. As written, it appears to be allowed as long as the parent group has at least one accessible child menuitem, but I'm not sure if that was the intended meaning.

giacomo-petri:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2483--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2483--wai-aria.netlify.app%2Findex.html)

Closes #2438

The current specifications for menu and menubar contradict those for menuitem, menuitemcheckbox, and menuitemradio.

Specifically, the following structure is allowed under the menu role but not when considering the menuitem role:
```
<div role="menu">
	<div role="group">
		<span role="menuitem">Item 1</span>
		<div role="group">
			<span role="menuitem">Item 2</span>
			<span role="menuitem">Item 3</span>
		</div>
	</div>
</div>
```
This PR clarifies that this structure is indeed allowed. 

Additionally, it explicitly clarifies in a note that the following example, where a group contains a button and an input as accessibility children, is not permitted. This prevents ambiguity by making it clear that having at least one menuitem as an accessibility child of the group does not justify including other unrelated elements, which was not clearly stated.
```
<div role="menu">
	<div role="group">
		<span role="menuitem">Item 1</span>
		<div role="group">
			<span role="menuitem">Item 2</span>
			<span role="menuitem">Item 3</span>
		</div>
		<button>Test</button>
		<input type="text aria-label="Test">
	</div>
</div>
```

This update refines the specifications for menu, menubar, menuitem, menuitemcheckbox, and menuitemradio roles consistently.

---

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2483.html" title="Last updated on Aug 7, 2025, 4:58 PM UTC (7048eb6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2483/f6dff36...7048eb6.html" title="Last updated on Aug 7, 2025, 4:58 PM UTC (7048eb6)">Diff</a>


index.html
```diff
@@ -6304,6 +6311,11 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
             </p>
             <p>To be <a>keyboard accessible</a>, authors SHOULD manage focus of descendants for all instances of this <a>role</a>, as described in <a href="#managingfocus">Managing Focus</a>.</p>
             <p>Elements with the role <code>menubar</code> have an implicit <pref>aria-orientation</pref> value of <code>horizontal</code>.</p>
+            <p class="note">
+              The <code>menubar</code> <a>role</a> doesn't allow for accessibility children roles other than <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, <rref>menuitemradio</rref>,
```
pkra, line 6315:
```suggestion
              The <code>menubar</code> <a>role</a> does not allow for accessibility children roles other than <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, <rref>menuitemradio</rref>,
```

index.html
```diff
@@ -6200,6 +6200,11 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
             </p>
             <p>To be <a>keyboard accessible</a>, authors SHOULD manage focus of descendants for all instances of this <a>role</a>, as described in <a href="#managingfocus">Managing Focus</a>.</p>
             <p>Elements with the role <code>menu</code> have an implicit <pref>aria-orientation</pref> value of <code>vertical</code>.</p>
+            <p class="note">
+              The <code>menu</code> <a>role</a> doesn't allow for accessibility children roles other than <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, <rref>menuitemradio</rref>,
+              <rref>separator</rref>, and <rref>group</rref>. In this context, the role <rref>group</rref> allows only <a>accessibility children</a> with a role <rref>group</rref>,
```
pkra, line 6205: The group role has 

> [...] when a group is used in the context of a listbox, for example, authors MUST limit its children to option elements.

I think this requirement here should be added there, too.

index.html
```diff
@@ -6304,6 +6311,11 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
             </p>
             <p>To be <a>keyboard accessible</a>, authors SHOULD manage focus of descendants for all instances of this <a>role</a>, as described in <a href="#managingfocus">Managing Focus</a>.</p>
             <p>Elements with the role <code>menubar</code> have an implicit <pref>aria-orientation</pref> value of <code>horizontal</code>.</p>
+            <p class="note">
+              The <code>menubar</code> <a>role</a> doesn't allow for accessibility children roles other than <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, <rref>menuitemradio</rref>,
+              <rref>separator</rref>, and <rref>group</rref>. In this context, the role <rref>group</rref> allows only <a>accessibility children</a> with a role <rref>group</rref>,
```
pkra, line 6316: The group role has 

> [...] when a group is used in the context of a listbox, for example, authors MUST limit its children to option elements.

I think this requirement here should be added there, too.

index.html
```diff
@@ -6520,7 +6535,7 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
             <p>A <rref>menuitem</rref> with a checkable state whose possible <span>values</span> are <code>true</code>, <code>false</code>, or <code>mixed</code>.</p>
             <p>
               Authors MUST ensure [=elements=] with <a>role</a> <code>menuitemcheckbox</code> are <a>accessibility children</a> of an element with role <rref>menu</rref>, <rref>menubar</rref>, or an
-              element with role <rref>group</rref> that is an <a>accessibility child</a> of an element with role <rref>menu</rref> or <rref>menubar</rref>.
+              element with role <rref>group</rref> that is an <a>accessibility child</a> of an element with role <rref>group</rref>, <rref>menu</rref> or <rref>menubar</rref>.
```
pkra, line 6538: I think this still allows for `group > group > menuItem`. 

If we resolve #2016, it might be enough to require "accessibility ancestor of menu or menubar".

index.html
```diff
@@ -6632,7 +6648,7 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
             <p>A checkable <rref>menuitem</rref> in a set of elements with the same role, only one of which can be checked at a time.</p>
             <p>
               Authors MUST ensure [=elements=] with <a>role</a> <code>menuitemradio</code> are <a>accessibility children</a> of an element with role <rref>menu</rref>, <rref>menubar</rref>, or an
-              element with role <rref>group</rref> that is an <a>accessibility child</a> of an element with role <rref>menu</rref> or <rref>menubar</rref>.
+              element with role <rref>group</rref> that is an <a>accessibility child</a> of an element with role <rref>group</rref>, <rref>menu</rref> or <rref>menubar</rref>.
```
pkra, line 6651: Same as above.

pkra (dismissed):

index.html
giacomo-petri: > I think this still allows for group > group > menuItem.
yes, it looks like people would like to use menu > group > group > menuitem, so I adjusted the PR accordingly

e.g.
```
<menu>
        <div role="group">
		<span role="menuitem">Item 1</span>
		<div role="group">
			<span role="menuitem">Item 2</span>
			<span role="menuitem">Item 3</span>
		</div>
	</div>
</menu>
```

index.html
giacomo-petri: answered above

index.html
giacomo-petri: Thanks @pkra 
Since it only contained an example, I initially didn't add this one to the group, but it’s probably worth including. I’ve added the menu and menubar scenario for better clarity as per your suggestion.

index.html
giacomo-petri: Added both menu and menubar with the commit above.

index.html
pkra: Thanks. I meant that this seems to be technically valid

```html
<div role="group"> 
  <div role="group">
    <span role="menuitem">Item</span>
  </div>
</div>
```

giacomo-petri: Right, sorry I didn't get it. 

What about adding at the end:

> All group elements containing menuitem (_for all the menuitem roles_) MUST be accessibility descendant an element with role menu or menubar.

pkra: That's one way. (I think adding a term for accessibility ancestor might make things easier to understand.)

giacomo-petri: I'll try using accessibility ancestor. However, both my approach and yours might be too broad. Accessibility child enforces stricter constraints by allowing only generic intervening elements, whereas accessibility ancestor applies a broader scope. Could using accessibility ancestor permit more than intended?

<div role="menu">
    <div role="navigation">
        <div role="group">
            <div role="group">
                <span role="menuitem">Item</span>
            </div>
        </div>
    </div>
</div>

This seems currently allowed.

Maybe we should add something like:

> Additionally, any top-level group element with at least one accessibility descendant <rref>menuitem</rref> MUST be an <a>accessibility child</a> of an element with role <rref>menu</rref> or <rref>menubar</rref>.

pkra: > This seems currently allowed.

But menu would not allow navigation as an accessibility child, right?

I think if we combine accessibility children with ancestor, we should be able to create this kind of scoping. But we might need more explicit content-model language.

pkra: > Exactly, menu would not allow navigation; that's why the "top-level group" wording.

I meant that menu already prevents the example, without the additional wording. 

But yes, the wording would work around 

```html
<body>
 <div role="navigation">
        <div role="group">
            <div role="group">
                <span role="menuitem">Item</span>
            </div>
        </div>
    </div>
</body>
```

being valid. 

I still think "menuitem MUST have an accessibility ancestor menu or menubar" would be easier to understand, especially since the opening line of the role basically says so: "An option in a set of choices contained by a menu or menubar."

Another term that could be helpful: "intervening" (as used for generics): menus (of various types) contain items (of various types) but allow intervening groups.

Anyway, we need another reviewer :wink:

pkra: Thanks, Giacomo. 

We currently have

>  The accessibility descendants of a DOM element are all DOM elements which correspond to descendants of the corresponding accessible object in the accessibility tree. 

Rough idea for ancestor:

> The accessibility ancestors of a DOM element are all DOM elements of which the given DOM element is an accessibility descendant.

giacomo-petri: Content updated as per items above.

index.html
```diff
@@ -16143,6 +16160,10 @@ <h2>Including Elements in the Accessibility Tree</h2>
       <section id="tree_relationships">
         <h2>Relationships in the Accessibility Tree</h2>
         <p>The following terms are used to describe relationships between <abbr title="Document Object Model">DOM</abbr> elements.</p>
+        <p>
+          The <dfn data-export="" data-lt="accessibility anestor|accessibility ancestors">accessibility ancestors</dfn> of a DOM element are all DOM elements of which the given DOM element is an
```
giacomo-petri, line 16164:
```suggestion
          The <dfn data-export="" data-lt="accessibility ancestor|accessibility ancestors">accessibility ancestors</dfn> of a DOM element are all DOM elements of which the given DOM element is an
```

spectranaut:
@np-at to review as well!

scottaohara:
actually, taking myself off as a reviewer since we have so many others.

giacomo-petri:
@pkra, I just realized the same issue affects tree/group/treeitem structures. Should we include this change in this PR or create a separate one?

pkra:
@giacomo-petri sounds good. Hopefully it's simpler since #2094 prevents group as child of tree.

index.html
```diff
@@ -4758,8 +4758,10 @@ <h2>Definition of Roles</h2>
...
-              However, when a <code>group</code> is used in the context of a <rref>listbox</rref>, for example, authors MUST limit its children to <rref>option</rref> elements. Therefore, proper
-              handling of <code>group</code> by authors and assistive technologies is determined by the context in which it is provided.
+              However, when a <code>group</code> is used in the context of a <rref>listbox</rref>, for example, authors MUST limit its children to <rref>option</rref> elements. Similarly, when a
+              <code>group</code> is used in the context of a <rref>menu</rref> or <rref>menubar</rref>, authors MUST limit its <a>accessibility descendants</a> to <rref>menuitem</rref>,
+              <rref>menuitemcheckbox</rref>, <rref>menuitemradio</rref> and <rref>separator</rref>, with only elements of role <rref>generic</rref>, <rref>group</rref>, and
+              <rref>none</rref> intervening. Therefore, proper handling of <code>group</code> by authors and assistive technologies is determined by the context in which it is provided.
```
pkra: Coming back to this I'm not sure about this addition. "MUST limit its accessibility descendants" seems too strong - groups can have other descendants because, e.g., menuitem does not have any restrictions on its descendants.

IIRC we're really after "donut scoping". If that's the case I'm wondering if this addition is necessary. Does donut scoping not follow from the new requirements on menuitem etc now that we have accessibility ancestor? 

If that's the case, then perhaps a note is more appropriate, describing this effect. (There could be a follow up issue to do the same with listbox and option.)

index.html
```diff
@@ -6200,6 +6202,10 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
             </p>
             <p>To be <a>keyboard accessible</a>, authors SHOULD manage focus of descendants for all instances of this <a>role</a>, as described in <a href="#managingfocus">Managing Focus</a>.</p>
             <p>Elements with the role <code>menu</code> have an implicit <pref>aria-orientation</pref> value of <code>vertical</code>.</p>
+            <p class="note">
+              The <code>menu</code> <a>role</a> allows only role <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, <rref>menuitemradio</rref>, and <rref>separator</rref> as
+              <a>accessibility descendants</a>, with only elements of role <rref>generic</rref>, <rref>group</rref>, and <rref>none</rref> intervening.
```
pkra: As above, I think menu can have other descendants (below e.g. menuitem). In which case this should be rephrased to describe the effects of donut scoping.

index.html
```diff
@@ -6304,6 +6312,10 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
             </p>
             <p>To be <a>keyboard accessible</a>, authors SHOULD manage focus of descendants for all instances of this <a>role</a>, as described in <a href="#managingfocus">Managing Focus</a>.</p>
             <p>Elements with the role <code>menubar</code> have an implicit <pref>aria-orientation</pref> value of <code>horizontal</code>.</p>
+            <p class="note">
+              The <code>menubar</code> <a>role</a> allows only role <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, <rref>menuitemradio</rref>, and <rref>separator</rref> as
+              <a>accessibility descendants</a>, with only elements of role <rref>generic</rref>, <rref>group</rref>, and <rref>none</rref> intervening.
```
pkra: Sames as above for menu.

index.html
pkra: Ah, I think what I wrote is not right either. Maybe children instead of descendants?  E.g., "if a group has an accessibility ancestor of role menu, then allowed children are ...".

index.html
```diff
@@ -6345,9 +6357,11 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
                 <td class="role-mustcontain">
                   <ul>
                     <!-- keep in sync with #menu -->
+                    <li><rref>group</rref> with <a>accessibility child</a> <rref>group</rref></li>
                     <li><rref>group</rref> with <a>accessibility child</a> <rref>menuitem</rref></li>
                     <li><rref>group</rref> with <a>accessibility child</a> <rref>menuitemradio</rref></li>
                     <li><rref>group</rref> with <a>accessibility child</a> <rref>menuitemcheckbox</rref></li>
+                    <li><rref>group</rref> with <a>accessibility child</a> <rref>separator</rref></li>
```
pkra: I think we should be able to reduce this to just group's again.

pkra (dismissed):

spectranaut:
Discussed in today's meeting: https://www.w3.org/2025/07/03-aria-minutes.html#3a9f

index.html
spectranaut: In the working group meeting, I suggested using "accessibility children", but now I realize that can't quite be used the way I suggested, so I took a try at re-writing this whole section, which doesn't solve Peter's problem of making it "less wordy" but it does make it more clear, at least to me:

> A set of user interface objects that is not intended to be included in a page summary or table of contents by assistive technologies.
> 
> Contrast with region, which is a grouping of user interface objects that will be included in a page summary or table of contents. If a section is significant enough to warrant inclusion in the web page's table of contents, the author SHOULD assign it a role of region or a standard landmark role.
> 
> Authors SHOULD use a group to form a logical collection of items in a widget, such as children in a tree widget forming a collection of siblings in a hierarchy. Authors MAY nest group elements.

> The proper handling of group by authors and assistive technologies is determined by the context in which it is provided. For example, when a group is the accessibility child of a menu or menubar, authors MAY nest additional groups, however, the accessibility children of the innermost group MUST be limited to menuitems, menuitemcheckboxs, menuitemradios and separators. Whereas when a group is the accessibility child of a listbox, authors MUST NOT nest groups and MUST limit the groups accessibility children to option elements.

I don't think "accessibility descendant" quite works here, because we never want to say something about all descendants.

index.html
spectranaut: Reading this I think I finally understand your concern @pkra. This is hard to write.

But since it's a note, maybe it can be more informal, here is a suggestion:
> Let's call the elements that are contained in a menu and have roles menuitems, menuitemcheckboxes, menuitemradios or separator the "items" of that menu. These items can be arranged for logical purposes with groups, or styling purposes with generics. However, no semantically meaningful elements other than group are allowed between the individual items of the menu and the element with role menu. This is similar to the definition and constraints of "accessibility children", however, along with allowing intervening elements of role generic or none, we allow intervening elements of role group.

Also, I realize, the "allowed accessibility children" isn't clear enough. "Allowed accessibility children" only tells you what the next accessibility child should be. So if we add to that list "group with the accessibility child of group", then it's not clear that the child of THAT group MUST be either a group or (....). Is it possible to say this without being recursive?

spectranaut (commented):
Thanks for this effort @giacomo-petri! I've contributed some ideas...

index.html
giacomo-petri: @pkra, @spectranaut

@spectranaut, I think your suggestion of replacing the term "accessibility descendants" with "accessibility children" reflects the intended requirement. The "first level of accessibility children" of a menu must consist only of menuitem, menuitemcheckbox, menuitemradio, or separator, and generic, group, or none are permitted as structural elements between these roles and the menu (but they do not count as "first-level accessibility children". While this is always true for generic and none, it's not inherently true for group, so we may want to add a note to clarify that explicitly).

To make this clearer, we may also add a note stating that other roles may appear as descendants of a menu, but they must not be accessibility children of the menu itself. We can refer to @spectranaut’s example involving the group role [in this comment](https://github.com/w3c/aria/pull/2483/files#r2183471596) with other examples to illustrate that point in a note.

index.html
giacomo-petri: Applied @spectranaut suggestion

index.html
```diff
@@ -4755,15 +4755,19 @@ <h2>Definition of Roles</h2>
           <rdef>group</rdef>
           <div class="role-description">
             <p>A set of user interface <a>objects</a> that is not intended to be included in a page summary or table of contents by <a>assistive technologies</a>.</p>
-            <p>Contrast with <rref>region</rref>, which is a grouping of user interface objects that will be included in a page summary or table of contents.</p>
             <p>
-              Authors SHOULD use a <code>group</code> to form a logical collection of items in a <a>widget</a>, such as children in a tree widget forming a collection of siblings in a hierarchy.
-              However, when a <code>group</code> is used in the context of a <rref>listbox</rref>, for example, authors MUST limit its children to <rref>option</rref> elements. Therefore, proper
-              handling of <code>group</code> by authors and assistive technologies is determined by the context in which it is provided.
-            </p>
+              Contrast with <rref>region</rref>, which is a grouping of user interface objects that will be included in a page summary or table of contents If a section is significant enough to
```
spectranaut:
```suggestion
              Contrast with <rref>region</rref>, which is a grouping of user interface objects that will be included in a page summary or table of contents. If a section is significant enough to 
```

Forgot a period :)

spectranaut (changes requested):
One minor thing!

index.html
spectranaut: Did you want to iterate on the note to add your suggestions, then? Would be nice to get this landed soon :)

giacomo-petri: @spectranaut @pkra,
please take another look to see if it works now.

index.html
giacomo-petri: updated together with menu

index.html
giacomo-petri: @pkra 

what about phrasing it as 
"<rref>group</rref> as intervening role"
?

My concern is that if we just list "group" without clarification, some authors, especially those who skip over the note, might incorrectly assume that this kind of structure is valid:
```
<div role="menu">
    <div role="group">
        <a href="...">...</a>
        <input type="checkbox" ...>
    </div>
</div>
```

Since WCAG notes are non-normative, there's a real chance the note will be overlooked entirely. By explicitly mentioning "intervening role," we at least raise a red flag and prompt them to think twice (and read the note).

spectranaut: I like that idea!

spectranaut (dismissed):
I approve, with removing the extra rows for group and just put "group as intervening role" -- thought I wonder now if we should define "intervening" somewhere...

index.html
spectranaut: looks good to me!

giacomo-petri:
@spectranaut, I agree that we likely need a definition for "intervening," but it's already used 7 times in the specs, though always referring to generic or none. That said, I think the current wording for group doesn't add complexity and still aligns with the intent of the others, even if the context and outcome differ. Should we consider trying adding that definition (since it also impacts generic intervening) in a separate PR?

index.html
giacomo-petri: Updated

index.html
```diff
@@ -4755,15 +4755,19 @@ <h2>Definition of Roles</h2>
...
-              handling of <code>group</code> by authors and assistive technologies is determined by the context in which it is provided.
-            </p>
+              Contrast with <rref>region</rref>, which is a grouping of user interface objects that will be included in a page summary or table of contents. If a section is significant enough to 
+	      warrant inclusion in the web page's table of contents, the author SHOULD assign it a <a>role</a> of<rref>region</rref> or a <a href="#landmark_roles">standard landmark role</a>.
+	    </p>
             <p>
-              Authors MAY nest <code>group</code> elements. If a section is significant enough to warrant inclusion in the web page's table of contents, the author SHOULD assign it a <a>role</a> of
-              <rref>region</rref> or a <a href="#landmark_roles">standard landmark role</a>.
+              Authors SHOULD use a <code>group</code> to form a logical collection of items in a <a>widget</a>, such as children in a tree widget forming a collection of siblings in a hierarchy.
```
```suggestion
              Authors SHOULD use a <code>group</code> to form a logical collection of items in a <a>widget</a>, such as children in a treeitem forming a collection of siblings in a hierarchy.
```

group is not allowed in tree, only treeitem (I realize this was not a change in this PR).

index.html
```diff
@@ -4755,15 +4755,19 @@ <h2>Definition of Roles</h2>
...
+	      Authors MAY nest <code>group</code> elements.
+	    </p>
+	    <p>
+	      The proper handling of <code>group</code> by authors and assistive technologies is determined by the context in which it is provided. For example, when a <code>group</code> is the 
+	      <a>accessibility child</a> of a <code>menu</code> or <code>menubar</code>, authors MAY nest additional <code>group</code> elements. However, the <a>accessibility children</a> of the
```
pkra: I think the new content (However, ...) is  good addition but  it should be a note and not using normative language. As my suggestion for tree widgets shows, it's easy for such language to be out of date.  I think normative statements on roles should be limited to their own sections.

index.html
```diff
@@ -6200,6 +6204,13 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
...
             <p>Elements with the role <code>menu</code> have an implicit <pref>aria-orientation</pref> value of <code>vertical</code>.</p>
+            <p class="note">
+	      The <code>menu</code> <a>role</a> allows only roles <rref>menuitem</rref>, <rref>menuitemcheckbox</rref>, <rref>menuitemradio</rref>, and <rref>separator</rref> as
+              <a>accessibility children</a>. These roles may be nested within structural elements with roles <rref>generic</rref>, <rref>group</rref>, or <rref>none</rref> that sit between them and the 
+	      <code>menu</code>. This allows authors to organize menus logically using groups or to style them using generic containers. However, aside from <rref>group</rref>, no other semantically 
+	      meaningful roles may intervene between the allowed child roles and the <code>menu</code>. This structure mirrors the concept of "accessibility children", but explicitly permits 
+	      <rref>group</rref> in addition to <rref>generic</rref> and <rref>none</rref> as intervening elements.
+            </p>
```
pkra, line 6183: I admit I'm still unhappy with this.

I don't like how the second sentence contradicts the first sentence's "allows only roles".  I don't like "structural" in "structural element" nor how generics are mentioned (which can always be added after all so here a reader might wonder what is special about mentioning them again).

I also do not like the duplication of this language in menubar; it will invariably fall out of sync.

I realize we cannot get to something as tight as we have for trees, but I think we should work on this concept rather than adding more complicated language like this throughout the spec.

Still, I don't want to keep blocking this.

index.html
```diff
@@ -6345,9 +6361,7 @@ <h4>Plain HTML or Polyfill DOM Result of the MathML Quadratic Formula</h4>
                 <td class="role-mustcontain">
                   <ul>
                     <!-- keep in sync with #menu -->
-                    <li><rref>group</rref> with <a>accessibility child</a> <rref>menuitem</rref></li>
-                    <li><rref>group</rref> with <a>accessibility child</a> <rref>menuitemradio</rref></li>
-                    <li><rref>group</rref> with <a>accessibility child</a> <rref>menuitemcheckbox</rref></li>
+                    <li><rref>group</rref> as intervening role</li>
```
pkra, line 6331: I second the idea of adding a definition (later).

pkra:
I'm not sure why my last review is marked "requesting changes" - I'm guessing the suggested change is to blame. 

I don't oppose merging this as is, despite remaining somewhat unhappy with the language.

MelSumner (dismissed):
I agree with @pkra reviews on this but maybe we can iterate and improve.

index.html
giacomo-petri: @pkra, I can try to tweak it a bit but the "MUST" condition is already in place for the listbox role. Should we move everything into a note and replace the "MUST" language with "has/have to"?

index.html
giacomo-petri: I think I can make it clearer and avoid the contradiction. I'll work on it.

> I also don’t like the duplication of this language in the menubar; it will inevitably fall out of sync.

@pkra, I believe we already have similar situations elsewhere. Honestly, I'm not sure how to resolve this, since it's very specific to the menu and menubar roles. The complexity here is higher than with the tree, and I don't really see a clean solution, without repeating stuff. We could add something unique to the group role, but that would significantly increase the complexity of the group role, and we'd end up with the same problem of pulling requirements from other roles into the group role.

giacomo-petri:
I've made some adjustments to address the concerns that were raised. This needs another round of review.

PR #2484, opened 2025-03-23
ARIA IDL updates: enumerated attribute conversions, new "enumerated" wai-aria type, IDL examples

### LINKED ISSUE #2279: Remove old note in section 6.3.4 ARIA nullable DOMString Attribute

### LINKED ISSUE #2281: Add missing/invalid value defaults to attribute characteristics tables

rahimabdi:
Closes #2281 
Closes #2279 

# Mini-Explainer

See [ARIA IDL updates mini-explainer](https://github.com/w3c/aria/wiki/ARIA-IDL-Updates-Mini%E2%80%90Explainer-(ARIA-PR-%232484)).

# Summary of Changes
* Convert following ARIA attributes to enumerated:
  * aria-atomic
  * aria-autocomplete
  * aria-busy
  * aria-checked
  * aria-current
  * aria-disabled
  * aria-expanded
  * aria-haspopup
  * aria-hidden
  * aria-invalid
  * aria-live
  * aria-modal
  * aria-multiline
  * aria-multiselectable
  * aria-orientation
  * aria-pressed
  * aria-readonly
  * aria-required
  * aria-selected
  * aria-sort
* Revise "Mapping WAI-ARIA Value types to languages" to include enumerated attributes
* Revise "6.2.4 Value" value types 
* Update "6.3 ARIA Attributes" section
* Remove obsolete note in "6.3.4 ARIA nullable DOMString Attributes" about ARIA transitioning to non-nullable DOMString?
* Update ARIA IDL usage examples in "6.3.4.1 Example Attribute Usage"
* Add note clarifying "Undefined" state vs. "undefined" string assignment 

# Feature Flag Testing in Safari Tech Preview

You can test the new ARIA IDL reflection behavior with a feature flag in [Safari Tech Preview 228](https://developer.apple.com/documentation/safari-technology-preview-release-notes/stp-release-228). Ensure that "Show features for developers" is enabled (Settings -> Advanced), and then enable the "Enumerated ARIA Attribute Reflection" flag (Settings -> Feature Flags, "JavaScript" section).

The following page can help with quick IDL testing: https://2553411.playcode.io.

* [x] "user agent MUST" tests: https://github.com/web-platform-tests/wpt/pull/54119
   * [x] WebKit: https://bugs.webkit.org/show_bug.cgi?id=296851  (Shipping behind feature flag)
   * [ ] Gecko:
   * [ ] Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:
* [ ] Further outreach?
   * [x] @rahimabdi outreach to Salesforce contributors (@gregwhitworth et al) re: potential webcompat
   * [ ] Salesforce response pending re: testing the new feature flag in Safari Tech Preview

# Meeting Notes
## ARIA WG & Salesforce Sync (Sept 30 2025)
- Presentation deck
  - [Salesforce/ARIA WG IDL sync (Google Slides)](https://docs.google.com/presentation/d/1exShOV7qZtRFztgP2V9so0kDuhcYCTS1O78e1CkBqfM/edit?usp=sharing)
- Attendees
    - @keithamus, @cookiecrook, @rahimabdi, @gregwhitworth (and other Salesforce team members: Caridy, Eva, Mai, Jason)
- On the LWC side (Lightning Web Components), no anticipated impact unless testing reveals issues
For first-party code, Salesforce framework doesn’t do significant normalization; primary areas of concern would be new behavior around null/undefined/default values
- The Salesforce ARIA IDL polyfill is currently being removed, and it’s only there for backwards compatibility
- There may be potential impact for code packages that aren’t being maintained. For customer code, need a way to communicate the changes (anticipating low impact); Salesforce to investigate an opt-in fallback behavior for those relying on old behavior
- Next steps
    - Salesforce folks to regroup with the ARIA WG in mid-October after completion of impact analysis to LWC and Salesforce first-party code
    - Ideal to wait for Chrome/Firefox to ship the ARIA IDL changes behind a flag as well, and to maintain the flag so it can be toggled off if needed. The goal is to ensure webcompat, so the ARIA WG does not wish to rush the change

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2484.html" title="Last updated on Oct 1, 2025, 3:40 PM UTC (5dc7486)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2484/38a1bfc...5dc7486.html" title="Last updated on Oct 1, 2025, 3:40 PM UTC (5dc7486)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -12000,6 +12000,7 @@ el.hidden; // false</pre
               <!-- ReSpec needs these examples to be unindented. -->
               <pre class="example highlight javascript">
 // aria-busy example (enumerated attribute).
+// The initial example's true/false tokens here are not true Boolean values.
 
 el.ariaBusy; // "false"
 
@@ -15285,7 +15286,7 @@ button.ariaPressed; // null</pre
                 <td class="value-description">The element is pressed.</td>
               </tr>
               <tr>
-                <th class="value-name" scope="row"><strong class="default">undefined (default), empty string ("")</strong></th>
+                <th class="value-name" scope="row"><strong class="default">null (default), empty string ("")</strong></th>
                 <th class="value-name" scope="row">Undefined</th>
                 <td class="value-description">The element does not support being pressed.</td>
               </tr>
@@ -15981,7 +15982,7 @@ button.ariaPressed; // null</pre
                 <td class="value-description">The selectable element is selected.</td>
               </tr>
               <tr>
-                <th class="value-name" scope="row"><strong class="default">undefined (default), empty string ("")</strong></th>
+                <th class="value-name" scope="row"><strong class="default">null (default), empty string ("")</strong></th>
                 <th class="value-name" scope="row">Undefined</th>
                 <td class="value-description">The element is not selectable.</td>
               </tr>
--- a/index.html
+++ b/index.html
@@ -11938,8 +11938,8 @@
             An ARIA attribute may reflect as one of several IDL attribute types, such as [=nullable type|nullable=] {{DOMString}}, [=nullable type|nullable=] {{FrozenArray}} or [=nullable
             type|nullable=] {{Element}}.
           </p>
-          <p>
-            For non-enumerated attributes, default values from the ARIA values tables MUST NOT reflect to IDL as the
+          <p class="note">
+            Per the HTML specification, for non-enumerated attributes, default values from the ARIA values tables MUST NOT reflect to IDL as the
             <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> nor the
             <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> for the attribute. On getting (a non-enumerated attribute), a missing ARIA attribute will
             return <code>null</code> as such ARIA attributes are not validated on get. If an ARIA value is invalid, on getting, it will return its set value as a literal string, and will not return an
@@ -11962,7 +11962,7 @@
             <aside class="example">
               <!-- ReSpec needs these examples to be unindented. -->
               <pre class="example highlight javascript">
-// aria-label example (non-enumerated attribute)
+// aria-label example (non-enumerated nullable DOMString attribute).
 
 el.ariaLabel; // null
 
@@ -11984,7 +11984,7 @@ el.ariaLabel; // null</pre
               <!-- ReSpec needs these examples to be unindented. -->
               <pre>
               <pre class="example highlight javascript">
-// HTML hidden="" example (not aria-hidden="true")
+// HTML hidden="" example (not aria-hidden="true").
 // Actual boolean type; defaults to false.
 
 // Note: Actual boolean assignment and return value.
--- a/index.html
+++ b/index.html
@@ -11883,21 +11883,13 @@
           <h3>Value</h3>
           <p>Value type of the <a>state</a> or [=ARIA/property=]. The value is one of the following types:</p>
           <dl>
-            <dt id="valuetype_true-false">true/false</dt>
-            <dd>Value representing either <code>true</code> or <code>false</code>. The default value for this value type is <code>false</code> unless otherwise specified.</dd>
-            <dt id="valuetype_tristate">tristate</dt>
+            <dt id="valuetype_enumerated">enumerated</dt>
             <dd>
-              Value representing <code>true</code>, <code>false</code>, <code>mixed</code>, or <code>undefined</code> values. The default value for this value type is <code>undefined</code> unless
-              otherwise specified.
-            </dd>
-            <dt id="valuetype_true-false-undefined">true/false/undefined</dt>
-            <dd>
-              Value representing <code>true</code>, <code>false</code>, or <code>undefined</code> (not applicable). The default value for this value type is <code>undefined</code> unless otherwise
-              specified. For example, an element with <sref>aria-expanded</sref> set to <code>false</code> is not currently expanded; an element with <sref>aria-expanded</sref> set to
-              <code>undefined</code> is not expandable.
+              Values are limited to a predefined set of states including two special states: <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and
+              <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a>.
             </dd>
             <dt id="valuetype_idref">ID reference</dt>
-            <dd>Reference to the ID of another <a>element</a> in the same document</dd>
+            <dd>Reference to the ID of another <a>element</a> in the same document.</dd>
             <dt id="valuetype_idref_list">ID reference list</dt>
             <dd>A list of one or more ID references.</dd>
             <dt id="valuetype_integer">integer</dt>
@@ -11922,47 +11914,50 @@
       </section>
       <section data-cite="webidl">
         <h2>ARIA Attributes</h2>
-        <section id="enumerated-attribute-values">
-          <h3>Multi-value Attribute Values</h3>
+        <section id="content-vs-idl-attributes">
+          <h3>Distinguishing between Content and IDL Attributes</h3>
           <p>
-            When the ARIA attribute definition includes a table listing the attribute's allowed <span>values</span>, that attribute is a multi-value nullable attribute. Each value in the table is a
-            keyword for the attribute, mapping to a state of the same name.
+            HTML content attributes including ARIA content attributes are set in HTML markup, whereas IDL attributes (or formerly DOM properties) are the corresponding JavaScript properties exposed on
+            DOM elements. See HTML [=reflect|reflection=] for more details.
+          </p>
+        </section>
+        <section id="enumerated-attribute-values">
+          <h3>Enumerated Attribute Values</h3>
+          <p>
+            When the ARIA attribute definition includes a table listing the attribute's allowed <span>values</span>, it is an [=enumerated attribute=]. Each value in the table is a keyword for the
+            attribute, usually mapping to a state of the same name and one or more keywords may map to the same state.
+          </p>
+          <p>
+            Some ARIA attributes support an Undefined state, which is equivalent to having no state (i.e., null). User agent implementations may omit the Undefined state since the lack of an explicit
+            state results in the same behavior. For more information about determining the state of enumerated attributes, see
+            <a data-cite="html/common-microsyntaxes.html#keywords-and-enumerated-attributes">Keyword and enumerated attributes</a>.
           </p>
         </section>
         <section id="idl-reflection-attribute-values">
-          <h3>IDL reflection of ARIA attributes</h3>
+          <h3>IDL Reflection of ARIA attributes</h3>
           <p>
-            All ARIA attributes reflect in IDL as [=nullable type|nullable=] {{DOMString}} attributes. This includes the boolean-like <a href="#valuetype_true-false">true/false</a> type, and all other
-            ARIA attributes.
+            An ARIA attribute may reflect as one of several IDL attribute types such as [=nullable type|nullable=] {{DOMString}}, [=nullable type|nullable=] {{FrozenArray}} or [=nullable
+            type|nullable=] {{Element}}.
           </p>
           <p>
-            Default values from the ARIA values tables MUST NOT reflect to IDL as the <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> or the
-            <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> for the attribute. On getting, a missing ARIA attribute will return <code>null</code>. ARIA
-            attributes are not validated on get. If an ARIA value is invalid, on getting, it will return its set value as a literal string, and will not return an invalid value default.
+            For non-enumerated attributes, default values from the ARIA values tables MUST NOT reflect to IDL as the
+            <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> nor the
+            <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> for the attribute. On getting (a non-enumerated attribute), a missing ARIA attribute will
+            return <code>null</code> as such ARIA attributes are not validated on get. If an ARIA value is invalid, on getting, it will return its set value as a literal string, and will not return an
+            invalid value default.
           </p>
         </section>
         <section id="os-aapi-attribute-mapping">
           <h3>Operating System Accessibility API mapping of multi-value ARIA attributes</h3>
           <p>
-            Unlike IDL reflection, operating system accessibility API mappings of ARIA attributes can have defaults. Any default values from the ARIA values tables are exposed to the operating system
+            Operating system accessibility API mappings of ARIA attributes can have defaults. Any default values from the ARIA values tables are exposed to the operating system
             accessibility API as described in [[[#supportedState]]], and in [[[CORE-AAM]]].
           </p>
         </section>
         <section id="enumerated-attribute-values-html">
-          <h3>ARIA nullable DOMString Attributes</h3>
+          <h3>ARIA IDL Attribute Types</h3>
           <p>As noted in [[[#typemapping]]], attributes are included in host languages, and the syntax for representation of WAI-ARIA types is governed by the host language.</p>
-          <p>The following algorithm should be used for ARIA nullable {{DOMString}} attributes in HTML:</p>
-          <p>
-            On getting, if the corresponding content attribute is not present, then the IDL attribute must return null, otherwise, the IDL attribute must get the value in a transparent,
-            case-preserving manner. On setting, if the new value is null, the content attribute must be removed, and otherwise, the content attribute must be set to the specified new value in a
-            transparent, case-preserving manner.
-          </p>
-          <p class="note">
-            Note: As of ARIA 1.2, all ARIA attributes exposed via IDL are defined as nullable {{DOMString}}s. This matches the current implementation of all major rendering engines. This specification
-            change should result in no implementation changes; it will merely represent the current reality of web engines. However, in a future draft, the ARIA Working Group intends to change several
-            ARIA attributes to non-nullable DOMStrings, and seek implementations. The proposed change will bring ARIA into alignment with the HTML’s usage of
-            <a data-cite="html/common-microsyntaxes.html#enumerated-attribute">enumerated attributes</a>.
-          </p>
+          <p>ARIA attributes [=reflect=] in IDL as defined in HTML.</p>
           <section class="informative" id="enumeration-example">
             <h4>Example Attribute Usage</h4>
             <aside class="example">
--- a/index.html
+++ b/index.html
@@ -629,9 +629,8 @@
           <dt><dfn>Indicates</dfn></dt>
           <dd>
             <p>
-              Used in an attribute description to denote that the value <a href="#propcharacteristic_value">type</a> is a named token or otherwise token-like, including the Boolean-like
-              <a href="#valuetype_true-false">true/false</a>, <a href="#valuetype_true-false-undefined">true/false/undefined</a>, <a href="#valuetype_tristate">tristate (true/false/mixed)</a>, a
-              single named <a href="#valuetype_token">token</a>, or a <a href="#valuetype_token_list">token list</a>.
+              Used in an attribute description to denote that the value <a href="#propcharacteristic_value">type</a> is a named token or otherwise token-like, including a single named
+              <a href="#valuetype_token">token</a>, a <a href="#valuetype_token_list">token list</a>, or <a href="#valuetype_enumerated">enumerated</a>.
             </p>
             <p>Related Terms: <a>Defines</a>, <a>Identifies</a></p>
           </dd>
--- a/index.html
+++ b/index.html
@@ -12337,6 +12337,7 @@ button.ariaPressed; // null</pre
             <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> is the Undefined state (null) that has no associated keyword, and its
             <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> is the False state.
           </p>
+          <p>The canonical keyword for <pref>aria-atomic</pref> is "false".</p>
         </div>
         <div class="property" id="aria-autocomplete">
           <pdef>aria-autocomplete</pdef>
@@ -12747,6 +12748,7 @@ button.ariaPressed; // null</pre
             <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and
             <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the False state.
           </p>
+          <p>The canonical keyword for <pref>aria-busy</pref> is "false".</p>
         </div>
         <div class="state" id="aria-checked">
           <sdef>aria-checked</sdef>
@@ -13324,6 +13326,7 @@ button.ariaPressed; // null</pre
             <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> is the False state, and its
             <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> is the True state.
           </p>
+          <p>The canonical keyword for <pref>aria-current</pref> is "false".</p>
         </div>
         <div class="property" id="aria-describedby">
           <pdef>aria-describedby</pdef>
@@ -13619,6 +13622,7 @@ button.ariaPressed; // null</pre
             <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and
             <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the False state.
           </p>
+          <p>The canonical keyword for <pref>aria-disabled</pref> is "false".</p>
         </div>
         <div class="property deprecated" id="aria-dropeffect">
           <pdef>aria-dropeffect</pdef>
@@ -14081,7 +14085,7 @@ button.ariaPressed; // null</pre
             </thead>
             <tbody>
               <tr>
-                <th class="value-name" scope="row"><strong class="default">false (default), empty string ("")</strong></th>
+                <th class="value-name" scope="row"><strong class="default">false (default)</strong></th>
                 <th class="value-name" scope="row">False</th>
                 <td class="value-description">Indicates the element does not have a popup.</td>
               </tr>
@@ -14224,6 +14228,7 @@ button.ariaPressed; // null</pre
             <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and
             <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the False state.
           </p>
+          <p>The canonical keyword for <pref>aria-hidden</pref> is "false".</p>
         </div>
         <div class="state" id="aria-invalid">
           <sdef>aria-invalid</sdef>
@@ -14313,6 +14318,7 @@ button.ariaPressed; // null</pre
             <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> is the False state, and its
             <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> is the True state.
           </p>
+          <p>The canonical keyword for <pref>aria-invalid</pref> is "false".</p>
         </div>
         <div class="property" id="aria-keyshortcuts">
           <pdef>aria-keyshortcuts</pdef>
@@ -14774,6 +14780,7 @@ button.ariaPressed; // null</pre
             <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and
             <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the False state.
           </p>
+          <p>The canonical keyword for <pref>aria-modal</pref> is "false".</p>
         </div>
         <div class="property" id="aria-multiline">
           <pdef>aria-multiline</pdef>
@@ -14845,6 +14852,7 @@ button.ariaPressed; // null</pre
             <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and
             <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the False state.
           </p>
+          <p>The canonical keyword for <pref>aria-multiline</pref> is "false".</p>
         </div>
         <div class="property" id="aria-multiselectable">
           <pdef>aria-multiselectable</pdef>
@@ -14914,6 +14922,7 @@ button.ariaPressed; // null</pre
             <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and
             <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the False state.
           </p>
+          <p>The canonical keyword for <pref>aria-multiselectable</pref> is "false".</p>
         </div>
         <div class="property" id="aria-orientation">
           <pdef>aria-orientation</pdef>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -12308,21 +12294,25 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
                 <td class="value-description">Assistive technologies will present only the changed node or nodes.</td>
               </tr>
               <tr>
                 <th class="value-name" scope="row">true</th>
+                <th class="value-name" scope="row">True</th>
                 <td class="value-description">Assistive technologies will present the entire changed region as a whole, including the author-defined label if one exists.</td>
               </tr>
             </tbody>
           </table>
+          <p>The attribute's <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the Missing state.</p>
```
rahimabdi, line 12315: Migrated comment from @keithamus ([original](https://github.com/w3c/aria/pull/2413/files#r1983915442)):

> HTML has a concept of states that don't map to a keyword, but each state must have a definition, the definition allows mapping back to a keyword, or null. As an example the `popover` attribute has a missing value default of [no-popover state](https://html.spec.whatwg.org/multipage/popover.html#attr-popover-none-state) which let's the spec say things like "if the popover attribute is in the no-popover state".
> 
> Additionally, traditional enumerated attributes will map an invalid state back to a particular keyword; for example `popover`'s invalid attribute state is the `manual` state, which maps to the keyword `"manual"`. It's important to know what the invalid state maps back to because it determines what value is returned in the IDL for those values.
> 
> Note: not all missing values return null (e.g. the `kind` attribute on `<track>` elements has a missing value default of the subtitles state, so a track's `.kind` property returns `"subtitles"` even if the attribute is not present, another example is the `type` attribute on `<input>` which defaults to `"text"`).
> 
> 
> With that said it's possible to be in "no state". In fact, an enumerated attribute without a missing value default will be in "no state", as will invalid. No state, AIUI, will always map to null.
> 
> So here, the simplest thing for us to do here, rather than the implied missing state is to remove this, and lean on the _lack of_ a missing/invalid state returning the no state. IOW this line could simply be removed:
> 
> ```suggestion
> ```
> 
> If you think missing the definitions create more ambiguity, then we can instead do something like:
> 
> ```suggestion
> 			<p>The attribute has no <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> or <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a>.</p>
> ```

index.html
```diff
@@ -12771,30 +12773,36 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
               </tr>
               <tr>
-                <th class="value-name" scope="row"><strong class="default">undefined</strong> (default)</th>
+                <th class="value-name" scope="row"><strong class="default">undefined</strong></th>
+                <th class="value-name" scope="row">Undefined</th>
                 <td class="value-description">The element does not support being checked.</td>
               </tr>
             </tbody>
           </table>
+          <p>The attribute's <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the Undefined state.</p>
```
rahimabdi, line 12805: Migrated comment from @keithamus ([original](https://github.com/w3c/aria/pull/2413/files#r1983923427)):

I don't think we want the missing value default here to return the Undefined state, that will mean the `.ariaChecked` property will return `"undefined"` when `aria-checked` isn't on the element. I presume we want to return `null` instead? Browsers currently return null, so missing should be "no state", i.e. removed:

```suggestion
			<p>The attribute has no <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> and its <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> are both the Undefined state.</p>
```

index.html
```diff
@@ -13963,41 +13991,50 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
                 <td class="value-description">Indicates the popup is a <rref>grid</rref>.</td>
               </tr>
               <tr>
                 <th class="value-name" scope="row">dialog</th>
+                <th class="value-name" scope="row">Dialog</th>
                 <td class="value-description">Indicates the popup is a <rref>dialog</rref>.</td>
               </tr>
             </tbody>
           </table>
+          <p>The attribute's <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> is the Undefined state, and its <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> is the Menu state.</p>
```
rahimabdi, line 14037: Migrated comment from @keithamus ([original](https://github.com/w3c/aria/pull/2413/files#r1983930085)):

Again, here, the missing value default being the `Undefined` state means `.ariaHasPopup` will return `"undefined"` which I don't think is desirable. Instead:

```suggestion
			<p>The attribute has no <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a>, and its <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> is the Menu state.</p>
```

rahimabdi:
@keithamus:

> @rahimabdi I've left a couple of review notes which I think apply in a lot of places, but rather than inundating you with review comments hopefully you can take the couple here and pattern match across the rest of the spec.

Thanks! I'll review those comments (which are migrated to this PR now) and respond.

index.html
rahimabdi: @keithamus My understanding is that, in order to keep backwards compatibility, we'll need to introduce a state without a keyword that defines the null/undefined value. I don't fully understand the difference between the lack of a missing/invalid value default state (such as "Missing") and explicitly defining it as a state (e.g., like popover's "no popover state" which is null) as they would both be null. Happy to change it for spec clarity but would appreciate you clarifying!

rahimabdi:
Re-added reviewers @pkra @spectranaut @cookiecrook @keithamus @scottaohara.

index.html
keithamus: Can you help me understand what backwards-compatibility you're concerned about? AFAICT setting the IDL reflection to `null`/`undefined` removes the attribute, e.g. if I run `.ariaAtomic = null` then `.hasAttribute('aria-atomic')` is false. This is in line with other enumerated attributes so I can't tell why we'd need this additional state.

annevk: Given the definition at https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#keywords-and-enumerated-attributes I think Keith is correct that there's no difference between an explicit "missing state" and HTML's "no state".

However, I think that for `DOMString?` IDL attributes we would end up hitting an assert in https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#limited-to-only-known-values without an explicit state.

So we would either have to change that wording or go with an explicit state. I personally rather like having an explicit state for all conditions of the attribute (and it seems to me code would have to deal with that regardless of whether it's "missing state" or "no state").

rahimabdi: Thanks for chiming in @annevk.

@keithamus I don't feel strongly either way but would like to hear your thoughts.

keithamus: I don't feel strongly on this. That @annevk mentions `DOMString?` hits an assert is valid, so I think keeping the explicit `Missing` state seems reasonable here. Perhaps we can add a note calling out the reason that the `Missing` state exists? Something to the effect of it providing an explicit alternative to the `no state`, but that implementers may elide the state as it's functionally identical to a `no state`.

rahimabdi: @keithamus I've added this note in the new "Enumerated Attribute Value" section and would welcome your thoughts:

> Some ARIA attributes support an Undefined state, which is equivalent to having no state (i.e., null). User agent implementations may omit the Undefined state since the lack of an explicit state results in the same behavior. For more information about determining the state of enumerated attributes, see Keyword and enumerated attributes (which links to ("html/common-microsyntaxes.html#keywords-and-enumerated-attributes").

index.html
rahimabdi: @keithamus Thanks! `aria-haspopup` doesn't actually support an undefined state so I've removed this. The updated enumerated attribute for aria-haspopup:

Keyword → State Mapping:
- false, (empty string) → False
- true, menu → Menu
- listbox → Listbox
- tree → Tree
- grid → Grid
- dialog → Dialog

Missing Value Default:
- (null)

Invalid Value Default:
- false

rahimabdi:
CC @pkra @spectranaut @cookiecrook @keithamus @scottaohara @annevk 

This PR is ready for review! I've made a number of updates, mainly:
- Cleaned up enumerated attribute keywords/states per @keithamus's feedback.
- Changed `aria-atomic`'s Missing state to Undefined, to align with other ARIA attributes that have an "Undefined" state
- Added more IDL usage examples
- Added a note clarifying that Undefined is equivalent to "no state"

If helpful, I also created this tool to simulate how the new ARIA IDL works for enumerated attributes: https://rahimabdi.github.io/aria-idl-simulator/.

index.html
```diff
@@ -11932,52 +11924,76 @@ <h3>Value</h3>
...
+          <h3>Enumerated Attribute Values</h3>
+          <p>
+            When the ARIA attribute definition includes a table listing the attribute's allowed <span>values</span>, it is an [=enumerated attribute=]. Each value in the table is a keyword for the
+            attribute, usually mapping to a state of the same name and one or more keywords may map to the same state.
+          </p>
           <p>
-            When the ARIA attribute definition includes a table listing the attribute's allowed <span>values</span>, that attribute is a multi-value nullable attribute. Each value in the table is a
-            keyword for the attribute, mapping to a state of the same name.
+            Some ARIA attributes support an Undefined state, which is equivalent to having no state (i.e., null). User agent implementations may omit the Undefined state since the lack of an explicit
+            state results in the same behavior. For more information about determining the state of enumerated attributes, see
```
annevk, line 11942: I think this sentence adds more confusion than it resolves. Especially since you cannot just leave it out at the moment without hitting an assert.

index.html
```diff
@@ -11932,52 +11924,76 @@ <h3>Value</h3>
...
-            All ARIA attributes reflect in IDL as [=nullable type|nullable=] {{DOMString}} attributes. This includes the boolean-like <a href="#valuetype_true-false">true/false</a> type, and all other
-            ARIA attributes.
+            An ARIA attribute may reflect as one of several IDL attribute types such as [=nullable type|nullable=] {{DOMString}}, [=nullable type|nullable=] {{FrozenArray}} or [=nullable
+            type|nullable=] {{Element}}.
           </p>
           <p>
-            Default values from the ARIA values tables MUST NOT reflect to IDL as the <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> or the
-            <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> for the attribute. On getting, a missing ARIA attribute will return <code>null</code>. ARIA
-            attributes are not validated on get. If an ARIA value is invalid, on getting, it will return its set value as a literal string, and will not return an invalid value default.
+            For non-enumerated attributes, default values from the ARIA values tables MUST NOT reflect to IDL as the
```
annevk, line 11953: This seems okay to state as a note, but it shouldn't be a requirement. HTML already requires this.

annevk (commented):
One thing that's nice in HTML is that all the states are a `<dfn>` and therefore also can be linked. But maybe that's best addressed in a follow-up.

index.html
```diff
@@ -12761,7 +12802,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
               </tr>
               <tr>
                 <th class="state-value-head" scope="row">Value:</th>
-                <td class="state-value"><a href="#valuetype_tristate">tristate</a></td>
+                <a href="#valuetype_enumerated">enumerated</a>
```
pkra, line 12805: This seems better
```suggestion
                <td class="state-value"><a href="#valuetype_enumerated">enumerated</a></td>
```

cyns:
@aleventhal you may want to review this

index.html
```diff
@@ -12412,32 +12435,42 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
                 <th class="value-name" scope="row">both</th>
+                <th class="value-name" scope="row">Both</th>
                 <td class="value-description">
                   When a user is providing input, an element containing a collection of values that could complete the provided input might be displayed. If displayed, one value in the collection is
                   automatically selected, and the text needed to complete the automatically selected value appears after the caret in the input.
                 </td>
               </tr>
               <tr>
-                <th class="value-name" scope="row"><strong class="default">none (default)</strong></th>
+                <th class="value-name" scope="row">none</th>
```
pkra, line 12463: Is it intentional that `(default)` is dropped here? (And in 18 other instances.)

spectranaut:
Discussed in today's meeting: https://www.w3.org/2025/05/08-aria-minutes.html#7d9a

cyns (approved):
lgtm! this is much more clear, thank you

index.html
```diff
@@ -12308,21 +12323,29 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
                 <th class="value-name" scope="row">true</th>
+                <th class="value-name" scope="row">True</th>
                 <td class="value-description">Assistive technologies will present the entire changed region as a whole, including the author-defined label if one exists.</td>
               </tr>
             </tbody>
           </table>
+          <p>
+            For <a data-cite="html/common-dom-interfaces.html#reflect">reflection</a> purposes, the attribute's
+            <a data-cite="html/common-microsyntaxes.html#missing-value-default">missing value default</a> is the Undefined state (null) that has no associated keyword, and its
+            <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> is the False state.
```
keithamus, line 12347: As the `False` state has two keywords we will need to specify the canonical keyword for this state;

```suggestion
            <a data-cite="html/common-microsyntaxes.html#invalid-value-default">invalid value default</a> is the False state. For the purposes of reflection, the canonical keyword for the False state is the `false` keyword.
```

(You might want to link to "reflection", "canonical keyword", etc).

This is the same for other properties like `aria-busy`, `aria-checked` and so on - I won't leave a note for each.

keithamus, line 12347: So I believe this kind of change to the spec is web observable, which means it will need changes to the implementations and we need to ensure compatibility for web developers who may depend on the existing behaviour. Unless I am missing that somewhere we clarify that reflected enumerated properties should only return the value of the content attribute?

In this example the web observable piece is that the "invalid value default" is the `False` state, which means that any invalid value would map to the `False` (for which I presume the keyword is `"false"` - see above).

This means pasting this HTML into your URL bar:

```
data:text/html,<body id=b><div id=d aria-atomic=foo><script>b.append(d.ariaAtomic)</script>
```

Should render `false` on the page. But right now it renders `foo`.

I think this will be the same for all of these properties, e.g. `aria-autocomplete`, `aria-checked` and so on.

index.html
```diff
@@ -12771,30 +12812,40 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
                 <td class="value-description">Indicates a mixed mode value for a tri-state checkbox or menuitemcheckbox.</td>
               </tr>
               <tr>
                 <th class="value-name" scope="row">true</th>
+                <th class="value-name" scope="row">True</th>
                 <td class="value-description">The element is checked.</td>
               </tr>
               <tr>
-                <th class="value-name" scope="row"><strong class="default">undefined</strong> (default)</th>
+                <th class="value-name" scope="row">undefined, empty string ("")</th>
```
keithamus, line 12838: We could reasonably drop the `undefined` keyword here:

```suggestion
                <th class="value-name" scope="row">empty string ("")</th>
```

If someone set `"undefined"` then it would map to the invalid value default, which is the undefined state. The only reason _not_ to do this is if you want the canonicalization steps to explicitly return `undefined` rather than the empty string.

The same is true for other properties, e.g. `aria-current`, and so on.

keithamus (changes requested):

index.html
keithamus: I think this is useful to call out, thanks for this. I'll mark this piece resolved.

index.html
keithamus: https://github.com/whatwg/html/pull/11303 has landed so we can consider this resolved, I believe.

annevk: Right, we should have tests for this and such, but this is very much the motivation for making this change. This makes these IDL attributes more meaningful and enables them to be used for feature testing going forward.

cookiecrook: Seems like we should add the tentative WPT tests, and implement in engines behind a runtime flag to review potential impact on the Salesforce polyfill (Keith recalled this example) and any other places this might affect webcompat. @nolanlawson FYI ^

cookiecrook: Just noticed Nolan is no longer at SalesForce. Please copy any relevant contact if you know who should review. Thanks.

nolanlawson: FYI @divmain @wjhsf ^

Thanks for the heads-up!

spectranaut:
Discussed briefly: https://www.w3.org/2025/06/26-aria-minutes.html#768c

Next step to write tentative tests?

index.html
```diff
@@ -623,9 +623,8 @@ <h1>Important Terms</h1>
           <dt><dfn>Indicates</dfn></dt>
           <dd>
             <p>
-              Used in an attribute description to denote that the value <a href="#propcharacteristic_value">type</a> is a named token or otherwise token-like, including the Boolean-like
-              <a href="#valuetype_true-false">true/false</a>, <a href="#valuetype_true-false-undefined">true/false/undefined</a>, <a href="#valuetype_tristate">tristate (true/false/mixed)</a>, a
-              single named <a href="#valuetype_token">token</a>, or a <a href="#valuetype_token_list">token list</a>.
+              Used in an attribute description to denote that the value <a href="#propcharacteristic_value">type</a> is a named token or otherwise token-like, including a single named
+              <a href="#valuetype_token">token</a>, a <a href="#valuetype_token_list">token list</a>, or <a href="#valuetype_enumerated">enumerated</a>.
```
cookiecrook: Might be good to leave a note about token true/false being Boolean-like but not a true Boolean? I recall adding this adjective to the above removal because some implementors and web devs assumed/wanted/thought the true/false values would/should behave like HTML boolean attributes.

I agree with your shorter definition as you have it, but consider whether an additional editorial note may head-off future questions.. For example, from @sideshowbarker regarding the LadyBird implementation.

index.html
cookiecrook: Part of this could be an informative note for authors/implementors, rather than in the normative text of the Enumerated Attribute definition. 

Do we ever anticipate a time where the explicit "undefined" content attribute values would need to map to an Undefined state, separately than the Null state? For example, I think the only scenarios where an `undefined` is allowed are with the `undefined/true/false/[/mixed]` enumerated values, but since the Missing Default and Invalid Default are the same one those, there is no impact... Can we foresee any scenario where those would be different? For example, if `undefined` was an allowed token value of `aria-invalid`? (Presumably we just need to always avoid that, which seems okay.)

index.html
```diff
@@ -11932,52 +11924,76 @@ <h3>Value</h3>
...
+            <a data-cite="html/common-microsyntaxes.html#keywords-and-enumerated-attributes">Keyword and enumerated attributes</a>.
           </p>
         </section>
         <section id="idl-reflection-attribute-values">
-          <h3>IDL reflection of ARIA attributes</h3>
+          <h3>IDL Reflection of ARIA attributes</h3>
           <p>
-            All ARIA attributes reflect in IDL as [=nullable type|nullable=] {{DOMString}} attributes. This includes the boolean-like <a href="#valuetype_true-false">true/false</a> type, and all other
-            ARIA attributes.
+            An ARIA attribute may reflect as one of several IDL attribute types such as [=nullable type|nullable=] {{DOMString}}, [=nullable type|nullable=] {{FrozenArray}} or [=nullable
```
```suggestion
            An ARIA attribute may reflect as one of several IDL attribute types, such as [=nullable type|nullable=] {{DOMString}}, [=nullable type|nullable=] {{FrozenArray}} or [=nullable
```

index.html
cookiecrook: _informative_ note

index.html
```diff
@@ -11932,52 +11924,76 @@ <h3>Value</h3>
...
-            <a data-cite="html/common-microsyntaxes.html#enumerated-attribute">enumerated attributes</a>.
-          </p>
+          <p>ARIA attributes [=reflect=] in IDL as defined in HTML.</p>
           <section class="informative" id="enumeration-example">
             <h4>Example Attribute Usage</h4>
             <aside class="example">
               <!-- ReSpec needs these examples to be unindented. -->
-              <pre class="example highlight javascript">
+              <pre>
+// aria-label example (non-enumerated attribute)
```
```suggestion
// aria-label example (non-enumerated nullable DOMString attribute)
```

index.html
```diff
@@ -11992,45 +12008,44 @@ <h4>Example Attribute Usage</h4>
             </aside>
             <aside class="example">
               <!-- ReSpec needs these examples to be unindented. -->
-              <pre class="example highlight javascript">
-// aria-busy example
-// true/false ~ boolean-like nullable string; returns null unless set
+              <pre>
+// aria-busy example (enumerated attribute)
```
```suggestion
// aria-busy example (enumerated attribute)
// the initial example's true/false tokens here are not true Boolean values.
```

index.html
cookiecrook: As a future enhancement for ARIA editors to avoid value error contradictions, perhaps the values table could be generated from the IDL? Or generate both the IDL and the values tables from the same pre-ReSpec source? (Not in this patch though. It's already massive.)

index.html
cookiecrook: good catch @pkra

index.html
cookiecrook: See @keithamus' related comment in https://github.com/w3c/aria/pull/2484/files#r2084175913

index.html
rahimabdi: Great callout! We can simplify many attributes, i.e., empty value and/or undefined map to an invalid value default:

I've also included the "empty value default" which is helpful for boolean-like declaration of ARIA attrs (e.g., &lt;div someAttr>; this could be a WebIDL extended attribute like [EmptyValueDefault] and where it's not null, it could provide an extra state for the attribute.

Simplifying enumeration for attributes:
- aria-atomic
    - Remove empty string keyword as it maps to “false” state
    - Empty value default = “false”
- aria-busy
    - Remove empty string keyword as it maps to “false” state
    - Empty value default = “false”
- aria-checked
    - Remove empty string and undefined keywords as they maps to null (invalid value default)
    - Empty value default = null
 - aria-current
     - Can’t remove empty string because IVD = “true”
     - Empty value default = “false”
     - Undefined string maps to “true”
     - For implementation, this one needs manual reflection (i.e., can't use a nice WebIDL extended attribute like [ReflectEnum]
- aria-disabled
    - Remove empty string keyword as it maps to “false” state
    - Empty value default = “false”
- aria-expanded
    - Both empty value and undefined map to null (invalid value default)
    - Empty value default = null
- aria-hidden
    - Both empty value and undefined map to false state "invalid value default”
    - Empty value default = “false”
    - Undefined maps to “false” which is IVD
- aria-invalid
    - Empty value default = “false”
- aria-modal
    - Empty value default is “false”, however it’s also covered by invalid value default so can be removed
- aria-multiline
    - Empty value default is “false”, however it’s also covered by invalid value default
- aria-multiselectable
    - Empty value default is “false”, however it’s also covered by invalid value default
- aria-orientation
    - Both “” and undefined could be covered by invalid value default (null)
    - Empty value default = null
 - aria-pressed
     - Both empty string and undefined could be covered by invalid value default (null)
    - Empty value default = null
- aria-readonly
    - Empty value default is “false”, however it’s also covered by invalid value default
- aria-required
    - Empty value default is “false”, however it’s also covered by invalid value default
- aria-selected
    - Both “” and undefined could be covered by invalid value default
    - Empty value default = null

rahimabdi:
It appears that many of these attributes have secondary/tertiary keywords that map to an existing state. For example, for `aria-checked`, both the empty string and the string "undefined" are both "false", which is the same as the invalid value default:

- Keywords: { "true", "false", "mixed",  "", "undefined" }
- No missing value default (value becomes null)
- No invalid value default (value becomes null)

Question for the ARIA WG: would it be preferable to explicitly state these keywords and their state mapping in spec, or should they be omitted? 

**Approach 1 - Explicitly state every single keyword/state for each enumerated attribute (simplified example)**:

> `aria-checked`: 
> - This attribute's keywords are "true" (True), "false" (False), "mixed" (Mixed), "" (False) and "undefined" (False). 
> - The attribute's missing value default and invalid value default are the Undefined state (null).
> 

**Approach 2 - Omit the extra keywords that have the same state mapping (simplified example)**:

> > `aria-checked`: 
> - This attribute's keywords are "true" (True), "false" (False), "mixed" (Mixed). 
> - The attribute's missing value default and invalid value default are the Undefined state (null).

**Approach 3 - Omit the extra keywords that have the same state mapping, but call out the extra keywords for clarity (simplified example)**:

> > `aria-checked`: 
> - This attribute's keywords are "true" (True), "false" (False), "mixed" (Mixed). 
> - The attribute's missing value default and invalid value default are the Undefined state (null).
> - **Note:** The empty string and "undefined" must return the False state (invalid value default).

annevk:
I think supporting the empty string as a valid keyword makes sense, but "`undefined`" not so much.

spectranaut:
Notes from today's meeting: https://www.w3.org/2025/08/14-aria-minutes.html#367f

index.html
rahimabdi: Agreed, I've added a note clarifying this.

index.html
rahimabdi: @annevk @cookiecrook Got it, I've removed this sentence: "User agent implementations may omit the Undefined state since the lack of an explicit state results in the same behavior."

@keithamus Hopefully the link to the HTML spec (which defines the state computation algorithm) is sufficient. 

> Do we ever anticipate a time where the explicit "undefined" content attribute values would need to map to an Undefined state, separately than the Null state? For example, I think the only scenarios where an undefined is allowed are with the undefined/true/false/[/mixed] enumerated values, but since the Missing Default and Invalid Default are the same one those, there is no impact... Can we foresee any scenario where those would be different? For example, if undefined was an allowed token value of aria-invalid? (Presumably we just need to always avoid that, which seems okay.)

As @keithamus also mentioned, the "undefined" content attribute value will map to the invalid value default (effectively treated as invalid), and fortunately there's no need for any ARIA attribute to have an explicit "undefined" keyword (nor an Undefined state mapping to "undefined" the keyword). I plan to add a note on this in this PR.

Here are all the attributes which currently support a value of "undefined" per spec, all mapping to their respective invalid value defaults and equivalent with behavior currently defined in spec:
- aria-checked ("undefined" maps to null)
- aria-expanded ("undefined" maps to null)
- aria-hidden ("undefined" maps to False state) 
- aria-orientation ("undefined" maps to null)
- aria-pressed ("undefined" maps to null)
- aria-selected ("undefined" maps to null)

Note: Each of these 6 attributes also supports the empty string as a keyword, so `""` and "undefined" are the same. 

Only two enumerated ARIA attributes have differing invalid/missing value defaults and they don't support "undefined" however, any invalid value provided to them should result in a value of "true" per spec (now that they're enumerated, this is handled by the invalid value default):
- aria-current
    - invalid value default = True state
    - missing value default = False state
- aria-invalid
    - invalid value default = True state
    - missing value default = False state

index.html
rahimabdi: Thanks! Updated this to be an informative note (added `class="note"` like other spec examples).

index.html
gregwhitworth: @nolanlawson @divmain @wjhsf @cookiecrook Following up here at SF to take a look into this change. Much appreciated.

index.html
rahimabdi: @keithamus Updated "Undefined state" to reference that it is equivalent to null.

index.html
rahimabdi: Fixed, thank you.

index.html
rahimabdi: I've reverted all of these to what is reflecting in main (showing default in parenthesis).

Nice idea @cookiecrook! @pkra perhaps an agenda topic for the Editor's meeting?

rahimabdi:
I've made many updates to this PR addressing outstanding feedback. I've also updated the PR description with a link to an unofficial explainer, details on how to test (via Safari feature flag) and more.

I've also reached out to the Salesforce folks (thanks @gregwhitworth!) to follow up on any impact related to their [LWC ARIA polyfill](https://www.npmjs.com/package/@lwc/aria-reflection).

I will Agenda+ to discuss these outstanding questions:
- `aria-haspopup` questions
    - Mapping of `aria-haspopup`: should "true" and "menu" be distinct states in terms of IDL?
    - Somebody asked why `aria-haspopup` has a missing value default of null rather than "false". Is this correct?
- Specifying undefined in the attribute Values table: does it look OK?
- Feedback on the new "Relationship Between the Undefined State and "undefined" String Value" section
- Should all ARIA enumerated attributes that support null be specified as an Undefined state? This question applies only to `aria-atomic` and `aria-haspopup`.
- New empty value default as part of [HTML enumerated attribute definition](https://html.spec.whatwg.org/#keywords-and-enumerated-attributes). Any concerns with using this for ARIA attrs that support an empty string keyword?

rahimabdi:
@keithamus @cookiecrook @rahimabdi met with the Salesforce team (@gregwhitworth and others) on Sept 30 2025 to discuss impact/risks for Salesforce frameworks, and to request testing. I've captured meeting notes/next steps in the "Meeting Notes" section in the PR description.

spectranaut:
Discussed Rahim's questions above in today's meeting: https://www.w3.org/2025/10/09-aria-minutes.html#31d8

spectranaut:
@rahimabdi did all of your questions get answered?

rahimabdi:
> @rahimabdi did all of your questions get answered?

@spectranaut Yes, I think there's some additional clarify that can be added to spec around the Undefined state. For example, for the [`popover` attribute's IDL definition](https://html.spec.whatwg.org/multipage/popover.html#attr-popover-auto):

> The attribute's [missing value default](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#missing-value-default) is the **No Popover state**...

The ARIA spec could have similar Undefined states (i.e., null) where an attribute supports it, e.g., No Pressed State. I'll investigate this further to ensure the Undefined state is properly defined for attributes.

pkra:
Discussed in the last editors' call https://www.w3.org/2026/05/13-aria-editors-minutes.html#a346

PR #2485, opened 2025-03-23
Editorial: Fix several instances of incorrect 'i.e.' and 'e.g.' references missing commas

rahimabdi:
The following i.e. and e.g. references were missing a comma and this PR adds them:

- [if the element does not [prohibit](https://w3c.github.io/aria/#prohibitedattributes) use of the attribute. If the label text is available in the DOM (i.e. typically visible text]
- [container element that is controlled by an element that has the [aria-activedescendant](https://w3c.github.io/aria/#aria-activedescendant) attribute (e.g. see [combobox](https://w3c.github.io/aria/#combobox)).]
- [SHOULD prefer full punctuation verbosity to ensure common symbols (e.g. "-") are spoken]
- [When trusted dictionary sources list both hyphenated or non-hyphenated spellings (e.g. "multi-line" 
- [but if future attributes include acronym usage, attempt to match existing DOM conventions (e.g. ID becomes Id).]]

### DIFF
--- a/index.html
+++ b/index.html
@@ -1052,7 +1052,7 @@
                   set <abbr title="Document Object Model">DOM</abbr> focus to the element that has the <pref>aria-activedescendant</pref> attribute.
                   <p class="note">
                     An element with an ID can be referenced when it is an <a>accessibility descendant</a> of a container element that has the <pref>aria-activedescendant</pref> attribute or by a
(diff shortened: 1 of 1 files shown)


spectranaut:
@front-endian to review!

front-endian (approved):
Looks good!

PR #2486, opened 2025-03-23
Editorial: Update accname step 2E (Host Language Label) to specify empty string treatment

rahimabdi:
Closes https://github.com/w3c/accname/issues/234.

This PR clarifies that the empty string is special as part of host language label calculation.

### DIFF
--- a/accname/index.html
+++ b/accname/index.html
@@ -629,7 +629,7 @@
                   <span id="step2E"><!-- Don't link to this legacy numbered ID. --></span><em>Host Language Label:</em> Otherwise, if the <code>current node</code>'s native markup provides an
                   [=attribute=] (e.g. <code>alt</code>) or [=element=] (e.g. HTML <code>label</code> or SVG <code>title</code>) that defines a text alternative, return that alternative in the form of
                   a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or
-                  <code>role="none"</code>).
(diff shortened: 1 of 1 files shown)


MelSumner (approved):

spectranaut:
Also related? https://github.com/w3c/aria/issues/2491

scottaohara (commented):
so i think this update helps - but per the discussion in #2491 i'm not sure if this content adequately indicates "unless .... [so stop here]" OR "unless .... [so keep going to find your accname!]

i understood it to be "so stop here" and more specifically "so stop here AND ALSO HTML AAM has more detailed guidance for images"

but maybe the conflicting wpts were contributing to the reading that the spec was actually saying 'unless ... [so move on until the tooltip step in the algorithm]"

rahimabdi:
> but maybe the conflicting wpts were contributing to the reading that the spec was actually saying 'unless ... [so move on until the tooltip step in the algorithm]"

Agreed, and a note pointing to HTML-AAMs somewhere in this step is a good idea. Should the note be complementary to this step, or should accname pick up in HTML-AAM after this step?

scottaohara:
me thinks this might be solved by however @MelSumner was planning to write that note for accName in general?

i'd welcome thoughts from @sideshowbarker though

accdc (approved):
I think as @MelSumner described the note yesterday it makes sense to me. I approve and welcome any additional feedback as well.

rahimabdi:
@spectranaut I've updated the language here so ready for another review, thanks!

spectranaut (approved):

PR #2498, opened 2025-03-26
Editorial: Updated list styles

### LINKED ISSUE #2494: accname: decide what to do with nested list styles
Split out of https://github.com/w3c/aria/issues/2462

>  accname's custom OL styles: do we want dedicated class-based styles (to avoid future problems)?

(Use a `style` block)

MelSumner:
Closes #2494 

Changes styles for numbered list elements (`<ol>`). Previously was a mix of different list styles; changed it to use multi-leveled numbered style (decimals). 

I think this approach is more appropriate for technical documents and may, perhaps, make algorithm steps a bit more logical for reference purposes. 

Before screenshot: 
CleanShot 2025-03-26 at 14 18 05

After screenshot: 
CleanShot 2025-03-26 at 14 18 23


accname/index.html
```diff
@@ -143,25 +143,34 @@
...
-        list-style: lower-roman;
+
+      ol.acc-ol > li:before {
+        content: counters(acclist, ".") " ";
+        font-weight: bold;
       }
-      ol ol ol ol {
-        list-style: lower-alpha;
+
+      ol.note-list,
```
MelSumner, line 161: This is for numbered lists inside of notes and examples.

spectranaut (approved):
I see your point, I think this is way better... it's not too much of a breaking change, you can easily figure out what "3.a.ii" means "3.1.2".

Also it doesn't break links, as the links come from the ids not the list numbers.

PR #2500, opened 2025-03-27
Normative update: if a tab is active a corresponding tabpanel must be rendered

### LINKED ISSUE #2497: Missing author requirement for use of tabs?
Consider the following markup:

```
<div role=tablist>
  <button role=tab aria-selected=false>foo</button>
  <button role=tab aria-selected=true>bar</button>
</div>

<article>
  bar content goes here
</article>
```

running a bunch of automated checkers, no issue is raised that there is no association between the selected tab and its content.  which is not unexpected, since the aria spec says authors SHOULD ensure there's an association between the currently active tab and tabpanel.  

But, the issue here is there's no explicit `role=tabpanel` element at all, and from my reading of the spec it seems more of an "well if you're creating a tab widget of course you would mark elements as tab panels" - but there is no overt author requirement to do so.

Is this a gap?  Or, is there an unstated opportunity for a tablist to not just be used for showing/hiding tabpanels - but other non-tabpanel content as well?  (i say this not because i think it's a good idea, but as the only logical explanation i can think of for why there is no overt requirement that if someone implements tabs within a tablist - that they could reasonably no also be required to declare at least one element in the UI as a tabpanel).

scottaohara:
closes #2497 initial stab at creating the new author must requirement so that if there is an activate tab, a corresponding tabpanel element must be rendered.

question is should this requirement be copied over to tabpanel as well?  i assume yes, but will wait to be told that before i add in more dupe content between these rules.

I've removed the checks for this PR that I don't think are applicable.  this should really just be an author requirement that should be picked up by a11y checkers / code validators

* [ ] "author MUST" tests:
* [X] Related APG Issue/PR: not necessary, basically indicates this
* [X] MDN Issue/PR: not necessay, basically indicates this

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2500.html" title="Last updated on Apr 17, 2025, 7:17 PM UTC (1e32f94)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2500/ab288d2...1e32f94.html" title="Last updated on Apr 17, 2025, 7:17 PM UTC (1e32f94)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -10233,7 +10233,7 @@
               for details on implementing a tab set design pattern.
             </p>
             <p>Authors MUST ensure [=elements=] with <a>role</a> <rref>tab</rref> are <a>accessibility children</a> of an element with the role <rref>tablist</rref>.</p>
-	    <p>Authors MUST ensure that if a <code>tab</code> is active, a corresponding <code>tabpanel</code> element that represents the active <code>tab</code> is rendered.</p>
+            <p>Authors MUST ensure that if a <code>tab</code> is active, a corresponding <code>tabpanel</code> element that represents the active <code>tab</code> is rendered.</p>
             <p>Authors SHOULD ensure the <rref>tabpanel</rref> associated with the currently active tab is <a>perceivable</a> to the user.</p>
             <!-- keep following para synced with its counterpart in #tablist -->
             <p>
@@ -10463,8 +10463,8 @@
               tabpanel. For a multi-selectable <rref>tablist</rref>, authors SHOULD ensure that the <rref>tab</rref> for each visible <rref>tabpanel</rref> has the <sref>aria-expanded</sref>
               <a>attribute</a> set to <code>true</code>, and that the <code>tabs</code> associated with the remaining [=element/hidden from all users=] <code>tabpanel</code> elements have their
               <sref>aria-expanded</sref> attributes set to <code>false</code>.
-            </p>  
-	
+            </p>
+
             <!-- keep following para synced with its counterpart in #tabpanel -->
             <p>
               <rref>tablist</rref> elements are typically placed near, and usually preceding, a series of <rref>tabpanel</rref> elements. See the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -10233,6 +10233,7 @@ <h5>Presentational Role Inheritance</h5>
               for details on implementing a tab set design pattern.
             </p>
             <p>Authors MUST ensure [=elements=] with <a>role</a> <rref>tab</rref> are <a>accessibility children</a> of an element with the role <rref>tablist</rref>.</p>
+            <p>Authors MUST ensure that if a <code>tab</code> is active, a corresponding <code>tabpanel</code> element that represents the active <code>tab</code> is rendered.</p>
```
scottaohara, line 10236: maybe this should also mention _not_ being in the collapsed state?

that's some multi-tab stuff that i'm not sure about though.... having an active tab but also having it be collapsed... is that common in this otherwise uncommon scenario?

MelSumner, line 10236:
```suggestion
            <p>Authors MUST ensure that if a <code>tab</code> is active, a corresponding <code>tabpanel</code> that represents the active <code>tab</code> is rendered.</p>
```

MelSumner (approved):
Left a suggestion based on convo today. Otherwise 👍

hidde-logius (approved):

smhigley (approved):

spectranaut:
I just took a look at the MDN article and Tabs pattern on APG -- they both seem too indicate this author requirement, I don't think  we need to update them.

I think we should merge this and open an issues on validators.

index.html
scottaohara: No one commented on my question here, so assuming it’s fine as is since it was merged

scottaohara:
Did people notice this question from my OP when reviewing?

“question is should this requirement be copied over to tabpanel as well? i assume yes, but will wait to be told that before i add in more dupe content between these rules.”

I don’t know if silence meant it wasn’t necessary or if my question wasn’t seen?

similar to the other question I posted as a comment.

spectranaut:
Hmm I saw the three reviews and merged, but I should've seen there were still open questions, sorry about that :/

> “question is should this requirement be copied over to tabpanel as well? i assume yes, but will wait to be told that before i add in more dupe content between these rules.”

Now paying closer attention, I agree this should be copied to tab panel, but i personally lean towards duplication is fine in specs if it helps the spec be more clear and specific.

Could you open another PR?

PR #2503, opened 2025-03-28
editorial: update html-aam note

MelSumner:
Closes [AccName 250](https://github.com/w3c/accname/issues/250)

I am making this change in an effort to make it more obvious that readers should be checking HTML-AAM in specific instances. Due to the recent issues filed against SVG-AAM wherein it's made clear that SVG-AAM should be referring to AccName instead, I've removed the note's reference to SVG-AAM.

Before: 
> See [HTML-AAM](https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation), [SVG-AAM](https://www.w3.org/TR/svg-aam-1.0/#mapping_additional_nd), or other host language documentation for more information on markup that defines a text alternative.

CleanShot 2025-03-28 at 16 23 32@2x

After: 
> Important! The accessible name for native HTML elements may be calculated differently. Make sure to review [HTML-AAM](https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation) for more information.

CleanShot 2025-03-28 at 16 34 33@2x

### DIFF
--- a/accname/index.html
+++ b/accname/index.html
@@ -642,7 +642,7 @@
                   a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or
                   <code>role="none"</code>).
                   <div class="note">
-                    Important! The accessible name for native HTML elements may be calculated a little differently. See <a href="https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation">HTML-AAM</a> for more information for how markup define a text alternative.
+                    Important! The accessible name for native HTML elements may be calculated differently. See <a href="https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation">HTML-AAM</a> for more information.
                   </div>
                   <div class="note">
                     <p>
--- a/accname/index.html
+++ b/accname/index.html
@@ -642,7 +642,7 @@
                   a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or
                   <code>role="none"</code>).
                   <div class="note">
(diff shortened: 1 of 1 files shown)


accname/index.html
```diff
@@ -639,11 +639,11 @@ <h4>Computation steps</h4>
...
-                  a <code>flat string</code> as defined by the host language, unless the <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or
-                  <code>role="none"</code>).
+                  a <code>flat string</code> as defined by the host language (e.g., <a href="https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation">HTML-AAM</a>), unless the
+                  <code>current node</code> is exposed as presentational (<code>role="presentation"</code> or <code>role="none"</code>).
                   <div class="note">
-                    See <a href="https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation">HTML-AAM</a>,
-                    <a href="https://www.w3.org/TR/svg-aam-1.0/#mapping_additional_nd">SVG-AAM</a>, or other host language documentation for more information on markup that defines a text alternative.
+                    Important! The accessible name for native HTML elements may be calculated differently. Make sure to review
```
rahimabdi, line 645:
```suggestion
                    Important! The accessible name for native HTML elements may be calculated differently depending on host language. Make sure to review
```

rahimabdi: I think clarifying why it may be calculated differently here would be helpful (commit suggestion is just an example).

rahimabdi (approved):

accdc (approved):

scottaohara (approved):

scottaohara:
approved, but calling out that https://github.com/w3c/aria/pull/2486 and this PR are making updates in the same area.  so it might make sense to combine these to avoid conflicts

PR #2504, opened 2025-04-01
Fix incorrect role specified for html element - 571

### LINKED ISSUE #571: DPub-AAM: UIA landmark mappings not consistent with Core AAM
If you look at the [role-mapping table in the Core AAM](https://rawgit.com/w3c/aria/master/core-aam/core-aam.html#mapping_role_table), you'll notice that all landmarks have UIA mappings in the form:

* Control Type is `$TYPE`.
* Localized Control Type is `$LOCALIZED_TYPE`.
* Landmark Type is `Custom`.
* Localized Landmark Type is `$LOCALIZED_TYPE`.

If you look at the [same table in the DPub AAM](https://rawgit.com/w3c/aria/master/dpub-aam/dpub-aam.html#mapping_role_table), you'll notice that all landmarks have UIA mappings in the form:

* Control Type is `Text`
* Localized Control Type is `$LOCALIZED_TYPE`.

Where the value of `$LOCALIZED_TYPE` is the name of the DPub ARIA role, minus the `doc-` prefix. As an example, see the mapping of [`doc-acknowledgments`](https://rawgit.com/w3c/aria/master/dpub-aam/dpub-aam.html#role-map-acknowledgments).

There are quite a few DPub ARIA roles which subclass the `landmark` role and which are mapped as landmarks by all the other platforms. Unless the lack of Landmark Type and Localized Landmark Type is a deliberate omission, we have a bunch of incorrect mappings for UIA.

@boggydigital: If the omission is indeed deliberate, please state so. If it's a mistake, please do a pull request fixing the mappings for your platform. Then we'll get @richschwer to review and commit the fixes. Thanks in advance!

scottaohara:
closes #571

changes the listed mapping for the html element to be the generic role, rather than the document role.

A note is added to clarify the document role of a web page is exposed by a parent `document node` created by the user agent.

From what i can tell, no implementation bugs need to be filed, as browsers have been doing the right thing - and this is a spec correction to match reality.

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -2974,7 +2974,7 @@
             </tr>
             <tr>
               <th>[[wai-aria-1.2]]</th>
-              <td><a class="core-mapping" href="#role-map-document">`document`</a> role</td>
+              <td><a class="core-mapping" href="#role-map-generic">`generic`</a> role</td>
             </tr>
             <tr>
               <th><a data-cite="core-aam-1.2/#roleMappingComputedRole">Computed Role</a></th>
@@ -3009,7 +3009,10 @@
             <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
             <tr>
               <th>Comments</th>
-              <td>User agents MUST ignore the `aria-hidden` attribute if specified on the `html` element.</td>
+              <td>
+                <p>User agents MUST ignore the `aria-hidden` attribute if specified on the `html` element.</p>
+                <p class="note">The `document` role of a web page is not exposed by the `html` element, but rather from a parent `document node` created by the user agent.</p>
+              </td>
             </tr>
           </tbody>
         </table>


jnurthen (approved):

html-aam/index.html
```diff
@@ -2974,7 +2974,7 @@ <h4 id="el-html">`html`</h4>
             </tr>
             <tr>
               <th>[[wai-aria-1.2]]</th>
-              <td><a class="core-mapping" href="#role-map-document">`document`</a> role</td>
+              <td><a class="core-mapping" href="#role-map-generic">`generic`</a> role</td>
```
cookiecrook, line 2977: Does this affect the Author and UA reqs [aria-hidden](https://w3c.github.io/aria/#aria-hidden)? Emphasis mine.

> Authors MUST NOT use aria-hidden to hide the root element or the _**host language element that [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) the contents of the primary document in view**_. For instance, the html or body elements in an HTML document.

and later:

> To prevent authors erroneously hiding entire window-rendered documents only to those using assistive technology, user agents MUST NOT expose the hidden state to assistive technologies if it is specified on the root element or the _**host language element that [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) the contents of the primary document in view**_. For instance, the html or body elements in an HTML document, or the root svg element if it is rendered as its own primary document in the browser window. If authors were to specify aria-hidden="true" on the opening tag for an embedded document, for instance on a math or svg embedded within an HTML document, user agents would still be expected to hide these elements from assistive technologies.

cookiecrook (changes requested):
Blocking the other approved reviews until we discuss this side effect of the change.

html-aam/index.html
scottaohara: this change is not intended to alter that requirement - no.  but, i could see a case for rewording those a bit to just more direct in stating user agents must not respect aria hidden on those elements.

cookiecrook: How about "represents or contains"?

scottaohara: that sounds good to me. i can make the change tomorrow - unless you beat me to it.  thank you.

scottaohara: OK I've made the change.  thank you @cookiecrook

spectranaut:
@cookiecrook it sounds like your issue was resolved, can you positively review this?

pkra:
I've edited the top card to point to the html-aam issue.

keithamus (approved):

PR #2506, opened 2025-04-02
editorial: tidying up "Characteristics of Roles"

### LINKED ISSUE #1263: Characteristics of Roles section needs tidying up

pkra:
Replaces #1733
Partially addresses #1263

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2506.html" title="Last updated on Apr 4, 2025, 10:10 AM UTC (1003238)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2506/932fc84...1003238.html" title="Last updated on Apr 4, 2025, 10:10 AM UTC (1003238)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -14261,10 +14261,11 @@ button.ariaPressed; // null</pre
               Most host languages provide an attribute that could be used to name the element (e.g., the <code>[^html-global/title^]</code> attribute in HTML), yet this could present a browser
               tooltip. In the cases where DOM content or a tooltip is undesirable, authors MAY set the accessible name of the element using <pref>aria-label</pref>, if the element does not
               <a href="#prohibitedattributes">prohibit</a> use of the attribute. If the label text is available in the DOM (i.e. typically visible text content), authors SHOULD use
-              <pref>aria-labelledby</pref> and SHOULD NOT use <pref>aria-label</pref>. There might be instances where the name of an element cannot be determined programmatically from the DOM, and 
+              <pref>aria-labelledby</pref> and SHOULD NOT use <pref>aria-label</pref>. There might be instances where the name of an element cannot be determined programmatically from the DOM, and
               there are cases where referencing DOM content is not the desired user experience. Authors MUST NOT specify <code>aria-label</code> on an element which has an explicit or implicit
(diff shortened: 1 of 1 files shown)


pkra:
@jnurthen ping.

Issue #2529, opened 2025-05-09
Clarify aria-placeholder's place in naming calculation


scottaohara:
This issue is being filed because @giacomo-petri [correctly pointed out](https://github.com/web-platform-tests/wpt/pull/44965#discussion_r2081462344) that while specs ( html aam nor accName ) clarify the order of precedence for instances where a `placeholder` attribute and an `aria-placeholder` attribute are specified on the same element.  

From my testing today, it fortunately looks like almost all browser implementations are in sync, where placeholder takes priority over aria-placeholder. The only outlier i found was when testing with Firefox on macOS, where that one instance combined the native placeholder value and the aria-placeholder value into a single string exposed to users.

I have created a PR to clarify these steps in HTML AAM

PR #2530, opened 2025-05-09
Clarification: include aria-placeholder in accName steps for HTML input elements

scottaohara:
Closes [#2529](https://github.com/w3c/aria/issues/2529)

clarifies that placeholder attribute value takes priority over aria-placeholder on HTML elements that allow for the use of these attributes.

* [ ] "author MUST" tests:
* [X] Browser implementations (link to issue or commit): all browsers do this
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2530.html" title="Last updated on May 15, 2025, 5:18 PM UTC (29a2a5d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2530/b085915...29a2a5d.html" title="Last updated on May 15, 2025, 5:18 PM UTC (29a2a5d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -13681,9 +13681,15 @@ button.ariaPressed; // null</pre
           <sdef>aria-expanded</sdef>
           <div class="state-description">
             <p><a>Indicates</a> whether a related element is expanded (shown) or collapsed (hidden).</p>
-            <p>The <sref>aria-expanded</sref> attribute is applied to a focusable, interactive element that toggles visibility of content of a different element. If the element with <sref>aria-expanded</sref> is also a <rref>treeitem</rref> in a <rref>tree</rref> or a <rref>row</rref> in a <rref>treegrid</rref>, then it SHOULD also be the <a>accessibility parent</a> of the content it expands and collapses. Otherwise, the element with <sref>aria-expanded</sref> SHOULD NOT be the <a>accessibility parent</a> of the content that is expanding or collapsing. Rather, identify that relationship between the interactive element and the element being controlled using <pref>aria-controls</pref>.</p>
+            <p>
+              The <sref>aria-expanded</sref> attribute is applied to a focusable, interactive element that toggles visibility of content of a different element. If the element with
+              <sref>aria-expanded</sref> is also a <rref>treeitem</rref> in a <rref>tree</rref> or a <rref>row</rref> in a <rref>treegrid</rref>, then it SHOULD also be the
+              <a>accessibility parent</a> of the content it expands and collapses. Otherwise, the element with <sref>aria-expanded</sref> SHOULD NOT be the <a>accessibility parent</a> of the content
+              that is expanding or collapsing. Rather, identify that relationship between the interactive element and the element being controlled using <pref>aria-controls</pref>.
+            </p>
             <p>For example, <sref>aria-expanded</sref> is applied to a parent <rref>treeitem</rref> to indicate whether its child branch of the tree is shown.</p>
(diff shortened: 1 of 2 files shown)


scottaohara:
might be able to jump onto https://github.com/web-platform-tests/wpt/pull/44965 for a wpt related to this update

scottaohara:
looks like prettier made updates to the ARIA spec.  mentioning that since those updates have nothing to do with this PR, and reviewers should be able to ignore them

giacomo-petri (approved):

giacomo-petri:
Eveything is fine on my side.

Just a minor point. First of all thanks @scottaohara for testing `placeholder` and `aria-placeholder`. Your results show that browsers are mostly consistent and that `placeholder` takes precedence over `aria-placeholder`. If the specs now confirm this behavior, that's fine, but it does feel inconsistent compared to other ARIA attributes: for example, `aria-label` and `aria-labelledby` override the accessible name, and `aria-describedby` or `aria-description` override the description. So it seems odd that `aria-placeholder` doesn't override `placeholder`.
I get that this is a bit of an edge case, and changing browser behavior might not be justified. That said, IMO it's important to recognize the inconsistency, especially since it can lead to incorrect assumptions by developers who expect `aria-placeholder` to override `placeholder` in a consistent way. If we decide to leave it as is, it might be worth including a note in the ARIA spec to clarify this specific behavior of `aria-placeholder`.

scottaohara:
per the discussion about how the priority of placeholder over aria-placeholder needed to be called out - the HTML AAM spec _already_ has the following stated in the mapping table for the placeholder attribute:

>When the `placeholder` and `aria-placeholder` attributes are both present, and the `placeholder` attribute's value is non-empty, user agents MUST expose the value of the `placeholder` attribute, and ignore `aria-placeholder`. If the `placeholder` attribute's value is empty, then user agents MUST expose the value of the `aria-placeholder` attribute.

Unless someone would like to suggest a short note to maybe accompany the updating naming steps to reference this existing content in the spec, I'm considering this PR a clarification to this existing text and otherwise overtly documenting reality.

index.html
```diff
@@ -13692,12 +13698,15 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+            >
             <p>Similarly, it can be applied to a <rref>button</rref> to control the visibility of another element and its content on the current page.</p>
-            <pre class="example highlight">&lt;button type="button" aria-controls="mangosteen" aria-expanded="false"&gt;Mangosteen&lt;/button&gt;
+            <pre class="example highlight">
+&lt;button type="button" aria-controls="mangosteen" aria-expanded="false"&gt;Mangosteen&lt;/button&gt;
 &lt;div id="mangosteen" hidden&gt;
   An edible fruit native to tropical lands surrounding the Indian Ocean.
-&lt;/div&gt;</pre>
+&lt;/div&gt;</pre
+            >
```
front-endian, line 13709: Is there a reason for this added newline and indent before the `>` of the `</pre>`?

front-endian (approved):
Everything reads very cleanly. Added a comment but it is exceedingly minor.

index.html
scottaohara: This is from the prettier script that runs when a build is made. This “change” and other changes to the aria spec index file have nothing to do with this pr. They just unfortunately get included when the PR gets updated to match the main branch

PR #2531, opened 2025-05-10
Remove redundant UA CSS requirement

### LINKED ISSUE #2521: Duplicates requirements of Selectors
# Describe your concern

This requirement

> If a CSS selector includes a WAI-ARIA attribute (e.g.,
input[aria-invalid="true"]
), user agents MUST update the visual display of any elements matching (or no longer matching) the selector any time the attribute is added/changed/removed in the DOM.

duplicates an existing requirement. We should not duplicate requirements like that as it will lead to people being confused. If ARIA has to call it out, why doesn't SVG have to call it out or MathML? Does that mean an invalid attribute cannot be matched?

We should only require things once, even across specifications.

Link to documentation: https://w3c.github.io/aria/#ua_noninterference and https://w3c.github.io/aria/#document-handling_css-selectors

Yes.

scottaohara:
closes #2521

removes the following from the spec:
>If a CSS selector includes a WAI-ARIA attribute (e.g.,
input[aria-invalid="true"]
), user agents MUST update the visual display of any elements matching (or no longer matching) the selector any time the attribute is added/changed/removed in the DOM.

as CSS already defines how UAs are to handle element/attribute selectors

### DIFF
--- a/index.html
+++ b/index.html
@@ -792,9 +792,7 @@
           language.
         </p>
         <p>
-          If a CSS selector includes a WAI-ARIA attribute (e.g., <code class="highlight css">input[aria-invalid="true"]</code>), user agents MUST update the visual display of any elements matching (or
-          no longer matching) the selector any time the attribute is added/changed/removed in the <abbr title="Document Object Model">DOM</abbr>. The user agent MAY alter the mapping of the host
-          language features into an <a>accessibility <abbr title="Application Programming Interface">API</abbr></a
+          User agents MAY alter the mapping of host language features into an <a>accessibility <abbr title="Application Programming Interface">API</abbr></a
           >, but the user agent MUST NOT alter the <abbr title="Document Object Model">DOM</abbr> in order to remap <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> markup into host
           language features.
         </p>


scottaohara:
i know the issue is marked for discussion in an upcoming ARIA editor's meeting.  We can discuss this PR / what may need to change about it then as well.  cc @pkra

pkra:
If we remove this, we should also remove the normative statement in 8.7.

I think it would be good to keep a note on the general topic anyway, with a link to somewhere in CSS where these normative statements are made.

I put it on the editors' agenda because I've been wondering about what "supporting host languages" means these days, i.e. what does ARIA expect from a "supporting host language"?

pkra:
From our conversation: I'll update the PR to rewrite the second occurrence to make a reference https://drafts.csswg.org/selectors/#attribute-selectors (via xref https://respec.org/xref/?term=attribute+selector).

pkra:
@scottaohara I finally got around to this. What do you think of the additional change?

pkra:
FWIW at some point I wondered why it's not ok to have a normative requirement from the ARIA end - "if you support some parts of CSS as well as ARIA, you MUST support selectors-4". That kind of thing seems like a legitimate requirement for ARIA. But it's also irrelevant in this particular case (since attribute selectors are universally supported).

index.html
```diff
@@ -901,10 +899,8 @@ <h2><abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> States a
         </p>
         <p>
           Most modern user agents support
-          <cite
-            ><a href="https://www.w3.org/TR/css3-selectors/#attribute-selectors"><abbr title="Cascading Style Sheets">CSS</abbr> attribute selectors</a></cite
-          >
-          ([[CSS3-SELECTORS]]), and can allow the author to create <abbr title="User Interface">UI</abbr> changes based on <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> attribute
+          
+          CSS [=attribute selectors=] ([[SELECTORS-4]]), and can allow the author to create <abbr title="User Interface">UI</abbr> changes based on <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> attribute
```
annevk:
```suggestion
          CSS [=attribute selectors=] ([[SELECTORS-4]]) and allow the author to create <abbr title="User Interface">UI</abbr> changes based on <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> attribute
```

index.html
```diff
@@ -16561,11 +16557,10 @@ <h3>ID Reference Error Processing</h3>
...
-        <p class="note">This section might be removed in a future version.</p>
-        <p>
-          Support for <a class="termref">attribute</a> selectors MUST include <abbr title="Accessible Rich Internet Application">WAI-ARIA</abbr> attributes. For example,
-          <samp>.fooMenuItem[aria-haspopup=&quot;true&quot;]</samp> would select all [=elements=] with class <code>fooMenuItem</code>, and
-          <abbr title="Accessible Rich Internet Application">WAI-ARIA</abbr> property <pref>aria-haspopup</pref> with value of <code>true</code>. The presentation MUST be updated for dynamic changes
+        <p class="note">
+          If a host language supports CSS, then <abbr title="Accessible Rich Internet Application">WAI-ARIA</abbr> attributes are valid for [=attribute selectors=] ([[SELECTORS-4]]). For example,
+          <samp>.fooMenuItem[aria-haspopup=&quot;true&quot;]</samp> would select all [=elements=] with both class <code>fooMenuItem</code> and
+          <abbr title="Accessible Rich Internet Application">WAI-ARIA</abbr> property <pref>aria-haspopup</pref> with value of <code>true</code>. Similarly, the presentation updates for dynamic changes
```
annevk: I would drop this section. You already have a reference to Selectors earlier, which is probably already more than is necessary.

annevk:
> FWIW at some point I wondered why it's not ok to have a normative requirement from the ARIA end

This is not okay because we should not duplicate requirements. That often results in confusion and makes people misinterpret what is actually required. It's fine to point out it is required elsewhere through a note or statement of fact, but duplicating conformance criteria is an anti-pattern.

pkra:
> This is not okay because we should not duplicate requirements.

What I was trying to describe was not a duplication.

annevk:
Oh sorry, I misread. I agree you could do that, but I'm not sure it's useful.

index.html
scottaohara: agreed.  removed the whole section.

PR #2533, opened 2025-05-12
[Core-AAM] Update role=scrollbar AXAPI platform mappings

rahimabdi:
Closes https://github.com/w3c/core-aam/issues/242

Update Core-AAM's role=scrollbar AXAPI mapping to use NSAccessibility scrollbar APIs where `aria-controls` is provided, e.g., https://developer.apple.com/documentation/appkit/nsaccessibility-swift.struct/attribute/verticalscrollbar?language=objc.

* [ ] For every spec that this PR edits, please add the appropriate `spec:<spec_name>` label. If you don't have privileges to do this, editors will do it for you.

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko: n/a
   * Blink: n/a
* [ ] Does this need AT implementations? n/a
* [ ] Related APG Issue/PR: n/a
* [ ] MDN Issue/PR: n/a

### DIFF
--- a/core-aam/index.html
+++ b/core-aam/index.html
@@ -4259,6 +4259,12 @@
                   ><br />
                   <span class="property">AXSubrole: <code>&lt;nil&gt;</code></span
                   ><br />
+                  <p>
+                    <span class="property">Property: <code>NSAccessibilityVerticalScrollBarAttribute</code>: pointer to accessible node matching IDREF of author-provided <code>aria-controls</code> for an element that represents a vertical scrollbar.</span>
+                  </p>
+                  <p>
+                    <span class="property">Property: <code>NSAccessibilityHorizontalScrollBarAttribute</code>: pointer to accessible node matching IDREF of author-provided <code>aria-controls</code> for an element that represents a horizontal scrollbar.</span>
+                  </p>
                 </td>
               </tr>
             </tbody>


core-aam/index.html
```diff
@@ -4259,6 +4259,18 @@ <h4 id="role-map-scrollbar"><code>scrollbar</code></h4>
                   ><br />
                   <span class="property">AXSubrole: <code>&lt;nil&gt;</code></span
                   ><br />
+                  <p>
+                    <span class="property"
+                      >Property: <code>NSAccessibilityVerticalScrollBarAttribute</code>: pointer to accessible node matching IDREF of author-provided <code>aria-controls</code> for an element that
+                      represents a vertical scrollbar.</span
```
spectranaut, line 4644: Hey @rahimabdi, with my read of the [attribute's documentation](https://developer.apple.com/documentation/appkit/nsaccessibility-swift.struct/attribute/verticalscrollbar?language=objc)... this property is not exposed scrollbar element, but on the element that the scroll bar controls? If so then these two properties don't belong here, because they aren't properties surfaced on the scrollbar.

Or maybe we could still say something about these properties here... something like: 

> If `aria-controls` is set on this the scrollbar element and references a valid element, and the scrollbar has `aria-orientation=vertical`, property <code>NSAccessibilityVerticalScrollBarAttribute</code> should be exposed on the accessible node matching IDREF of author-provided <code>aria-controls</code>.  If the scrollbar has `aria-orientation=horizontal`, property <code>NSAccessibilityHorizontalScrollBarAttribute</code> should be exposed on the accessible node matching IDREF of author-provided <code>aria-controls</code>.

index.html
```diff
@@ -13681,9 +13681,15 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+            <p>
+              The <sref>aria-expanded</sref> attribute is applied to a focusable, interactive element that toggles visibility of content of a different element. If the element with
+              <sref>aria-expanded</sref> is also a <rref>treeitem</rref> in a <rref>tree</rref> or a <rref>row</rref> in a <rref>treegrid</rref>, then it SHOULD also be the
+              <a>accessibility parent</a> of the content it expands and collapses. Otherwise, the element with <sref>aria-expanded</sref> SHOULD NOT be the <a>accessibility parent</a> of the content
+              that is expanding or collapsing. Rather, identify that relationship between the interactive element and the element being controlled using <pref>aria-controls</pref>.
+            </p>
```
spectranaut, line 13721: this looks like formatting changes that snuck in because of prettier... :/ maybe try rebasing?

spectranaut (changes requested):
Hi Rahim! A couple of things.

PR #2542, opened 2025-05-30
Add placeholder iOS mapping rows

scottaohara:
We are planning to add iOS mappings to the html element and attribute tables.  This PR adds the rows for the iOS mappings, but everything is marked as TBD for now.

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -349,10 +349,14 @@
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


scottaohara:
per brief mention on the ARIA call today
many of the TBDs will likely need to be changed to "use wai-aria mapping"
and the real meat of this addition would need to be in core aam - and updating that spec should also be part of this PR

PR #2543, opened 2025-06-06
Editorial: Add <header> (sectionheader) and <footer> (sectionfooter) html-aam mappings

rahimabdi:
🚀 **Netlify Preview**:
- [html-aam preview](https://deploy-preview-2543--wai-aria.netlify.app/html-aam/index.html) &mdash; [html-aam diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fhtml-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-2543--wai-aria.netlify.app%2Fhtml-aam%2Findex.html)

Closes https://github.com/w3c/html-aam/issues/585

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2543.html" title="Last updated on Sep 8, 2025, 8:52 PM UTC (d581d10)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2543/0256e79...d581d10.html" title="Last updated on Sep 8, 2025, 8:52 PM UTC (d581d10)">Diff</a>


html-aam/index.html
```diff
@@ -2513,7 +2513,7 @@ <h4 id="el-footer">`footer` (scoped to the `main` element,a sectioning content e
             </tr>
             <tr>
               <th>[[wai-aria-1.2]]</th>
-              <td><a class="core-mapping" href="#role-map-generic">`generic`</a> role</td>
+              <td><a class="core-mapping" href="#role-map-sectionfooter">`sectionfooter`</a> role if the <a data-cite="html">`footer`</a> element has an <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>. Otherwise, the <a class="core-mapping" href="#role-map-generic">`generic`</a> role.</td>
```
scottaohara: the intent of the role change would be to do away with generic all together, so i think that needs to be removed.

aardrian: So the presence of an accName is not a factor?

giacomo-petri: I cannot agree more with Scott.
In a real-world scenario (e.g., article context)  there's no need to assign an accessible name to these elements (and an author does not expect to do that), as the article context already conveys the meaning of the nested header and footer. 
This also seems consistent with the ARIA specifications, which do not require an accessible name for such roles.

rahimabdi: @scottaohara @giacomo-petri Agreed, the computed role for `<header>` and `<footer>` should be `sectionheader`/`sectionfooter` in a sectioning element context in terms of mapping and for interop purposes so will remove the `generic` clause.

However, when `<header>`/`<footer>` are descendants of sectioning elements, my understanding is that they should be exposed as navigable containers only when they are supplied with accname (like `role="region"`, which requires disambiguation). The Webkit and Chrome implementations both behave in this way so if this is incorrect, I can submit bugs to get this fixed.

giacomo-petri: @rahimabdi,

do you have a concrete example to illustrate your point? I'm having trouble seeing how `<header>` and `<footer>` could be descendants of other elements in the way you are suggesting.

From my understanding, and based on my testing, `<header>` and `<footer>` are either scoped to the nearest sectioning content (e.g. main, article, etc.), in which case they are mapped to sectionheader/sectionfooter, or they default to the page level, scoped to body, and get mapped to banner/contentinfo respectively.

Here’s what I mean with an example:

```
<body>
    <table>
        <tbody>
            <tr>
                <th>test</th>
                <th>test</th>
            </tr>
            <tr>
                <td><header>test</header></td>
                <td><footer>test</footer></td>
            </tr>
        </tbody>
    </table>
</body>
```

assuming it's somehow meaningful, hedaer and footer still retain the banner/contentinfo mapping, while in 

```
<body>
    <main>
        <table>
            <tbody>
                <tr>
                    <th>test</th>
                    <th>test</th>
                </tr>
                <tr>
                    <td><header>test</header></td>
                    <td><footer>test</footer></td>
                </tr>
            </tbody>
        </table>
    </main>
</body>
```

they are mapped as sectionheader/sectionfooter

I don't see any other scenarios, unless we explicitly decide that, under certain conditions, these elements should instead map to a generic role.

giacomo-petri: Sorry, I think I misunderstood the comment. In my view, since `<header>` and `<footer>` elements used as descendants of sectioning elements (like `<article>`) usually have a well-defined context, IMO authors typically don't need to provide an accessible name. Doing so would often just repeat the context that's already implied. I don't think developers are likely to add aria-label/aria-labelledby to native `<header>` or `<footer>` elements within an article, because their meaning is already clear from the structure.

This is also why, in my opinion, we differentiate roles like banner/contentinfo from sectionheader/sectionfooter. We want to avoid placing extra burdens on authors.

Moreover, `<header>` and `<footer>` elements, when used within sectioning elements, carry a more specific contextual meaning than generic `<section>` elements. While sections are often used as generic containers and may not always represent regions (reason why we require acc name), headers and footers usually indicate a defined part of a sectioning element, and not a generic one. If their use doesn't carry that semantic weight, they are better replaced with more neutral containers.

scottaohara: This was discussed at length when these roles were created. That they are essentially role=group with platform localized role descriptions


So if they are named, the role is exposed by screen readers

if they are not named, then the role is treated the same as the group role and screen readers ignore it.  BUT the role doesn’t change. It isn’t generic or none…

does this reminder of what these are supposed to act like ring any bells / squash any doubts?

scottaohara: Per the example of the header/footer in the table - further updates to when these elements map to the section* vs landmark roles is going to be handled in another pr

rahimabdi: Thanks @scottaohara. I've updated the PR so ready for another review.

@giacomo-petri Please let me know if there's any further changes, thanks!

giacomo-petri: I'm okay with it, but I'm still not entirely convinced that these elements should be exposed by ATs as generic if they don't have an accessible name. Unlike other elements like articles, dialogs, regions, etc., where we are working on deriving names from headings (#1860), I believe these items often don't have a heading, and the content itself shouldn't be used as the name. 
In some cases, the heading of the container (e.g., the article heading) might make sense as the name for the contained footer, but that won't always be true.

I also don't see a strong case for requiring developers to write markup like:
```
<article>
  <h1>Title</h1>
  <div>Content</div>
  <footer aria-label="name">Footer</footer>
</article>
```
just to have ATs expose it as a sectionfooter.

That said, I don't want to block this PR.

index.html
```diff
@@ -16774,8 +16773,8 @@ <h2>Interface Mixin <dfn>ARIAMixin</dfn></h2>
 				[CEReactions] attribute DOMString? ariaValueText;
 			};
       Element includes ARIAMixin;
-		</pre>
-
+		</pre
+        >
```
keithamus, line 16810: This doesn't look right.

keithamus (commented):
Looks like this needs rebasing.

index.html
pkra: It's prettier formatting. Just ignore it.

spectranaut:
Agenda because old and needs reviews, Scott thought people might need clarifications/reminders about it.

spectranaut:
This is approved by the working group according to the discussion in yesterday's meeting: https://www.w3.org/2025/09/04-aria-minutes.html#a095

But it needs reviews.

rahimabdi:
@scottaohara @spectranaut @giacomo-petri I undid the prettier changes but would definitely request a final review before merging.

html-aam/index.html
cookiecrook: #1860 is now superseded by:
- #2650

pkra:
WPT tests: https://wpt.fyi/results/html-aam/roles-contextual.tentative.html?label=experimental&label=master&aligned 

@rahimabdi should we file an issue with WPT to change the name of the test file (to remove "tentative")?

pkra:
Discussed on today's editors' meeting https://www.w3.org/2026/02/25-aria-editors-minutes.html#7738

Issue #2433, opened 2025-02-12
follow up to css revision


pkra:
Follow up to https://github.com/w3c/aria/pull/2410.

Reviewing the spec css identified various other issues. Let's collect them here

- [ ] nested `.rfc2119` -- look like simple authoring errors 
- [ ] highlight.js failing to identify snippet language -- add explicit source info
- [ ] aria.js: dl with `.compact` (for lists of roles and states&props - 5.4, 6.8) -  class no longer used 
- [ ] ...

pkra:
Some of these items are good first issues (e.g., fixing the HTML source).

pkra:
I couldn't figure out a way to configure respec for a more accessible highlightjs theme so [the respec issue I already filed](https://github.com/speced/respec/issues/4871) has to suffice.

pkra:
The remaining PRs have been merged.

PR #2553, opened 2025-06-13
editorial: remove redundant rfc2119 markup

pkra:
Prevents duplicate/nested em elements since respec generates those.

Part of #2433


pkra:
As is hopefully obvious, the RFC checker is giving a false positive.

PR #2566, opened 2025-06-24
Editorial: Set DPUB 1.1 to finished

daniel-montalvo:
This prevents further pushes to TR space from DPUB 1.1 specs, which are now finished.


pkra (approved):

daniel-montalvo:
This doesn't need to happen now. Added new shortname in the w3c_override key in the actions is sufficient to stop publication.

PR #2564, opened 2025-06-20
Uses custom action to run prettier on PRs and on pushes to main

daniel-montalvo:
Closes #????

Describe Change Here!

And a few other todo items (delete this section after performing them):
* [ ] For every spec that this PR edits, please add the appropriate `spec:<spec_name>` label. If you don't have privileges to do this, editors will do it for you.
* [ ] If the change is [editorial](https://github.com/w3c/aria/blob/main/documentation/process.md#editorial-changes), please add "Editorial:" at the start of your PR name, and delete the "Test, Documentation and Implementation tracking" section below.

# Test, Documentation and Implementation tracking
Once this PR has been reviewed and has consensus from the working group, tests should be written and issues should be opened on browsers. Add N/A and check when not applicable.

* [ ] "author MUST" tests:
* [ ] "user agent MUST" tests:
* [ ] Browser implementations (link to issue or commit):
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Does this need AT implementations?
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

PR #2563, opened 2025-06-20
editorial: definition of "host language"

pkra:
Adds and exports a definition for host language.
Also incorporates some of the discussion in #2261.

Closes #2035, closes #2261

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2563.html" title="Last updated on Aug 18, 2025, 7:48 AM UTC (739db24)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2563/34306ed...739db24.html" title="Last updated on Aug 18, 2025, 7:48 AM UTC (739db24)">Diff</a>

PR #2562, opened 2025-06-20
UA SHOULD guidance for roles outside of required a11y parent role

pkra:
Replicates #2418 by @rahimabdi

Closes https://github.com/w3c/aria/issues/2012, closes https://github.com/w3c/aria/pull/2418

This PR adds additional user agent guidance for how orphan roles SHOULD be treated outside of the context of a required accessibility parent role.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2562.html" title="Last updated on Jul 9, 2025, 4:27 PM UTC (ab43d35)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2562/3030dd8...ab43d35.html" title="Last updated on Jul 9, 2025, 4:27 PM UTC (ab43d35)">Diff</a>

PR #2568, opened 2025-07-03
Editorial: Trying to trick prettier into running

spectranaut:


spectranaut:
@pkra should we just get all these prettier changes checked in that keep showing up in other PRs?

pkra:
I reviewed it but now I'm worried that merging it will create merge conflicts. I'll check the other PRs.

pkra:
@spectranaut I'm closing this because I'm worried that this will mess up #2484 - which is difficult enough as is (and on a fork). #2562 and #2563 would also get merge conflicts but are less complex - removing the prettier commit and rebasing should be straight forward for them.

pkra:
<s>I think #2563 is the cleanest prettier commit we'll get (cleaner than this one because of the added&removed empty lines).</s>

pkra:
Ugh, I meant #2562 is better. #2563 has its own sloppy authoring being prettified. This makes me feel like we should just stop doing this on PR. Sigh. I hope things calm down now that we have #2564.

PR #2211, opened 2024-05-23
add arianotify draft

keithamus:
This adds some mechanics for introducing the `ariaNotify` API on Elements, which is a function that will map to various AT APIs, as defined in core-aam https://github.com/w3c/core-aam/pull/232

# PR tracking
Check these when the relevant issue or PR has been made, OR after you have confirmed the
related change is not necessary (add N/A). Leave unchecked if you are unsure. Read the
[Process Document](https://github.com/w3c/aria/wiki/ARIA-WG-Process-Document/_edit) or
[Test Overview](https://github.com/w3c/aria/wiki/ARIA-Test-Overview) for more information.

* [ ] Related Core AAM Issue/PR:
* [ ] Related AccName Issue/PR:
* [ ] Any other dependent changes?

# Test, Documentation and Implementation tracking
Once this PR and all related PRs have been  approved by the working group, tests
should be written and issues should be opened on browsers. Add N/A and check when not
applicable.

* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:
* [ ] "author MUST" tests:
* [ ] "user agent MUST" tests:
* [ ] Browser implementations (link to issue or when done, link to commit):
   * WebKit:
   * Gecko:
   * Blink:
* [ ] Does this need AT implementations?

Issue #2499, opened 2025-03-26
[ariaNotify] avoid pestering/annoyance of notifications, either too many or too irrelevant


cookiecrook:
Moved at @alisonmaher's request from: 
- https://github.com/w3c/aria/discussions/1958#discussioncomment-10363453

## AriaNotify should avoid pestering/annoyance of notifications, either too many or too irrelevant

I understand many aspects of the ariaNotify proposal (types, notification IDs, element-scopes, priority, division of critical vs interruptible text) are intended to reduce the annoyance of live regions and allow some level of verbosity or relevance control, particularly in the scenarios where irresponsible misuse equates to the level of abuse.

However, I also see a primary goal of the API to be “simpler,” easier to use than Live Regions, etc, which could lead us (even with the best of intentions) to an even-more abusable API for speech and braille announcements.

What follows is a list of known severe misuse scenarios that have proven annoying to screen reader users, so annoying that Apple engineers have implemented VoiceOver prefs to shut off live region notifications entirely. We’d like to provide a path to give users incentives to enable some types of helpful notifications while also giving them sufficient control to disallow abusive misuse, but the live regions API is vague enough that it’s differentiate these unwanted notifications from genuinely useful notifications.

### Examples of misuse/abuse of Live Regions that should be disincentivized with a new API
- live regions in advertising containers announcing irrelevant or unwanted content
- timers or progress indicators using polite or assertive notifications, rather than “off”
- extra chatty “log” containers, particularly that conflict with the primary content… For example, a social media live stream (Twitch, Instagram, etc) that displays excessive amounts of “likes” or comments, such that the main audio is trampled by live regions notifications

#### Specific examples:
- This [Texas Monthly article](https://www.texasmonthly.com/being-texan/forgotten-female-sharpshooter-surpassed-annie-oakley/) misuses an assertive live region to count down an ad, making the webpage unusable.
- Numerous examples of custom audio and video players on the web have added `role=status` to the timecode element (0:42) which updates every second, making the video and other page content unusable.

### Miscellaneous musings of how to address these and other misuse scenarios.
- Some ideas already mentioned in the ariaNotify discussion thread (types, notification IDs, element scoping, priority, division of critical vs interruptible text, etc.) would allow more granular control by the web engine or downstream AT
- “ignore ongoing notifications from [this labeled control] or [from the site altogether]” similar to social media site (Slack, etc) features for “mute notifications for this thread”
[...]

PR #2577, opened 2025-07-16
Add ariaNotify

janewman:
This adds the mechanics for introducing the ariaNotify API on Elements, which is a function that will map to various AT APIs, as defined in core-aam.

[Aria Notify Explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/Accessibility/AriaNotify/explainer.md#iframes-and-use-in-subresources)

Original PR: https://github.com/w3c/aria/pull/2211

* [x] "author MUST" tests: not needed
* [x] "user agent MUST" tests: not available (waiting for WPT support)
* [x] At least two browser implementations (link to issue or commit):
   * [x] WebKit: implemented. ([standards position: Positive, with concerns](https://github.com/WebKit/standards-positions/issues/370))
   * [ ] Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=2005588 ([standards position: Positive](https://github.com/mozilla/standards-positions/issues/1049))
   * [x] Blink: implemented in 140, see [issue 326277796](https://issues.chromium.org/issues/326277796)
* [x] ACT review? n/a
* [ ] Does this need AT implementations? Yes.
  * [x] Work has been done and basic announcements function in NVDA, JAWS, Narrator, Orca, and VoiceOver
  * [ ] However, the v2 roadmap includes ideas that are not yet fully implementable. New platform API is needed on Mac, for example, probably more platforms. [Edit by @cookiecrook]
* [x] Related APG Issue/PR: https://github.com/w3c/aria-practices/issues/3414
* [x] MDN Issue/PR: => https://developer.mozilla.org/en-US/docs/Web/API/Document/ariaNotify

***
<a href="https://pr-preview.s3.amazonaws.com/janewman/aria/pull/2577.html" title="Last updated on Jan 14, 2026, 1:00 PM UTC (20c4ad4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2577/4c60da4...janewman:20c4ad4.html" title="Last updated on Jan 14, 2026, 1:00 PM UTC (20c4ad4)">Diff</a>

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -3160,7 +3160,10 @@
             </tr>
           </tbody>
         </table>
-        <h4 id="el-img-empty-alt">`img` <span class="el-context">(`alt` attribute value is the empty string, i.e., `alt=""` or `alt` with no value in the markup)</span></h4>
+        <h4 id="el-img-empty-alt">
+          `img`
+          <span class="el-context">(`alt` attribute value, when trimmed of [=ascii whitespace|whitespace=], is the empty string, i.e., `alt=""`, `alt=" "`, or `alt` with no value in the markup)</span>
+        </h4>
         <table class="data" aria-labelledby="el-img-empty-alt">
           <tbody>
             <tr>
@@ -3168,7 +3171,8 @@
               <td>
                 <a data-cite="html">`img`</a>
                 <span class="el-context"
-                  >(<a data-cite="html/embedded-content.html#attr-img-alt">`alt`</a> attribute value is the empty string, i.e., `alt=""` or `alt` with no value in the markup)</span
+                  >(<a data-cite="html/embedded-content.html#attr-img-alt">`alt`</a> attribute value, when trimmed of [=ascii whitespace|whitespace=], is the empty string, i.e., `alt=""`, `alt=" "`,
+                  or `alt` with no value in the markup)</span
                 >
               </td>
             </tr>
@@ -16620,8 +16624,8 @@
               </p>
             </li>
             <li>
-              If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty: use the value of the control's `alt` attribute if present and its value is not the empty
-              string.
+              If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty: use the value of the control's `alt` attribute if present and its value, when trimmed of
+              [=ascii whitespace|whitespace=], is not the empty string.
             </li>
             <!-- NOTE: use of value attribute is invalid on input type=image, but it DOES contribute to the name if used.
                        if this were to be specified, it'd belong here in the steps -->
@@ -16749,9 +16753,10 @@
               <a href="" class="accname">Accessible Name and Description: Computation and API Mappings</a>.
             </li>
             <li>
-              If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty: use `alt` attribute, even if its value is the empty string.
+              If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty: use `alt` attribute, even if its value, when trimmed of [=ascii whitespace|whitespace=], is the
+              empty string.
               <div class="note">
-                An `img` with an `alt` attribute whose value is the empty string is mapped to the
+                An `img` with an `alt` attribute whose value, when trimmed of [=ascii whitespace|whitespace=], is the empty string is mapped to the
                 <a class="core-mapping" href="#role-map-presentation">`presentation`</a> role. It has no accessible name.
               </div>
             </li>
@@ -16902,7 +16907,7 @@
       </section>
     </section>
 
-    <!-- obsolete features 
+    <!-- obsolete features
     <tr tabindex="-1" id="el-rb">
         <th>
           <a href="https://html.spec.whatwg.org/#rb">`rb`</a>
@@ -16986,8 +16991,8 @@
               <td class="atk"><div class="general">Not mapped</div></td>
               <td class="ax"><div class="general">Not mapped</div></td>
               <td class="comments"></td>
-            </tr> 
-         
+            </tr>
+
 
          other obsolete HTML elements
          applet
--- a/pdf-aam/index.html
+++ b/pdf-aam/index.html
@@ -285,6 +285,209 @@
 
         <section id="mapping_role_table">
           <h3>Role Mapping Tables</h3>
+          <h4 id="role-art"><code> Art</code></h4>
+          <table class="data" aria-labelledby="role-art">
+            <tbody>
+              <tr>
+                <th>ARIA Specification</th>
+                <td>
+                  <a class="core-mapping" href="#role-map-article">Role: <code>article</code></a>
+                </td>
+              </tr>
+              <tr>
+                <th>Analogous HTML Element</th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th>MSAA + IAccessible2</th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="User Interface Automation">UIA</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th>
+                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+                </th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Notes">Notes:</abbr></th>
+                <td>
+                  <span class="property">Note from PDFa: <code>This is the default handling unless a different ARIA role is present</code></span>
+                </td>
+              </tr>
+            </tbody>
+          </table>
+
+          <!-- I'm unsure here how we want to denote use the WAI-ARIA mapping or should we say use HTML-AAM mapping? -->
+          <h4 id="role-aside"><code> Aside</code></h4>
+          <table class="data" aria-labelledby="role-aside">
+            <tbody>
+              <tr>
+                <th>ARIA Specification</th>
+                <td>
+                  <a class="core-mapping" href="#role-map-complementary">Role: <code>complementary</code></a>
+                </td>
+              </tr>
+              <tr>
+                <th>Analogous HTML Element</th>
+                <td>
+                  <a class="html-mapping" href="#el-aside">Role: <code>aside (scoped to the body or main element)</code></a>
+                </td>
+              </tr>
+              <tr>
+                <th>MSAA + IAccessible2</th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="User Interface Automation">UIA</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th>
+                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+                </th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Notes">Notes:</abbr></th>
+                <td>
+                  <span class="property">Note from PDFa: <code>Use WAI-ARIA mapping</code></span>
+                </td>
+              </tr>
+            </tbody>
+          </table>
+
+          <h4 id="role-attribute-textdecorationtype-underline"><code> Attribute (TextDecorationType Underline)</code></h4>
+          <table class="data" aria-labelledby="role-attribute-textdecorationtype-underline">
+            <tbody>
+              <tr>
+                <th>ARIA Specification</th>
+                <td>
+                  <a class="core-mapping" href="#role-map-generic"><code>generic </code></a>
+                </td>
+              </tr>
+              <tr>
+                <th>Analogous HTML Element</th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th>MSAA + IAccessible2</th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="User Interface Automation">UIA</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th>
+                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+                </th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Notes">Notes:</abbr></th>
+                <td>
+                  <span class="property">Note from PDFa: <code>Exposed by platform specific underline text styles.</code></span>
+                </td>
+              </tr>
+            </tbody>
+          </table>
+
+          <h4 id="role-div-has-attributes"><code> Div (has attributes)</code></h4>
+          <table class="data" aria-labelledby="role-div-has-attributes">
+            <tbody>
+              <tr>
+                <th>ARIA Specification</th>
+                <td>
+                  <a class="core-mapping" href="#role-map-group">Role: <code>group</code></a>
+                </td>
+              </tr>
+              <tr>
+                <th>Analogous HTML Element</th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th>MSAA + IAccessible2</th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="User Interface Automation">UIA</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th>
+                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+                </th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Notes">Notes:</abbr></th>
+                <td>
+                  <span class="property">Note from PDFa: <code>Use WAI-ARIA mapping</code></span>
+                </td>
+              </tr>
+            </tbody>
+          </table>
+
+          <h4 id="role-div-no-attributes"><code> Div (no attributes)</code></h4>
+          <table class="data" aria-labelledby="role-div-no-attributes">
+            <tbody>
+              <tr>
+                <th>ARIA Specification</th>
+                <td>
+                  <a class="core-mapping" href="#role-map-generic">Role: <code>generic</code></a>
+                </td>
+              </tr>
+              <tr>
+                <th>Analogous HTML Element</th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th>MSAA + IAccessible2</th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="User Interface Automation">UIA</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th>
+                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+                </th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Notes">Notes:</abbr></th>
+                <td>
+                  <span class="property">Note from PDFa: <code>Use WAI-ARIA mapping</code></span>
+                </td>
+              </tr>
+            </tbody>
+          </table>
+
           <h4 id="role-document"><code> Document (Child of Structure tree root)</code></h4>
           <table class="data" aria-labelledby="role-document">
             <tbody>
@@ -363,68 +566,23 @@
                 <td>
                   <span class="property"
                     >Note from PDFa:
-                    <code
-                      >Added note - question for the group... A screen reader shouldn't "announce" when it hits the first Document tag but should it "announce" when it hits "nested" (subsequent)
-                      Document tags, to tell someone they are in a "second" document? AND... the same question applies to Document Fragment, too. If "Yes," how do we indicate that in the AAM?)</code
-                    ></span
-                  >
+                    <code>
+                      Added note - question for the group... A screen reader shouldn't "announce" when it hits the first Document tag but should it "announce" when it hits "nested" (subsequent)
+                      Document tags, to tell someone they are in a "second" document? AND... the same question applies to Document Fragment, too. If "Yes," how do we indicate that in the AAM?)
+                    </code>
+                  </span>
                 </td>
               </tr>
             </tbody>
           </table>
 
-          <h4 id="role-sect"><code> Sect</code></h4>
-          <table class="data" aria-labelledby="role-sect">
+          <h4 id="role-document-fragment-block"><code> Document Fragment as Block-Level element</code></h4>
+          <table class="data" aria-labelledby="role-document-fragment-block">
             <tbody>
               <tr>
                 <th>ARIA Specification</th>
                 <td>
-                  <!-- Guidance needed on how to handle this entry.-->
-                  <a class="core-mapping" href="role-map-region"> Role: <code>region IF there is an accessible name. Otherwise use generic</code></a>
-                </td>
-              </tr>
-              <tr>
-                <th>Analogous HTML Element</th>
-                <td>
-                  <a class="html-mapping" href="el-section">Role: <code>section</code></a>
-                </td>
-              </tr>
-              <tr>
-                <th>MSAA + IAccessible2</th>
-                <td>Use WAI-ARIA mapping</td>
-              </tr>
-              <tr>
-                <th><abbr title="User Interface Automation">UIA</abbr></th>
-                <td>Use WAI-ARIA mapping</td>
-              </tr>
-              <tr>
-                <th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
-                <td>Use WAI-ARIA mapping</td>
-              </tr>
-              <tr>
-                <th>
-                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
-                </th>
-                <td>Use WAI-ARIA mapping</td>
-              </tr>
-              <tr>
-                <th><abbr title="Notes">Notes:</abbr></th>
-                <td>
-                  <span class="property"
-                    >Note from PDFa: <code>To calculate the "accessible name": The first Hn descendant that is not a descendant of another Sect element provides the accessible name.</code></span
-                  >
-                </td>
-              </tr>
-            </tbody>
-          </table>
-
-          <h4 id="role-P"><code> P</code></h4>
-          <table class="data" aria-labelledby="role-P">
-            <tbody>
-              <tr>
-                <th>ARIA Specification</th>
-                <td>
-                  <a class="core-mapping" href="#role-map-paragraph">Role: <code>paragraph </code></a>
+                  <a class="core-mapping" href="#role-map-generic">Role: <code>generic</code></a>
                 </td>
               </tr>
               <tr>
@@ -451,7 +609,49 @@
               </tr>
               <tr>
                 <th><abbr title="Notes">Notes:</abbr></th>
+                <td>
+                  <span class="property">Note from PDFa: <code>Use WAI-ARIA mapping</code></span>
+                </td>
+              </tr>
+            </tbody>
+          </table>
+
+          <h4 id="role-document-fragment-grouping"><code> Document Fragment as Grouping</code></h4>
+          <table class="data" aria-labelledby="role-document-fragment-grouping">
+            <tbody>
+              <tr>
+                <th>ARIA Specification</th>
+                <td>
+                  <a class="core-mapping" href="#role-map-group">Role: <code>group</code></a>
+                </td>
+              </tr>
+              <tr>
+                <th>Analogous HTML Element</th>
                 <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th>MSAA + IAccessible2</th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="User Interface Automation">UIA</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th>
+                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+                </th>
+                <td>Use WAI-ARIA mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Notes">Notes:</abbr></th>
+                <td>
+                  <span class="property">Note from PDFa: <code>Use WAI-ARIA mapping</code></span>
+                </td>
               </tr>
             </tbody>
           </table>
@@ -462,7 +662,6 @@
               <tr>
                 <th>ARIA Specification</th>
                 <td>
-                  <!-- need guidance on this mapping-->
                   <a class="core-mapping" href="#role-map-heading">Role: <code>ARIA heading (role)</code></a>
                   <a class="core-mapping" href="#ariaLevel">Property: <code>ARIA level (property)</code></a>
                 </td>
@@ -493,46 +692,6 @@
                 <th><abbr title="Notes">Notes:</abbr></th>
                 <td>
                   <span class="property">Note from PDFa: <code>heading with the aria-level property set to the number in the element's tag name.</code></span>
-                </td>
-              </tr>
-            </tbody>
-          </table>
-
-          <h4 id="role-lbl-forms"><code> Lbl</code> (associated with a Form element)</h4>
-          <table class="data" aria-labelledby="role-lbl-forms">
-            <tbody>
-              <tr>
-                <th>ARIA Specification</th>
-                <td>No corresponding role</td>
-              </tr>
-              <tr>
-                <th>Analogous HTML Element</th>
-                <td>
-                  <a class="html-mapping" href="#el-label">Role: <code>label</code></a>
-                </td>
-              </tr>
-              <tr>
-                <th>MSAA + IAccessible2</th>
-                <td>Use HTML-AAM mapping</td>
-              </tr>
-              <tr>
-                <th><abbr title="User Interface Automation">UIA</abbr></th>
-                <td>Use HTML-AAM mapping</td>
-              </tr>
-              <tr>
-                <th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
-                <td>Use HTML-AAM mapping</td>
-              </tr>
-              <tr>
-                <th>
-                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
-                </th>
-                <td>Use HTML-AAM mapping</td>
-              </tr>
-              <tr>
-                <th><abbr title="Notes">Notes:</abbr></th>
-                <td>
-                  <span class="property"><code></code></span>
                 </td>
               </tr>
             </tbody>
@@ -613,6 +772,46 @@
                 <th><abbr title="Notes">Notes:</abbr></th>
                 <td>
                   <span class="property">Note from PDFa: <code> Use WAI-ARIA mapping - BUT - "computed role" may change</code></span>
+                </td>
+              </tr>
+            </tbody>
+          </table>
+
+          <h4 id="role-lbl-forms"><code> Lbl</code> (associated with a Form element)</h4>
+          <table class="data" aria-labelledby="role-lbl-forms">
+            <tbody>
+              <tr>
+                <th>ARIA Specification</th>
+                <td>No corresponding role</td>
+              </tr>
+              <tr>
+                <th>Analogous HTML Element</th>
+                <td>
+                  <a class="html-mapping" href="#el-label">Role: <code>label</code></a>
+                </td>
+              </tr>
+              <tr>
+                <th>MSAA + IAccessible2</th>
+                <td>Use HTML-AAM mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="User Interface Automation">UIA</abbr></th>
+                <td>Use HTML-AAM mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
+                <td>Use HTML-AAM mapping</td>
+              </tr>
+              <tr>
+                <th>
+                  <abbr title="macOS Accessibility Protocol">AX API</abbr><sup>[<a href="#ftn.note1">Note 1</a>]</sup>
+                </th>
+                <td>Use HTML-AAM mapping</td>
+              </tr>
+              <tr>
+                <th><abbr title="Notes">Notes:</abbr></th>
+                <td>
+                  <span class="property"><code></code></span>
                 </td>
               </tr>
             </tbody>
@@ -704,13 +903,13 @@
             </tbody>
           </table>
 
-          <h4 id="role-attribute-textdecorationtype-underline"><code> Attribute (TextDecorationType Underline)</code></h4>
-          <table class="data" aria-labelledby="role-attribute-textdecorationtype-underline">
+          <h4 id="role-nonstruct"><code> NonStruct</code></h4>
+          <table class="data" aria-labelledby="role-nonstruct">
             <tbody>
               <tr>
                 <th>ARIA Specification</th>
                 <td>
-                  <a class="core-mapping" href="#role-map-generic"><code>generic </code></a>
+                  <a class="core-mapping" href="#role-map-generic">Role: <code>generic</code></a>
                 </td>
               </tr>
               <tr>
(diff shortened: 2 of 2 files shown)


janewman:
@cookiecrook, @jcsteh, @front-endian, thanks in advance for reviewing this change!
Please let me know if you have any questions, suggestions, or clarifications that would help you review. 

While I'll be out of the office from tonight (July 29th) until next Tuesday (Aug 5th), I'll be able to answer questions or make clarifications, just with a delay. Once I am back on July 11th, I can apply any feedback to this PR. Looking forward to hearing your thoughts!

@scottaohara, while I'll be out, would it make sense for this to be added to the agenda for this or next week's meeting, or would that only make sense if reviewers had something they would like to discuss specifically? I just want to ensure this keeps momentum.

Adding a few relevant links in case they are helpful:
* [I have published a chromium intent to ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/QCtWzIPgcCY), but we don't intend on moving forwards until we get positive signals from this forum.
* [Here is the link to the approved TAG review.](https://github.com/w3ctag/design-reviews/issues/1075)
* [And here is a polyfill implementation Github put together and a few folks are already using today.](https://github.com/github/arianotify-polyfill)

spectranaut:
@keithamus why is there also another PR with the same name, should we close it as a duplicate? https://github.com/w3c/aria/pull/2211

janewman:
@cookiecrook, @front-endian, friendly ping to PTAL at this when you get the chance!

core-aam/index.html
```diff
@@ -11198,6 +11198,115 @@ <h3>Special Events for Menus</h3>
...
+            <td>
+              <a class="method-reference" href="#arianotify"><code>aria notify</code></a>
+            </td>
+          </tr>
+          <tr>
+            <th>MSAA + IAccessible2</th>
+            <td>
+              <ol>
+                <li>
+                  <p>Return.</p>
```
front-endian: Just "Return." is less clear to me. I think I understand what is intended given the note below, but is it possible to add a little elaboration or clarification in the actual step here for **MSAA + IAccessible2**?

index.html
```diff
@@ -16795,6 +16795,48 @@ <h2>Interface Mixin <dfn>ARIAMixin</dfn></h2>
...
+          interface mixin ARIANotifyMixin {
+            undefined ariaNotify(DOMString announcement, optional AriaNotificationOptions options = {});
+          };
+          Element includes ARIANotifyMixin;
+          Document includes ARIANotifyMixin;
+        </pre>
+
+        <p>
+          User Agent processing for the
+          <code><dfn method for="ARIANotifyMixin" >ariaNotify</dfn>(<var>announcement</var>, <var>options</var>)</code>
```
front-endian, line 16829: Minor nit, but I had to read this sentence a couple times due to the phrasing. Might be easier to read as something like the following?

> The User Agent processing steps for the `ariaNotify(announcement, options)` method are:

front-endian (commented):
Left some minor suggestions and questions around two bits that might be able to be written a bit clearer, but everything else looked good to me.

core-aam/index.html
janewman: The intention here was to say that this won't work for MSAA + IAccessible2, as there isn't an appropriate API to call. Since chrome is moving forward with supporting UIA by default anyways, I didn't feel like this gap made sense to fill by attempting to add new functionality to MSAA/IA2, especially as all major screen readers are supporting the UIA call detailed here.

**I can clarify that this is a no-op, and detail the reasoning.**

Since we are here, I do want to mention that have a fallback implemented in chromium, but I don't think this should be standard behavior, thus didn't feel it belonged in the spec. I'll detail that below in case you were interested, or felt that this might make sense to add in some form:

The implementation that exists today behind a flag in chromium browsers has a "MSAA/IA2 Fallback" that exposes a temporary live region (exposed to screen readers only, not visible to web) so that the API can continue to function in a basic sense, even if on a chromium instance that doesn't support UIA.
In case you really wanted to dig in, here is the implementation: [6663991: Reland "[AriaNotify] Add IA2 fallback for AriaNotify"](https://chromium-review.googlesource.com/c/chromium/src/+/6663991)

core-aam/index.html
```diff
@@ -11198,6 +11198,115 @@ <h3>Special Events for Menus</h3>
...
+      <h3>ARIANotifyMixin Algorithm Mapping Tables</h3>
+      <p>
+        The <code>ARIANotifyMixin</code> provides methods for announcing content
+        to assistive technologies. The following algorithm mappings specify how
+        these announcements will be implemented across different accessibility
+        APIs to ensure consistent behavior for users of assistive technologies.
+      </p>
+      <h4 id="arianotifymixin-map-arianotify">ariaNotify</h4>
+      <p>To <code>aria notify</code> given <var>node</var>, <var>announcement</var>, and <var>priority</var>:</p>
+      <table aria-labelledby="arianotifymixin-map-arianotify">
```
spectranaut:
```suggestion
      <table aria-labelledby="arianotifymixin-map-arianotify" class="data">
```

For better formatting :)

core-aam/index.html
spectranaut: We say "not mapped" for attributes which aren't mapped in an API. We could write "No implementation specified.", and remove the order list.

Because if the specification is that MSAA/IA2 should do nothing, but chrome does this live region fall back, it would be incorrect according to the spec :)

I think it might be worth mentioning the fallback in the in the note... I don't think I understand what scenarios would cause the fall back to be be enacted? If a screen reader didn't support UIA?

index.html
```diff
@@ -16795,6 +16795,48 @@ <h2>Interface Mixin <dfn>ARIAMixin</dfn></h2>
...
+        </p>
+        <ol>
+          <li>
+            <p>If the document is not [=allowed to use=] the feature identified by <a>"aria-notify"</a> abort these steps.</p>
+          </li>
+          <li>
+            <p>Let <var>priority</var> be <var>options</var>["<code>priority</code>"].</p>
+          </li>
+          <li>
+            <p>Run the <a data-cite="core-aam-1.2#arianotify">aria notify steps</a> given <var>node</var>, <var>announcement</var>, and <var>priority</var>.</p>
```
spectranaut, line 16846: I think we need to be more clear what `node` is... maybe add a step:

> Let <var>node</var> be the accessible object that represents the Element.

... is "Element" the the right IDL way to refer to the element `el` of `el.ariaNotify(...)` ?

keithamus: I think `node` wants to be `this` here.

core-aam/index.html
janewman: I've updated this with a note calling out that user agents can do this optional fallback in scenarios where there isn't a applicable API to call.

>I don't think I understand what scenarios would cause the fall back to be be enacted? If a screen reader didn't support UIA?

Its not for if the screen reader doesn't support UIA (JAWS, NVDA, and Narrator all correctly handle the UIA notification today), but for when _the browser_ doesn't support using UIA. For chrome, we are still working on enabling UIA to 100%, and even then, there is an incompatibility with older versions of JAWS where chrome must stop using UIA altogether to ensure correct functionality.

I also added a note that we do the same kind of fallback on older versions of Linux/ATK. Here, it is a different case where ATK doesn't support a viable notification API to call.

Please let me know if this is the right place to put the note, and if there is anything I can clarify.

index.html
janewman: Thanks for the feedback! I agree this new version is clearer.

index.html
janewman: I applied Keith's suggestion, let me know if there is any additional clarification needed.

index.html
```diff
@@ -16827,6 +16827,8 @@ <h2>Interface Mixin <dfn>ARIANotifyMixin</dfn></h2>
           </li>
         </ol>
 
+        <p class="note">Assistive technologies may provide features that let users review past <code>ariaNotify</code> announcements. For example, a command could repeat the most recent announcement or open a chronological list of recent announcements. These capabilities can be implemented without additional author markup so users who miss or need to re-hear an announcement can easily do so.</p>
```
Heydon: This is very clear, succinct. Just wondering if we can strengthen it a bit, as I think it is very important for UX. Instead of "may provide features" can we not have "should provide features"? Or perhaps "It is recommended assistive technologies provide features..." is a good middle ground?

Heydon: Perhaps we can start with the problem: "Since `ariaNotify` announcements are not appended to the DOM, it's recommended assistive technologies provide their own features to let users review past announcements.”

?

janewman: Great suggestion, I updated the wording here, let me know what you think!

core-aam/index.html
```diff
@@ -11198,6 +11198,109 @@ <h3>Special Events for Menus</h3>
...
+            <td>
+              <ol>
+                <li>
+                  <p>Let <var>mapped_priority</var> be <code>Atk.Live.ATK_LIVE_ASSERTIVE</code> if <var>priority</var> is "high", otherwise <code>Atk.Live.ATK_LIVE_POLITE</code>.</p>
+                </li>
+                <li>
+                  <p>Call <code>g_signal_emit_by_name</code> with <var>node</var>, <code>"notification"</code>, <var>announcement</var>, and <var>mapped_priority</var>.</p>
+                </li>
+              </ol>
+              <p>On older Linux accessibility stacks prior to ATK 2.50.0, user agents may use the fallback; <a href="#arianotify-fallback-note">see fallback note</a>.</p>
```
spectranaut:
```suggestion
              <p>On older Linux accessibility stacks prior to ATK 2.50.0, user agents MAY use the fallback; <a href="#arianotify-fallback-note">see fallback note</a>.</p>
```

RFC-2119 words need all caps. And it seems right to use MAY.

core-aam/index.html
```diff
@@ -11198,6 +11198,109 @@ <h3>Special Events for Menus</h3>
...
+                </li>
+                <li>
+                  <p>Call <code>NSAccessibilityPostNotificationWithUserInfo</code> with <var>window</var>, <var>NSAccessibilityAnnouncementRequestedNotification</var>, and <var>userInfo</var>.</p>
+                </li>
+              </ol>
+            </td>
+          </tr>
+        </tbody>
+      </table>
+      <p id="arianotify-fallback-note" class="note">When no suitable platform notification API is available (for example, older Linux accessibility stacks prior to ATK 2.50.0 that lack the newer notification signal, or on Windows when the user agent is not able to use UIA), a user agent might synthesize a temporary, assistive-technology-only live region in its accessibility tree to convey an <code>aria notify</code> announcement. Such fallback nodes are not exposed to or detectable by web content and this behavior is not required.</p>
```
spectranaut:
```suggestion
      <p id="arianotify-fallback-note" class="note">When no suitable platform notification API is available (for example, older Linux accessibility stacks prior to ATK 2.50.0 that lack the newer notification signal, or on Windows when the user agent is not able to use UIA), a user agent MAY synthesize a temporary, assistive-technology-only live region in its accessibility tree to convey an <code>aria notify</code> announcement. Such fallback nodes are not exposed to or detectable by web content and this behavior is not required.</p>
```

I think its appropriate to use an RFC-2110 keyword here.

core-aam/index.html
spectranaut: This looks great now, thanks!

index.html
spectranaut: I think I'd prefer to assign `this` to `node` as a step in the algorithm rather than rely on essentially positional arguments here :)

essentially:
```
          <li>
            <p>Let <var>node</var> be <var>this</var>.</p>
          </li>
          <li>
            <p>Run the <a data-cite="core-aam-1.2#arianotify">aria notify steps</a> given <code>node</code>, <var>announcement</var>, and <var>priority</var>.</p>
          </li>
```

janewman: Edit applied, thanks for helping make this section clear.

spectranaut (approved):

core-aam/index.html
```diff
@@ -11198,6 +11198,109 @@ <h3>Special Events for Menus</h3>
...
+      </p>
+      <h3>ARIANotifyMixin Algorithm Mapping Tables</h3>
+      <p>
+        The <code>ARIANotifyMixin</code> provides methods for announcing content
+        to assistive technologies. The following algorithm mappings specify how
+        these announcements will be implemented across different accessibility
+        APIs to ensure consistent behavior for users of assistive technologies.
+      </p>
+      <h4 id="arianotifymixin-map-arianotify">ariaNotify</h4>
+      <p>To <code>aria notify</code> given <var>node</var>, <var>announcement</var>, and <var>priority</var>:</p>
```
jcsteh, line 12462: This is probably a broader issue, but it occurs to me that it's not entirely clear what to do here if `node` is hidden, aria-hidden, etc. The spec says that such elements should generally be excluded from the accessibility tree. However, I can't see any explicit specification that such elements must not fire accessibility events. Most of the time, this doesn't matter because events on hidden nodes aren't useful. But in this case, clients might not check whether the node is hidden.

intuitively, I would think that hidden nodes cannot fire notification events. I certainly hope that's the case because Gecko doesn't include hidden nodes in the accessibility tree at all.

What does Chromium do for ariaNotify called on a hidden or aria-hidden element?

core-aam/index.html
```diff
@@ -11198,6 +11198,109 @@ <h3>Special Events for Menus</h3>
...
+          </tr>
+          <tr>
+            <th><abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
+            <td>
+              <ol>
+                <li>
+                  <p>Let <var>mapped_priority</var> be <code>ATSPI_LIVE_ASSERTIVE</code> if <var>priority</var> is "high", otherwise <code>ATSPI_LIVE_POLITE</code>.</p>
+                </li>
+                <li>
+                  <p>Send a DBUS signal <code>ATSPI_DBUS_INTERFACE_EVENT_OBJECT</code> with <var>node</var>, <code>"announcement"</code>, <var>announcement</var>, and <var>mapped_priority</var>.</p>
```
jcsteh: Is it intentional that ATK uses "notification" but AT-SPI uses "announcement" for the signal name?

core-aam/index.html
```diff
@@ -11198,6 +11198,109 @@ <h3>Special Events for Menus</h3>
...
+                </li>
+                <li>
+                  <p>Let <var>userInfo</var> be a <code>NSDictionary</code> with the following keys:</p>
+                  <ol>
+                    <li><code>NSAccessibilityAnnouncementKey</code> as <var>announcement</var></li>
+                    <li><code>NSAccessibilityPriorityKey</code> as <var>mapped_priority</var></li>
+                  </ol>
+                </li>
+                <li>
+                  <p>Call <code>NSAccessibilityPostNotificationWithUserInfo</code> with <var>window</var>, <var>NSAccessibilityAnnouncementRequestedNotification</var>, and <var>userInfo</var>.</p>
```
jcsteh: It's rather unfortunate that the info about the source node is lost here. Can we use `node` instead of `window` or do the API and/or clients expect a window; e.g. do they ignore the announcement if we use the node instead? This isn't a blocking concern - for now, we're limited by the APIs we have available - but I thought this worth clarifying explicitly.

core-aam/index.html
janewman: Like firefox, chrome does not fire notifications on nodes it considers hidden, so I can add a line here about not considering hidden elements.

core-aam/index.html
janewman: I am a little fuzzy here, but after some reading, I believe this is correct.
[`DBUS_INTERFACE_EVENT_OBJECT` maps to` org.a11y.atspi.Event.Object`](https://docs.gtk.org/atspi2/const.DBUS_INTERFACE_EVENT_OBJECT.html)
and we are wanting to use [`org.a11y.atspi.Event.Object::Announcement`](https://gnome.pages.gitlab.gnome.org/at-spi2-core/devel-docs/doc-org.a11y.atspi.Event.Object.html#org-a11y-atspi-event-object-announcement).

However, for ATK, ["announcement" has been deprecated](https://docs.gtk.org/atk/signal.Object.announcement.html) in favor of ["notification"](https://docs.gtk.org/atk/signal.Object.notification.html). 


Let me know if something doesn't follow here.

jnurthen (approved):

cookiecrook (dismissed):
Wedging the door to prevent the merge (sorry), since there are now two approving reviews. I am trying to get a Standards Position answer for you by later this week. Won't be much longer.

core-aam/index.html
jcsteh: Perhaps this could refer to the section of the ARIA spec regarding [Excluding Elements from the Accessibility Tree](https://w3c.github.io/aria/#tree_exclusion).

core-aam/index.html
jcsteh: I guess AT-SPI maps from notification to announcement internally? I'm not familiar enough with the code to comment with any authority. Anyway, it sounds like you're across this, so I'll take your assurance that we're all good here.

core-aam/index.html
janewman: Looking into this, have to get my MacBook up and running to test any changes. I'll let you know what I find.

index.html
```diff
@@ -16795,6 +16795,61 @@ <h2>Interface Mixin <dfn>ARIAMixin</dfn></h2>
...
+          <li>
+            <p>Run the <a data-cite="core-aam-1.2#arianotify">aria notify steps</a> given <var>node</var>, <var>announcement</var>, and <var>priority</var>.</p>
+          </li>
+        </ol>
+
+        <p class="note">
+          Because <code>ariaNotify</code> announcements are not appended to the DOM, assistive technologies are encouraged to provide
+          features that let users review previous announcements. For example, a command could repeat the most recent announcement or open
+          a chronological list of recent announcements. These capabilities can be implemented without additional author markup so users
+          who miss or need to re-hear an announcement can easily do so.
```
cookiecrook: What's the expected expiration of previous notifications from the encouraged list feature? Clear on any navigation? What about in-page history mods? What about screen lock? 

Presumably there may be privacy implications to repetition of announcements from a tab or window that's been closed. What about use cases that may cross this privacy or security threshold? What if a user wants to wait for an announcement, but their screen locks in the meantime? The mainstream Notifications API accounts for this, but ariaNotify does not. Do the announcements continue through the lock screen? Regardless of the prior answer, are they reviewable when the user unlocks the screen?

jcsteh: FWIW, I had some reservations about this encouragement too. I can live with it because it's not a requirement, just an encouragement, but i honestly wonder whether it'd be better to strike this completely until it is more clearly defined and discussed with AT vendors (if it hasn't been already).

janewman: I'm fine with moving this out of this PR and into an issue we can use for further discussion. (I'll link the issue here once I create it.)

janewman: I removed the note from this PR, we can continue the discussion in https://github.com/w3c/aria/issues/2611

core-aam/index.html
```diff
@@ -11248,6 +11248,14 @@ <h4 id="arianotifymixin-map-arianotify">ariaNotify</h4>
                   </p>
                 </li>
               </ol>
+              <p class="note">
```
janewman, line 12513: Apologies for the late addition, I was reading over the steps for UIA and realized that this was missing.

In chromium, we consider "grouping-only" elements to not be considered as part of the UIA Content or Control view, but they are still exposed to ATs as part of the "raw view" so this is separate from the "hidden" consideration.

As Narrator does not listen to most notifications fired on elements that are _only_ in the raw view, finding and firing the event on the closest ancestor that is in the Content or Control view ensures that the notification is read out. This only impacts situations where developers call the API on an element with nothing that the user agent considers "interesting" enough for it to consider them as part of the Content or Control View. For example, this can happen in cases where it is fired on the `body`, when the body is purely used for grouping. e.g. `document.body.ariaNotify("announcement");`
See: [Views of the UI Automation Tree](https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-treeoverview#views-of-the-ui-automation-tree)

core-aam/index.html
janewman: This seems to be possible, at least in my own local testing. I'm putting together a chromium change, if nothing comes up for why we shouldn't use `node` here I'll update the spec as well.

@cookiecrook, let me know if you have thoughts here.

core-aam/index.html
jcsteh: The body element part of this is somewhat Chromium specific, since in Firefox, the body element always maps to the document Accessible (UIA_DocumentControlTypeId), rather than being a separate grouping child of the document. I guess this example just doesn't apply to Firefox, but I wonder whether it's worth removing the specific mention of the body element altogether to avoid any confusion.

janewman: Thanks for this feedback, updated the note accordingly.

core-aam/index.html
cookiecrook: Yes, developers can post the announcementNotification on any AXElement... There currently isn't any user-side difference in VoiceOver between announcements coming form different nodes in the same window, but there could be in the future. Probably best to use DOM `element` here rather than DOM `node`? I'm not sure what the behavior should be if a notification were posted on an unrendered comment node, for example. 

The spec should probably also document what should happen if notifications were posted on unrendered (E.g. hidden) elements, too. I'm guessing no notification should be sent. Is that right?

index.html
```diff
@@ -16795,6 +16795,54 @@ <h2>Interface Mixin <dfn>ARIAMixin</dfn></h2>
           <a href="#idl_attr_exceptions">IDL Attribute Name Notes or Exceptions</a>.
         </p>
       </section>
+      <section id="ARIANotifyMixin" class="normative" data-dfn-for="ARIANotifyMixin" data-link-for="ARIANotifyMixin">
+        <h2>Interface Mixin <dfn>ARIANotifyMixin</dfn></h2>
+        <pre class="idl" data-cite="webidl">
+          enum AriaNotifyPriority { "normal", "high" };
+
+          dictionary AriaNotificationOptions {
+            AriaNotifyPriority priority = "normal";
```
cookiecrook, line 16818: lang/loc seems necessary for i18n/l10n

cookiecrook (commented):
[WebKit Standards position for ariaNotify](https://github.com/WebKit/standards-positions/issues/370) coming in later today will be **Supports** (with remaining concerns).

index.html
janewman: Thanks for calling this out, I agree this is missing here. (I'll write an update shortly). 
This can be handled by asking the AT to use the document's language, or, when this API is called on an element, the nearest ancestor's lang attribute. This ensures we don't add complexity to the API itself, while ensuring authors have control. Do you think this will be sufficient?

This was outlined in [the explainer here](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/Accessibility/AriaNotify/explainer.md#faq):
>**Are Element-level notifications really necessary?**
Adding `ariaNotify` to Elements was driven by several goals: 
 >- Resolve the question of how language input should be provided. To keep the API simple, we are able to leverage the lang attribute that is used to override the document language for specific subtrees. `ariaNotify` can use the nearest ancestor element's lang attribute as a language hint (or the document's default language).

cookiecrook: If the author leaves it unspecified, I think DOM lang or user defaults are a reasonable way for the implementation to determine a language, but there will also be use cases where the announcement will be a different language than the page contents. 

Likewise, there could be use cases where all announcements are posted through a single delegate element like `body`. It would be inappropriate to change this element's HTML `lang` value for the announcement and then change it back.

I would not want to make the web author hunt for or have to artificially create an announcement courier node with the appropriate lang in those cases. Problems with courier elements was one of the Live Regions shortcomings that this new API was trying to solve.

It can just be another parameter option, right? Undefined by default; settable if needed.
```js
AriaNotifyLanguageLocale langloc = undefined; // lang: ISO 639-1 and optionally loc: ISO 3166-1, E.g. "en-us"
```

jcsteh: > [Language/locale] can just be another parameter option, right? Undefined by default; settable if needed.

This isn't going to be feasible for most existing APIs, since they have no way to pass that language information for a single notification message. Mac might be the exception, assuming that NSAccessibilityAnnouncementKey can be an NSAttributedString (I'm not sure if it can).

cookiecrook: @jcsteh wrote:
> This isn't going to be feasible for most existing APIs, since they have no way to pass that language information for a single notification message. 

There are other things in ariaNotify that aren't yet supported by platform APIs. Is this Windows API change fundamentally any different from the API changes needed for macOS? Any API that can't support a string language/locale in 2025 should change, right?

Pronunciation is important; ask @mhakkinen for the ETS use cases regarding language testing.
- "gift" (en-us) means a present.
- "gift" (de-de) means a poison. 

> Mac might be the exception, assuming that NSAccessibilityAnnouncementKey can be an NSAttributedString

https://developer.apple.com/documentation/foundation/nsattributedstring/key/accessibilitylanguage

core-aam/index.html
cookiecrook: > The spec should probably also document what should happen if notifications were posted on unrendered (E.g. hidden) elements, too. I'm guessing no notification should be sent. Is that right?

Was there prior discussion on this point? If not, we should discuss in an ARIA WG and align. This one seems critical to get in before the spec merge.

jcsteh: > The spec should probably also document what should happen if notifications were posted on unrendered (E.g. hidden) elements, too.

Is https://github.com/w3c/aria/pull/2577/files#diff-530e7cfb7275cd5fa6f81ae8347807f09511988b2ebb97b9a5a22c4973fe9068R11228 insufficient?

index.html
jcsteh: > There are other things in ariaNotify that aren't yet supported by platform APIs.

For this spec PR, are there? There are planned features - e.g. interrupt - that aren't supported by some APIs, but my understanding is that all features are supported by all APIs (Mac included) for this first version as currently written.

> Is this Windows API change

This isn't just Windows. It also applies to ATK.

> fundamentally any different from the API changes needed for macOS?

As above, are any API changes needed for MacOS for the initial spec here?

> Any API that can't support a string language/locale in 2025 should change, right?

Yes, except the issue isn't that the APIs don't support string languages/locales. They do. The issue is that this information is normally retrieved by querying the language of the node. Other APIs don't attribute the text string. Rather, you query attributes from the node separately from the text content.

> Pronunciation is important; ask @mhakkinen for the ETS use cases regarding language testing.

I agree, and I absolutely agree that it at least needs to be explicitly specified that the language/locale should be determined from the node. I'm just less convinced that there are enough use cases where the announcement will be a different language than the node contents to warrant blocking the initial set of ariaNotify functionality covered here.

> > Mac might be the exception, assuming that NSAccessibilityAnnouncementKey can be an NSAttributedString
> 
> https://developer.apple.com/documentation/foundation/nsattributedstring/key/accessibilitylanguage

Right, but I'm less clear that [NSAccessibilityAnnouncementKey](https://developer.apple.com/documentation/appkit/nsaccessibility-swift.struct/notificationuserinfokey/announcement?language=objc) can be an NSAttributedString. It just says "The announcement as a localized string" but doesn't specify that it can be an NSAttributedString. But maybe there's some convention where any string can also be an attributed string that I'm unaware of? For the text marker stuff, there are separate "AttributedString" methods, so I assumed this would also be true for something like NSAccessibilityAnnouncementKey if it supported attributed strings.

core-aam/index.html
cookiecrook: That seems okay to me. I must've missed it in review, and I don't recall any prior discussion. It'd be good to call out that plan in the ARIA WG meeting tomorrow in case others have insights.

index.html
cookiecrook: I'm not suggesting that the Web API need to require an attributed string on ATK. That might not even be the best way to implement it on Mac... That's just an implementation detail. 

I'm suggesting the web author needs a way to specify the language f the announcement, and the implementations can pass it along in whatever way is best for the implementation.

cookiecrook: @jcsteh wrote:

> For this spec PR, are there [other things not yet implemented]?

Apple's prior concern in the discussions list was that this is intended to be a better replacement for Live Regions, but the spec and test implementation does not address the annoyance factor that was the primary complaint of Live Regions. 

We've discussed not shipping the API in Safari until the user has more control over these announcements... That could be in-browser (like the mainstream Notification API or Location API settings that are per-domain), or perhaps in-AT (perhaps like Activities but more discoverable for AT announcements). Without this, ariaNotify today includes similar user risks as Live Regions. IMO, shipping it without this user control is premature, but I also don't want my user control complaint to be the roadblock that hinders iterative progress. I've called this out as one of a few concerns in the WebKit Standards Position.

jcsteh: Understood. My point (which I perhaps could have been clearer in making) is that although it's entirely possible with all current APIs to specify a language by way of the source node (and that should probably be explicitly specified for absolute clarity), it's not possible with the current APIs on Windows or Linux (and possibly even Mac depending on whether NSAccessibilityAnnouncementKey can accept an NSAttributedString instead of just an NSString) to specify a language for a notification which differs to that of the source node. So, the question is: are there a sufficient number of use cases where the language of a notification differs to that of the source node to warrant blocking this first push, potentially for years while all the APIs are updated? It seems to me that this is somewhat of an edge case, but I can certainly be convinced otherwise.

cookiecrook: Actually maybe @janewman implemented the permissions policy UI in Edge or Chrome already? Did you?

cookiecrook: > My point is that although it's entirely possible with all current APIs to specify a language by way of the source node (and that should probably be explicitly specified for absolute clarity), it's not possible with the current APIs on Windows or Linux to specify a language for a notification which differs to that of the source node. 

That's fair. I'll discuss with the team and get back to you on this point.

cookiecrook: I should probably also be more clear that I'm concerned about a JavaScript API design that wisely seeks to distance itself from DOM nodes, due to the problems with node-based Live Regions, but then ships with design patterns that keep it chained to the nodes, perhaps permanently, if there isn't enough momentum to get platform APIs updated later.

So we've got this clean caller:
```js
myElement.ariaNotify("Hello World!", { priority: "high" });
```
…but if I want to set the language explicitly, I've got a hang a content attribute off the node, and let the implementation implicitly wrap it up for me under the hood, rather than passing it along with the priority? It feels like an oversight.

Maybe a slightly less concerning compromise would be to pass an optional contextual node with the options, which would allow something like delegation, but would still unfortunately be tied to a content attribute.

```js
notificationDelegateElement.ariaNotify("Hello World!", { priority: "high", contextElement: optDifferentElement });
```

jcsteh: > I should probably also be more clear that I'm concerned about a JavaScript API design that wisely seeks to distance itself from DOM nodes, due to the problems with node-based Live Regions, but then ships with design patterns that keep it chained to the nodes, perhaps permanently, if there isn't enough momentum to get platform APIs updated later.

Yeah, that's fair. I think we've already made that compromise to some extent for v12 by dropping the identifier/type/whatever attribute from notifications, since now the source of the notification can only be identified using its source node. On the flip side, if we really want to divorce this from nodes altogether, wouldn't it make more sense for ariaNotify to only be available on the document object?

> Maybe a slightly less concerning compromise would be to pass an optional contextual node with the options

I'm not quite following how this would be better semantically. Either way, you have to hang a content attribute off the node, which I do acknowledge is not ideal for some use cases, even if they're less common.

> ```js
> notificationDelegateElement.ariaNotify("Hello World!", { priority: "high", contextElement: optDifferentElement });
> ```

To clarify, which node would be the source node for the accessibility notification events here and which node would have the content attribute?

index.html
```diff
@@ -16795,6 +16795,54 @@ <h2>Interface Mixin <dfn>ARIAMixin</dfn></h2>
...
+        </p>
+        <ol>
+          <li>
+            <p>If the document is not [=allowed to use=] the feature identified by <a>"aria-notify"</a> abort these steps.</p>
+          </li>
+          <li>
+            <p>Let <var>node</var> be <code>this</code>.</p>
+          </li>
+          <li>
+            <p>Let <var>priority</var> be <var>options</var>["<code>priority</code>"].</p>
```
keithamus: It might be useful here to link `["<code>priority</code>"]` to the `priority` IDL, which is currently unreferenced in the document. Though I don't think this is blocking.

keithamus (approved):

keithamus:
There is an official position of Support on Mozilla's side. https://github.com/mozilla/standards-positions/issues/1049#issuecomment-3222429917.

index.html
cookiecrook: > To clarify, which node would be the source node for the accessibility notification events here and which node would have the content attribute?

In the light of day, I'm wondering if what I'm attempting is even possible in an IDL dictionary. It's not a formal proposal, so I should have first floated the brainstorm somewhere other than the PR. (Sorry @janewman)

I was attempting to make the contextElement passed in `AriaNotificationOptions` be entirely optional, so that, only if necessary, node-specific checks could be placed on an existing or contrived element other than the posting element. Let me rewrite to hopefully make the intent more clear. 

#### HTML
```html
<!-- notifications sent on the body element -->
<body lang="de">
  …
</body>
```

#### JavaScript
```js
// optional: create an element on the fly as a hook for a lang value that differs from the posting body element
let languageCourier = document.createElement("div");
languageCourier.lang = "en"; // unattached <div lang="en"></div>

// document.body remains the rendered, posting element, but passes the notification lang through this contrived languageCourier element.
document.body.ariaNotify("Hello World!", { priority: "high", contextElement: languageCourier });
```

It seems like the element reference may be lost though, so the point may be moot. Likewise if any API needs a callback to check the language from the posting element, this also may not be possible.

core-aam/index.html
cookiecrook: Just FYI for completeness that today's ARIA meeting was cancelled. Leaving this thread open for @janewman's diff.

index.html
janewman: > Actually maybe @janewman implemented the permissions policy UI in Edge or Chrome already? Did you?

Chrome and Edge allow authors to opt out child frames using permissions policy, but don't have UI for users to do the same. Implementation matches the spec here and the explainer: [iframes and use in subresources](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/Accessibility/AriaNotify/explainer.md#iframes-and-use-in-subresources) From my understanding, a user agent surfacing this permission policy to users be also aligned with this spec and sounds like something we should consider as well.
___
Regarding lang, I can understand the desire for authors to be able to fire a notification on the document with a specified language and absolutely agree that pronunciation is important. But, as James Teh already mentioned, existing APIs on other platforms don't already allow for such functionality within their notification APIs, while the element supplying the language is already an established pattern.
While I feel that this behavior is sufficient for V1, I also feel like there is room for improvement here and would like to discussion to nail down what we think the right approach is for language specification moving forwards, and work with ATs and/or platform APIs to fill the existing gaps.

core-aam/index.html
janewman: Catching up with all of the conversations 😅
Thanks @cookiecrook for the clarification on announcementNotification's behavior here, that matches my observations.

Sounds like the question about hidden nodes has already been addressed but let me know if you feel like there is anything that should be called out or discussed that hasn't been already.

I've now updated this section to prefer calling this API with the node that ariaNotify was called on. Continuing to use `node` as this could be an element or a document.

janewman: > Yes, developers can post the announcementNotification on any AXElement... There currently isn't any user-side difference in VoiceOver between announcements coming form different nodes in the same window, but there could be in the future. Probably best to use DOM `element` here rather than DOM `node`? I'm not sure what the behavior should be if a notification were posted on an unrendered comment node, for example.

`node` is used here as ariaNotify could be called on either the document or an element.

index.html
janewman: good callout, done.

index.html
janewman: The latest version of this PR now includes a section on how language determination should be made.

jcsteh: > It seems like the [courier] element reference may be lost though, so the point may be moot.

Yeah, unfortunately, I don't think this solves the problem for any existing API. Aside from the posting node, existing APIs don't have any way to pass another node. In addition, existing APIS cannot access a node in any way if it is excluded from the accessibility tree (which it would be in this example because it isn't attached to the document).

cookiecrook: @janewman wrote:
> While I feel that [element-based lang determination] is sufficient for V1, I also feel like there is room for improvement here and would like to discussion to nail down what we think the right approach is for language specification moving forwards, and work with ATs and/or platform APIs to fill the existing gaps.

Apple DRIs have discussed more and agreed to drop the request to add a lang IDL parameter for v1 with the understanding this should be pursued for v2. The i18n concern remains in the standards position but we do not want to delay the progress made here. 

@janewman it's okay to mark this thread as resolved. Thanks.

core-aam/index.html
cookiecrook: Right. What should the behavior be if called on a non-element node within the document? For example:
- If called on a comment node, it's unrendered, so the algo would exit. No problem.
- If called on a rendered text node, is it expected that the announcement would make it through to AT on all platforms?

cookiecrook (approved):
A few outstanding questions, but no blocking changes needed. Thanks again for the follow-through.

core-aam/index.html
janewman: Seems like I may need to make this more explicit (thanks for bearing with me), I believe that ariaNotify should **only** be able to be called on Elements and on the Document, not any text node.
This is how it is implemented in chromium, so calling it on some rendered text node would be an error at the JS-level.

cookiecrook: Ah yes... I see that now...

```diff
+          Element includes ARIANotifyMixin;
+          Document includes ARIANotifyMixin;
```

Thanks for bearing with me.

cookiecrook: Okay to mark as resolved. Thanks.

core-aam/index.html
```diff
@@ -11199,6 +11199,132 @@ <h3>Special Events for Menus</h3>
...
+    <section id="mapping_algorithms">
+      <h2>Algorithms</h2>
+      <p>
+        Some <abbr title="Application Programming Interfaces">APIs</abbr> provide
+        methods which require specific algorithms to be followed. The following
+        sections provide the algorithm mapping tables for these methods.
+      </p>
+      <h3>ARIANotifyMixin Algorithm Mapping Tables</h3>
+      <p>
+        The <code>ARIANotifyMixin</code> provides methods for announcing content
```
smockle: Non-blocking: Is it accurate to say that this mixin provides “methods” (plural)?

```suggestion
        The <code>ARIANotifyMixin</code> provides a method for announcing content
```

core-aam/index.html
```diff
@@ -11199,6 +11199,132 @@ <h3>Special Events for Menus</h3>
...
+            <th>Preconditions</th>
+            <td>
+              <p>If <var>node</var> is <a class="specref" href="#tree_exclusion">excluded from the accessibility tree</a>, then abort these steps.</p>
+            </td>
+          </tr>
+          <tr>
+            <th>Language</th>
+            <td>
+              <p>
+                User agents and assistive technologies MUST determine the language of the announcement by taking the first valid BCP&nbsp;47 language tag from the following sources:
```
smockle, line 12481: Non-blocking: Would it be useful to link to a definition of “BCP 47 language tag”?

I looked for something we could xref with respec, but https://respec.org/xref/?term=BCP+47 returns no results.

index.html
```diff
@@ -16797,6 +16797,54 @@ <h2>Interface Mixin <dfn>ARIAMixin</dfn></h2>
...
+        </pre>
+
+        <p>
+          The User Agent processing steps for the
+          <code><dfn method for="ARIANotifyMixin" >ariaNotify</dfn>(<var>announcement</var>, <var>options</var>)</code>
+          method are:
+        </p>
+        <ol>
+          <li>
+            <p>If the document is not [=allowed to use=] the feature identified by <a>"aria-notify"</a> abort these steps.</p>
```
smockle, line 16834: If disabled via policy, will (attempting to) call `ariaNotify` throw an error or fail silently? Or will `ariaNotify` be `undefined`, i.e. uncallable in the first place?

janewman: Fail silently, let me know if you think this section needs additional clarification.

janewman:
> Where can folks review “the v2 roadmap”?
> 
Other than the [open issues](https://github.com/w3c/aria/issues?q=is%3Aissue%20state%3Aopen%20%22%5BAriaNotify%5D%22), the "future considerations" section of the [edge explainer ](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/Accessibility/AriaNotify/explainer.md#future-considerations) goes over the kinds of things we think aria notify should support, though as mentioned, there isn't consensus on the details yet. Much of the v2+ functionality will require additional API or screen reader support.

index.html
smockle: Thank you for clarifying! “Fail silently” is what I had assumed based on the text, so an edit isn’t necessary.

spectranaut:
Discussed during PR triage today: https://www.w3.org/2025/10/09-aria-minutes.html#31d8

Waiting for second implementation or implementation commitment to merge.

janewman:
Adding this to the agenda to discuss progress of other implementations, and timeline of when this should be merged.

index.html
```diff
@@ -16818,6 +16818,54 @@ <h2>Interface Mixin <dfn>ARIAMixin</dfn></h2>
...
+          </li>
+          <li>
+            <p>Run the <a data-cite="core-aam-1.2#arianotify">aria notify steps</a> given <var>node</var>, <var>announcement</var>, and <var>priority</var>.</p>
+          </li>
+        </ol>
+
+        <section id="arianotify-permissions-policy">
+          <h3>Permissions Policy Integration</h3>
+          <p>This specification defines a [=policy-controlled feature=] identified by the string <code><dfn class="permission">"aria-notify"</dfn></code>.
+          It has a [=policy-controlled feature/default allowlist=] of <code class=permissionspolicy>"*"</code>.
```
jcsteh, line 16853: I'm working on implementing this in Gecko and [I was asked about why this is `*` and not `self`](https://phabricator.services.mozilla.com/D278955#inline-1519850):

> Usually new APIs should be using 'self'.

I replied:

> I'm not sure of the background behind this decision. This is about messages that are reported to users, so I guess a site would normally not want to interfere with that for iframes it chooses to load, since that effectively breaks some of the functionality of the iframe. The only time you'd want to disable it is if an iframe you're loading behaves very badly and spams users with annoying messages.

Replies:

> But why would a site then load such iframe at all? Would they even notice that a bad site is spamming users.

IMO, this is a fair argument and I don't have a good response for it. @janewman, could you give me some background on why the decision was made to default this to `*` rather than `self`?

jcsteh:
Another question that came up as I'm implementing this: should we restrict this so it only fires when the document has focus/is in the foreground? I think we should. I don't think background tabs should be able to fire announcements at the user. There may be reasons this happens implicitly - e.g. a browser might not fire events for background tabs, the tab might be a background OS widget and so screen readers ignore it because of that, etc. - but I think it's worth calling this out explicitly so there's no room for interpretation.

spectranaut:
@janewman Now that WebKit landed, this is ready for merge! but there are outstanding questions from @jcsteh that would be nice to resolve first :)

index.html
janewman: While I can contrive a situation where the badly behaving iframe is still needed for xyz reason, this isn't a strong opinion at all. 
I think this came from wanting to be as permissive as possible, tracked down to this thread: https://github.com/whatwg/html/issues/11004#issuecomment-2641306814

janewman:
> Another question that came up as I'm implementing this: should we restrict this so it only fires when the document has focus/is in the foreground? I think we should. I don't think background tabs should be able to fire announcements at the user. There may be reasons this happens implicitly - e.g. a browser might not fire events for background tabs, the tab might be a background OS widget and so screen readers ignore it because of that, etc. - but I think it's worth calling this out explicitly so there's no room for interpretation.

Agree this would be worth being explicit about. I just did a sanity check and verified that this is the current chromium behavior, I'll update this PR with language that restricts this shortly.

janewman:
Here is the issue tracking the issue about what user agents should do with notifications from background tabs in a follow up, or what note to add to help authors understand filtering, in a follow up.
#2721

janewman:
@spectranaut, @jcsteh, I believe I've addressed any remaining questions/concerns, as discussed today we will do a follow up PR to document how user agents and ATs can choose to filter notifications, #2721

Let me know if there is anything left to do here, and if not, I'd love to see this merged 🍾

jcsteh:
My security peers still have some concerns about the permissions policy being `"*"` and not `"self"`. It's a really tricky issue and I think there are valid arguments on both sides. I think the argument for `"self"` is that we should assume iframes are untrustworthy by default and make exceptions otherwise, and that's how most new permission policies work. On the other hand, the main argument for `"*"` is that iframe content might otherwise be inaccessible because authors mightn't realise they broke something, which is also valid. The "dodgy ad" scenario tends to make me lean more towards being less permissive, but I could equally see an argument the other way.

It's also been argued that live regions can already be abused by iframes, but I don't think this should be a significant factor: we should be trying to make new APIs better, not saddle them with the same problems the old APIs had. So I'd suggest that we dismiss that particular argument unless there are strong reasons it should matter.

I'm trying to work out the best way to get this resolved. Personally, I could go either way, but it feels like we have a bit of an impasse here.

If there's some appetite to discuss this further, perhaps we can merge this and agree that it needs to be discussed further in a timely fashion. That said, the decision we make here does have an impact on how authors use this, so I think it is substantive from a web compat perspective.

janewman:
> On the other hand, the main argument for `"*"` is that iframe content might otherwise be inaccessible because authors mightn't realise they broke something, 

This is my main concern. The author of the top level frame may not even be aware that ariaNotify exists, is being used, or is important for the accessibility of the hosted content. I think of the permission policy being more as a failsafe - hopefully it will never need to be used, but at least it is there if we need it.

That said, if there is consensus in the other direction, this isn't something I'll hold out on.
@spectranaut, @cookiecrook, @smockle, I'd be really curious to get your opinions here.

smockle:
I re-read https://github.com/whatwg/html/issues/11004 and https://github.com/w3ctag/design-reviews/issues/1075, and I agree there are valid arguments on both sides.

Two things stand out to me:

1. If we decide on a default now, then decide to change it later, it would be less disruptive to go from `self` to `*` than from `*` to `self`. That gives me a slight preference for `self`.

2. I share @janewman’s concern that _“the author of the top level frame may not even be aware that ariaNotify exists [or] is being used…”_ but I want to make sure we don’t prioritize authors over users[^1]. Ensuring users get notifications is part of this, but not all of it: Would either `*` or `self` lead to better filtering ([w3c/aria#2499](https://github.com/w3c/aria/issues/2499))? If one is better for filtering, then I have a strong preference for that one. If they’re entirely orthogonal, then disregard this point.

[^1]: https://www.w3.org/TR/design-principles/#priority-of-constituencies

janewman:
> 2. I share @janewman’s concern that _“the author of the top level frame may not even be aware that ariaNotify exists [or] is being used…”_ but I want to make sure we don’t prioritize authors over users[1](#user-content-fn-1-bbdc4ba48c748f2f2888a33b5a67d8fc). 

My concern isn't just about authors, but users of that site. Here's an example:
Site "A" hosts content from site "B"
In this example, site "B" learns about ariaNotify, and refactor's its implementation to use it in place of live regions or other techniques. Site "A" makes no change, does not make any adjustments, as they aren't aware that anything has changed.
From the perspective of a user of site "A", their experience has degraded after the update. 

This could be argued that this is just a problem of site A's authors, but this scenario is why I have preference for opt-out rather than opt-in. (Notifications allowed unless author chooses to block a child frame).

> If one is better for filtering, then I have a strong preference for that one. If they’re entirely orthogonal, then disregard this point.

In either case, authors have the option to block notifications from child frames via permission policy, this is just a question of what the default should be: should notifications be blocked by default, with authors able to add a policy to allow, or should they be allowed by default, with authors able to add a policy to block.

jcsteh:
Sounds like we have consensus to go with "*" for the policy, so I think that addresses the main substantive outstanding question on this initial version of the feature. While I accept the consensus decision, for the record, there are two points I think worth highlighting:

> I think allowing by default and then changing to denying by default would be an option

As per https://github.com/w3c/aria/pull/2577#issuecomment-3819158842, this is *possible*, but it creates a web compat problem, since something that was previously working in some cases would now stop working. In contrast, going the other way (more restrictive to less restrictive) is far less of a concern.

> What we are resolving on today is not to give somebody something new that hey couldn't do with live-regions
> ...
> Yes, this is not going to work worse than live regions

I don't think "this isn't worse than live regions" is necessarily a good reason to do something with ariaNotify. The whole point of ariaNotify is to improve on live regions, not to inherit its problems.

janewman:
@spectranaut at this point all issues have been addressed, should we merge? (I don't have the permission to do so myself)

PR #2583, opened 2025-07-24
Removed aria-controls from the states and properties table for the combobox role to align with the combobox role definition.

### LINKED ISSUE #2582: Remove typo of required aria-controls for combobox in states and properties table
While removing `aria-controls` for the scrollbar, I realized that although we had already removed it as a requirement for `combobox` (https://w3c.github.io/aria/#combobox), we forgot to remove it from the table of states and properties (https://w3c.github.io/aria/#document-handling_author-errors_states-properties).

giacomo-petri:
Closes #2582

Removed aria-controls from the states and properties table for the combobox role to align with the combobox role definition.

### DIFF
--- a/index.html
+++ b/index.html
@@ -16626,7 +16626,6 @@ button.ariaPressed; // null</pre
             </tr>
             <tr>
               <td><rref>combobox</rref></td>
-              <td><sref>aria-controls</sref></td>
               <td>no mapping</td>
             </tr>
             <tr>


jnurthen (approved):
LGTM

index.html
```diff
@@ -16624,11 +16624,6 @@ <h3>States and Properties</h3>
               <td><sref>aria-checked</sref></td>
               <td><code>false</code></td>
             </tr>
-            <tr>
-              <td><rref>combobox</rref></td>
-              <td><sref>aria-controls</sref></td>
```
smhigley, line 16629: `aria-controls` should be conditionally required on combobox, right? That is, my understanding is if it's expanded it's required, and not if collapsed. Should this be kept, since while it's not absolutely required, it may still need a fallback defined for the expanded state?

pkra: @smhigley I've extracted this to https://github.com/w3c/aria/issues/2604 to get an answer. But given the current spec, I think this change is the correct.

pkra (approved):

PR #2598, opened 2025-08-14
[html-aam] Set heading level based on HTML computed heading level

keithamus:
Refs https://github.com/whatwg/html/pull/11086.

This wires up HTML-AAM to the `headingoffset` feature of the html spec which defines an algorithm for computing a heading level.

* [ ] "author MUST" tests:
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=295092
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1974383
   * Blink: https://issues.chromium.org/issues/333628468
* [ ] ACT review?
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:


rahimabdi:
Should the WebKit bug link to: https://bugs.webkit.org/show_bug.cgi?id=295092?

html-aam/index.html
```diff
@@ -2657,8 +2657,7 @@ <h4 id="el-h1-h6">`h1`, `h2`, `h3`, `h4`, `h5`, and `h6`</h4>
             <tr>
               <th>[[wai-aria-1.2]]</th>
               <td>
-                <a class="core-mapping" href="#role-map-heading">`heading`</a> role, with the <a class="core-mapping" href="#ariaLevel">`aria-level`</a> property set to the number in the element's tag
-                name.
+                <a class="core-mapping" href="#role-map-heading">`heading`</a> role, with the <a class="core-mapping" href="#ariaLevel">`aria-level`</a> property set to the <a data-cite="html/sections.html#get-an-element's-computed-heading-offset">the computed heading level</a>.
```
scottaohara, line 2659: initial reaction to this is "awesome! this is moving forward!" 

But, in regards to how these new features should be specified in this spec, it might be best to combine the proposed updated text with the original text.  e.g., indicate that the default level of a heading element would correlate to the number used in the tag name.  But, the computed heading level of the element can be adjusted if also used with the new headingoffset and headingreset attributes.

Then those attributes would need to be added to the attributes table (so then their mention in the heading element mapping table can then in-doc link to those attributes) indicating how they can adjust the heading level (aria-level) mapping for in-scope heading elements.

Probably also worth adding a note stating that specifying an `aria-level` attribute will take priority over whatever the implicit aria-level would be exposed as.

pkra:
The upstream PR has landed. Should this move out of draft, @keithamus ?

keithamus:
Closing this in favour of https://github.com/w3c/aria/pull/2720 which is both updated but also pushed to the right remote, so that we get deploy previews.

PR #2599, opened 2025-08-14
Add transparent generic definition and update accessibility parent and child defs

### LINKED ISSUE #2423: A generic element (with aria-live) intervening with the parent/descendant relationship.
I was checking one of the Failed examples in the ACT Rules about "ARIA required context role" ([Failed example 3](https://w3.org/WAI/content-assets/wcag-act-rules/testcases/ff89c9/52508dc0ac389108301d7cbd7f931be45a45741f.html)). I'm unsure about the expected violation of the acc parent/descendant relationship in the following case:

```
<div role="list">
	<div aria-live="polite">
		<div role="listitem">List item 1</div>
		<div role="listitem">List item 2</div>
	</div>
</div>
```

The [accessibility parent](https://www.w3.org/TR/wai-aria-1.3/#dfn-accessibility-parent) si defined as:

> The accessibility parent of a DOM element is the parent of the corresponding [accessible object](https://www.w3.org/TR/wai-aria-1.3/#dfn-accessible-object) in the [accessibility tree](https://www.w3.org/TR/wai-aria-1.3/#dfn-accessibility-tree). In terms of the DOM, the accessibility parent is one of the following:
> 
> - The DOM parent of the [element](https://dom.spec.whatwg.org/#concept-element).
> - The DOM ancestor of the [element](https://dom.spec.whatwg.org/#concept-element) with only elements of role [generic](https://www.w3.org/TR/wai-aria-1.3/#generic) or [none](https://www.w3.org/TR/wai-aria-1.3/#none) intervening.
> - A DOM element with [aria-owns](https://www.w3.org/TR/wai-aria-1.3/#aria-owns) set to the DOM ID of the DOM element in question.
> - A DOM element with [aria-owns](https://www.w3.org/TR/wai-aria-1.3/#aria-owns) set to the DOM ID of an ancestor of the DOM element in question, with only elements of role [generic](https://www.w3.org/TR/wai-aria-1.3/#generic) or [none](https://www.w3.org/TR/wai-aria-1.3/#none) intervening.

So, it looks like the code above is valid.

On the other hand, the [generic role](https://www.w3.org/TR/wai-aria-1.3/#generic) states:

> However, unlike elements with role presentation, user agents expose generic elements in [accessibility APIs](https://www.w3.org/TR/wai-aria-1.3/#dfn-accessibility-api) when permitted accessibility attributes have been specified. User agents MAY otherwise ignore generic elements if such permitted attributes have not been specified.

so, the generic element with aria-live="polite" is exposed. I don’t believe this impacts the validity of the code above, but I’d appreciate confirmation.

Last, but not least, in [Presentational Roles Conflict Resolution](https://www.w3.org/TR/wai-aria-1.3/#conflict_resolution_presentation_none):

> If an element has global WAI-ARIA states or properties, user agents MUST ignore the [none](https://www.w3.org/TR/wai-aria-1.3/#none)/[presentation](https://www.w3.org/TR/wai-aria-1.3/#presentation) role and instead expose the element's implicit role.

we have a generic element with a global property, causing the UA to ignore the none role and expose the implicit role (which is still none). This shouldn’t impact the validity of the code.

Thoughts?

Thanks

smhigley:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2599--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2599--wai-aria.netlify.app%2Findex.html)

Closes #2423

Updates the accessibility parent/child relationships to reference "transparent" elements rather than simply generics or `role="none/presentation"`. Transparent is defined based on the language in the previous section (including elements in the a11y tree), and I added an explicit reference to live regions in the examples of semantics that cause an accessibility event.

* [X] Browser implementations (link to issue or commit):
   * WebKit: n/a
   * Gecko: n/a
   * Blink: n/a
* [ ] ACT review?
* [x] Does this need AT implementations? no
* [x] MDN Issue/PR: N/A

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2599.html" title="Last updated on Aug 20, 2025, 5:37 PM UTC (d76f68a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2599/34306ed...d76f68a.html" title="Last updated on Aug 20, 2025, 5:37 PM UTC (d76f68a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -172,7 +172,7 @@
 
         preProcess: [linkCrossReferences, getParticipants],
         postProcess: [ariaAttributeReferences],
-        xref: ["dom", "accname-1.2", "core-aam-1.2", "infra", "HTML"],
+        xref: ["dom", "accname-1.2", "core-aam-1.2", "infra", "HTML", "permissions-policy"],
         definitionMap: [],
       };
       function respecChangelogFilter(commit) {
@@ -2809,7 +2809,8 @@
             </p>
             <p>
               Elements with the role <code>combobox</code> have an implicit <pref>aria-haspopup</pref> value of <code>listbox</code>. If the <code>combobox</code> popup element has a role other than
-              <rref>listbox</rref>, authors MUST specify an <pref>aria-haspopup</pref> value of <rref>tree</rref>, <rref>grid</rref>, or <rref>dialog</rref> that corresponds to the role of its popup.
+              <rref>listbox</rref>, authors MUST specify an <pref>aria-haspopup</pref> value of <rref>tree</rref>, <rref>grid</rref>, <rref>menu</rref>, or <rref>dialog</rref> that corresponds to the
+              role of its popup.
             </p>
             <p>
               If the user interface includes an additional icon that allows the visibility of the popup to be controlled via pointer and touch events, authors SHOULD ensure that element has role
@@ -10208,7 +10209,13 @@
               for details on implementing a tab set design pattern.
             </p>
             <p>Authors MUST ensure [=elements=] with <a>role</a> <rref>tab</rref> are <a>accessibility children</a> of an element with the role <rref>tablist</rref>.</p>
+            <!-- keep following paragraphs in sync with counterpart in #tabpanel -->
             <p>Authors MUST ensure that if a <code>tab</code> is active, a corresponding <code>tabpanel</code> that represents the active <code>tab</code> is rendered.</p>
+            <p>
+              Authors SHOULD associate a <code>tabpanel</code> <a>element</a> with its <rref>tab</rref>, by using the <pref>aria-controls</pref> attribute on the tab to reference the tab panel, and/or
+              by using the <pref>aria-labelledby</pref> attribute on the tab panel to reference the tab.
+            </p>
+            <!-- /end sync with counterpart in #tabpanel -->
             <p>Authors SHOULD ensure the <rref>tabpanel</rref> associated with the currently active tab is <a>perceivable</a> to the user.</p>
             <!-- keep following para synced with its counterpart in #tablist -->
             <p>
@@ -10542,10 +10549,13 @@
           <rdef>tabpanel</rdef>
           <div class="role-description">
             <p>A container for the resources associated with a <rref>tab</rref>, where each <rref>tab</rref> is contained in a <rref>tablist</rref>.</p>
+            <!-- keep following paragraphs in sync with counterpart in #tab -->
+            <p>Authors MUST ensure that if a <code>tab</code> is active, a corresponding <code>tabpanel</code> that represents the active <code>tab</code> is rendered.</p>
             <p>
               Authors SHOULD associate a <code>tabpanel</code> <a>element</a> with its <rref>tab</rref>, by using the <pref>aria-controls</pref> attribute on the tab to reference the tab panel, and/or
               by using the <pref>aria-labelledby</pref> attribute on the tab panel to reference the tab.
             </p>
+            <!-- /end sync with counterpart in #tab -->
             <!-- keep following para synced with its counterpart in #tablist -->
             <p>
               <rref>tablist</rref> elements are typically placed near, and usually preceding, a series of <rref>tabpanel</rref> elements. See the
@@ -11878,7 +11888,7 @@
               <code>undefined</code> is not expandable.
             </dd>
             <dt id="valuetype_idref">ID reference</dt>
-            <dd>Reference to the ID of another <a>element</a> in the same document</dd>
+            <dd>Reference to the ID of another <a>element</a> in the same document.</dd>
             <dt id="valuetype_idref_list">ID reference list</dt>
             <dd>A list of one or more ID references.</dd>
             <dt id="valuetype_integer">integer</dt>
@@ -16522,7 +16532,10 @@ button.ariaPressed; // null</pre
         </p>
         <section id="mapping_additional_relations_error_processing">
           <h3>ID Reference Error Processing</h3>
-          <p>[=user agents=] SHOULD ignore ID references that do not match the ID of another <a>element</a> in the same document.</p>
+          <p>
+            The author MAY set either an empty value or a value that does not match the ID of another <a>element</a> in the same document. This flexibility acknowledges the dynamic nature of modern
+            websites, where the DOM can be populated accordingly when necessary. [=user agents=] SHOULD ignore ID references that do not match the ID of another <a>element</a> in the same document.
+          </p>
           <p>
             It is the author's responsibility to ensure that IDs are unique. If more than one element has the same ID, the user agent SHOULD use the first element found with the given ID. The behavior
             will be the same as <code>getElementById</code>.
@@ -16750,8 +16763,8 @@ button.ariaPressed; // null</pre
       </section>
     </section>
     <section id="idl-interface" class="normative">
-      <h2>IDL Interface</h2>
-      <p>Conforming user agents MUST implement the following IDL interface.</p>
+      <h2>IDL Interfaces</h2>
+      <p>Conforming user agents MUST implement the following IDL interfaces.</p>
       <section id="ARIAMixin" class="normative" data-dfn-for="ARIAMixin" data-link-for="ARIAMixin">
         <h2>Interface Mixin <dfn>ARIAMixin</dfn></h2>
         <pre class="idl" data-cite="webidl">
@@ -16812,6 +16825,56 @@ button.ariaPressed; // null</pre
       Element includes ARIAMixin;
 		</pre
         >
+      </section>
+      <section id="ARIANotifyMixin" class="normative" data-dfn-for="ARIANotifyMixin" data-link-for="ARIANotifyMixin">
+        <h2>Interface Mixin <dfn>ARIANotifyMixin</dfn></h2>
+        <pre class="idl" data-cite="webidl">
+          enum AriaNotifyPriority { "normal", "high" };
+
+          dictionary AriaNotificationOptions {
+            AriaNotifyPriority priority = "normal";
+          };
+          interface mixin ARIANotifyMixin {
+            undefined ariaNotify(DOMString announcement, optional AriaNotificationOptions options = {});
+          };
+          Element includes ARIANotifyMixin;
+          Document includes ARIANotifyMixin;
+        </pre>
+
+        <p>
+          The User Agent processing steps for the
+          <code><dfn data-df-type="method" data-dfn-for="ARIANotifyMixin">ariaNotify</dfn>(<var>announcement</var>, <var>options</var>)</code>
+          method are:
+        </p>
+        <ol>
+          <li>
+            <p>If the document is not [=allowed to use=] the feature identified by <a>"aria-notify"</a> abort these steps.</p>
+          </li>
+          <li>
+            <p>Let <var>node</var> be <code>this</code>.</p>
+          </li>
+          <li>
+            <p>
+              Let <var>priority</var> be <var>options</var>["<code><a href="#dom-arianotificationoptions-priority">priority</a></code
+              >"].
+            </p>
+          </li>
+          <li>
+            <p>If <var>node</var> is <a href="#tree_exclusion">excluded from the accessibility tree</a>, then abort these steps.</p>
+          </li>
+          <li>
+            <p>Run the <a data-cite="core-aam-1.2#arianotify">aria notify steps</a> given <var>node</var>, <var>announcement</var>, and <var>priority</var>.</p>
+          </li>
+        </ol>
+
+        <section id="arianotify-permissions-policy">
+          <h3>Permissions Policy Integration</h3>
+          <p>
+            This specification defines a [=policy-controlled feature=] identified by the string <code><dfn class="permission">"aria-notify"</dfn></code
+            >. It has a [=policy-controlled feature/default allowlist=] of <code class="permissionspolicy">"*"</code>.
+          </p>
+          <p class="note">When disabled in a document, any notifications applied using <code>ariaNotify</code> within that document will not be sent.</p>
+        </section>
       </section>
 
       <section id="accessibilityroleandproperties-correspondence" class="normative" data-dfn-for="ARIAMixin" data-link-for="ARIAMixin">
--- a/core-aam/index.html
+++ b/core-aam/index.html
@@ -12444,6 +12444,133 @@
       </section>
     </section>
 
+    <section id="mapping_algorithms">
+      <h2>Algorithms</h2>
+      <p>
+        Some <abbr title="Application Programming Interfaces">APIs</abbr> provide methods which require specific algorithms to be followed. The following sections provide the algorithm mapping tables
+        for these methods.
+      </p>
+      <h3>ARIANotifyMixin Algorithm Mapping Tables</h3>
+      <p>
+        The <code>ARIANotifyMixin</code> provides a method for announcing content to assistive technologies. The following algorithm mappings specify how these announcements will be implemented across
+        different accessibility APIs to ensure consistent behavior for users of assistive technologies.
+      </p>
+      <h4 id="arianotifymixin-map-arianotify">ariaNotify</h4>
+      <p>To <code>aria notify</code> given <var>node</var>, <var>announcement</var>, and <var>priority</var>:</p>
+      <table aria-labelledby="arianotifymixin-map-arianotify" class="data">
+        <tbody>
+          <tr>
+            <th>ARIA Specification</th>
+            <td>
+              <a class="method-reference" href="#arianotify"><code>aria notify</code></a>
+            </td>
+          </tr>
+          <tr>
+            <th>Preconditions</th>
+            <td>
+              <p>If <var>node</var> is <a class="specref" href="#tree_exclusion">excluded from the accessibility tree</a>, then abort these steps.</p>
+            </td>
+          </tr>
+          <tr>
+            <th>Language</th>
+            <td>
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -16156,6 +16156,7 @@ <h2>Including Elements in the Accessibility Tree</h2>
             <ul>
               <li>Elements that are currently focused, even if the element or one of its ancestor elements has its <sref>aria-hidden</sref> attribute set to <code>true</code>.</li>
               <li>Elements that are a valid target of an <pref>aria-activedescendant</pref> attribute.</li>
+              <li>Elements that have <a>live region</a> roles or attributes.</li>
```
smhigley, line 16186: This addition is editorial, since the section already specifies elements that can fire an accessibility API event. I thought it should be added explicitly since it was the original example that kicked this off (and is probably the other most common attribute authors use that would trigger an a11y event)

index.html
```diff
@@ -16174,17 +16175,21 @@ <h2>Including Elements in the Accessibility Tree</h2>
       <section id="tree_relationships">
         <h2>Relationships in the Accessibility Tree</h2>
         <p>The following terms are used to describe relationships between <abbr title="Document Object Model">DOM</abbr> elements.</p>
+        <p>A <dfn data-export="" data-lt="transparent|transparent element|transparent elements">transparent element</dfn> is an element that meets one of the following criteria:</p>
+        <ul>
+          <li>It has a role of <rref>presentation</rref> or <rref>none</rref>, and does not meet one of the exceptions documented in the <a href="#conflict_resolution_presentation_none">Presentational Roles Conflict Resolution</a> section.</li>
+          <li>It is a <a>generic</a> element that does not have any attributes such as <pref>aria-live</pref> that would require it to be included in the accessibility tree, as covered in <a href="#tree_inclusion">Including Elements in the Accessibility Tree</a>.</li>
```
spectranaut:
```suggestion
          <li>It is a <rref>generic</rref> element that does not have any attributes such as <pref>aria-live</pref> that would require it to be included in the accessibility tree, as covered in <a href="#tree_inclusion">Including Elements in the Accessibility Tree</a>.</li>
```

Fixed link ^

spectranaut (approved):
This looks great! And I think it is clear.

About whether or not this change is normative -- I'm not sure. In some respects, I think this "transparent" definition is what the "accessible children" work was originally trying to achieve, a case was just missed.

But this change in definition does change "Authors MUST statements", like: [Authors MUST only add child element with allowed roles.](https://deploy-preview-2599--wai-aria.netlify.app/#mustContain). Because it changes `author MUST` statements we should have ACT review.

But it doesn't effect browser implementations right, because we don't specify how to exposes roles outside of their required context?

index.html
```diff
@@ -16174,17 +16175,21 @@ <h2>Including Elements in the Accessibility Tree</h2>
...
+        </ul>
         <p>
           The <dfn data-export="" data-lt="accessibility child|owned child|child element|child|children|child elements">accessibility children</dfn> of a DOM element are all of the children of that
           element's corresponding <a>accessible object</a> in the <a>accessibility tree</a>. In terms of the DOM, that includes the following (with exclusions listed blow):
```
adampage:
```suggestion
          element's corresponding <a>accessible object</a> in the <a>accessibility tree</a>. In terms of the DOM, that includes the following (with exclusions listed below):
```

Just caught this unrelated typo in the neighborhood.

adampage (approved):
This is so great, @smhigley. 🤩  Very helpful to have a clear term — dad joke intended. 🤓

index.html
smhigley: oops, thanks for catching that! 😅

smhigley:
@spectranaut thanks for the review! For this:

> But it doesn't effect browser implementations right, because we don't specify how to exposes roles outside of their required context?

That's also my understanding, it doesn't affect browser implementations b/c we don't specify how to expose the intervening generics, and also this should bring the spec in line with the current behavior (i.e. aria-live on an intervening generic didn't work).

jnurthen (approved):

index.html
```diff
@@ -1179,6 +1179,7 @@ <h3>Allowed Accessibility Child Roles</h3>
             example, an element with the role <rref>list</rref> allows <a>accessibility children</a> with the role <rref>listitem</rref>, but does not allow <a>accessibility children</a> with the role
             <rref>option</rref>.
           </p>
+          <p>To determine whether an element is the <a>child</a> of an element, <a>user agents</a> MUST ignore any intervening elements with the role <rref>generic</rref> or <rref>none</rref> which are <a>transparent</a>.</p>
```
```suggestion
          <p>To determine whether an element is the <a>accessibility child</a> of an element, <a>user agents</a> MUST ignore any intervening elements with the role <rref>generic</rref> or <rref>none</rref> which are <a>transparent</a>.</p>
```

index.html
```diff
@@ -1247,8 +1248,9 @@ <h3>Required Accessibility Parent Role</h3>
             current role is an <a>accessibility child</a> of an element with a role from this list. For example, an element with role <code>listitem</code> is only meaningful when it is a child of an
             element with role <code>list</code>.
           </p>
-          <p><a>User agents</a> SHOULD ignore the current role if the role of the <a>accessibility parent</a> is not in this list.</p>
-          <p class="note">An <a>accessibility parent</a> with the appropriate <a href="#implicit_semantics">implicit WAI-ARIA semantic</a> fulfills this requirement.</p>
+          <p>To determine whether an element has a parent with the required role, <a>user agents</a> MUST ignore any elements with the role <rref>generic</rref> or <rref>none</rref> which are <a>transparent</a>.</p>
```
```suggestion
          <p>To determine whether an element has an accessibility parent with the required role, <a>user agents</a> MUST ignore any elements with the role <rref>generic</rref> or <rref>none</rref> which are <a>transparent</a>.</p>
```

index.html
```diff
@@ -1179,6 +1179,7 @@ <h3>Allowed Accessibility Child Roles</h3>
             example, an element with the role <rref>list</rref> allows <a>accessibility children</a> with the role <rref>listitem</rref>, but does not allow <a>accessibility children</a> with the role
             <rref>option</rref>.
           </p>
+          <p>To determine whether an element is the <a>accessibility child</a> of an element, <a>user agents</a> MUST ignore any intervening elements with the role <rref>generic</rref> or <rref>none</rref> which are <a>transparent</a>.</p>
```
pkra, line 1182: This seems redundant (now).

index.html
```diff
@@ -1247,8 +1248,9 @@ <h3>Required Accessibility Parent Role</h3>
             current role is an <a>accessibility child</a> of an element with a role from this list. For example, an element with role <code>listitem</code> is only meaningful when it is a child of an
             element with role <code>list</code>.
           </p>
-          <p><a>User agents</a> SHOULD ignore the current role if the role of the <a>accessibility parent</a> is not in this list.</p>
-          <p class="note">An <a>accessibility parent</a> with the appropriate <a href="#implicit_semantics">implicit WAI-ARIA semantic</a> fulfills this requirement.</p>
+          <p>To determine whether an element has an accessibility parent with the required role, <a>user agents</a> MUST ignore any elements with the role <rref>generic</rref> or <rref>none</rref> which are <a>transparent</a>.</p>
```
pkra: The older version of this paragraph was just recently removed in #2713.

index.html
```diff
@@ -1247,8 +1248,9 @@ <h3>Required Accessibility Parent Role</h3>
...
+          <p>Also, <a>user agents</a> SHOULD ignore the role if it occurs outside the context of a required accessibility parent role.</p>
```
pkra: This would revert the changes from #2713.

index.html
```diff
@@ -1247,8 +1248,9 @@ <h3>Required Accessibility Parent Role</h3>
...
+          <p class="note">An element with the appropriate <a href="#implicit_semantics">implicit WAI-ARIA semantic</a> fulfills this requirement.</p>
```
pkra: Ditto.

index.html
```diff
@@ -16192,17 +16195,21 @@ <h2>Including Elements in the Accessibility Tree</h2>
       <section id="tree_relationships">
         <h2>Relationships in the Accessibility Tree</h2>
         <p>The following terms are used to describe relationships between <abbr title="Document Object Model">DOM</abbr> elements.</p>
+        <p>A <dfn data-export="" data-lt="transparent|transparent element|transparent elements">transparent element</dfn> is an element that meets one of the following criteria:</p>
```
pkra, line 16205: On revisiting this a year later, I'm not sure about this additional definition anymore.

index.html
```diff
@@ -16192,17 +16195,21 @@ <h2>Including Elements in the Accessibility Tree</h2>
...
+        <ul>
+          <li>It has a role of <rref>presentation</rref> or <rref>none</rref>, and does not meet one of the exceptions documented in the <a href="#conflict_resolution_presentation_none">Presentational Roles Conflict Resolution</a> section.</li>
+          <li>It is a <rref>generic</rref> element that does not have any attributes such as <pref>aria-live</pref> that would require it to be included in the accessibility tree, as covered in <a href="#tree_inclusion">Including Elements in the Accessibility Tree</a>.</li>
```
pkra, line 16208: I think it would be simpler to add this rule for generics to the list at the beginning of Section 7.1 instead (adding the same sentence we have in the item for role=none that descendants are not excluded).

index.html
```diff
@@ -16192,17 +16195,21 @@ <h2>Including Elements in the Accessibility Tree</h2>
...
         <p>
-          The <dfn data-export="" data-lt="accessibility child|owned child">accessibility children</dfn> of a DOM element are all of the children of that element's corresponding
-          <a>accessible object</a> in the <a>accessibility tree</a>. In terms of the DOM, that includes the following (with exclusions listed blow):
+          The <dfn data-export="" data-lt="accessibility child|owned child|child element|child|children|child elements">accessibility children</dfn> of a DOM element are all of the children of that
+          element's corresponding <a>accessible object</a> in the <a>accessibility tree</a>. In terms of the DOM, that includes the following (with exclusions listed below):
         </p>
         <ul>
           <li>The <abbr title="Document Object Model">DOM</abbr> children of the <a>element</a>.</li>
-          <li>All DOM descendants of the <a>element</a> with only elements of role <rref>generic</rref> or <rref>none</rref> intervening.</li>
+          <li>All DOM descendants of the <a>element</a> with only <a>transparent elements</a> intervening.</li>
```
pkra, line 16216: Then here we would only have to say "DOM descendants that are not excluded as per Section 7.1".

pkra: I suppose we don't even have to say that much since this list is preceded by "with exclusions listed below" and those exclusions follow immediately after.

index.html
```diff
@@ -16192,17 +16195,21 @@ <h2>Including Elements in the Accessibility Tree</h2>
...
           <li>All DOM elements specified via an <pref>aria-owns</pref> relationship to the element.</li>
           <li>
-            All DOM descendants of an element with role <rref>generic</rref> or <rref>none</rref> specified via <pref>aria-owns</pref> with only elements of role <rref>generic</rref> or
-            <rref>none</rref> intervening.
+            All DOM descendants of an element with role <rref>generic</rref> or <rref>none</rref> specified via <pref>aria-owns</pref> with only <a>transparent elements</a> intervening.
```
pkra, line 16219: This paragraph was already redundant, I think.

index.html
```diff
@@ -16246,10 +16253,10 @@ <h2>Relationships in the Accessibility Tree</h2>
         </p>
         <ul>
```
pkra, line 16261: I think this entire list should be removed as redundant. As shown here, it only adds maintenance burden.

pkra:
@smhigley sorry for changing my mind a year later :pensive:  I'll volunteer if we decide on a different direction.

index.html
smhigley: The issue is that this category of "transparent" elements are distinct from both generics (which may or may not be in the a11y tree at the browser's discretion), and elements which must be excluded entirely, which are outlined in 7.1.

This addition is specifically for elements that do not interfere with parent/chlid relations, but do not rise to the level of MUST NOT or SHOULD NOT a11y tree exclusion. When we were discussing the original a11y parent/child work, I remember some browser folks being clear that this should not add a new requirement for whether the intervening generics are part of the a11y tree; the only requirement is that they do not interfere with e.g. list/listitem or listbox/option hierarchy.

index.html
smhigley: hm, I know transparent is also in the definition of acc child, but this is the only place with browser MUST wording related to it

pkra:
I'm worried about even messier inline comments, so let me try a global comment.

> I remember some browser folks being clear that this should not add a new requirement for whether the intervening generics are part of the a11y tree; the only requirement is that they do not interfere with e.g. list/listitem or listbox/option hierarchy.

Right. I struggle to see how this change achieves that outcome. 

Re-reading today, the changes even seem a little contradictory.

The addition to Section 5.2.6 (Allowed Accessibility Child Roles):

> To determine whether an element is the accessibility child of an element, user agents MUST ignore any intervening elements with the role generic or none which are transparent .

The addition to Section 7.3 (Relationships in the Accessibility Tree)

>  All DOM descendants of an element with role generic or none specified via aria-owns with only transparent elements intervening. 

The latter change seems to add all "transparent" children as accessibility children while the former change forces UAs to ignore them. 

I think there is a conflict between the spec's needs and UA needs. 

I feel a more concise definition of "accessibility children" alongside an acknowledgment that UAs may keep excluded nodes as "ignored nodes" (or some such) might be a better path. (But I also admit that, at this point, I'm just getting more and more confused by Section 7. :pensive: )

cookiecrook (changes requested):
The Interop Accessibility Investigation has had Tree Walker investigation on the list for two or three years now, because once this is testable in WebDriver, we know there are going to be major differences between implementations. 

A key details I brought up in 2023 is that each implementation starts with different trees:
- Gecko starts with the DOM tree
- WebKit starts with the Layout tree
- Chromium was originally WebKit, and so started with Layout, but has moved to a hybrid model somewhere between the others

As such there are known and unknown places where important and potentially not-easily-reconcilable differences are left undefined and left up to the implementation. 
- relationships between elements and nearby CSS generated content (pseudo elements)
- style or other changes that result in new in-betweener layout elements such as extra scroll views due to CSS `overflow` which can change on the fly based on layout changes like newly loaded content or the browser window changing size.
- other style properties and emerging featuers, like responsive layouts via media queries, and newer display values that may result in new layout objects or hierarchical ordering differences

As such, the investigations have purposefully avoided testing or defining these relationships. The 2026 investigation is obviously farthest along.

> ### Tree Walker investigation using GetAccessibilityProperties
> - Author investigation tests for which parts of the Tree or Subtrees might be standardized in 2026 or later?
>   - Purposefully write failing tests (tentative) that expose the areas where the internal accessibility tree differ between engines
>   - File AOM or ARIA issues if it appears are some tree relationships that might be standardized as part of a 2027 Investigation

Citation: https://github.com/web-platform-tests/interop-accessibility/issues/202#issuecomment-3862051095 is a carryover every year, but this may be the year we can finally ship tests and investigate further.

Sorry to throw a wrench in after you've spent some time on this already, but I'm hopeful that the list above makes it clear why I can't support specifying accessibility parent/child definitions prior to these testability prerequisites.

cookiecrook:
As an example: 
> To determine whether an element is the [ accessibility child ](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2599--wai-aria.netlify.app%2Findex.html#dfn-accessibility-child) of an element, [ user agents ](https://infra.spec.whatwg.org/#user-agent) MUST ignore any intervening elements with the role [ generic ](https://services.w3.org/aria/#generic) or [ none ](https://services.w3.org/aria/#none) which are [ transparent ](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2599--wai-aria.netlify.app%2Findex.html#dfn-transparent).

```html
<main>
  <!-- start of in-betweener scroll view -->
  <div role="group" style="overflow:scroll;">
    contents
  </div>
  <!-- end of in-betweener scroll view -->
</main
```

Is the div a accessibility child of main? It may be in Windows. It would be unexpected for it to be in Mac, since the generated scroll view is a layout object (for example, in native AppKit view) that is the child of main and parent of the div. But even on the Mac platform, Safari and Firefox may disagree due to the implementation details. And that extra AX tree element scroll view may not conform to the ARIA definitions of "transparent" or "generic", even if we are able to agree on a definition for DOM elements.

That's just one example why I think we need to tread cautiously in anything that seems close to defining a formalized accessibility tree in the browser. We have to ease up against the borders of platform-specific API differences. We'll get there, but not IMO until more reliable test methods become available.

index.html
smhigley: We chased this down in the git blame, so adding it here for reference -- this isn't a new UA MUST, it's re-adding the UA requirement that was added in the original PR, but removed in this one: https://github.com/w3c/aria/pull/2697

smhigley:
@cookiecrook I think you're interpreting this the same way Peter did, which is to read it as a general definition of how to calculate a full tree. In reality, this currently only pertains to specific role hierarchical requirements defined in the spec, such as `listbox` > `option` or `tablist` > `tab`.

CSS pseudo elements would never come up because they (currently) cannot have a role. For things like `overflow: scroll` styles on divs, we should discuss it if they would currently disrupt things like `listbox` > `option` associations. Right now we need a way to define allowed authoring patterns for intervening generics for roles with parent/child limitations.

PR #2615, opened 2025-09-02
Editorial: Add back accessibility subtree definition

spectranaut:
In a rebase, I lost the part of the commit which moved the definition of subtree: https://github.com/w3c/aria/pull/2293 Then it got merged, now there is an error in the build about not finding the definition. I added it back in this PR.


spectranaut:
@jnurthen or @daniel-montalvo can you take a quick look?

PR #2617, opened 2025-09-02
Add interestfor attribute related mappings

mfreed7:
This adds the mapping for the `interestfor` attribute.

See [the explainer](https://open-ui.org/components/interest-invokers.explainer/) for the overall API description and use cases.

* [ ] For every spec that this PR edits, please add the appropriate `spec:<spec_name>` label. If you don't have privileges to do this, editors will do it for you.

* [ ] "author MUST" tests:
* [X] Browser implementations (link to issue or commit):
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=297786
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1984699
   * Blink: https://crbug.com/326681249
* [ ] ACT review?
* [X] Does this need AT implementations? Yes
* [ ] Related APG Issue/PR:
* [X] MDN Issue/PR: https://github.com/mdn/mdn/issues/721
* [X] Other: https://github.com/w3c/a11y-request/issues/118

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -11625,6 +11625,87 @@
             </tr>
           </tbody>
         </table>
+        <h4 id="att-interestfor-attribute">`interestfor`</h4>
+        <table class="data" aria-labelledby="att-interestfor-attribute">
+          <tbody>
+            <tr>
+              <th>HTML Specification</th>
+              <td>
+                <span class="el-context"> `interestfor` </span>
+              </td>
+            </tr>
+            <tr>
+              <th>Element(s)</th>
+              <td>
+                <a data-cite="html/form-elements.html#the-button-element">`button`</a>;
+                <a data-cite="html/text-level-semantics.html#the-a-element">`a`</a>;
+                <a data-cite="html/image-maps.html#the-area-element">`area`</a>
+              </td>
+            </tr>
+            <tr>
+              <th>[[WAI-ARIA-1.2]]</th>
+              <td>
+                <p>A popover is <dfn>"plain"</dfn> if it contains only elements with a computed role of generic, text and images.</p>
+                <p>A popover is <dfn>"rich"</dfn> if it is not "plain".</p>
+                <p>If the `interestfor`-associated element is invalid, or is not a not a popover: no `aria-expanded` mapping</p>
+                <p>
+                  If the `interestfor`-associated element is an accessibility ancestor of the interest invoker:
+                  <a class="core-mapping" href="#ariaExpandedUndefined">`aria-expanded=undefined`</a>
+                </p>
+                <p>If the `interestfor`-associated element is a <a>"plain"</a> popover:
+                  <ul>
+                    <li><p>Set <a class="core-mapping" href="#ariaHiddenTrue">`aria-hidden="true"`</a> on the target popover.</p></li>
+                    <li><p>If no other <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is available, use the popover's inner text for
+                      the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.</p></li>
+                    <li><p>If the popover is not used as the <a data-cite="accname-1.2/#dfn-accessible-name">name</a>, then use it to compute the
+                      <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a>, setting the
+                      <a data-cite="wai-aria-1.2/#aria-describedby">`aria-describedby`</a> relation to point from the interest invoker to the popover</p></li>
+                  </ul>
+                </p>
+                <p>If the `interestfor`-associated element is a <a>"rich"</a> popover:
+                  <ul>
+                    <li><p>If the popover is displayed: <a class="core-mapping" href="#ariaExpandedTrue">`aria-expanded=true`</a></p></li>
+                    <li><p>If the popover is hidden: <a class="core-mapping" href="#ariaExpandedFalse">`aria-expanded=false`</a></p></li>
+                    <li><p>Set a <a>minimum role</a> of <a class="core-mapping" href="#role-map-tooltip">`tooltip`</a> on the target popover</p></li>
+                    <li><p>Set `aria-details` pointing to the target popover</p></li>
+                  </ul>
+              </td>
+            </tr>
+            <tr>
+              <th>
+                <a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a>
+              </th>
+              <td>
+                <div class="general">Use WAI-ARIA mapping</div>
+                <div class="objattrs"><span class="type">Object attributes:</span> `details-roles:popover`</div>
+              </td>
+            </tr>
+            <tr>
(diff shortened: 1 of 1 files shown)


mfreed7:
Uber-caveat: I cargo-culted much of what I added here, e.g. the sections with "MSAA" and "UIA". Please tell me what I did wrong, and give me a hint at how to fix it, and I'll do my best.

html-aam/index.html
```diff
@@ -11625,6 +11625,87 @@ <h4 id="att-indeterminate">`indeterminate [IDL]`</h4>
...
+                <p>A popover is <dfn>"rich"</dfn> if it is not "plain".</p>
+                <p>If the `interestfor`-associated element is invalid, or is not a not a popover: no `aria-expanded` mapping</p>
+                <p>
+                  If the `interestfor`-associated element is an accessibility ancestor of the interest invoker:
+                  <a class="core-mapping" href="#ariaExpandedUndefined">`aria-expanded=undefined`</a>
+                </p>
+                <p>If the `interestfor`-associated element is a <a>"plain"</a> popover:
+                  <ul>
+                    <li><p>Set <a class="core-mapping" href="#ariaHiddenTrue">`aria-hidden="true"`</a> on the target popover.</p></li>
+                    <li><p>If no other <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is available, use the popover's inner text for
```
front-endian, line 11668: It is good that this would catch a common source of missing or repeated accnames on some elements, but there are plenty of real world situations where a `plain` popover has text that wouldn't make sense as the accname.

Also, the calculation of accnames already has many non-intuitive parts. I'm concerned that this would be another behavior that is opaque to authors, making it harder to reason about accnames.

- Has there been discussion about this choice of behaviors and the tradeoffs involved? If yes, is it documented somewhere so I can read through it?
- How does this interact with some of the other ways an accname can be implicitly added to an element? Does it make sense to be the bottom of the list?
- Would mapping it to the ARIA `description` unless it is explicitly marked for use as the accname be safer? (Maybe not since that that isn't as well exposed to users by ATs.)

front-endian, line 11668: Separate question: is there a matching PR to update the accname spec to reflect this change as well?

html-aam/index.html
```diff
@@ -11625,6 +11625,87 @@ <h4 id="att-indeterminate">`indeterminate [IDL]`</h4>
...
+                <span class="el-context"> `interestfor` </span>
+              </td>
+            </tr>
+            <tr>
+              <th>Element(s)</th>
+              <td>
+                <a data-cite="html/form-elements.html#the-button-element">`button`</a>;
+                <a data-cite="html/text-level-semantics.html#the-a-element">`a`</a>;
+                <a data-cite="html/image-maps.html#the-area-element">`area`</a>
+              </td>
```
front-endian, line 11653: What is the reasoning for making the list of applicable elements so restricted? What about things like elements with `input` roles?

front-endian (changes requested):

html-aam/index.html
scottaohara: this is called out in the intro to the explainer and there are links to related issues about this topic.
https://open-ui.org/components/interest-invokers.explainer/#introduction

tldr; these are the elements that had consensus for initial implementation, and further updates could come to extend to other elements.

html-aam/index.html
scottaohara: this may fall under step 2.5 in accName which is meant to broadly call out 'other' host language features which could provide accNames.

ideally, we shouldn't need to add this to that spec, as part of the benefit of having specific AAM specs is to identify these host language details, which allows accName to be more general. 

but leaving this comment unresolved in case this does warrant unique consideration.

html-aam/index.html
```diff
@@ -11622,6 +11622,87 @@ <h4 id="att-indeterminate">`indeterminate [IDL]`</h4>
...
+                    <li><p>If the popover is not used as the <a data-cite="accname-1.2/#dfn-accessible-name">name</a>, then use it to compute the
+                      <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a>, setting the
+                      <a data-cite="wai-aria-1.2/#aria-describedby">`aria-describedby`</a> relation to point from the interest invoker to the popover</p></li>
+                  </ul>
+                </p>
+                <p>If the `interestfor`-associated element is a <a>"rich"</a> popover:
+                  <ul>
+                    <li><p>If the popover is displayed: <a class="core-mapping" href="#ariaExpandedTrue">`aria-expanded=true`</a></p></li>
+                    <li><p>If the popover is hidden: <a class="core-mapping" href="#ariaExpandedFalse">`aria-expanded=false`</a></p></li>
+                    <li><p>Set a <a>minimum role</a> of <a class="core-mapping" href="#role-map-tooltip">`tooltip`</a> on the target popover</p></li>
```
rahimabdi: A link to [minimum role](https://w3c.github.io/html-aam/#minimum-role) may be helpful here.

html-aam/index.html
mfreed7: > It is good that this would catch a common source of missing or repeated accnames on some elements, but there are plenty of real world situations where a `plain` popover has text that wouldn't make sense as the accname.

Interesting! The typical use case is a "tooltip" which seems to always make sense as an accname, right? If not, please help me understand the cases where it doesn't make sense. Maybe we can tweak something.

> * Has there been discussion about this choice of behaviors and the tradeoffs involved? If yes, is it documented somewhere so I can read through it?

It was mostly in discussions with a few a11y experts (which I am not). Perhaps @scottaohara and/or @aleventhal have more of the context somewhere publicly-visible?

> * How does this interact with some of the other ways an accname can be implicitly added to an element? Does it make sense to be the bottom of the list?

It's definitely written so that the `interestfor` accname comes "last". I do worry (as in your initial comment) that it might not make sense in some cases, but it seems safe to use this accname if there isn't another one from any other mechanism.

> * Would mapping it to the ARIA `description` unless it is explicitly marked for use as the accname be safer? (Maybe not since that that isn't as well exposed to users by ATs.)

I'm not the right person to answer that question - I'd happily take advice.

html-aam/index.html
mfreed7: Hmm, I was using `<a>minimum role</a>` for that link, perhaps it doesn't work? Looking through the source, there is one definition of "minimum role", five references with no link (`minimum role`), one with just an `<a>` (`<a>minimum role</a>`), and four with `<a href="termref">minimum role</a>`. I've just changed to that last incantation - hopefully that fixes the link?

html-aam/index.html
front-endian: > Interesting! The typical use case is a "tooltip" which seems to always make sense as an accname, right? If not, please help me understand the cases where it doesn't make sense. Maybe we can tweak something.

There are applications where a tooltip provides additional info. It could be help text for an input, additional data about an item in a graph or a node in a network chart, the status of an item, etc. In a some data-heavy web apps, that use case of tooltips is just as common as showing an element's name on hover.

However, after spending some time reviewing real world examples of those, the only time this would cause a problem is with elements that lack a proper accname anyway, so they are going to be missing a name no matter what.

mfreed7: Great! Thanks for the extra info.

mfreed7:
> Approving after further review/feedback. Thank you for providing additional information @mfreed7!

Great, thanks! I appreciate the review.

There's one open question about the `<a>minimum role</a>` link - I'm not sure if anything needs to be fixed there?

PR #2633, opened 2025-09-17
Editorial: Fix id value references and add h2s where appropriate

daniel-montalvo:
Fix some markup issues for the PDF-AAM spec to build. Others need to be fixed via scripts

### DIFF
--- a/pdf-aam/index.html
+++ b/pdf-aam/index.html
@@ -201,7 +201,7 @@
       </div>
     </section>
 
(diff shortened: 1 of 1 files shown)


spectranaut (approved):

daniel-montalvo:
Thanks @spectranaut for approving this. Actually, i think we should just merge this and get next PRs more focused.

PR #2215, opened 2024-05-23
[Monorepo] [html-aam PR 529] Addition: Name from heading

pkra:
Moved from https://github.com/w3c/html-aam/pull/529/

PR #2650, opened 2025-10-09
Add namefrom: heading.

### LINKED ISSUE #138: ARIA Spec could be more flexible when elements with "nameFrom:author" are left unlabeled by the author
I received the following bug report from a member of my team.

> Summary:
> Web AX, VoiceOver: When browsing list of articles using VO Articles rotor on [site redacted], VO announces each item as "Article" without speaking headline
> 
> Steps To Reproduce:
> 1. Turn on VoiceOver.
> 2. Open Safari and browse to [site redacted].
> 3. Press VO-U to bring up rotors.
> 4. Press VO-Right arrow repeatedly to select “Articles” rotor.
> 5. Explore rotor items by pressing VO-Down arrow.
> 
> Results:
> When using VO’s Articles rotor, VO does show all the articles presented on [site redacted], but only speaks each item as “Article” without providing any additional info, such as the article’s headline. For better clarity, VO should speak some additional information here.

Upon closer inspection, the articles are not labeled but they all have a heading in close proximity to the start of the article element. For the sake of the user, we could probably work around this web page error heuristically in either WebKit or VoiceOver, but the ARIA spec appears to explicitly discourage this.

The ARIA spec could add some language explicitly allowing user agents to heuristically determine the label for unlabeled elements whose label has not been set by the author. In this case, the most logical choice would be to find the first heading in the article.

Perhaps: "User Agents MAY use heuristics to determine the label of an unlabeled element, if the role  characteristics include `nameFrom: author`. For example, browsers may set the label of an unlabeled dialog to be the first heading in that dialog."

### LINKED ISSUE #182: Add `nameFrom: heading` steps to computation after spec addition #w3c/aria/pull/1860 is reviewed.
Add `nameFrom: heading` steps to computation after spec addition #w3c/aria/pull/1860 is reviewed.

cookiecrook:
🚀 **Netlify Preview**:
- [accname preview](https://deploy-preview-2650--wai-aria.netlify.app/accname/index.html) &mdash; [accname diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faccname%2F&doc2=https%3A%2F%2Fdeploy-preview-2650--wai-aria.netlify.app%2Faccname%2Findex.html)
- [html-aam preview](https://deploy-preview-2650--wai-aria.netlify.app/html-aam/index.html) &mdash; [html-aam diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fhtml-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-2650--wai-aria.netlify.app%2Fhtml-aam%2Findex.html)
- [ARIA preview](https://deploy-preview-2650--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2650--wai-aria.netlify.app%2Findex.html)

Closes https://github.com/w3c/accname/issues/138
Closes #1018
Closes #1860
Closes #2209
Closes https://github.com/w3c/accname/issues/182
Closes https://github.com/w3c/accname/pull/229

Related to https://github.com/w3c/aria/pull/2215

Add namefrom: heading again; Will year 6 be the year? 

Roles that are affected: now `namefrom: author, heading`:
- alertdialog
- article
- dialog

History and Readiness:
- Jon Gunderson started this change in 2019 with #1018.
- I attempted it again in 2023 with #1860.
- Latter PR was already reviewed/approved but had irreconcilable merge conflicts, in part due to the Monorepo shift and editorial addition of Prettier that happened in the meantime.
- A few other issues and pull requests (linked above) were also filed in the meantime.
- This PR (2650) supersedes 1860 which superseded 1018.
- [WebKit is shipping already](https://github.com/WebKit/WebKit/commit/6eb2b6af4dd8b9c0bd8810a7f6c1dd7e9dc24583) (@rahimabdi)
- [WPT Tests are live](https://wpt.fyi/results/accname/name/comp_name_from_heading.tentative.html?label=experimental&label=master&aligned).

I was worried this might never make it into the spec, but we now have:
- approval from ARIA WG (including prior PR approval from @accdc, @scottaohara, and @aleventhal)
- a no-conflict, single-monorepo pull request
- live WPT tests
- a shipping implementation (WebKit)
- support from a second implementation (Chromium)
- support (IIRC @keithamus?) from a third implementation (Gecko)

Please re-review and merge this ASAP before this layered PR stagnates again... Thanks!

# Implementation tracking

* [x] WPT tests: https://wpt.fyi/results/accname/name/comp_name_from_heading.tentative.html
   * [x] [WebKit 257186](https://webkit.org/b/257186)
   * [ ] [Gecko 1834463](https://bugzilla.mozilla.org/show_bug.cgi?id=1834463)
   * [ ] [Blink 1447981](https://bugs.chromium.org/p/chromium/issues/detail?id=1447981)
* [x] Does this need AT implementations? No. Engine-specific change.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2650.html" title="Last updated on Dec 18, 2025, 6:01 PM UTC (72c1c15)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2650/55751aa...72c1c15.html" title="Last updated on Dec 18, 2025, 6:01 PM UTC (72c1c15)">Diff</a>


cookiecrook:
Tagging **F2FCandidate** because if this third-attempt PR (or maybe it’s the fifth?) hasn't merged by TPAC, we should lock the doors until it is.

scottaohara:
@cookiecrook i had done this work as well for html aam https://github.com/w3c/aria/pull/2215

cookiecrook:
Taking a note from the WG call today to double-check the history for `form` and `region`. @rahimabdi asked for additional tests, and @mcking65 wondered if `region` should be left out due to a more recent spec change related to that role.

spectranaut:
Discussed during triage today: https://www.w3.org/2025/10/09-aria-minutes.html#31d8

I believe @cookiecrook has some follow up tasks and we are primarily waiting for a second implementation or stronger implementation commitment.

cookiecrook:
Confirmed `nameFrom:heading` was removed from `region` (and `complementary` FWIW) in the prior PR in [this commit](https://github.com/w3c/aria/pull/1860/commits/b4f7124ba3bd55e55e33652095c3a10cddcb014f) and determined we would address it elsewhere. Will pull that role here, too, but not yet sure wondering why the related prose remained. Probably just an oversight, so thanks Matt...

cookiecrook:
Both `region` and `form` had [seemingly related prose about using heading as the label](https://github.com/w3c/aria/pull/1860/files#r1188975118). Removing it was technically an editorial change, but in the new PR, I accidentally made it substantive with the addition of `namefrom: heading` on those two. I will correct that error, and also NOT remove the duplicative text for the roles that should not be related to this PR, and that should resolve both Rahim and Matt's comments.

index.html
```diff
@@ -3721,7 +3743,12 @@ <h2>Definition of Roles</h2>
               </tr>
               <tr>
                 <th class="role-namefrom-head" scope="row">Name From:</th>
-                <td class="role-namefrom">author</td>
```
cookiecrook, line 3720: Reviewer context: this is on the `dialog` role

index.html
```diff
@@ -1886,7 +1903,12 @@ <h2>Definition of Roles</h2>
...
```
cookiecrook, line 1885: Reviewer context: this is on the `article` role

index.html
```diff
@@ -1670,7 +1682,12 @@ <h2>Definition of Roles</h2>
...
```
cookiecrook, line 1669: Reviewer context: this is on the `alertdialog` role

cookiecrook (commented):
Adding comments to make review easier.

accname/index.html
```diff
@@ -646,6 +646,10 @@ <h4>Computation steps</h4>
                     </p>
                   </div>
                 </li>
+                <li id="comp_name_from_heading">
+                  <em>Name From Heading:</em> Otherwise, if the <code>current node</code> has a role that supports <a class="specref" href="#namefromheading">nameFrom: heading</a>, 
+                  return the text alternative of the first descendant <a>element</a> node matching the role of <code>heading</code> in an <em>depth-first search.</em>
```
rahimabdi:
```suggestion
                  return the text alternative of the first descendant <a>element</a> node matching the role of <code>heading</code> in a <em>depth-first search.</em>
```

index.html
```diff
@@ -1265,15 +1265,27 @@ <h3>Name From</h3>
               having the lowest precedence for specifying a text alternative.
             </li>
             <li>
-              contents: name comes from the text value of the <a>element</a> node. Although this might be allowed in addition to "author" in some <a>roles</a>, this is used in content only if higher
-              priority "author" features are not provided. Priority is defined by the <cite><a href="" class="accname">Accessible Name and Description Computation</a></cite> [[ACCNAME-1.2]].
+              heading: name comes from the text alternative of the first descendant <a>element</a> node (depth first search) matching the role of <code>heading</code> as defined in the AccName
```
rahimabdi:
```suggestion
              heading: name comes from the text alternative of the first descendant <a>element</a> node (depth-first search) matching the role of <code>heading</code> as defined in the AccName 
```

index.html
```diff
@@ -3631,7 +3653,7 @@ <h2>Definition of Roles</h2>
               focus to dialogs. See the <a href="https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/">WAI-ARIA Authoring Practices: Dialog (modal) pattern</a> for additional details on implementing
               modal dialog design patterns.
             </p>
-            <p>Authors SHOULD provide an accessible name for a dialog, which can be done with the <pref>aria-label</pref> or <pref>aria-labelledby</pref> attribute.</p>
+            <p>Authors SHOULD provide an accessible name for a dialog, using either <a href="#namecalculation">namefrom</a>: author or <a href="#namecalculation">namefrom</a>: heading.</p>
```
rahimabdi, line 3652: Given a dialog that contains an `<h2>` for example, would it be preferable for authors to use `aria-labelledby` or rely on nameFrom: heading? Should nameFrom: author be preferred over nameFrom: heading?

rahimabdi (changes requested):

rahimabdi:
@cookiecrook One thought that came to mind during implementation was around the descendant element with `role="heading"`, and that it could potentially be a link or image (which is common practice) rather than just a text node. I think nameFrom: heading still applies here but should this be specified in spec (e.g., "first descendant element node matching the role of heading (including headings that contain images or links.")?

cookiecrook:
Lucas, in [this comment](https://issues.chromium.org/issues/40914051#comment16), indicated potential Q4 2025 support for Chromium/Edge.

index.html
cookiecrook: Yes, the AccName computation prioritizes `namefrom:author` first... `namefrom:heading` is only used as a fallback.

cookiecrook: The direct author reference is probably slightly more performant, but I don't think a change is needed here unless you have one to suggest.

cookiecrook:
@rahimabdi wrote:

> @cookiecrook One thought that came to mind during implementation was around the descendant element with `role="heading"`, and that it could potentially be a link or image (which is common practice) rather than just a text node. I think nameFrom: heading still applies here but should this be specified in spec (e.g., "first descendant element node matching the role of heading (including headings that contain images or links.")?

What's special about a heading containing a an image or link that would not also apply to some other type of content? The label derived for the element that supports `namefrom:heading` will be the heading's label. But then that element's label will be derived from running its contents through the computation... `link` gets its name from author or contents... `image` from author via alt or aria-label. Wouldn't the same be true for any included [fill in the blank] element? If so, why the need for special mention?

scottaohara (approved):

index.html
rahimabdi: Just a personal clarification to make sure I'm using this new nameFrom technique properly 😄

rahimabdi (approved):
LGTM 🚀

jcsteh:
In general, this makes sense to me. However, I'm wondering if there has been any discussion anywhere regarding mutations, events and the performance thereof.

When the name of anything changes, the browser should ideally fire a name change event. The importance of this differs between operating systems and browser implementations, but it's still generally the right thing to do. For Firefox (and I assume Chromium), failure to fire an event could result in a stale cross-process cache, so it's particularly important there. However, clients may depend on the event as well.

There are two particular challenges here:

1. When the name of a heading changes, we need to potentially check to see if it's being used as the name of a NameFrom: heading element (`<dialog>`, etc.). Because the rules specify that we must search for the first descendant heading, this could get a bit expensive, especially if the heading is somewhat deep or far in the subtree. I guess we could cache the heading associated with a NameFrom: heading element, but that introduces its own complexity.
2. When a heading is moved in the tree, we need to do the same.

@rahimabdi, does WebKit fire name change events for these cases? Did you run into any issues implementing those?

I'd also be curious as to whether anyone has done any design thinking on this for Chromium and considered these issues.

Thanks!

html-aam/index.html
```diff
@@ -16816,7 +16816,43 @@ <h4>`iframe` Element Accessible Name Computation</h4>
           </p>
         </section>
         <section>
-          <h4>Section and Grouping Element Accessible Name Computation</h4>
+          <h4>`dialog` Element Accessible Name Computation</h4>
```
jcsteh, line 16839: Just an editorial question for my understanding, here and below:
Why do we need this information here? Shouldn't the mapping to the ARIA role mean that we rely on the name computation for that role, which specifies NameFrom: author, heading and thus covers the heading requirement here? Is there something that doesn't cover?

html-aam/index.html
```diff
@@ -16816,7 +16816,43 @@ <h4>`iframe` Element Accessible Name Computation</h4>
...
+        <section>
+          <h4>`article` Element Accessible Name Computation</h4>
+          <ol>
+            <li>
+              If the `article` element has an <a data-cite="wai-aria-1.2/#aria-label">`aria-label`</a> or an <a data-cite="wai-aria-1.2/#aria-labelledby">`aria-labelledby`</a> 
+              attribute the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is to be calculated using the algorithm defined in
+              <a class="accname">Accessible Name and Description: Computation and API Mappings</a>.
+            </li>
+            <li>
+              If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty, then: if the `dialog` element has a
```
jcsteh:
```suggestion
              If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty, then: if the `article` element has a
```

html-aam/index.html
```diff
@@ -16816,7 +16816,43 @@ <h4>`iframe` Element Accessible Name Computation</h4>
...
+              <a href="https://dom.spec.whatwg.org/#concept-tree-descendant">descendant</a> element with a role of `heading`, then use the
```
jcsteh, line 16866: This refers explicitly to a DOM descendant. Does this mean aria-owns content should explicitly be excluded from this check? For example, should we skip NameFrom: heading in this case?

`data:text/html,<article aria-owns="content"></article><div id="content"><h1>Some heading</h1>`

Does that also mean we shouldn't consider aria-owns content when doing name computation for a dialog? For example, in this case, do we take first or second as the label of the dialog?

`data:text/html,<article aria-owns="first second"><h1 id="second">second</h1></article><h1 id="first">first</h1>`

I think we should respect aria-owns for consistency. That means we should be using [accessibility descendants](https://w3c.github.io/aria/#dfn-accessibility-descendant).

html-aam/index.html
scottaohara: these can be considered redundant, sure.

the spec has always had these element-specific steps that, to not include these changes for article and dialog, would seem an oversight.

if we want to have a larger conversation about if we actually need to maintain this practice of making element-specific naming steps in this spec, we can do that.  but i'd rather not do it now, in this pr.

cookiecrook:
As a status update, this stalled due to a comment at W3C TPAC (Nov 2025) from @jcsteh pointing out that we haven't fully considered what happens in the case of DOM mutations.

jcsteh:
> As a status update, this stalled due to a comment at W3C TPAC (Nov 2025) from @jcsteh pointing out that we haven't fully considered what happens in the case of DOM mutations.

Just in case this isn't clear, I documented my concerns shortly prior to TPAC in https://github.com/w3c/aria/pull/2650#issuecomment-3429958401. My comment at TPAC was just flagging this comment.

minorninth:
> > As a status update, this stalled due to a comment at W3C TPAC (Nov 2025) from @jcsteh pointing out that we haven't fully considered what happens in the case of DOM mutations.
> 
> Just in case this isn't clear, I documented my concerns shortly prior to TPAC in [#2650 (comment)](https://github.com/w3c/aria/pull/2650#issuecomment-3429958401). My comment at TPAC was just flagging this comment.

I feel like we already have plenty of similar issues, where we need to fire a name change event because an element uses aria-labelledby and the target ID changes its effective name, sometimes in some roundabout way like a descendant changing.

Chromium and WebKit have both fixed bugs along those lines before. They both already keep track of reverse maps for all IDREF relations to make that sort of thing more efficient.

I'm not arguing that this sort of thing is easy; some of those fixes are tricky. Rather, I'm just arguing that it's already the status quo that browsers have to keep track of a lot of complex relationships in order to implement accessible names and to keep them updated as the DOM mutates.

My feeling is that these scenarios don't actually come up that often in the real world - or rather, that most accessibility bugs we spend time fixing aren't of this variety. However, when they do they're treated as bugs and fixed.

jcsteh:
> I feel like we already have plenty of similar issues, where we need to fire a name change event because an element uses aria-labelledby and the target ID changes its effective name, sometimes in some roundabout way like a descendant changing.

That's true. However, existing "name from subtree" cases are for roles which are expected to have relatively few descendants; e.g. button, link, heading, etc. In contrast, things like dialogs will very possibly have many, many descendants. That makes the cost of calculating (and maintaining) name from subtree potentially much more expensive.

> My feeling is that these scenarios don't actually come up that often in the real world - or rather, that most accessibility bugs we spend time fixing aren't of this variety. However, when they do they're treated as bugs and fixed.

Fair. On the other hand, I think we should be cautious about introducing things into the spec where the probability of hitting a performance cliff which is difficult to mitigate might be quite high.

cookiecrook:
@jcsteh wrote:
> That makes the cost of calculating (and maintaining) name from subtree potentially much more expensive.

The only scenarios I anticipate where the perf would be bad would be scenarios where web authors have made multiple mistakes... e.g. a dialog, with no label, no labelled-by, no nearby included heading, with excessively long contents... so the cost would only be on sites where that dialog was already unlabeled... 

maybe there's a lazy-load label compromise, where after some reasonable delay trying to find the fallback heading (500ms? 700ms?), NVDA or VO would just speak "dialog" (as if it's unlabeled) and then update that assumption for the user later when or if the label search comes back before the user moves on to another element.

minorninth:
> Fair. On the other hand, I think we should be cautious about introducing things into the spec where the probability of hitting a performance cliff which is difficult to mitigate might be quite high.

As you mentioned, the existing "name from subtree" cases are supposed to have few descendants, but in obscure cases there might be lots, so I know some browsers already enforce a limit that's not in the spec, for performance reasons.

In this particular case, would a limit on how far to search for a heading descendant mitigate it?

html-aam/index.html
```diff
@@ -16836,7 +16836,43 @@ <h4>`iframe` Element Accessible Name Computation</h4>
...
+              attribute the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is to be calculated using the algorithm defined in
+              <a class="accname">Accessible Name and Description: Computation and API Mappings</a>.
+            </li>
+            <li>
+              If the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is still empty, then: if the `article` element has a
+              <a href="https://dom.spec.whatwg.org/#concept-tree-descendant">descendant</a> element with a role of `heading`, then use the 
+              <a data-cite="accname-1.2/#mapping_additional_nd_te">text equivalent computation</a> of the element as defined in
+              <a data-cite="accname-1.2/#comp_name_from_heading">accname: Name from Heading</a>.
+            </li>
+            <li>Otherwise, use the `title` attribute.</li>
```
rahimabdi, line 16870: @scottaohara @cookiecrook Shouldn't `title` take precedence over nameFrom: heading? 

My understanding is that nameFrom: heading is a fallback where nameFrom: author is not specified. If an author has a robust heading structure inside an alertdialog/dialog/article, and a `title` is specified on these elements, I think `title` would be honored over nameFrom: heading.

rahimabdi (changes requested):

html-aam/index.html
cookiecrook: I don't have a strong preference for either, but was trying to follow the AccName general pattern of the tooltip feature being used as the final fallback content. I recall it was set up this way to avoid forcing authors to include workarounds if they wanted to use the visible tooltip for something else, but the easy workaround here is just to use aria-label or aria-labelledby, so I think it's fine either way.

rahimabdi (approved):

PR #2651, opened 2025-10-10
Editorial: Change failures to warnings for RFC2119 keywords

daniel-montalvo:
We currently have this check that simply fails if it finds RFC2119 keywords. I think that's too hard for this check, especially if we plan to move to a situation where we do want to actively have failing checks prevent merges.

I think the best for us here is for this to just throw a warning to make sure the keyword uses are intended.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -14,7 +14,7 @@ jobs:
           fetch-depth: 0
       
       - name: Set up Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v6
         with:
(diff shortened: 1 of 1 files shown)


pkra:
Looks ok to me but I don't know enough about github actions to say this will work as desired.

daniel-montalvo:
@pkra 

I have tried this at https://github.com/daniel-montalvo/aria/actions/runs/18534545894/job/52825998393?pr=9

It still throws the error from the github action we are using but also adds the adhoc warning

pkra (approved):

Issue #2501, opened 2025-03-28
continue aria.js work


pkra:
Continuation of #2408 

- [ ] review questions & TODO items noted in the code comments
- [ ] revisit #1823, #2073, #2234, #2445
- [ ] return roleInfo to make nodejs-generation of roleInfo.js easier
  - [ ] and automatically update roleInfo.js
- [ ] turn everything into ESMs (if respec allows it)
- [ ] review other JS files (e.g., aria-child.js)  to prepare their refactoring (with shared modules)

pkra:
Note to self: re-review https://github.com/w3c/aria-common/issues/104.

pkra:
A first experiment with SSG roleInfo.js updates https://github.com/pkra/aria/pull/2 @jnurthen @spectranaut @daniel-montalvo  Note that the action is set to update the PR itself but we talked about updating on push to main.

pkra:
https://github.com/w3c/aria/issues/2549 is a good example of how aria.js uses class names and text Content as hooks which seems risky and outdated.

pkra:
We discussed the rolinfo SSG in yesterday's editors call https://www.w3.org/2025/07/09-aria-editors-minutes.html#4908

pkra:
Sigh. #2625 didn't work because the main branch is protected (and we all forgot about that).

PR #2654, opened 2025-10-14
editorial: fix roleInfo.yml

pkra:
Enables roleInfo.yml to update roleInfo on the main branch (based on prettier.yaml).

Part of #2501


pkra:
Discussed during today's editors' call https://www.w3.org/2025/10/22-aria-editors-minutes.html#cf95

PR #2677, opened 2025-11-20
updating aria-live=off to match the aligned implementations.

### LINKED ISSUE #2144: `marquee` and `timer` listed as live region roles but have intrinsic `aria-live="off"`
In https://www.w3.org/TR/wai-aria-1.3/#live_region_roles `marquee` and `timer` are listed as live region roles. However, when then following through to their definitions, it states:

> Elements with the role `marquee` have an implicit `aria-live` value of `off`.

https://www.w3.org/TR/wai-aria-1.3/#marquee

> Elements with the role `timer` have an implicit `aria-live` value of `off`.

https://www.w3.org/TR/wai-aria-1.3/#timer

As `aria-live="off"` is the default for all elements, isn't this contradictory? I'd have expected that roles listed as being live regions would have an implicit value of either `polite` or `assertive`.

Unless there's some extra "magic" here that is not explained - "they're `off` but still behave differently in some unspecified way compared to other elements"? But then that aspect should be clarified explicitly... looking at https://www.w3.org/TR/wai-aria-1.3/#aria-live there's a lot of prose explaining `assertive` and `polite`, but for `off` you only get the short definition in the table 

> Indicates that updates to the region should not be presented to the user unless the user is currently focused on that region.

But again, if this is the *default*, I would not expect a difference in an element that either has no `aria-live` attribute at all, and one that has `aria-live="off"`.

### LINKED ISSUE #2653: Should the default value of aria-live be undefined instead of off?
A number of ARIA properties have a default/implicit value of "undefined" which means the property is not supported on that element. `aria-live` is not one of those properties, and its default value of "off" has this definition:

> Indicates that updates to the region should not be presented to the user unless the user is currently focused on that region.

That, to me, suggests that the element is a live region, it just shouldn't be automatically announced. 

Given that `aria-live` is a global property, and that most elements are not live regions, I think the default value of `aria-live` should be "undefined" rather than "off."

### LINKED ISSUE #2664: Synonym for `aria-live="off"` that is less confusing to authors

In a similar vein to precedents set by `role="presentation"`/`role="none"` and `role="img"`/`role="image"`, I'd love to see some kind of synonym for the `off` value of `aria-live`. This would be a nice one to do in parallel with https://github.com/w3c/aria/issues/2653

Many developers are still confused by the fact that `off` doesn't *actually* mean "this isn't a live region", but that it has this special "it IS actually a live region, but only announced when the user's focus/regard/cursor is on/in that element" meaning. This was one of the take-aways that came out of https://tetralogical.com/blog/2024/05/01/why-are-my-live-regions-not-working/ (when I actually had to go and edit MDN to clarify this very aspect, as MDN had documented it wrong as well).

I understand that trying to find a good synonym for this particular thing is hard. A naive `aria-live="focus"` is probably too loaded and incorrect a term, as it's not necessarily *just* about focus, and certainly not just focus on the element itself but also inside it. The exact best name for the synonym may need some wordsmithing/bikeshedding, but anything may be clearer than `off` and the confusion this caused/stil causes for developers. `aria-live="current"` ... or is that too close to `aria-current`? something more abstract like `aria-live="context"`?

## Will this require a change to [CORE-AAM](https://w3c.github.io/core-aam/)?

Yes

## Will this require a change to the [ARIA authoring guide](https://w3c.github.io/aria-practices/)?

Possibly yes (don't think there's actually any good standalone examples of live regions at the moment in the APG, which is also a gap that should really be filled ... as quirks such as "in most cases, make sure the element exists in the DOM *before* you populate it" would be good to have some authoritative examples for.

### LINKED ISSUE #2674: Update marquee, timer, aria-live=off to match the currently reality supported by all major browsers and screen readers
## IDL questions re: aria-live=off.

https://w3c.github.io/aria/

The spec defines aria-live as a global attribute in https://w3c.github.io/aria/#aria-live
> Used in Roles:	All elements of the base markup

it also defines `off` as the default value:
> `off` (default):	Indicates that updates to the region should not be presented to the user unless the user is currently focused on that region.

The definition "should not be presented to the user _unless_ the user is currently focused on that region" accurately reflects the intention based on my recollection at the time, as well as based on other prose in the spec about order prioritization, etc. 

> Priority levels (none[sic, errata, should be 'off'] , polite, assertive) act as an ordering mechanism for updates and serve as a recommendation to user agents or assistive technologies. 

However, the combination of this default value, its definition, and the global attribute means that all elements are live regions at all times... And there is no real way to stop live region announcements from the element the AT user is focused on. "off" doesn't really mean "off"... 

I don't believe the editors intended to make aria-live affect elements globally this way, at least not in the context of `aria-live=off`, or didn't consider the consequences of making a value other then undefined the default value for a global attribute. (I was one of the spec editors around that time and admit to not considering this consequence in the context of an IDL implementation.)

I expect this might be addressed as errata... 

~~The easiest and most web-compatible option may be to change `aria-live` to no longer be global. Instead, defining it and other live region attributes as allowed only on elements with an explicitly defined live region role, and ignored on other elements.~~

~~Another solution may be to add another `undefined` value as the default, but I think this path has more risk to webcompat.~~

[Update: I struck the previous two options because the [third option](https://github.com/w3c/aria/issues/2674#issuecomment-3555965835) below seems best now]

cookiecrook:
Closes #2144 
Closes #2653 
Closes #2664 
Closes #2674 

Updating `aria-live="off"` to match the aligned implementations:… Effectively that the following are not live regions, because `aria-live="off"` negates all live region functionality.
- `role="marquee"`
- `role="timer"`
- `<div aria-live="off">` (`aria-live` is global, and the default is `off` on everything, so this is identical to `<div>`)

See the issues for more detail.
- #2144 
- #2653 
- #2664 
- #2674

Implementations:
- [ ] [Chrome issue](https://issues.chromium.org/issues/477317119)
- [ ] [Firefox issue](https://bugzilla.mozilla.org/show_bug.cgi?id=2011535)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2677.html" title="Last updated on Dec 10, 2025, 8:34 AM UTC (da754d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2677/a1a7a0d...da754d9.html" title="Last updated on Dec 10, 2025, 8:34 AM UTC (da754d9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -12047,7 +12047,6 @@ button.ariaPressed; // null</pre
         <ol>
           <li><a href="#attrs_widgets">Widget Attributes</a></li>
           <li><a href="#attrs_liveregions">Live Region Attributes</a></li>
-          <li><a href="#attrs_dragdrop">Drag-and-Drop Attributes</a></li>
           <li><a href="#attrs_relationships">Relationship Attributes</a></li>
         </ol>
         <section id="attrs_widgets">
@@ -12100,17 +12099,6 @@ button.ariaPressed; // null</pre
             <li><sref>aria-busy</sref></li>
             <li><pref>aria-live</pref></li>
             <li><pref>aria-relevant</pref></li>
-          </ul>
-        </section>
-        <section id="attrs_dragdrop">
-          <h3>Drag-and-Drop Attributes</h3>
-          <p>
-            This section lists [=attributes=] which indicate information about drag-and-drop interface [=elements=], such as draggable elements and their drop targets. Drop target information will be
-            rendered visually by the author and provided to <a>assistive technologies</a> through an alternate modality.
-          </p>
-          <ul>
-            <li><pref>aria-dropeffect</pref></li>
-            <li><sref>aria-grabbed</sref></li>
           </ul>
         </section>
         <section id="attrs_relationships">
@@ -16925,7 +16913,6 @@ button.ariaPressed; // null</pre
             <td><sref>aria-disabled</sref></td>
             <td><a href="#valuetype_true-false">true/false</a></td>
           </tr>
-          <!-- <tr><td><dfn>ariaDropeffect</dfn></td><td><pref>aria-dropeffect</pref></td></tr> -->
           <tr>
             <td><dfn>ariaErrorMessageElements</dfn></td>
             <td><pref>aria-errormessage</pref></td>
@@ -16941,7 +16928,6 @@ button.ariaPressed; // null</pre
             <td><pref>aria-flowto</pref></td>
             <td><a href="#valuetype_idref_list">ID reference list</a></td>
           </tr>
-          <!-- <tr><td><dfn>ariaGrabbed</dfn></td><td><pref>aria-grabbed</pref></td></tr> -->
           <tr>
             <td><dfn>ariaHasPopup</dfn></td>
             <td><pref>aria-haspopup</pref></td>
--- a/.github/workflows/accname.yml
+++ b/.github/workflows/accname.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout monorepo
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           ref: main
           path: aria
@@ -47,7 +47,7 @@ jobs:
             specStatus: WD
           ARTIFACT_NAME: accname
       - name: Checkout accname repo
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           repository: w3c/accname
           ref: gh-pages
--- a/.github/workflows/aria.yml
+++ b/.github/workflows/aria.yml
@@ -13,7 +13,7 @@ jobs:
     permissions:
       contents: write
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: w3c/spec-prod@v2
         with:
           GH_PAGES_BRANCH: gh-pages
--- a/.github/workflows/core-aam.yml
+++ b/.github/workflows/core-aam.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout monorepo
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           ref: main
           path: aria
@@ -48,7 +48,7 @@ jobs:
             specStatus: CRD
           ARTIFACT_NAME: core-aam
       - name: Checkout core-aam repo
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           repository: w3c/core-aam
           ref: gh-pages
--- a/.github/workflows/dpub-aam.yml
+++ b/.github/workflows/dpub-aam.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout monorepo
-        uses: actions/checkout@v5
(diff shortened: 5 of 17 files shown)


cookiecrook:
Added Agenda+ to [the related issue](https://github.com/w3c/aria/issues/2674), just to make sure this doesn't slip by any WG members accidentally. Removing the duplicate keyword here.

spectranaut:
Assigning myself to test orca.

cookiecrook:
Braille tests revealed no unexpected results. 

@spectranaut Assuming Orca aligns with the others, you can merge since this has two other approvals.

index.html
```diff
@@ -14555,7 +14546,7 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
               </tr>
               <tr>
                 <th class="value-name" scope="row"><strong class="default">off (default)</strong></th>
-                <td class="value-description">Indicates that updates to the region should not be presented to the user unless the user is currently focused on that region.</td>
+                <td class="value-description">Indicates that updates to the region should not be presented to the user.</td>
```
cookiecrook: Correcting this errant RFC-2119 term usage after it was caught by the PR check.

```suggestion
                <td class="value-description">Indicates that updates to the region will not be presented to the user.</td>
```

spectranaut (approved):
From my testing in chrome on linux, Orca ignores updates in `timer` and `marquee`, like expected.

However, this does actually need bugs on browsers -- at least chrome exposes aria-live:none on these roles: https://source.chromium.org/chromium/chromium/src/+/main:content/test/data/accessibility/aria/aria-timer-expected-auralinux.txt

And if we remove it from the AAM, then it shouldn't.

spectranaut:
@joanmarie -- this change to Core-AAM removes "container-live:off" and "live:off" from the object attributes for `marquee` and `timer`: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fcore-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-2677--wai-aria.netlify.app%2Fcore-aam%2F#role-map-marquee

But chrome exposes this information on linux for both regardless of whether aria-live is present. I assume, but wanted to make sure, that Orca doesn't need that information (in the case that aria-live is not present) before opening the bug on chrome to remove it?

joanmarie:
Just checked. Orca's definition is:

```python
    @staticmethod
    def is_live_region(obj: Atspi.Accessible, _role: Atspi.Role | None = None) -> bool:
        """Returns True if obj is a live region."""

        attrs = AXObject.get_attributes_dict(obj)
        return "container-live" in attrs and attrs.get("container-live") in ["polite", "assertive"]
```

So removing the attribute is no different than exposing it with a value of "off". Remove away! :) And thanks for checking!!

spectranaut:
Ok, I added bugs.

Because (1) change to Core-AAM is not technically normative -- we do not say that you "MUST NOT" expose attributes not listed in the AAM tables and (2) exposing container-live:off and live:off (container-live-role:.. is already not exposed) has no downstream effect, I think this is fine to merge.

patrickhlauke:
sorry for the naive question, but: does this change now mean that `aria-live=off` isn't the old magical "it's supposed to be announced, but only when the AT cursor/focus is inside/on the element", and instead "off" now means what most people would assume..."this is not a live region"?

(i.e., quite self-servingly, can i now update the first part of that final note in https://tetralogical.com/blog/2024/05/01/why-are-my-live-regions-not-working/ that rants about "off" being weird? and do we need to make updates to MDN and such?)

PR #2680, opened 2025-11-25
[core-aam]: add Android to sections 1.1 and 1.2.

dtsengchromium:


dtsengchromium:
Thanks, Cynthia! (great to see your name).
    
I suspect the CI failure[1] is happening because the workflow is configured to fetch the branch name directly from origin (the main repo), but my branch exists on a fork.
    
Some AI-suggested fixes[2] to this issue. I haven't spent time looking at this but can given appropriate perms. Probably simpler if someone with write access can do it :).
    
1. +refs/tags/android_frontmatter*:refs/tags/android_frontmatter*
      Error: The process '/usr/bin/git' failed with exit code 1
    
     https://github.com/w3c/aria/actions/runs/19685387275/job/56389556047?pr=2680
    
2.  Suggestion: Could you verify the workflow configuration? Usually, removing the ref input allows actions/checkout to handle forks automatically by checking out the PR's merge commit instead of the branch name.
    
    Example fix:
```
    YAML
    
    # Change this:
    - uses: actions/checkout@v4
      with:
        ref: ${{ github.head_ref }}
    
    # To this (use default behavior):
    - uses: actions/checkout@v4
```

core-aam/index.html
```diff
@@ -250,6 +250,7 @@ <h3>Accessibility <abbr title="Application Programming Interfaces">APIs</abbr></
           <li><cite>User Interface Automation</cite> [[UI-AUTOMATION]]</li>
           <li><cite>ATK - Accessibility Toolkit</cite> [[ATK]] and <cite>Assistive Technology Service Provider Interface</cite> [[AT-SPI]], referred to hereafter as "ATK/AT-SPI"</li>
           <li><cite>macOS Accessibility Protocol</cite> [[AXAPI]]</li>
+          <li><cite>Android</cite> [[AccessibilityNodeInfo]]</li>
```
spectranaut: This link doesn't work, it should bring us to an entry in the references section: https://w3c.github.io/core-aam/#references

However, I looked around a bit and I don't see how to do this. Also I noticed that there is a strange split between normative and informative references... which I suppose I'll make an issue about: https://github.com/w3c/core-aam/issues/255

Also I wonder about the name, "AccessibilityNodeInfo", I think from looking at the documents that maybe it should be the umbrella [android.view.accessibility
](https://developer.android.com/reference/android/view/accessibility/package-summary) ... because might we want to reference AccessibilityEvents as well?

cyns:
> Thanks, Cynthia! (great to see your name).
> 
Hi David, was great to see your name too!

core-aam/index.html
```diff
@@ -58,6 +58,13 @@
             mailto: "buphie@gmail.com",
             w3cid: 11897,
           },
+          {
+            name: "David Tseng",
+            company: "Google LLC",
+            mailto: "dtseng@google.com",
+            note: "AccessibilityNodeInfo", // Or android.view.accessibilty or whatever the name of this API is
```
daniel-montalvo: @dtsengchromium  could you please update this accordingly?

spectranaut: Hi Daniel, the idea was to update the "Platform Mapping Maintainers" list, not the editors list -- the list that is called "authors" shows up in the "Platform Mapping Maintainers" list in the rendered specification. I'm not asking David to be an editor, just a platform mapping maintainer :)

dtsengchromium:
Thanks all! I believe I've address comments. Please have another look.

PR #2681, opened 2025-11-25
[core-aam]: add Android TBD entries

dtsengchromium:


cyns (approved):

dtsengchromium:
Thanks Cynthia! I think, similar to the other pr, we need to try and fix
the check failure or someone with admin can push for me.

core-aam/index.html
```diff
@@ -10757,6 +12041,12 @@ <h3>Changes to document content or node visibility</h3>
                 <td><code>text_changed::delete</code> and <code>text_changed::insert</code></td>
                 <td>If in a <a class="termref">live region</a>, <code>AXLiveRegionChanged</code>.<br />If in <code>aria-errormessage</code>, <code>AXValidationErrorChanged</code>.</td>
               </tr>
+              <tr>
+                <th>Android</th>
+                <td>
+                  <p>TBD</p>
+                </td>
+              </tr>
```
spectranaut: This table, and all of the ones following, shouldn't have an android row -- the format of these tables are a bit different. I think just remove the rows for now, we can update the tables later. 

See the table in the PR preview here: https://deploy-preview-2681--wai-aria.netlify.app/core-aam/#mapping_events_visibility

PR #2682, opened 2025-11-25
core aam roles known entries

dtsengchromium:
- **Add initial Android mappings**
- **[core-aam]: fill in known Android role mappings**


cyns (commented):
Great start! 
I noticed that a lot of roles map to android.view.View. Is this sufficient to express their semantics and behavior? Similarly, is GridView sufficient to express the semantics of a data table using the table role?

dtsengchromium:
Thanks! This can be a topic at the next wg meeting and I can try to attend
and help fill the group in on the general effort.

tl;dr mappings to
android.view.View
reflect the current usage in Chrome.

As background, classnames represent existing Android widgets. This can be
anything from the Android sdk or perhaps beyond. The widget itself defines
the more granular details of how the "role" works on Android.

For cases where we have no mapping or no closest match, we may default to
android.view.View
which at least surfaces a base level of support and interpretation by ATs.

For the specification process, we can also default to
android.view.View
or simply leave as
TBD
if the group prefers.

daniel-montalvo:
@dtsengchromium Can you please fix the conflicts here? It's probably just a matter of accepting what is currently on this branch, it's just that since we merged #2681 we created the conflicts.

dtsengchromium:
@daniel-montalvo I'm not sure how the merge was done. I've tried to see the conflicts by syncing the branch, editing its upstream, etc. I'd rather not have to go through the web UI if possible since this should have been a clean merge.
  
  I'm also wondering if there's a preferred workflow the group is used to. I probably should have just created one pr since it looks like these changes are not being merged to main.

dtsengchromium:
That looks to have fixed the conflict. Please take another look.

Still curious about preferred workflows.

daniel-montalvo:
Looking great now @dtsengchromium 

Thanks!

spectranaut:
> For cases where we have no mapping or no closest match, we may default to android.view.View which at least surfaces a base level of support and interpretation by ATs. For the specification process, we can also default to android.view.View or simply leave as TBD if the group prefers.

Hi @dtsengchromium, sorry for the late reply, I finally looked at this PR more closely. About this question -- I'm curious, are there some roles in ARIA that you think will always map to android.view.View? Roles for which android.view.View is actually the correct mapping? It sounds like in at least some cases it is a stopgap.

As for the AAM, I would say, if Android currently maps the role to android.view.View, then that is what we should have in the AAM. If it is generally or always a stopgap, we could have a note somewhere that says something about android.view.View being a stopgap in some cases, explaining why, and explaining that that mapping may change. That could belong in the [android summary section](https://deploy-preview-2682--wai-aria.netlify.app/core-aam/#android-accessibility-api) or maybe even the [Exposing attributes that do not directly map to accessibility API properties](https://deploy-preview-2682--wai-aria.netlify.app/core-aam/#mapping_nodirect) section.

Also, I'm curious, for things that map to `android.view.View`, are there other things that should be exposed or set in this class? Just looking around at the chromium tests, the mapping for [columnheader](https://deploy-preview-2682--wai-aria.netlify.app/core-aam/#role-map-columnheader), should it say something about the [`table_header` like in this test](https://source.chromium.org/chromium/chromium/src/+/main:content/test/data/accessibility/aria/aria-table-expected-android.txt)?

If the answer is yes, then I think we should only land things in the AAM that have a full description of everything a implementer would need to expose for a core-aam role to be accessible, including these attributes. If the class alone is sufficient, then we can land those.

dtsengchromium:
Hi @cyns, thanks for the feedback! You are right that many roles default to `android.view.View`. In many cases, this is a base mapping, and the specific semantics are carried by attached attributes like `CollectionInfo`, `CollectionItemInfo`, `RangeInfo`, or methods like `setHeading` and `setLiveRegion`. In the latest PR updates, I've added notes to clarify these specific exposures where applicable. Some of these are still subject to change as the Android accessibility support evolves.

dtsengchromium:
Hi @spectranaut, thanks for the detailed review! To answer your points:

1. **On `android.view.View` as a stopgap**: It's a bit of both. For generic containers or roles without a direct Android equivalent, `View` is often the correct mapping. For others, it might serve as a stopgap.
2. **On adding a note to the summary**: I actually have already added a note to the Android summary section (around line 356) explaining that `android.view.View` is used in some cases as a stopgap and may change.
3. **On additional exposures**: Yes, absolutely. Many roles rely on extra attributes or specific methods to be fully exposed.
4. **On landing full descriptions**: I agree! In this latest PR update, I have added specific notes for many roles to document these extra exposures (like `RangeInfo` for range roles, `setHeading` for headers, and `CollectionItemInfo` for items). This should make the mappings much more complete and useful for implementers.

Let me know if this covers what you were thinking!

core-aam/index.html
```diff
@@ -2527,7 +2529,9 @@ <h4 id="role-map-heading"><code>heading</code></h4>
               <tr>
                 <th>Android</th>
                 <td>
-                  <p>TBD</p>
+                  <p>android.view.View</p>
+                  <p>Note: Exposed as <code>android.widget.TextView</code> if the element has only text children.</p>
+                  <p>Note: Marked as a heading via <code>AccessibilityNodeInfo#setHeading</code>.</p>
```
spectranaut: Since in general we describe how the API is consumed, I wonder if we should reference isHeading instead of setHeading: https://developer.android.com/reference/android/view/accessibility/AccessibilityNodeInfo#isHeading()

Also, for other APIs, when describing the result of an API call, `Method:` is used instead of `Note:`

```suggestion
                  <p>Method: <code>isHeading</code>: <code>true</code></p>
```

core-aam/index.html
```diff
@@ -3296,7 +3301,8 @@ <h4 id="role-map-math"><code>math</code></h4>
               <tr>
                 <th>Android</th>
                 <td>
-                  <p>TBD</p>
+                  <p>android.view.View</p>
+                  <p>Note: Mathematical content is indicated by the presence of a <a href="https://developer.android.com/reference/android/view/accessibility/AccessibilityNodeInfo.MathInfo"><code>MathInfo</code></a> object.</p>
```
spectranaut: I also feel like this isn't quite a "Note" if it is always necessary. Maybe something like: 
```suggestion
                  <p>Method: <code>getStructuredDataInfo</code> returns mathematical content in a <a href="https://developer.android.com/reference/android/view/accessibility/AccessibilityNodeInfo.MathInfo"><code>MathInfo</code></a> object.</p>
```

core-aam/index.html
```diff
@@ -3358,7 +3364,8 @@ <h4 id="role-map-menu"><code>menu</code></h4>
               <tr>
                 <th>Android</th>
                 <td>
-                  <p>TBD</p>
+                  <p>android.view.View</p>
+                  <p>Note: Exposed with <code>CollectionInfo</code>.</p>
```
spectranaut: Maybe similar to the MathInfo?

core-aam/index.html
```diff
@@ -3420,7 +3427,8 @@ <h4 id="role-map-menubar"><code>menubar</code></h4>
...
```
spectranaut: Similar to MathInfo suggestion?

core-aam/index.html
```diff
@@ -3660,7 +3668,8 @@ <h4 id="role-map-meter"><code>meter</code></h4>
               <tr>
                 <th>Android</th>
                 <td>
-                  <p>TBD</p>
+                  <p>android.widget.ProgressBar</p>
+                  <p>Note: Exposed with <code>RangeInfo</code> if it represents a numeric range without <code>aria-valuetext</code>.</p>
```
spectranaut: I think this should also be written like the MathInfo suggestion. I think you can leave the prose: 
```suggestion
                  <p>Method: <code>getStructuredDataInfo</code> returns <code>RangeInfo</code> if it represents a numeric range without <code>aria-valuetext</code>.</p>
```

core-aam/index.html
```diff
@@ -3960,7 +3969,8 @@ <h4 id="role-map-option-in-combobox"><code>option</code> inside <code>combobox</
               <tr>
                 <th>Android</th>
                 <td>
-                  <p>TBD</p>
+                  <p>android.view.View</p>
+                  <p>Note: Exposed with <code>CollectionItemInfo</code>.</p>
```
spectranaut: Same as MathInfo

core-aam/index.html
```diff
@@ -5192,7 +5208,8 @@ <h4 id="role-map-status"><code>status</code></h4>
               <tr>
                 <th>Android</th>
                 <td>
-                  <p>TBD</p>
+                  <p>android.view.View</p>
+                  <p>Note: Marked as a live region via <code>AccessibilityNodeInfo#setLiveRegion</code>.</p>
```
spectranaut: Same as `isHeading`?

core-aam/index.html
```diff
@@ -5250,7 +5267,8 @@ <h4 id="role-map-strong"><code>strong</code></h4>
               <tr>
                 <th>Android</th>
                 <td>
-                  <p>TBD</p>
+                  <p>android.view.View</p>
+                  <p>Note: Bold styling is exposed via a <code>StyleSpan</code> with <code>Typeface.BOLD</code> in <code>AccessibilityNodeInfo#setText</code>.</p>
```
spectranaut: I'm not sure how to write this and all the other span-related content (i.e. role link) in a way that seems more aligned with the rest of the spec... The only similarly concept I can think of is the AX API mappings for role mark: https://w3c.github.io/core-aam/#role-map-mark

Is there a way to document this as the return of some specific API call?

core-aam/index.html
```diff
@@ -5914,7 +5936,8 @@ <h4 id="role-map-textbox-multiline"><code>textbox</code> when <code>aria-multili
               <tr>
                 <th>Android</th>
                 <td>
-                  <p>TBD</p>
+                  <p>android.widget.EditText</p>
+                  <p>Note: Marked as multiline via <code>AccessibilityNodeInfo#setMultiLine</code>.</p>
```
spectranaut: See comment on role heading.

core-aam/index.html
```diff
@@ -6132,7 +6155,8 @@ <h4 id="role-map-tooltip"><code>tooltip</code></h4>
               <tr>
                 <th>Android</th>
                 <td>
-                  <p>TBD</p>
+                  <p>android.view.View</p>
+                  <p>Note: The tooltip text is typically exposed via <code>AccessibilityNodeInfo#setTooltipText</code> on the associated anchor element.</p>
```
spectranaut: This is trickier, because we are talking about another element/node. I think this makes sense as a "Note:" so I would leave it.

spectranaut (changes requested):
Hi David, Thanks for the work here and the answers in the previous comment! This looks really good. There are a few patterns I'd like changed to match the way the other APIs are specified -- related to the *Info objects, and methods in the form of `setFoo`, and the metadata on spans of text.

cyns:
> Hi @cyns, thanks for the feedback! You are right that many roles default to `android.view.View`. In many cases, this is a base mapping, and the specific semantics are carried by attached attributes like `CollectionInfo`, `CollectionItemInfo`, `RangeInfo`, or methods like `setHeading` and `setLiveRegion`. In the latest PR updates, I've added notes to clarify these specific exposures where applicable. Some of these are still subject to change as the Android accessibility support evolves.

core-aam/index.html
dtsengchromium: Agreed. Made similar to other platforms with formatting.

core-aam/index.html
dtsengchromium: Agreed and made similar in formatting to other platforms.

core-aam/index.html
dtsengchromium: Agreed. Updated to use the `method` span and described the mapping via `AccessibilityNodeInfo.getCollectionInfo`.

core-aam/index.html
dtsengchromium: Agreed. Updated to use the `method` span and `AccessibilityNodeInfo.getCollectionInfo` to match the other collection roles.

core-aam/index.html
dtsengchromium: Agreed. Updated to use the `method` span and standard Android API `AccessibilityNodeInfo.getRangeInfo` to describe the mapping.

core-aam/index.html
dtsengchromium: Agreed. Updated to use the `method` span and standard Android API `AccessibilityNodeInfo.getCollectionItemInfo`.

core-aam/index.html
dtsengchromium: Agreed. Updated to describe the consumed API using the correct constant: `Method: AccessibilityNodeInfo.getLiveRegion: View.ACCESSIBILITY_LIVE_REGION_POLITE`.

core-aam/index.html
dtsengchromium: Agreed. Documented this as the return of `AccessibilityNodeInfo.getText` using the `method` span to align with other platforms. I have also updated the other span-related roles (`link`, `subscript`, `suggestion`, `superscript`) to use this same `getText` pattern for consistency.

core-aam/index.html
dtsengchromium: Agreed. Updated to describe the consumed API using the `method` span to match the heading role.

core-aam/index.html
dtsengchromium: Agreed. Kept as a `Note:` but updated the code reference to `AccessibilityNodeInfo.getTooltipText` to consistently describe the consumed API.

dtsengchromium:
@spectranaut updated everything to use method spans and fully qualified names. Let me know how this looks!

spectranaut (changes requested):
Thank you for the changes and I'm sorry to say I just caught a formatting issue! Sorry I didn't catch this earlier -- I must have not looked at the rendered file...

For the android mappings, you added paragraph tags inside the `td` -- this causes extra space around each entry in the android row that doesn't exist around the other API's mappings.

All other API mappings are in the format:
```
<span>Role: foo</span><br />
<span>Role: foo</span><br />
<span>Note: some stuff</span>
```

dtsengchromium:
@spectranaut Good catch on the rendered spacing! I've updated all Android mapping cells to use `<span>` tags and `<br />` line breaks instead of paragraphs. Base class names are now consistently wrapped in `<code>` as well. Let me know how this renders for you!

spectranaut (approved):
Ok, this looks good!!

spectranaut:
@cyns or @lucasradaelli it would be nice to get another +1, if either of you (or both) could take a look.

dtsengchromium:
Thanks Val! Perhaps we should have a ping at the meeting tomorrow for this change?

PR #2697, opened 2025-12-11
Editorial: Normalize on using “accessibility child” — not “child elements”

sideshowbarker:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2697--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2697--wai-aria.netlify.app%2Findex.html)

This makes the following changes:

- Replace all occurrences of _“child element(s)”_ or _“child(ren)”_ with _“accessibility child(ren)”._
- Make every occurrence of _“accessibility child(ren)”_ a hyperlink to where that term is defined in the spec.
- Do similarly for _“allowed child(ren)”_ — make those all actual hyperlinks to `#mustContain`.

Rationale: To typical readers of the spec, it is ambiguous and misleading for the spec to be referring to something as a _“child element”_ — or simply a _“child”_ — in cases where that might not actually be a child _in the DOM_.

For a typical reader who sees the term _“child element”_ or _“child”_ on its own, it is completely _reasonable_ for that reader to assume that what’s meant by that term is the same as what _“child element”_ and _“child”_ mean in the DOM spec and the HTML spec — and in basically every single other spec for the web platform, everywhere.

Readers cannot reasonably be expected to realize that _only_ when they are reading this spec, _“child element”_ and _“child”_ do _not_ have their normal meaning but instead actually mean _“accessibility child”_.

So instead _always_ consistently using the (hyperlinked) term _“accessibility child”_ makes the intended meaning unambiguous to readers.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria/pull/2697.html" title="Last updated on Jan 13, 2026, 8:11 PM UTC (0fc105d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria/2697/0b40412...0fc105d.html" title="Last updated on Jan 13, 2026, 8:11 PM UTC (0fc105d)">Diff</a>


index.html
```diff
@@ -1175,21 +1175,21 @@ <h3><dfn>Prohibited</dfn> States and Properties</h3>
         <section id="mustContain">
           <h3>Allowed Accessibility Child Roles</h3>
           <p>
-            A list of roles which are allowed on an <a>accessibility child</a> (simplified as "child") of the element with this <a>role</a>. Authors MUST only add child element with allowed roles. For
-            example, an element with the role <rref>list</rref> can own child elements with the role <rref>listitem</rref>, but cannot own elements with the role <rref>option</rref>.
+            A list of roles which are allowed on an <a>accessibility child</a> of the element with this <a>role</a>. Authors MUST only add <a>accessibility children</a> with allowed roles. For
+            example, an element with the role <rref>list</rref> can own <a>accessibility children</a> with the role <rref>listitem</rref>, but cannot own elements with the role <rref>option</rref>.
```
```suggestion
            example, an element with the role <rref>list</rref> allows <a>accessibility children</a> with the role <rref>listitem</rref>, but does not allow <a>accessibility children</a> with the role <rref>option</rref>.
```

index.html
```diff
@@ -1175,21 +1175,21 @@ <h3><dfn>Prohibited</dfn> States and Properties</h3>
...
           </p>
-          <p>To determine whether an element is the <a>child</a> of an element, <a>user agents</a> MUST ignore any intervening elements with the role <rref>generic</rref> or <rref>none</rref>.</p>
+          <p>To determine whether an element is the <a>accessibility child</a> of an element, <a>user agents</a> MUST ignore any intervening elements with the role <rref>generic</rref> or <rref>none</rref>.</p>
```
```suggestion
```

This paragraph is redundant / covered by the definition for accessibility child.

index.html
```diff
@@ -6998,7 +6998,7 @@ <h5>Presentational Role Inheritance</h5>
               explicit role of <rref>none</rref>/<rref>presentation</rref>, the user agent MUST apply an inherited role of <rref>none</rref> to any <a>accessibility descendants</a> that do not have an
               explicit role defined. Also, when an explicit or inherited role of <rref>none</rref>/<rref>presentation</rref> is applied to a host language element which has specifically allowed
               children as defined by the host language specification, in addition to the element with the explicit role of <rref>none</rref>/<rref>presentation</rref>, the user agent MUST apply an
-              inherited role of <rref>none</rref> to any specifically allowed children that do not have an explicit role defined.
+              inherited role of <rref>none</rref> to any specifically <a href="#mustContain">allowed accessibility children</a> that do not have an explicit role defined.
```
pkra: I'm not sure about this one but I'm struggling with this paragraph. The "specifically allowed children" refers to host languages content models (I think?). 

I'm not sure what should (or does) happen if such a (host language) child ends up elsewhere in the accessibility tree (i.e. has aria-owns). I think that needs some investigation.

I'd suggest to revert this for the time being.

pkra: I've filed https://github.com/w3c/aria/issues/2711. It seems this change would not match implementations.

sideshowbarker: > ```diff
> --- a/index.html
> +++ b/index.html
> -              inherited role of <rref>none</rref> to any specifically allowed children that do not have an explicit role defined.
> +              inherited role of <rref>none</rref> to any specifically <a href="#mustContain">allowed accessibility children</a> that do not have an explicit role defined.
> ```
>
> @pkra wrote:
>
> I'd suggest to revert this for the time being.

Made it so.

sideshowbarker:
> Thank you, @sideshowbarker!

Thank you for the review+merge~

PR #2717, opened 2026-01-20
Implement mappings for popover hint

keithamus:
🚀 **Netlify Preview**:
- [html-aam preview](https://deploy-preview-2717--wai-aria.netlify.app/html-aam/index.html) &mdash; [html-aam diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fhtml-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-2717--wai-aria.netlify.app%2Fhtml-aam%2Findex.html)

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -3180,7 +3180,10 @@
             </tr>
           </tbody>
         </table>
-        <h4 id="el-img-empty-alt">`img` <span class="el-context">(`alt` attribute value is the empty string, i.e., `alt=""` or `alt` with no value in the markup)</span></h4>
+        <h4 id="el-img-empty-alt">
+          `img`
+          <span class="el-context">(`alt` attribute value, when trimmed of [=ascii whitespace|whitespace=], is the empty string, i.e., `alt=""`, `alt=" "`, or `alt` with no value in the markup)</span>
+        </h4>
         <table class="data" aria-labelledby="el-img-empty-alt">
           <tbody>
             <tr>
@@ -3188,7 +3191,8 @@
               <td>
                 <a data-cite="html">`img`</a>
                 <span class="el-context"
-                  >(<a data-cite="html/embedded-content.html#attr-img-alt">`alt`</a> attribute value is the empty string, i.e., `alt=""` or `alt` with no value in the markup)</span
+                  >(<a data-cite="html/embedded-content.html#attr-img-alt">`alt`</a> attribute value, when trimmed of [=ascii whitespace|whitespace=], is the empty string, i.e., `alt=""`, `alt=" "`,
+                  or `alt` with no value in the markup)</span
                 >
               </td>
             </tr>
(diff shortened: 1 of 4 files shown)


html-aam/index.html
cookiecrook, line 9403: May need something else here depending on the outcome of https://webkit.org/b/306041

html-aam/index.html
```diff
@@ -237,6 +237,26 @@ <h3 id="minimum-role">Exposing HTML Features That Require a Minimum Role</h3>
           require a minimum role be returned as the computed role for the element, prioritize the more specific role in the ARIA taxonomy.
         </p>
       </section>
+      <section>
+        <h3 id="plain-content">Plain Content</h3>
```
cookiecrook, line 241: Is there an HTML spec concept here we can reference instead? This overlaps the similar concept of block/flow content in a way that leaves ambiguity in my understanding of your intention.

cookiecrook: For example, images aren't always plain, as in the context of an animated SVG. Nor can an image (not even a bitmap) be referenced as plaintext, unless you bring in another concept like an object replacement character. This seems to blur the line, esp re: attributed strings and other similar rich text concepts like Flow contents.

html-aam/index.html
cookiecrook, line 13722: May need something else here depending on the outcome of https://webkit.org/b/271485 and https://webkit.org/b/306041

html-aam/index.html
keithamus: No this is somewhat of an ARIA semantic. At least [in Chrome](https://source.chromium.org/chromium/chromium/src/+/main:ui/accessibility/ax_role_properties.cc;l=1126-1143;drc=796e6996a524533ef2af2ab09c8dd541368f491b;bpv=1;bpt=1) and how [I've also implemented this in Firefox](https://phabricator.services.mozilla.com/D276568#change-NVNgxwA4MRea) is to iterate over the popover & it's descendants and determine if the role matches.

I don't know if the heuristic is worth changing or the spec doesn't properly express this idea, but I am open to suggestions.

html-aam/index.html
```diff
@@ -13760,18 +13808,20 @@ <h4 id="att-popovertarget">`popovertarget`</h4>
...
+                  User Agents MUST expose a relation with the associated popover element <strong>except</strong> under the following conditions:
                 </p>
                 <ul>
                   <li>The element's `popovertargetaction` attribute value is "`hide`"</li>
                   <li>The associated popover element is the next immediate accessibility sibling to the invoking element,</li>
                   <li>The element's implicit or explicit `popovertargetaction` is the "`auto`" state and the element is a descendant of the `popover` it is associated with.</li>
                   <!-- accessibility sibling will be defined in the ARIA specification -->
                 </ul>
+                <p>
+                  Author-provided `aria-describedby` or `aria-details` attributes on the invoking element take precedence over these implicit relations.
```
scottaohara, line 13828: we may want to mention that even if they resolve to the empty string or result in no associated details relation (e.g., aria-details="").

i remember talking a lot about this with aaron, and how to allow for people to opt out of an automatic description/details relation, if they needed to.

this comment applies to other instances of this text in this PR

spectranaut:
Discussed today: https://www.w3.org/2026/02/12-aria-minutes.html#630f

Conclusions:
* Keith will address/resolve Scott's and James' comments
* Keith will find some examples for Matt King and the APG to look at, see linked issue in card
* Sarah will still review
* @aardrian do you still want to/have time to review this?

PR #2726, opened 2026-02-05
Add aria-placeholder for combobox role

### LINKED ISSUE #2689: Missing availability of aria-placeholder property with combobox role
Just as using the `textbox` role in the absence of a `<input>` element makes a `<div>` element with `contenteditable` an input for ATs, the same applies to the `combobox` role in the last spec version: 

Encoded on a `<div>` or `<span>` element with `contenteditable`, this role also signals to ATs that it is an input with additional values offered for selection.

Since the HTML `placeholder` attribute is not available in this case, the `aria-placeholder` should be available on elements with the `combobox` role.

FabasoftAccessibility:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2726--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2726--wai-aria.netlify.app%2Findex.html)

Closes #2689

Add aria-placeholder as a property of the combobox.

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:
* [ ] ACT review?
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

### DIFF
--- a/index.html
+++ b/index.html
@@ -2930,6 +2930,7 @@
                     <li><pref>aria-errormessage</pref></li>
                     <li><pref>aria-haspopup</pref></li>
                     <li><sref>aria-invalid</sref></li>
+                    <li><sref>aria-placeholder</sref></li>
                     <li><pref>aria-readonly</pref></li>
                     <li><pref>aria-required</pref></li>
                   </ul>


index.html
```diff
@@ -2930,6 +2930,7 @@ <h2>Definition of Roles</h2>
                     <li><pref>aria-errormessage</pref></li>
                     <li><pref>aria-haspopup</pref></li>
                     <li><sref>aria-invalid</sref></li>
+                    <li><sref>aria-placeholder</sref></li>
```
scottaohara, line 2933: per my comments in the issue that prompted this PR, does just adding the attribute without any caveats about usage make sense?

i still don't see the practical use here (nor is there support) for a select-only combobox that isn't based on a text input field

smhigley: +1, I think adding this should be coupled with something in (probably) Core AAM on how it's handled when the `aria-placeholder` string matches the current text content, which is how I assume a select-only-combobox's placeholder would be actually authored to render.

pkra:
Closing this since the authors have left the group. If somebody else wants to take this up, they can re-open or create a new PR.

PR #2736, opened 2026-02-19
Editorial: Editorial updates to SVG-AAM after re-publication as a WD

daniel-montalvo:
🚀 **Netlify Preview**:
- [svg-aam preview](https://deploy-preview-2736--wai-aria.netlify.app/svg-aam/index.html) &mdash; [svg-aam diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fsvg-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-2736--wai-aria.netlify.app%2Fsvg-aam%2Findex.html)

This is mainly to bring back to the Editor's Draft the updated refs that we caught while publishing the Working Draft under ARIA WG

### DIFF
--- a/svg-aam/index.html
+++ b/svg-aam/index.html
@@ -31,7 +31,7 @@
         //previousDiffURI:      "http://www.w3.org/TR/2014/REC-wai-aria-implementation-20140320/",
 
         // if there a publicly available Editors Draft, this is the link
-        ED: "https://github.com/svg-aam/",
+        ED: "https://w3c.github.io/svg-aam/",
         github: "w3c/svg-aam",
 
         // if this is a LCWD, uncomment and set the end of its review period


daniel-montalvo:
@dd8 @cyns This incorporates our work to publish the 17 Feb Working Draft into our main branch. And also updates the token to hopefully enable auto-publishing to TR as a Working Draft.

cyns:
Is the content from old section 3 somewhere else now?
And, I do miss the big table from the old-style aam docs, but I think I may be the only one who does! No need to block on that.

daniel-montalvo:
> Is the content from old section 3 somewhere else now? And, I do miss the big table from the old-style aam docs, but I think I may be the only one who does! No need to block on that.

It's removed. The [Core-aam](https://deploy-preview-2736--wai-aria.netlify.app/core-aam/index.html) preview doesn't have it. I think we decided we didn't want to duplicate these terms and only have "important terms" when we uniquely define a term on a child spec.

I'll remove the empty container.

daniel-montalvo:
@pkra  as to your point related to w3c/html-aria#568, the new id role-map-svg is kept in this PR, so it wouldn't be changing back.

My understanding after having done some digging into this is that these ids `details-id-##` where generated by scripts when we still used the old tables. It's a while now since we moved away from them, thus the ids we are generating now are way more meaningful, which I think makes things better :-)

PR #2738, opened 2026-02-21
editorial: tweak group role description

### LINKED ISSUE #1796: Clarify that group role may be used with non-operable content
# Describe the change
I'm concerned that the phrase "user interface objects" mentioned in the spec for `group` might incorrectly imply "operable elements", yet a `group` may certainly contain non-operable content, at least that's how I understand the relationship to the 'section' superclass.

I dare say the current spec needlessly discourages the use of "group" for non-operable content, and that's a bit of a shame.

I note that "user interface objects" which appears several times in the spec does not have a formal definition. Is this a problem?

For example, in the section on Accessibility Tree, it says 
> The accessibility tree includes the user interface objects of the user agent **and** the objects of the document.

... suggesting (to me) that "user interface objects" and "objects of the document" (also undefined) are different.

Link to documentation: https://w3c.github.io/aria/#group

Does the issue exist in [the editors draft](https://w3c.github.io/aria/) (the editors draft is the most recent draft of the specification)?
Yes

pkra:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2738--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2738--wai-aria.netlify.app%2Findex.html)

Matches language in other roles in that other information is permitted in elements with group role.

Resolves #1796

### DIFF
--- a/index.html
+++ b/index.html
@@ -4739,7 +4739,7 @@
           <rdef>group</rdef>
           <div class="role-description">
             <p>A set of user interface <a>objects</a> and information that is not intended to be included in a page summary or table of contents by <a>assistive technologies</a>.</p>
-            <p>Contrast with <rref>region</rref>, which is a grouping of user interface objects that will be included in a page summary or table of contents.</p>
+            <p>Contrast with <rref>region</rref>, which is a grouping of user interface objects and information that will be included in a page summary or table of contents.</p>
             <p>
               Authors SHOULD use a <code>group</code> to form a logical collection of items in a <a>widget</a>, such as children in a tree widget forming a collection of siblings in a hierarchy.
               However, when a <code>group</code> is used in the context of a <rref>listbox</rref>, for example, authors MUST limit its children to <rref>option</rref> elements. Therefore, proper


index.html
```diff
@@ -4738,7 +4738,7 @@ <h2>Definition of Roles</h2>
         <div class="role" id="group">
           <rdef>group</rdef>
           <div class="role-description">
-            <p>A set of user interface <a>objects</a> that is not intended to be included in a page summary or table of contents by <a>assistive technologies</a>.</p>
+            <p>A set of user interface <a>objects</a> and information that is not intended to be included in a page summary or table of contents by <a>assistive technologies</a>.</p>
```
scottaohara, line 4741: so as to not read as "a group is a set of objects and information" but instead "a group is a set of objects which could be controls, information, or both?" - what about the following text?

```suggestion
            <p>A set of user interface <a>objects</a> representing interactive controls, informative content, or both that is not intended to be included in a page summary or table of contents by <a>assistive technologies</a>.</p>
```

pkra: I don't mind either way. 

I was copying from other roles. Should I make similar changes there?

scottaohara: naw.  i failed to double check on that.  this matches the other content (some of which i even wrote) so if it was good enough for me then, it should be now too :D

index.html
```diff
@@ -4738,7 +4738,7 @@ <h2>Definition of Roles</h2>
...
             <p>Contrast with <rref>region</rref>, which is a grouping of user interface objects that will be included in a page summary or table of contents.</p>
```
scottaohara: probably fine as is, but just in case we want to keep this 'object and information' consistent
```suggestion
            <p>Contrast with <rref>region</rref>, which is a grouping of user interface objects and information that will be included in a page summary or table of contents.</p>
```

pkra: Thanks, Scott!

PR #2744, opened 2026-03-10
Experiment with JSON to generate Core-AAM mapping tables

### LINKED ISSUE #258: Create more machine readable source of truth for mapping tables - use for tests also
What if we had a more machine readable source of truth for the mapping tables, which the tables were generated from, and also the AAM tests?

Then we edit the "source of truth" instead of the HTML tables.

spectranaut:
🚀 **Netlify Preview**:
- [core-aam preview](https://deploy-preview-2744--wai-aria.netlify.app/core-aam/index.html) &mdash; [core-aam diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fcore-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-2744--wai-aria.netlify.app%2Fcore-aam%2Findex.html)

Closes: https://github.com/w3c/core-aam/issues/258

Extracting the data wasn't too hard -- especially with the help of generative AI to write the extraction and generation scripts. I added the extraction script but in the long run I don't think anything like that would be checked in.

Some questions are:
* Is editing a JSON like this better than editing tables?
* Can we actually use this JSON to write tests?

About test writing:
* We would need to add more metadata in order to to this. For example, the "test" html, like `<div id="test" role="blockquote"></div>`. Sometimes we will want to test variations in the markup as well. 
* Also, not everything will be so easy to programmatically write tests from the prose alone, so we would have to mark/know which things could auto generate tests and which can't?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -220,3 +220,9 @@ pip-log.txt
 
 # Local Netlify folder
 .netlify
+
+#################
+## Node
+#################
+
+tools/node_modules/
--- a/core-aam/index.html
+++ b/core-aam/index.html
@@ -2193,19 +2193,19 @@
               <tr>
                 <th>MSAA + IAccessible2</th>
                 <td>
-                  <span class="property">Do not expose the <a class="termref">element</a> as a landmark. Use the native host language role of the element instead.</span>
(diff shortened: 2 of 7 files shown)


pkra:
Discussed in today's editors call https://www.w3.org/2026/03/11-aria-editors-minutes.html#bc81

PR #2745, opened 2026-03-10
editorial: aria.js clean-up

pkra:
Removes no longer needed respec roleInfo download button. 

Note: `ariaAttributeReferences` is called by respec configuration (postProcess)

Part of #2501


pkra:
Ugh. I broke the roleInfo builder.

PR #2747, opened 2026-03-11
editorial: fix buildRoleInfo.js

pkra:
Removes hack 1 of 3 (which is no longer needed after #2745)

Fixes a regression from #2745


pkra:
@daniel-montalvo I tested the fix locally.

Issue #2768, opened 2026-04-19
roleInfo update via github action failing


pkra:
The action `roleInfo.yml` (to update roleInfo.js) keeps failing with:

"Updates were rejected because the remote contains work that you do not"

For example, after I merged 

https://github.com/w3c/aria/pull/2766

this failure

https://github.com/w3c/aria/actions/runs/24632916199/workflow

I'm fairly sure this used to work so I'm not sure what's happened to break it.

pkra:
@daniel-montalvo could you please take a look?

daniel-montalvo:
Hi @pkra 

I am leaning towards thinking the message is a genuine one. In having a look at [all the recent runs](https://github.com/w3c/aria/actions/workflows/roleInfo.yml) I found out that this one you pointed out failed, but [the subsequent one](https://github.com/w3c/aria/actions/runs/24632922952) succeeded. It makes sense to me that this workflow started running, and it tried to push the changes after the other prettier workflows finished its job.

I'll try to see if we can avoid this by fetching origin/main first to minimize the chance of having the roleinfo updates pushed after the prettier commit.

pkra:
@daniel-montalvo thank you for looking into it. That makes sense. 

In #2769 I'm proposing a different approach for roleInfo altogether - which I hope resolves this issue as well (but would conflict with your proposal in #2774 - I'll leave a note there).

pkra:
This was resolved via https://github.com/w3c/aria/pull/2769

PR #2769, opened 2026-04-19
editorial: [aria.js] generate roleInfo after respec runs

pkra:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2769--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2769--wai-aria.netlify.app%2Findex.html)


.github/workflows/roleInfo.yml
```diff
@@ -3,7 +3,7 @@ name: roleInfo update
 on:
   push:
     branches:
-      - main
+      - gh-pages
```
pkra, line 6: I don't know how to test this without merging. Also, this may help with or may be blocked by #2768.

common/script/roleInfo.js
```diff
@@ -862,13 +862,6 @@
         disallowed: false,
         deprecated: false,
       },
```
pkra, line 864: This diff is worse than my local diff was. Reviewers may want to try to paste this version into their local editor for diffing.

There are a few simple de-duplications as well as a few larger groups (gridcell changes are the most "confusing" part).

spectranaut:
Hoping to take a look at this next week!

.github/workflows/roleInfoCheck.yml
```diff
@@ -17,6 +17,6 @@ jobs:
         with:
           node-version: "latest"
       - run: npm i linkedom prettier@3.6.0
-      - run: node ./common/script/buildRoleInfo.js > ./common/script/roleInfo.js
+      - run: node ./common/script/buildRoleInfo.js
```
daniel-montalvo, line 21: I am not clear how this check would work. In PRs you don't have the respec esported spec, previews are built at run time.

.github/workflows/roleInfo.yml
daniel-montalvo: Only way would be through setting up a fork

common/script/buildRoleInfo.js
```diff
@@ -1,17 +1,153 @@
...
+import fs from "node:fs";
+import path from "node:path";
+
 import { parseHTML } from "linkedom";
+const __dirname = import.meta.dirname;
+
+// NOTE: this script expects compiled respec output (e.g. gh-pages branch). 
+// If you are working on this file, you will need to run, e.g. get a copy from gh-pages or
+// run `$ npx respec --src index.html --out index.html`
+const inputFilename = path.resolve('index.html');
```
daniel-montalvo, line 10: I was just wondering. Wouldn't it be cleaner to have this scrit to the respec export itself? This way this could be run on PRs too.

.github/workflows/roleInfo.yml
pkra: I thought this is the cleanest way: when a version is published to github, we update roleinfo.

.github/workflows/roleInfoCheck.yml
pkra: Oh! You're right. My brain clearly stopped working at some point.

I'll fix it.

common/script/buildRoleInfo.js
pkra: My motivation was to separate things from respec.

daniel-montalvo: Not feeling strongly about this one

common/script/buildRoleInfo.js
```diff
@@ -1,17 +1,153 @@
...
+    value.fragID = key; //TODO: [minor] is this duplication really worth it? Role names should be valid IDREFS, no?
+    value.parentRoles = [];
+    roleSection.querySelectorAll('.role-parent .role-reference').forEach(node => value.parentRoles.push(node.textContent));
+    value.localprops = [];
+
+    roleSection.querySelectorAll(':is(.role-required-properties, .role-properties, .role-disallowed) :is(.property-reference, .state-reference)').forEach(
+        link => {
+            const name = link.textContent.split(' ')[0]; //TODO: hack because roletype has (state) inside link (as the only role to have that), cf. TODO: in ariaPreprocessing.js
+            const prop = structuredClone(statesAndProps[name]);
+            if (key !== 'roletype') prop.deprecated = false; // TODO: should roletype have deprecated=true when the spec lists everything as supported?
```
spectranaut, line 48: Can you explain this comment/situation to me... probably a silly question but I'm having trouble wrapping around why is deprecated always set to false, when there are entries in roleInfo.js where deprecated is true?

common/script/buildRoleInfo.js
```diff
@@ -1,17 +1,153 @@
...
+    if (currentyBroken.indexOf(key) > -1) return; //TODO: remove this HACK (cf. above)
 
-const script = fs.readFileSync("./common/script/aria.js").toString();
-const prescript = fs.readFileSync("./common/script/ariaPreprocessing.js").toString();
+    value.allprops = structuredClone(value.localprops); //TODO: why do we duplicate them? Does ariaChild.js need this duplication? (I understand its "allprops" but just "inherited" seems cleaner.)
+    roleSection.querySelectorAll('.role-inherited :is(.property-reference, .state-reference)').forEach(
+        link => {
+            const name = link.textContent.split(' ')[0]; //TODO: hack because roletype has (state) inside link (as the only role to have that), cf. TODO: in ariaPreprocessing.js
+            const prop = structuredClone(statesAndProps[name]);
+            if (localPropNames.indexOf(prop.name) > -1) prop.deprecated = false; // NOTE: ignores the fact that localProps may be disallowed (which is ok at time of writing)
```
spectranaut: also confused about this line... so if this state or property is inherited but listed as deprecated, and it exists in the localProps (not inherited) list, then you make it as not deprecated?

doesn't this mean that the state or property already exists in `value.allprops` and and shouldn't be added again?

and now -- I just ran this code (against respected index.html) and it doesn't seem like this if statement is every triggered?

spectranaut (approved):
Ok, I just read through this all and ran it locally, these changes overall seem fine! Lolcry at the hacks though. Are you thinking of removing "currentlyBroken" as a next step?

I remain confused about the extra logic for marking things are deprecated thought, see comments.

common/script/buildRoleInfo.js
pkra: If an attribute shows up in a characteristics table, then it's not deprecated on this role. 

(For `role=roletype`, roleInfo lists the 4 attributes "deprecated as global in 1.2" as `deprecated: true` - hence the TODO).

pkra: Oh, I should add: I'm not saying this is correct (looking again now, I think it's not but I haven't had enough coffee yet). 

I'm just trying to reproduce roleInfo.js as much as possible right now, with code that's hopefully easier to reason about. (And if we have identified a bug here, then I'll take that as a win 🥳 )

common/script/buildRoleInfo.js
pkra: Thanks! I think you're right - this line is no longer necessary. 

I think this is leftover from an older version. It tries to deal with the same issue as your earlier comment: if an attribute is listed in the characteristics table, it's not deprecated on the role.

pkra:
@spectranaut 
> Are you thinking of removing "currentlyBroken" as a next step?

Yes. I hope to discuss it on today's editors call.

.github/workflows/roleInfoCheck.yml
pkra: @daniel-montalvo I've updated this to run respec before buildRoleInfo (and resetting index.html before diffing roleInfo.js)

pkra:
Discussed in the last editors' call https://www.w3.org/2026/05/13-aria-editors-minutes.html#e0aa

pkra:
@spectranaut as discussed I've removed the "main" hack and regenerated roleInfo.js.

spectranaut (approved):
looks good to me!

.github/workflows/roleInfoCheck.yml
daniel-montalvo: lgtm -- I'd say I had put a thumbs up here some days ago but it seems I didn't.

.github/workflows/roleInfo.yml
pkra: @daniel-montalvo I think I misunderstood you in my previous comment. If you think this needs a test on a fork, let me know.

.github/workflows/roleInfoCheck.yml
pkra: @daniel-montalvo I had requested a re-review. It sounds like github's UI does not surface that information accessibly - I'll keep that in mind going forward.

.github/workflows/roleInfo.yml
```diff
@@ -12,15 +12,18 @@ jobs:
...
       - name: Use Node.js
         uses: actions/setup-node@v6
         with:
           node-version: "latest"
       - run: npm i linkedom prettier@3.6.0
-      - run: node ./common/script/buildRoleInfo.js > ./common/script/roleInfo.js
+      - run: git checkout gh-pages index.html # checkout (respec'ed) aria spec from gh-pages (WARNING: changes get staged!)
+      - run: node ./common/script/buildRoleInfo.js  # build roleInfo.js
       - run: npx prettier --write --print-width 200 ./common/script/roleInfo.js
+      - run: git reset index.html # reset spec source (to avoid comittting, cf. wanrning above)
```
daniel-montalvo:
```suggestion
      - run: git reset index.html # reset spec source (to avoid comittting, cf. warning above)
```

.github/workflows/roleInfoCheck.yml
```diff
@@ -16,7 +16,9 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version: "latest"
```
daniel-montalvo:
```suggestion
          node-version: "lts/*"
```
I have seen the pr preview workflow fail in latest versions, so I'd figure we should also err on the side of caution here

.github/workflows/roleInfoCheck.yml
```diff
@@ -16,7 +16,9 @@ jobs:
...
-      - run: npm i linkedom prettier@3.6.0
-      - run: node ./common/script/buildRoleInfo.js > ./common/script/roleInfo.js
+      - run: npm i linkedom prettier@3.6.0 respec
+      - run: npx respec --src index.html --out index.html
```
daniel-montalvo:
```suggestion
       - run: npx respec --src index.html --out index.html --localhost
```

I think if we don't use `--localhost` the respec imports will fail. The document still builds and I don't think the funders are needed for this, but just flagging this in case you want to consider this.

common/script/buildRoleInfo.js
```diff
@@ -1,17 +1,74 @@
...
+
+const roleInfo = {};
+
+
+const generateRoleInfoEntry = (roleSection) => {
+    const key = roleSection.id;
+    const value = {};
+    roleInfo[key] = value;
+    value.name = key;
+    value.fragID = key; //TODO: [minor] is this duplication really worth it? Role names should be valid IDREFS, no?
```
daniel-montalvo, line 39: +1 to this being an unnecessary duplication. Would be good to know if there's people relying on this already, otherwise should probably just remove

pkra: Thanks. I would wait until ariaChild.js can be refactored.

pkra:
@daniel-montalvo I've merged your suggestions -- thank you! Do you think this can be merged now?

.github/workflows/roleInfo.yml
daniel-montalvo: No worries. Yes, we may want to merge here and see what happens to avoid complications, but the only way to fully test this would be through a fork for the actions to run properly.

pkra: Thanks, Daniel!

PR #2771, opened 2026-04-20
Update definition of drop-down vs listbox rendering

josepharhar:
HTML's definition of how the select element is rendered as a listbox vs drop-down box has changed, so we should link to the more specific definition instead of repeating what makes a select a drop-down box vs a listbox.

Closes #????

* [x] For every spec that this PR edits, please add the appropriate `spec:<spec_name>` label. If you don't have privileges to do this, editors will do it for you.
* [x] If the change is [editorial](https://github.com/w3c/aria/blob/main/documentation/process.md#editorial-changes), please add "Editorial:" at the start of your PR name, and delete the "Test, Documentation and Implementation tracking" section below.

* [ ] "author MUST" tests:
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=309606
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1988022
   * Blink: https://issues.chromium.org/issues/439964654
* [ ] ACT review?
* [ ] Does this need AT implementations? I don't think so.
* [ ] Related APG Issue/PR: I'm not sure that the APG needs to be updated for this?
* [ ] MDN Issue/PR: There are existing MDN issues for this change [here](https://github.com/mdn/content/issues/43823) and [here](https://github.com/mdn/content/issues/43403). Are they good enough?


josepharhar:
Very similar html-aria PR: https://github.com/w3c/html-aria/pull/555

keithamus (approved):
This is a useful change. The data below still mentions the attributes, but they align with the display modes. Perhaps we can also modify those? (So L6349 & L6403)

spectranaut (commented):
lgtm, thanks!

josepharhar:
> This is a useful change. The data below still mentions the attributes, but they align with the display modes. Perhaps we can also modify those? (So L6349 & L6403)

Done, thank you!

keithamus (approved):

PR #2778, opened 2026-04-29
[html-aam] Add focusgroup attribute mapping

### LINKED ISSUE #2602: Scoping focusgroup to scenarios defined by aria roles
[Scoped Focus group explainer on open ui](https://github.com/openui/open-ui/pull/1260)

While this explainer/proposal is for an HTML attribute I wanted to discuss this in ARIA to get the group's thoughts on the feature.
I will not get into all the specifics of the feature - that's why I linked the explainer - but I want to ensure that this new behavior is correctly scoped, encouraging developers to use this in a way that works well with keyboard and assistive technology users, improving their experience rather than frustrating them.

I specifically wanted to discuss the scoping mechanism -
The current explainer outlines a couple of different approaches to scope usage to a specific set of scenarios, but I don't have strong opinions on how we scope, just that we are taking steps to ensure we are encouraging developers to do the right thing, e.g. paving the cow paths, rather than encouraging this powerful attribute to be used in any way.

The Explainer's current scoping mechanism is to have `focusgroup` take a dependency on the role of the element, and the attribute is designed to not function when a supported role is not present. This breaks the expectation that aria is used for semantics, not behavior, so an [alternative scoping mechanism](https://github.com/janewman/open-ui/blob/scoped-focusgroup/site/src/pages/components/scoped-focusgroup.explainer.mdx#alternative-scoping-mechanism) is described that flips the script, with `focusgroup` instead informing the element's role. This allows us to keep the requirement without breaking norms.

I would love to hear what this group thinks of the various approaches outlined, as well as gather thoughts about alternative approaches that fit the needs this Explainer raises.

janewman:
Companion to whatwg/html#11723.

Closes #2602

Adds an `att-focusgroup` entry to html-aam documenting the mapping for the `focusgroup` content attribute being specified in [whatwg/html#11723](https://github.com/whatwg/html/pull/11723). The attribute itself has no direct accessibility API mapping, so all six platform rows are `Not mapped`. The Comments cell explains the details.

* [ ] "author MUST" tests:
* [x] "user agent MUST" tests: https://wpt.live/html/interaction/focus/focusgroup/tentative/
   * WebKit: No Implementation, [Standards Position](https://github.com/WebKit/standards-positions/issues/171)
   * Gecko: No Implementation, [Standards Position](https://github.com/mozilla/standards-positions/issues/1348)
   * Blink: https://issues.chromium.org/issues/40210717
* [ ] ACT review?
* [ ] Does this need AT implementations? No, the feature is designed to match existing patterns on the web today, moving authors away from a javascript-based implementation towards a user-agent provided one.
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

### DIFF
--- a/html-aam/index.html
+++ b/html-aam/index.html
@@ -11103,6 +11103,119 @@
             </tr>
           </tbody>
         </table>
+        <h4 id="att-focusgroup">`focusgroup`</h4>
+        <table class="data" aria-labelledby="att-focusgroup">
+          <tbody>
+            <tr>
+              <th>HTML Specification</th>
+              <td>`focusgroup`</td>
+            </tr>
+            <tr>
+              <th>Element(s)</th>
+              <td>
+                <a data-cite="html/interaction.html#attr-focusgroup">HTML elements</a>
+              </td>
+            </tr>
+            <tr>
+              <th>[[WAI-ARIA-1.2]]</th>
+              <td>
+                <div class="general">Not mapped</div>
+              </td>
+            </tr>
+            <tr>
+              <th>
+                <a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a>
+              </th>
+              <td>
+                <div class="general">Not mapped</div>
+              </td>
+            </tr>
+            <tr>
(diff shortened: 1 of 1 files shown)


spectranaut:
@scottaohara from triage in the meeting today, @janewman is not sure this is necessary?

janewman:
HTML folks are all feeling this is the right home for this, let me know what you folks think of this change, especially the format.

PR #2781, opened 2026-04-30
Add support for role-namerequired-caveat

### LINKED ISSUE #2767: Clarify when listbox role should require accessible name
The `listbox` role currently says that it requires an accessible name, which is a bit strict.

For listboxes that appear as inline form controls with other content, it absolutely makes sense for a name to be required - and that name should likely be derived from its visible label.

However, for listboxes that are invoked as popups (e.g., from comboboxes), naming likely shouldn't be required, let alone necessary.  Assuming the invoking element has been given an accessible name, the user will already have context to the purpose of the listbox.

MelSumner:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2781--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2781--wai-aria.netlify.app%2Findex.html)

Closes #2767 (maybe)

So examining the issue, I think that we will need to expand this a bit to include caveats. @pkra hinted as much in discussion, so for now this is a draft with a potential implementation of both support for caveats and the `listbox` caveat in particular.

I'll update below once we've examined and discussed this approach. 

* [ ] "author MUST" tests:
   * WebKit:
   * Gecko:
   * Blink:
* [ ] ACT review?
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:

### DIFF
--- a/common/script/aria.js
+++ b/common/script/aria.js
@@ -133,6 +133,25 @@ const buildStatesProperties = function (item) {
 };
 
 /**
+ * Parse a role-namerequired cell while excluding caveat span text.
+ * @param {HTMLTableCellElement} cell
+ * @returns {{ primary: string, caveat: string, combined: string }}
+ */
+const parseNameRequiredCell = (cell) => {
+  const caveatNode = cell?.querySelector("span.role-namerequired-caveat");
+  const caveat = (caveatNode?.innerText || "").trim();
+  const clone = cell?.cloneNode(true);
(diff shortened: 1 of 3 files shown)


pkra:
@MelSumner this will likely clash with https://github.com/w3c/aria/pull/2769 and other work planned in #2501. I would suggest to revisit this later.

PR #2798, opened 2026-05-27
editorial: Fix the role attribute definition

jnurthen:
🚀 **Netlify Preview**:
🔄 **this PR updates the following sspecs**:
- [ARIA preview](https://deploy-preview-2798--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2798--wai-aria.netlify.app%2Findex.html)

Closes #2145, 

export attr-role, MUST allow, glossary data-noexport, role-attribute term linking.
related work: [whatwg/html#6890](https://github.com/whatwg/html/issues/6890).

And a few other todo items (delete this section after performing them):
* [x] For every spec that this PR edits, please add the appropriate `spec:<spec_name>` label. If you don't have privileges to do this, editors will do it for you.
* [ ] If the change is [editorial](https://github.com/w3c/aria/blob/main/documentation/process.md#editorial-changes), please add "Editorial:" at the start of your PR name, and delete the "Test, Documentation and Implementation tracking" section below.

# Test, Documentation and Implementation tracking
Once this PR has been reviewed and has consensus from the working group, tests should be written and issues should be opened on browsers. Add N/A and check when not applicable.

* [ ] "author MUST" tests: N/A
* [ ] "user agent MUST" tests: N/A
* [ ] Browser implementations (link to issue or commit):
   * WebKit: no change needed
   * Gecko: no change needed
   * Blink: no change needed
* [ ] ACT review? N/A
* [ ] Does this need AT implementations? N/A
* [ ] Related APG Issue/PR: N/A
* [ ] MDN Issue/PR: N/A

PR #2799, opened 2026-05-27
Chore: Ensure system chrome  is detected by puppeteer and stops using `--disable-sandbox`

jnurthen:
Builds are failing - apparently this is safe within a github action

### DIFF
--- a/.github/workflows/roleInfoCheck.yml
+++ b/.github/workflows/roleInfoCheck.yml
@@ -10,6 +10,8 @@ on:
 jobs:
   roleInfo:
     runs-on: ubuntu-latest
+    env:
+      PUPPETEER_EXECUTABLE_PATH: '/usr/bin/google-chrome'
     steps:
       - uses: actions/checkout@v6
       - name: Use Node.js
@@ -17,7 +19,7 @@ jobs:
         with:
           node-version: "lts/*"
       - run: npm i linkedom prettier@3.6.0 respec
-      - run: npx respec --src index.html --out index.html --localhost --no-sandbox
+      - run: npx respec --src index.html --out index.html --localhost
       - run: node ./common/script/buildRoleInfo.js
       - run: npx prettier --write --print-width 200 ./common/script/roleInfo.js
       - run: git checkout index.html
--- a/.github/workflows/roleInfoCheck.yml
+++ b/.github/workflows/roleInfoCheck.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           node-version: "lts/*"
       - run: npm i linkedom prettier@3.6.0 respec
-      - run: npx respec --src index.html --out index.html --localhost
+      - run: npx respec --src index.html --out index.html --localhost --no-sandbox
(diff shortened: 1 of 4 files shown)


.github/workflows/roleInfoCheck.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         with:
           node-version: "lts/*"
       - run: npm i linkedom prettier@3.6.0 respec
-      - run: npx respec --src index.html --out index.html --localhost
+      - run: npx respec --src index.html --out index.html --localhost --no-sandbox
```
pkra: Has anybody checked why respec suddenly needs this?

Also, which builds are failing?

jnurthen:
https://github.com/w3c/aria/pull/2798 is failing

.github/workflows/roleInfoCheck.yml
jnurthen: This was a suggested fix - I was concerned about security of a no-sandbox but from reading it appears that within a github action its kind of sandboxed already so it may not be an issue. However I'd be very happy to find a different solution

daniel-montalvo:
> #2798 is failing

Because puppeteer is not able to find the browser in  its cached directory itseems. Exporting the environment variable  to point to the system browsers seems to fix it.

I added the --disabled-sandbox to the preview scripts a while ago but I am really not sure if it was an ubuntu update which prevented the preview to run or something else. I'll probably have a look at that later.

.github/workflows/roleInfoCheck.yml
jnurthen: ok - lets do the same thing here rather than adding the --no-sandbox

jnurthen:
Thanks @daniel-montalvo — the system Chrome approach looks like a better fix than --no-sandbox. I notice roleInfoCheck.yml still has the --no-sandbox from my original commit. Should we apply the same PUPPETEER_EXECUTABLE_PATH detection there too, so both workflows are consistent and we can drop --no-sandbox entirely?

daniel-montalvo:
> Thanks @daniel-montalvo — the system Chrome approach looks like a better fix than --no-sandbox. I notice roleInfoCheck.yml still has the --no-sandbox from my original commit. Should we apply the same PUPPETEER_EXECUTABLE_PATH detection there too, so both workflows are consistent and we can drop --no-sandbox entirely?


Yes, but I think there's a more elegant way to do this by including these environment variables at the top of the workflow file, avoiding these convoluted "if". I'll take a pass at this PR and make the necessary changes here.

daniel-montalvo:
After taking a closer look at this I've made the following changes:

- Export the PUPPETEER_EXECUTABLE_PATH environment variable to ensure puppeteer detects the system chrome,
- ensure all workflows run node lts for better stability, and
- consistently install required dependencies (including respec).

daniel-montalvo:
@pkra @jnurthen  I will merge this, because again the previews were not working across the repo, and this seems to at least momentarily fix them.

PR #2800, opened 2026-05-28
Core-AAM: Add Computed Accessibility Properties section

spectranaut:
🚀 **Netlify Preview**:
- [core-aam preview](https://deploy-preview-2800--wai-aria.netlify.app/core-aam/index.html) &mdash; [core-aam diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fcore-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-2800--wai-aria.netlify.app%2Fcore-aam%2Findex.html)


core-aam/index.html
```diff
@@ -6469,6 +6516,12 @@ <h4 id="ariaActiveDescendant"><code>aria-activedescendant</code></h4>
                   <a class="property-reference" href="#aria-activedescendant"><code>aria-activedescendant</code></a>
                 </td>
               </tr>
+              <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  Not mapped, or, referenced element should have `focus: true` ?
```
cookiecrook, line 6555: does focus here mean focused or focusable?

core-aam/index.html
```diff
@@ -6506,6 +6559,12 @@ <h4 id="ariaAtomicTrue"><code>aria-atomic</code>=<code>true</code></h4>
                   >=<code>true</code>
                 </td>
               </tr>
+              <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  <code>atomic: true</code>
```
cookiecrook, line 6598: I think we should TBD this one... Not sure every ARIA attr should flow into props.

core-aam/index.html
```diff
@@ -6564,6 +6623,12 @@ <h4 id="ariaAtomicFalse"><code>aria-atomic</code>=<code>false</code></h4>
                   >=<code>false</code>
                 </td>
               </tr>
+              <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  <code>atomic: false</code>
```
cookiecrook, line 6662: ditto

core-aam/index.html
```diff
@@ -6670,6 +6741,12 @@ <h4 id="ariaAutocompleteNone"><code>aria-autocomplete</code>=<code>none</code></
                   >=<code>none</code>
                 </td>
               </tr>
+              <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  Do not expose <code>autocomplete</code>.
```
cookiecrook, line 6780: Let's treat this as an Interop/WICG issues, too. No need to land all of them at once.

core-aam/index.html
spectranaut: removed this one for now!

spectranaut:
Hi @jcsteh, @cookiecrook, @rahimabdi, @lucasradaelli -- this is ready for review based on the feedback from yesterday's meeting. [Here is a link to the added section](https://deploy-preview-2800--wai-aria.netlify.app/core-aam/#computed_accessibility_properties)

I only included "Computed Accessibility Properties" for the following rows:
* aria-description
* aria-checked
* aria-pressed
* aria-label
* aria-labelledby
* aria-level
* aria-multiline
* aria-orientation
* aria-required
* aria-roledecription
* aria-selected

Let me know what you think of this list, we can remove or add things!

core-aam/index.html
cookiecrook: Not removed? "referenced element should have `focus: true`"

core-aam/index.html
cookiecrook: ditto. doesn't seem resolved.

core-aam/index.html
cookiecrook: All these are marked as resolved but I don't see a diff that would resolve them???

core-aam/index.html
cookiecrook: Nm... the comment was not obvious from prior diff I was looking at. Can the value be commented instead of the row?

spectranaut: Oh is that what we discussed? Do you think it's less odd to have it table cell empty rather than the whole row missing? I suppose I don't have a strong opinion.

core-aam/index.html
```diff
@@ -6445,6 +6445,42 @@ <h2>General rules</h2>
             </li>
           </ol>
         </section>
+        <section id="computed_accessibility_properties">
+          <h2>Computed Accessibility Properties</h2>
+          <p>
+            The computed accessibility properties of an element are a set of cross-platform key/value pairs of strings that represent the states and attributes of any <a href="#dfn-accessible-object" class="specref">accessible object</a> in the <a>accessibility tree</a>. The computed accessibility properties are used primarily for the purposes of developer tools, specification conformance, and interoperability testing.
```
jcsteh: Should we go with "platform independent" here rather than "cross-platform" to be consistent with the WebDriver wording? Personally, I'm happy with either and tend to use them as synonyms, but I believe @cookiecrook objected to cross-platform?

core-aam/index.html
```diff
@@ -6445,6 +6445,42 @@ <h2>General rules</h2>
...
+          <p>
+            Not all ARIA states and attributes will be surfaced in the computed accessibility properties. To see if a state or attribute should be surfaced, and the name of the key and it's expected value, check the "Computed Accessibility Property" row of the tables below. If a host language state or property maps to an ARIA state or property, use the key and value specified here.
+          </p>
+          <aside class="example">
+            <p>For both of the following examples, the computed accessibility properties will include `checked: true`.</p><pre>
+&lt;div role=checkbox aria-checked=true&gt;&lt;/div&gt;
+&lt;input type="checkbox" checked /&gt;</pre>
+            <p>
+              For both of the following examples, the computed accessibilty properties will include `multiselectable: true`.</p><pre>
+&lt;div role=listbox aria-multiselectable&gt;...&lt;/div&gt;
```
jcsteh: I think this needs to be `aria-multiselectable="true"`? (In contrast, `multiple` is correct below.)

core-aam/index.html
```diff
@@ -6445,6 +6445,42 @@ <h2>General rules</h2>
...
+&lt;select multiple&gt;...&lt;/select&gt;</pre>
+          </aside>
+          <p>
+             When the state or property is required on a role but the attribute is missing, undefined, or set to a invalid token, user agents MUST expose the property with the appropriate <a href="#authorErrorDefaultValuesTable" class="specref">fallback values for missing required attributes</a>. When a state or property has a <a href="#implictValueForRole" class="specref">implicit value for role</a> and the attribute is missing, undefined, or set to a invalid token, user agents MUST expose the default value for the role.
```
jcsteh: nit:
```suggestion
             When the state or property is required on a role but the attribute is missing, undefined, or set to an invalid token, user agents MUST expose the property with the appropriate <a href="#authorErrorDefaultValuesTable" class="specref">fallback values for missing required attributes</a>. When a state or property has a <a href="#implictValueForRole" class="specref">implicit value for role</a> and the attribute is missing, undefined, or set to an invalid token, user agents MUST expose the default value for the role.
```

core-aam/index.html
```diff
@@ -6445,6 +6445,42 @@ <h2>General rules</h2>
...
+          </p>
+          <aside class="example">
+            <ol>
+              <li><pre>&lt;div role=checkbox&gt;&lt;/div&gt;</pre> <a href="#aria-checked" class="specref">aria-checked</a> on a <a href="#checkbox" class="specref">checkbox</a> has the fallback value of `false`, therefore the computed accessibility value would be: `checked: false`.</li>
+              <li><pre>&lt;div role=tab&gt;&lt;/div&gt;</pre> <a href="#aria-selected" class="specref">`selected`</a> on a <a href="#tab" class="specref">tab</a> has the implicit value of `false`, therefore the computed accessibility property will be `selected: false`.</li>
+            </ol>
+          </aside>
+          <p>
+            When the state or property does not have a default value other than undefined, is not required on a given role, and is missing, undefined, or set to a invalid token, user agents MUST NOT include it the computed accessibility properties. When a state or property has a default value other than undefined and is supported but not required on a given role, user agents MAY include it in the computed accessibility properties.
```
jcsteh: nit:
```suggestion
            When the state or property does not have a default value other than undefined, is not required on a given role, and is missing, undefined, or set to an invalid token, user agents MUST NOT include it in the computed accessibility properties. When a state or property has a default value other than undefined and is supported but not required on a given role, user agents MAY include it in the computed accessibility properties.
```

jcsteh: An alternative (or addition) to the "MAY" case here is something like:

"When a state or property has a default value other than undefined, is supported but not required on a given role and is not mapped on all platforms, user agents MUST NOT include it in the computed accessibility properties."

This would mean that we have a clear rule for aria-required="false", for example: we do not include it in the property bag.

That said, I see you already have "Do not expose required" for aria-required="false". Does that already fit one of the rules here and I'm just missing it?

core-aam/index.html
spectranaut: Whoops! thanks for catching.

core-aam/index.html
spectranaut: Hmm.. right. So below, in the tables, when something is not mapped, it seems to make sense to write "Do not expose".  I guess, without thinking about it too clearly, it seemed like I didn't have to put the case you suggested into the prose because it would always be covered below.

But I suppose it's fine to be overly pedantic, this is a technical specification, I'm going to include your sentence (and the "may") for now.

core-aam/index.html
```diff
@@ -6445,6 +6445,42 @@ <h2>General rules</h2>
...
+              <li><pre>&lt;div role=tab&gt;&lt;/div&gt;</pre> <a href="#aria-selected" class="specref">`selected`</a> on a <a href="#tab" class="specref">tab</a> has the implicit value of `false`, therefore the computed accessibility property will be `selected: false`.</li>
+            </ol>
+          </aside>
+          <p>
+            When the state or property does not have a default value other than undefined, is not required on a given role, and is missing, undefined, or set to an invalid token, user agents MUST NOT include it in the computed accessibility properties. When a state or property has a default value other than undefined, is supported but not required on a given role, and is not mapped on all platforms, user agents MUST NOT include it in the computed accessibility properties. If the state or property with the previous conditions is mapped on one or more platforms, user agents MAY include it in the computed accessibility properties.
```
lolaodelola: > When a state or property has a default value other than undefined, is supported but not required on a given role, and is not mapped on all platforms, user agents MUST NOT include it in the computed accessibility properties. If the state or property with the previous conditions is mapped on one or more platforms, user agents MAY include it in the computed accessibility properties.

The "is not mapped on all platforms" and "is mapped on one or more platforms" reads as contradictory to me. A state/role can be mapped on one out of four platforms and meet both these conditions. What about:


```suggestion
            When the state or property does not have a default value other than undefined, is not required on a given role, and is missing, undefined, or set to an invalid token, user agents MUST NOT include it in the computed accessibility properties. When a state or property has a default value other than undefined, is supported but not required on a given role, and is not mapped on any platforms, user agents MUST NOT include it in the computed accessibility properties. If the state or property with the previous conditions is mapped on one or more platforms, user agents MAY include it in the computed accessibility properties.
```

core-aam/index.html
cookiecrook: I recalled ReSpec removed (or can be configured to remove) the empty rows. Am I mistaken?

cookiecrook:
> Hi @jcsteh, @cookiecrook, @rahimabdi, @lucasradaelli -- this is ready for review based on the feedback from yesterday's meeting. [Here is a link to the added section](https://deploy-preview-2800--wai-aria.netlify.app/core-aam/#computed_accessibility_properties)
> 
> I only included "Computed Accessibility Properties" for the following rows:
> 
> * aria-description
> * aria-checked
> * aria-pressed
> * aria-label
> * aria-labelledby
> * aria-level
> * aria-multiline
> * aria-orientation
> * aria-required
> * aria-roledecription
> * aria-selected
> 
> Let me know what you think of this list, we can remove or add things!

Thoughts individually.

- **level**: Should we add a clarifying note on some of the ones that have known correlate reconciliation with the host languages? For example, `<h3>` gets a `level` of `3` event though the `ariaLevel` IDL reflection should be… null for now unless it changes to Integer, and then would be `0` I think? 
- **required**: note to cross-ref `html:required`?
- **multiline**: This is another we should reconsider. ARIA's choice to use `textbox` and `textbox multiline` instead of the existing `input type=text` and `textarea` was weird, especially in the context of `contenteditable`. IIRC, this was one I raised as a problem early as an ARIA editor, but [corp rep] didn't want to pay [AT dev] to change the implementation, so we left it in the spec. Had I been a more experienced editor at the time, I may have pushed back more firmly.
- **selected**: We should discuss the name of this one in the context of ambiguity. "selected" can be confused with text range selection, so maybe we want to disambiguate the name further? Also, some implementations don't make as much of a distinction between checked, pressed, and selected in the API... I'm not opposed to this one as-is. Jut trying to anticipate future problems.

The rest look good to me. Thanks for trimming the big list down and proceeding more deliberately.

core-aam/index.html
spectranaut: perfect catch, thanks, applying.

spectranaut:
> * **level**: Should we add a clarifying note on some of the ones that have known correlate reconciliation with the host languages? For example, `<h3>` gets a `level` of `3` event though the `ariaLevel` IDL reflection should be… null for now unless it changes to Integer, and then would be `0` I think?

I don't think we need to make this note in the table... however, we could say something about this in the definition of computed accessibility properties, perhaps adding this after the first sentence (first sentence included):

`The computed accessibility properties of an element are a set of platform independent key/value pairs of strings that represent the states and attributes of any accessible object in the accessibility tree. Many different things affect the value of the key in the accessibility tree, including but not limited to: defaults values for the state or attribute, default values when used on roles, host language states and attributes, and whether the host language overrides the ARIA values. These values do not always align with IDL reflection of the aria state or attribute. `

No objections to keeping this in the list, right?

> * **required**: note to cross-ref `html:required`?

I don't think we should cross ref to html required, because html required already references Core-AAM: https://w3c.github.io/html-aam/#att-required

No objections to keeping this in the list, right?

> 
> * **multiline**: This is another we should reconsider. ARIA's choice to use `textbox` and `textbox multiline` instead of the existing `input type=text` and `textarea` was weird, especially in the context of `contenteditable`. IIRC, this was one I raised as a problem early as an ARIA editor, but [corp rep] didn't want to pay [AT dev] to change the implementation, so we left it in the spec. Had I been a more experienced editor at the time, I may have pushed back more firmly.

Interesting history :) I'm fine with leaving this one out. But I think that means I need to update the example that uses it....

> * **selected**: We should discuss the name of this one in the context of ambiguity. "selected" can be confused with text range selection, so maybe we want to disambiguate the name further? Also, some implementations don't make as much of a distinction between checked, pressed, and selected in the API... I'm not opposed to this one as-is. Jut trying to anticipate future problems.

I would like to leave this one it, but I'm totally fine if you think we should drop it for now. Chrome does differentiate between select and checked/pressed, it's just checked/pressed that are combined, and we are already surfacing that information. Since we are already sorting out the checked/pressed problem, and we might as well deal with all three?

As for the name, if we wanted to do something different then selected... I'm not sure, the whole web does have this name collision, not just accessibility... I wonder how it's dealt with in other contexts? I feel like if we want to surface selected text we should say "selectedText".

core-aam/index.html
spectranaut: I just tried this, when the second column is empty, it does not remove the row, and I didn't see any settings related to this in the respec documentation.

I don't think I like the empty column look, so I'd like to stick to commenting out the row. Alternatively, we can write something like "Not specified." or "Intentionally not specified."

core-aam/index.html
```diff
@@ -6445,6 +6445,38 @@ <h2>General rules</h2>
...
           </ol>
         </section>
+        <section id="computed_accessibility_properties">
+          <h2>Computed Accessibility Properties</h2>
+          <p>
+            The computed accessibility properties of an element are a set of platform independent key/value pairs of strings that represent the states and attributes of any <a href="#dfn-accessible-object" class="specref">accessible object</a> in the <a>accessibility tree</a>. The computed accessibility properties are used primarily for the purposes of developer tools, specification conformance, and interoperability testing.
+          </p>
+          <p>
+            Not all ARIA states and attributes will be surfaced in the computed accessibility properties. To see if a state or attribute should be surfaced, and the name of the key and it's expected value, check the "Computed Accessibility Property" row of the tables below. If a host language state or property maps to an ARIA state or property, use the key and value specified here.
+          </p>
```
cookiecrook: Okay @spectranaut, let's go with the shorter list you narrowed it down to. The following is your prose addition suggestion, with a few more editorial changes.

```suggestion
          <p>
            The computed accessibility properties of an element are a set of platform-independent and UA-independent key/value pairs that represent the properties of any <a href="#dfn-accessible-object" class="specref">accessible object</a> in the <a>browser accessibility tree</a>. The computed accessibility properties are used primarily for the purposes of developer tools, specification conformance, and interoperability testing.
          </p>
          <p>
            Not all ARIA states and attributes will be surfaced in the computed accessibility properties. To see if a state or attribute should be surfaced, and the name of the key and it's expected value, check the "Computed Accessibility Property" row of the tables below. If a host language state or property maps to an ARIA attribute, use the key and value specified here. However, the final value in the accessibility tree can be affected by multiple things, including but not limited to the following: default IDL values for the DOM attribute, default ARIA values when used on specific ARIA roles, and host language features with conflicting semantics, depending on whether the host language overrides the ARIA value. These final computed properties values do not always align with IDL reflection of the ARIA attribute.
          </p>
          
          
          
          
```

cookiecrook (approved):

core-aam/index.html
```diff
@@ -6445,6 +6445,38 @@ <h2>General rules</h2>
             </li>
           </ol>
         </section>
+        <section id="computed_accessibility_properties">
+          <h2>Computed Accessibility Properties</h2>
+          <p>
+            The computed accessibility properties of an element are a set of platform independent key/value pairs of strings that represent the states and attributes of any <a href="#dfn-accessible-object" class="specref">accessible object</a> in the <a>accessibility tree</a>. The computed accessibility properties are used primarily for the purposes of developer tools, specification conformance, and interoperability testing.
```
rahimabdi: > are a set of platform independent

Should this word be hyphenated? "platform-independent"

rahimabdi: > represent the states and attributes

What might be the distinction between states and attribute? Thinking of the existing state vs. attribute vs. property potential confusion 😄

core-aam/index.html
```diff
@@ -6445,6 +6445,38 @@ <h2>General rules</h2>
...
+              <li><pre>&lt;div role=tab&gt;&lt;/div&gt;</pre> <a href="#aria-selected" class="specref">`selected`</a> on a <a href="#tab" class="specref">tab</a> has the implicit value of `false`, therefore the computed accessibility property will be `selected: false`.</li>
+            </ol>
+          </aside>
+          <p>
+            When the state or property does not have a default value other than undefined, is not required on a given role, and is missing, undefined, or set to an invalid token, user agents MUST NOT include it in the computed accessibility properties. When a state or property has a default value other than undefined, is supported but not required on a given role, and is not mapped on any platforms, user agents MUST NOT include it in the computed accessibility properties. If the state or property with the previous conditions is mapped on one or more platforms, user agents MAY include it in the computed accessibility properties.
```
rahimabdi: > When the state or property does not have a default value other than undefined, is not required on a given role, and is missing, undefined, or set to an invalid token, user agents MUST NOT include it in the computed accessibility properties.

Does this mean that the value would be null (or empty string)? If someone checks `accProps.checked` in such a scenario for example, what would be the value of `checked` since it's not included?

core-aam/index.html
```diff
@@ -6445,6 +6445,38 @@ <h2>General rules</h2>
             </li>
           </ol>
         </section>
+        <section id="computed_accessibility_properties">
+          <h2>Computed Accessibility Properties</h2>
+          <p>
+            The computed accessibility properties of an element are a set of platform independent key/value pairs of strings that represent the states and attributes of any <a href="#dfn-accessible-object" class="specref">accessible object</a> in the <a>accessibility tree</a>. The computed accessibility properties are used primarily for the purposes of developer tools, specification conformance, and interoperability testing.
+          </p>
+          <p>
+            Not all ARIA states and attributes will be surfaced in the computed accessibility properties. To see if a state or attribute should be surfaced, and the name of the key and it's expected value, check the "Computed Accessibility Property" row of the tables below. If a host language state or property maps to an ARIA state or property, use the key and value specified here.
```
rahimabdi: Should the spec made clearer that there is a one-to-many mapping between computed accessibility property and ARIA attribute? For example, for `label`, it maps to `aria-label`, `aria-labelledby`, `title`, etc.

rahimabdi (approved):
LGTM with some minor comments, thank you @spectranaut!

I noticed some additional properties that weren't discussed in initial browser implementations:
- `description`
- `level`
- `multiline`
- `orientation`
- `required`
- `roledescription`
- `selected`

Also, `parent`/`children` appear to be missing.

core-aam/index.html
cyns: This is kind of a nit, but aren't the computed accessibility properties also used by browsers as the basis for building the platform-specific trees? (in addition to developer tools, specification conformance, and interoperability testing)

cyns (approved):
Overall, looks good! It's really helpful to have this layer documented. I added one small comment, ok with merging either way.

core-aam/index.html
spectranaut: taking suggestion as is and applying thanks!

core-aam/index.html
spectranaut: Hmmm I just had to research grammar rules to know whether this was the case, and it seems like yeah, it should be hyphenated, ty!

spectranaut: omg! thank you for catching this, it should have been "states and properties" everywhere, not "states and attributes". But also, on closer look at the ARIA spec language, I think I can just say "attributes" instead of "states and properties" -- so I'll make that switch, which should make it easier to read.

Edit: I'm no longer sure I can just use attribute. I think attribute might be too confusable with the actual aria attribute in the DOM (https://dom.spec.whatwg.org/#concept-attribute) where as states and properties have things like default values for roles and there for might not have an HTML attribute set. So, I'm reverting to using "states and properties", but adding "ARIA states and properties" for most of them.

In the case where it's clear we are talking about a DOM attribute, I left attribute.

I'd appreciate you taking another look with this change, however!

core-aam/index.html
spectranaut: In terms of a JSON object, the key "checked" would simply not be in the JSON object. Does that make sense? If you are talking about a javascript object, there is no "checked" attribute, so "accProps.checked" would result in undefined.

core-aam/index.html
spectranaut: I think this might be more clear now that I've applied James Craig's suggestion: https://github.com/w3c/aria/pull/2800#discussion_r3501379353

core-aam/index.html
spectranaut: Hmmm interesting point, it's true. I changed the wording here slightly with this reflection, from

> The computed accessibility properties **are used** primarily for the purposes of developer tools, specification conformance, and interoperability testing.

To:
> The computed accessibility properties **are exposed** primarily for the purposes of developer tools, specification conformance, and interoperability testing.

spectranaut:
I've converted this to draft because I'm working on moving the definition of accessibilityID, parent and child here per webdriver reviewer's request:  https://github.com/w3c/webdriver/pull/1960#discussion_r3507966410

spectranaut:
Ok, @cookiecrook @janewman @rahimabdi @cyns  - I'd really appreciate another look at this section: https://deploy-preview-2800--wai-aria.netlify.app/core-aam/index.html#computed_accessibility_properties

As mentioned above, the webdriver reviewer requested that the definition of these concepts be owned outside of webdriver, so that they can be referred to by webdriverBiDi as well. So here is my first attempt at doing that. I had to add a definition of "accessibility node" and "accessibility node ID" which I'm particularly unsure about, so review of those specifically would be appreciated.

core-aam/index.html
```diff
@@ -6445,6 +6445,81 @@ <h2>General rules</h2>
...
+          <p>
+            The computed accessibility properties are a set of platform-independent key/value pairs of strings that represent the states and properties of any <a href="#dfn-accessible-object" class="specref">accessible object</a> in the <a>accessibility tree</a>. The computed accessibility properties are exposed primarily for the purposes of developer tools, specification conformance, and interoperability testing. Because the term <a href="#dfn-accessible-object" class="specref">accessible object</a> is used to refer to a node in a specific platform accessibility API, and the computed accessibility properties are properties of an intermediate platform-independent, implementation-defined abstraction used to back the various platform APIs, the term <dfn>accessibility node</dfn> will be used to refer to a node with the computed accessibility properties defined here. The tree of <a>accessibility nodes</a> follows the same rules as the <a>accessibility tree</a> and it is an exactly parallel structure to the platform accessibility API trees except where the Core-AAM specifies a difference for a specific accessibility API.
```
lolaodelola: The third sentence is long and a little confusing, and ends up burying the lede a little. You've also already defined what the computed accessibility properties are in the first sentence. What about:

```suggestion
The computed accessibility properties are a set of platform-independent key/value pairs of strings that represent the states and properties of any <a href="#dfn-accessible-object" class="specref">accessible object</a> in the <a>accessibility tree</a>. The computed accessibility properties are exposed primarily for the purposes of developer tools, specification conformance, and interoperability testing.
The term <dfn>accessibility node</dfn> will be used to refer to a node with the computed accessibility properties defined here, as the term <a href="#dfn-accessible-object" class="specref">accessible object</a> is currently used to refer to a node in a specific platform accessibility API. 
The tree of <a>accessibility nodes</a> follows the same rules as the <a>accessibility tree</a> and it is an exactly parallel structure to the platform accessibility API trees except where the Core-AAM specifies a difference for a specific accessibility API.
```

lolaodelola (approved):
LGTM bar one editorial comment I have about readability below.

core-aam/index.html
```diff
@@ -6445,6 +6445,81 @@ <h2>General rules</h2>
...
+            <aside class="example">
+              <ol>
+                <li><pre>&lt;div role=button&gt;&lt;/div&gt;</pre> <a href="#aria-description" class="specref">`aria-description`</a> does not have a default value and it is allowed but not required on a <a href="#button" class="specref">`button`</a>, therefore the computed accessibility properties will not include `aria-description` .</li>
+                <li><pre>&lt;div role=textbox&gt;&lt;/div&gt;</pre> <a href="#aria-required" class="specref">`aria-required`</a> has a default value of `false` (not `undefined`) and it is allowed but not required on a <a href="#textbox" class="specref">`textbox`</a>, therefore the computed accessibility properties may include `aria-required: false`.</li>
+              </ol>
+            </aside>
+            </p>
+          </section>
+          <section id="computed_accessibility_properties_aria_states_and_properties">
```
cookiecrook: Editorial: Reconcile with duplicate ID on line 6453.

```suggestion
          <section id="additional_computed_accessibility_properties">
```

core-aam/index.html
```diff
@@ -6445,6 +6445,81 @@ <h2>General rules</h2>
...
+          </p>
+          <section id="computed_accessibility_properties_aria_states_and_properties">
```
cookiecrook: Editorial: Reconcile with duplicate ID on line 6482.
```suggestion
          <section id="computed_accessibility_properties_from_aria">
```

core-aam/index.html
```diff
@@ -7453,6 +7660,12 @@ <h4 id="ariaDescribedBy"><code>aria-describedby</code></h4>
                   <a class="property-reference" href="#aria-describedby"><code>aria-describedby</code></a>
                 </td>
               </tr>
+              <!-- <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  <code>description: &lt;value&gt;</code> where value is calculated according to the <a>Accessible Name and Description specification<a>.
```
cookiecrook: Editorial: 
```suggestion
                  <code>description: &lt;value&gt;</code> where value is calculated according to the <a>Accessible Name and Description specification</a>.
```

core-aam/index.html
```diff
@@ -6445,6 +6445,81 @@ <h2>General rules</h2>
...
+            <p>
+              To see if an ARIA state or property should be surfaced, and the name of the key and it's expected value, check the "Computed Accessibility Property" row of the tables below. Not all ARIA states and properties will be surfaced in the computed accessibility properties. If a host language feature maps to an ARIA state or property, use the key and value specified here. However, the final value in the accessibility tree can be affected by multiple things, including but not limited to the following: default IDL values for the DOM attribute, default ARIA values when used on specific ARIA roles, and host language features with conflicting semantics, depending on whether the host language overrides the ARIA value. These final computed properties values do not always align with IDL reflection of the ARIA attribute.
+            </p>
+            <aside class="example">
+              <p>For both of the following examples, the computed accessibility properties will include `checked: true`.</p><pre>
+&lt;div ro  le=checkbox aria-checked=true&gt;&lt;/div&gt;
+&lt;input   type="checkbox" checked /&gt;</pre>
+            </aside>
```
cookiecrook: Editorial: flow nesting
```suggestion
            <p>For both of the following examples, the computed accessibility properties will include `checked: true`.</p>
            <aside class="example">
              <pre>
&lt;div role=checkbox aria-checked=true&gt;&lt;/div&gt;
&lt;input type="checkbox" checked /&gt;</pre>
            </aside>
```

core-aam/index.html
```diff
@@ -7067,6 +7220,12 @@ <h4 id="ariaCheckedUndefined"><code>aria-checked</code> is undefined</h4>
                   <a class="state-reference" href="#aria-checked"><code>aria-checked</code></a> is undefined
                 </td>
               </tr>
+              <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  Do not expose <code>checked</code>.
```
cookiecrook, line 7231: @rahimabdi @spectranaut **Substantive:** would this conflict with the IDL structure? Should we treat this as exposing the default value rather than not exposing? 

Ditto for several listed in this PR as "Do not expose [x]."

core-aam/index.html
```diff
@@ -6445,6 +6445,81 @@ <h2>General rules</h2>
...
+              </ol>
+            </aside>
+            <p>
+              When the ARIA state or property does not have a default value other than undefined, is not required on a given role, and is missing, undefined, or set to an invalid token, user agents MUST NOT include it in the computed accessibility properties. When a state or property has a default value other than undefined, is supported but not required on a given role, and is not mapped on any platforms, user agents MUST NOT include it in the computed accessibility properties. If the state or property with the previous conditions is mapped on one or more platforms, user agents MAY include it in the computed accessibility properties.
+            <aside class="example">
+              <ol>
+                <li><pre>&lt;div role=button&gt;&lt;/div&gt;</pre> <a href="#aria-description" class="specref">`aria-description`</a> does not have a default value and it is allowed but not required on a <a href="#button" class="specref">`button`</a>, therefore the computed accessibility properties will not include `aria-description` .</li>
```
cookiecrook:
```suggestion
                <li><pre>&lt;div role=button&gt;&lt;/div&gt;</pre> <a href="#aria-description" class="specref">`aria-description`</a> does not have a default value and it is allowed but not required on a <a href="#button" class="specref">`button`</a>, therefore the computed accessibility properties will not include `description`.</li>
```

core-aam/index.html
```diff
@@ -6445,6 +6445,81 @@ <h2>General rules</h2>
...
+                <li><pre>&lt;div role=textbox&gt;&lt;/div&gt;</pre> <a href="#aria-required" class="specref">`aria-required`</a> has a default value of `false` (not `undefined`) and it is allowed but not required on a <a href="#textbox" class="specref">`textbox`</a>, therefore the computed accessibility properties may include `aria-required: false`.</li>
```
cookiecrook: Editorial:
```suggestion
                <li><pre>&lt;div role=textbox&gt;&lt;/div&gt;</pre> <a href="#aria-required" class="specref">`aria-required`</a> has a default value of `false` (not `undefined`) and it is allowed but not required on a <a href="#textbox" class="specref">`textbox`</a>, therefore the computed accessibility properties may include `required: false`.</li>
```

core-aam/index.html
```diff
@@ -6445,6 +6445,81 @@ <h2>General rules</h2>
...
+          </p>
+          <section id="computed_accessibility_properties_aria_states_and_properties">
+            <h3>Computed Accessibility Properties derived from ARIA States and Attributes</h3>
+            <p>
+              To see if an ARIA state or property should be surfaced, and the name of the key and it's expected value, check the "Computed Accessibility Property" row of the tables below. Not all ARIA states and properties will be surfaced in the computed accessibility properties. If a host language feature maps to an ARIA state or property, use the key and value specified here. However, the final value in the accessibility tree can be affected by multiple things, including but not limited to the following: default IDL values for the DOM attribute, default ARIA values when used on specific ARIA roles, and host language features with conflicting semantics, depending on whether the host language overrides the ARIA value. These final computed properties values do not always align with IDL reflection of the ARIA attribute.
```
cookiecrook: Editorial:
```suggestion
              To see if an ARIA state or property should be surfaced, and the name of the key and its expected value, check the "Computed Accessibility Property" row of the tables below. Not all ARIA states and properties will be surfaced in the computed accessibility properties. If a host language feature maps to an ARIA state or property, use the key and value specified here. However, the final value in the accessibility tree can be affected by multiple things, including but not limited to the following: default IDL values for the DOM attribute, default ARIA values when used on specific ARIA roles, and host language features with conflicting semantics, depending on whether the host language overrides the ARIA value. These final computed properties values do not always align with IDL reflection of the ARIA attribute.
```

core-aam/index.html
```diff
@@ -6445,6 +6445,81 @@ <h2>General rules</h2>
...
+              </ol>
+            </aside>
+            </p>
+          </section>
+          <section id="computed_accessibility_properties_aria_states_and_properties">
+            <h3>Additional Computed Accessibility Properties</h3>
+            <p>For the purposes developer tools, specification conformance, and interoperability testing, the set of Computed Accessibility Properties exposed for an accessible object also includes information about the relationship between <a>accessibility nodes</a> in the <a>accessibility tree</a>. In order to surface this information, the term <dfn>accessibility node ID</dfn> is used to refer to a implementation-defined unique ID used as a handle to an <a>accessibility node</a>. (TODO: add something about the stability of this ID? How long it should be available for?) The following keys and values MUST be included in the Computed Accessibility Properties:</p>
```
cookiecrook: Editorial: (two changes)
```suggestion
            <p>For the purposes of developer tools, specification conformance, and interoperability testing, the set of Computed Accessibility Properties exposed for an accessible object also includes information about the relationship between <a>accessibility nodes</a> in the <a>accessibility tree</a>. In order to surface this information, the term <dfn>accessibility node ID</dfn> is used to refer to an implementation-defined unique ID used as a handle to an <a>accessibility node</a>. (TODO: add something about the stability of this ID? How long it should be available for?) The following keys and values MUST be included in the Computed Accessibility Properties:</p>
```

cookiecrook: Maybe also change the TODO to an editorial note block so the TODO doesn't block merging?

core-aam/index.html
```diff
@@ -10134,6 +10677,12 @@ <h4 id="ariaRelevant"><code>aria-relevant</code></h4>
                   <a class="property-reference" href="#aria-relevant"><code>aria-relevant</code></a>
                 </td>
               </tr>
+              <!-- <tr>
+                <th>Computed Accessibility Property</th>
+                  <code>relevant: &lt;value&gt;</code>
+                <td>
```
cookiecrook: Editorial: invalid nesting... currently commented so linting didn't catch.
```suggestion
                <th>Computed Accessibility Property</th>
                <td>
                  <code>relevant: &lt;value&gt;</code>
```

core-aam/index.html
```diff
@@ -7509,6 +7722,12 @@ <h4 id="ariaDescription"><code>aria-description</code></h4>
                   <a class="property-reference" href="#aria-description"><code>aria-description</code></a>
                 </td>
               </tr>
+              <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  <code>description: &lt;value&gt;</code>
```
cookiecrook: Editorial: matching with the other mapping to description to avoid someone implementing a different conflict resolution.
```suggestion
                  <code>description: &lt;value&gt;</code> where value is calculated according to the <a>Accessible Name and Description specification</a>
```

cookiecrook (commented):
Mostly editorial

cyns (approved):
lgtm

core-aam/index.html
spectranaut: Thanks Lola! I incorporated your suggestion and modified it, if you want to take a look.

core-aam/index.html
spectranaut: I think the idea is that if there is a default value, like "false", then that should be exposed, and that is covered by the table for aria-checked="false". So this is only the scenario where it is explicitly not set, or missing on something that doesn't have a default value of a specific token.

ChrisC (approved):

core-aam/index.html
```diff
@@ -6445,6 +6445,86 @@ <h2>General rules</h2>
             </li>
           </ol>
         </section>
+        <section id="computed_accessibility_properties">
+          <h2>Computed Accessibility Properties</h2>
+          <p>
+            The <dfn>computed accessibility properties</dfn> are a set of platform-independent key/value pairs of strings that represent the states and properties of any <a href="#dfn-accessible-object" class="specref">accessible object</a> in the <a>accessibility tree</a>. The computed accessibility properties are exposed primarily for the purposes of developer tools, specification conformance, and interoperability testing.
```
jcsteh, line 6451: We define "computed accessibility properties" as representing platform-independent information about "accessible objects", except that "accessible objects" (by definition) are nodes in a platform accessibility API. We then define "accessibility node" as being the platform-independent container of "computed accessibility properties". This feels a bit circular and maybe slightly backward to me: the platform-specific tree is generated from the platform-independent tree, not the other way around. That said, it's hard to come up with a solution because of the existing spec definition of "accessible object" (and by extension "accessibility tree") as being platform-specific. I guess this might just be one of those cases where the spec language doesn't quite match real world architecture, which feels weird, but might be unavoidable.

core-aam/index.html
```diff
@@ -6445,6 +6445,86 @@ <h2>General rules</h2>
...
+          </p>
+          <p>
+            The term <dfn>accessibility node</dfn> will be used to refer to a node with the computed accessibility properties defined here. These platform-independent, implementation-defined accessibility nodes are expected to back all of the platform-specific <a href="#dfn-accessible-object" class="specref">accessible objects</a> that a user agent exposes in supported <a class="termref">accessibility APIs</a>.
```
jcsteh, line 6454: "are expected to back *all* of the platform-specific accessible objects" (emphasis mine)

There are some cases where a node is only exposed in a single platform API. In that case, a browser might expose a platform-specific object, but not a have a platform-independent accessibility node. The sentence below covers this, but I wonder if it's in conflict with this sentence above it.

I wonder if we can just drop "all of" here and let the sentence below cover the general case and exceptional cases?

core-aam/index.html
```diff
@@ -6445,6 +6445,86 @@ <h2>General rules</h2>
...
+          </p>
+          <section id="computed_accessibility_properties_from_aria">
+            <h3>Computed Accessibility Properties derived from ARIA States and Attributes</h3>
+            <p>
+              To see if an ARIA state or property should be surfaced, and the name of the key and its expected value, check the "Computed Accessibility Property" row of the tables below. Not all ARIA states and properties will be surfaced in the computed accessibility properties. If a host language feature maps to an ARIA state or property, use the key and value specified here. However, the final value in the accessibility tree can be affected by multiple things, including but not limited to the following: default IDL values for the DOM attribute, default ARIA values when used on specific ARIA roles, and host language features with conflicting semantics, depending on whether the host language overrides the ARIA value. These final computed properties values do not always align with IDL reflection of the ARIA attribute.
```
jcsteh, line 6460: > These final computed properties values

Should this be "computed *property* values"?

core-aam/index.html
```diff
@@ -6445,6 +6445,86 @@ <h2>General rules</h2>
...
+                <li><pre>&lt;div role=tab&gt;&lt;/div&gt;</pre> <a href="#aria-selected" class="specref">`selected`</a> on a <a href="#tab" class="specref">tab</a> has the implicit value of `false`, therefore the computed accessibility property will be `selected: false`.</li>
+              </ol>
+            </aside>
+            <p>
+              When the ARIA state or property does not have a default value other than undefined, is not required on a given role, and is missing, undefined, or set to an invalid token, user agents MUST NOT include it in the computed accessibility properties. When a state or property has a default value other than undefined, is supported but not required on a given role, and is not mapped on any platforms, user agents MUST NOT include it in the computed accessibility properties. If the state or property with the previous conditions is mapped on one or more platforms, user agents MAY include it in the computed accessibility properties.
```
jcsteh, line 6478: I think "with the previous conditions" means "has a default value other than undefined, is supported but not required on a given role". However, I wonder whether "the previous conditions" could be misinterpreted as encompassing both of the previous two sentences, not just the one immediately prior. I think explicit would be better here, despite the repetition.

nit: This paragraph has a combination of "a state or property" and "the state or property". I suggest changing "the" to "a".

core-aam/index.html
```diff
@@ -6711,6 +6821,12 @@ <h4 id="ariaBraillelabel"><code>aria-braillelabel</code></h4>
                   <a class="property-reference" href="#aria-braillelabel"><code>aria-braillelabel</code></a>
                 </td>
               </tr>
+              <!-- <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  <code>braillelabel: &lt;value&gt;</code>
```
jcsteh, line 6827: Here and later, we're not confined to lower case for these and there's precedent for using camel case in the IDL attribute names. Should we camel case these properties like we do for IDL?

```suggestion
                  <code>brailleLabel: &lt;value&gt;</code>
```

core-aam/index.html
```diff
@@ -6752,6 +6868,12 @@ <h4 id="ariaBrailleroledescription"><code>aria-brailleroledescription</code></h4
                   <a class="property-reference" href="#aria-brailleroledescription"><code>aria-brailleroledescription</code></a>
                 </td>
               </tr>
+              <!-- <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  <code>brailleroledescription: &lt;value&gt;</code>
```
jcsteh:
```suggestion
                  <code>brailleRoleDescription: &lt;value&gt;</code>
```

core-aam/index.html
```diff
@@ -6793,6 +6915,12 @@ <h4 id="ariaBrailleroledescriptionUndefined"><code>aria-brailleroledescription</
                   <a class="property-reference" href="#aria-brailleroledescription"><code>aria-brailleroledescription</code></a> is undefined or the empty string
                 </td>
               </tr>
+              <!-- <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  Do not expose <code>brailleroledescription</code>.
```
jcsteh:
```suggestion
                  Do not expose <code>brailleRoleDescription</code>.
```

core-aam/index.html
```diff
@@ -7281,6 +7469,12 @@ <h4 id="ariaControls"><code>aria-controls</code></h4>
                   <a class="property-reference" href="#aria-controls"><code>aria-controls</code></a>
                 </td>
               </tr>
+              <!-- <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  TODO: No exposed property?
```
jcsteh, line 7475: I think we could expose this as `controls: [nodeId1, nodeId2, ...]`
That would also be how I would expose other relations. But that might be better dealt with in a follow-up.

core-aam/index.html
```diff
@@ -7453,6 +7665,12 @@ <h4 id="ariaDescribedBy"><code>aria-describedby</code></h4>
                   <a class="property-reference" href="#aria-describedby"><code>aria-describedby</code></a>
                 </td>
               </tr>
+              <!-- <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  <code>description: &lt;value&gt;</code> where value is calculated according to the <a>Accessible Name and Description specification</a>.
```
jcsteh, line 7671: In addition, I think we'll want to expose this as `describedBy: [nodeId1, ...]`, but again, maybe that should go in a follow-up.

core-aam/index.html
```diff
@@ -10870,6 +11514,12 @@ <h4 id="ariaSortNone"><code>aria-sort</code>=<code>none</code></h4>
                   >=<code>none</code>
                 </td>
               </tr>
+              <!-- <tr>
+                <th>Computed Accessibility Property</th>
+                <td>
+                  <code>sort: none</code>
```
jcsteh, line 11520: Ug, this one is a bit annoying. It's effectively not mapped on all platforms, so we would ideally be able to just not expose it, except that apparently IA2 and ATK differentiate between "none" and empty here, even though they both mean the same thing.

PR #2808, opened 2026-06-05
Editorial: CORE-AAM small fixes

spectranaut:
🚀 **Netlify Preview**:
- [core-aam preview](https://deploy-preview-2808--wai-aria.netlify.app/core-aam/index.html) &mdash; [core-aam diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fcore-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-2808--wai-aria.netlify.app%2Fcore-aam%2Findex.html)

Editorial:
* Makes clear that an interface is an ATK interface, not ATSPI
* Corrects a typo in a role name.

### DIFF
--- a/core-aam/index.html
+++ b/core-aam/index.html
@@ -718,7 +718,7 @@
                 <td>
                   <span class="property">Role: <code>ROLE_ALERT</code></span
                   ><br />
-                  <span class="property">Interface: <code>Window</code></span
+                  <span class="property">ATK Interface: <code>Window</code></span
                   ><br />
                   <span class="event"
                     >Event: The user agent SHOULD fire a system alert <a class="termref">event</a>. <sup>[<a href="#ftn.note2">Note 2</a>]</sup></span
@@ -1822,7 +1822,7 @@
                 <td>
                   <span class="property">Role: <code>ROLE_DIALOG</code></span
                   ><br />
-                  <span class="property">Interface: <code>Window</code></span>
+                  <span class="property">ATK Interface: <code>Window</code></span>
                 </td>
               </tr>
               <tr>
--- a/core-aam/index.html
+++ b/core-aam/index.html
@@ -5213,7 +5213,7 @@
               <tr>
                 <th><abbr title="Accessibility Toolkit">ATK</abbr>/<abbr title="Assistive Technology - Service Provider Interface">AT-SPI</abbr></th>
                 <td>
-                  <span class="property">Role: <code>ROLE_STATUSBAR</code></span
+                  <span class="property">Role: <code>ROLE_STATUS_BAR</code></span
                   ><br />
                   <span class="property">Object Attribute: <code>container-live:polite</code></span
                   ><br />


spectranaut:
@joanmarie -- for some reason, in Core-AAM, both the [alertdialog](https://www.w3.org/TR/core-aam-1.2/#role-map-alertdialog) and [dialog](https://www.w3.org/TR/core-aam-1.2/#role-map-dialog) mention a "window" interface.. which does not seem to exist in the Atspi: https://docs.gtk.org/atspi2/class.Accessible.html

And also your old manual tests don't mention it... so I'm guess it doesn't exist and it's a mistake that it was ever included, but I thought I'd check with you just in case :)

joanmarie:
> @joanmarie -- for some reason, in Core-AAM, both the [alertdialog](https://www.w3.org/TR/core-aam-1.2/#role-map-alertdialog) and [dialog](https://www.w3.org/TR/core-aam-1.2/#role-map-dialog) mention a "window" interface.. which does not seem to exist in the Atspi: https://docs.gtk.org/atspi2/class.Accessible.html

It exists in ATK: https://docs.gtk.org/atk/iface.Window.html, but as you found not in AT-SPI.

AT-SPI listens for the ATK events and passes (most of) them to assistive technologies. See [this code](https://github.com/GNOME/at-spi2-core/blob/main/atk-adaptor/event.c#L760-L790).

> And also your old manual tests don't mention it... so I'm guess it doesn't exist and it's a mistake that it was ever included, but I thought I'd check with you just in case :)

My old manual tests don't mention it because they come from the AT-SPI side.

Having said that, [this Chromium code](https://source.chromium.org/chromium/chromium/src/+/main:ui/accessibility/platform/ax_platform_node_auralinux.cc;l=2420-2421;drc=95cdd5f7fa8df49707e6faffb071f57ec3c6b307;bpv=0;bpt=0) suggests that there's not a Chromium implementation. I *think* Gecko does implement it. And I have no idea about WebKitGTK.

**EDIT: Please see my next comment in which I clarify the above*.*

This is a long way of saying: I'm honestly not sure what you should do. :grin: I'd have to give it some more thought, test Orca with various ARIA alerts and dialogs, and ponder what the ideal solution is. I can add that to my TODO list...

joanmarie:
I really should drink more energy drinks before submitting comments....

The Chromium code from my previous comment, combined with [this other Chromium code](https://source.chromium.org/chromium/chromium/src/+/main:ui/accessibility/platform/ax_platform_node_auralinux.cc;l=2360-2361;bpv=1;bpt=0;drc=95cdd5f7fa8df49707e6faffb071f57ec3c6b307) which I forgot to include in my previous comment, suggests that there's not a Chromium implementation **for the ARIA roles**.

And my Gecko and WebKitGTK comments were also about the ARIA roles; not about the native browser window.

spectranaut:
Ahhh thanks @joanmarie awesome, that is so helpful. Once the new aam tests can test events I guess we will be able to test if the window interface is implemented by testing for events...? I'm also going to make a to do to look into this later.

And for now I'll update this PR to specify it is an ATK interface, somehow.

Edit: Made issue for adding tests https://github.com/w3c/core-aam/issues/260

spectranaut:
@cyns can you review this one?

jnurthen (approved):

PR #2818, opened 2026-06-16
Add mappings for CSS display property.

jcsteh:
🚀 **Netlify Preview**:
- [css-aam preview](https://deploy-preview-2818--wai-aria.netlify.app/css-aam/index.html) &mdash; [css-aam diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fcss-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-2818--wai-aria.netlify.app%2Fcss-aam%2Findex.html)

### DIFF
--- a/css-aam/index.html
+++ b/css-aam/index.html
@@ -163,8 +163,7 @@
                         <tr>
                             <th>CSS Specification</th>
                             <td>
-                                <a href="https://www.w3.org/TR/css-display-3/#box-generation">`display: none`</a>
-                                <!-- fix link -->
+                                <code>[[[css-display#box-generation|display: none]]]</code>
                             </td>
                         </tr>
                         <tr>
@@ -204,8 +203,7 @@
                         <tr>
                             <th>CSS Specification</th>
                             <td>
-                                <a href="https://www.w3.org/TR/css-display-3/#box-generation">`display: contents`</a>
-                                <!-- fix link -->
+                                <code>[[[css-display#box-generation|display: none]]]</code>
                             </td>
                         </tr>
                         <tr>
@@ -252,8 +250,7 @@
                         <tr>
                             <th>CSS Specification</th>
                             <td>
-                                <a href="https://www.w3.org/TR/css-display-3/#the-display-properties">`display`</a>
-                                <!-- fix link -->
+                                <code>[[[css-display#the-display-properties|display]]]</code>
                             </td>
                         </tr>
                         <tr>
--- a/css-aam/index.html
+++ b/css-aam/index.html
@@ -156,6 +156,137 @@
             </section>
             <section id="css-property-mappings">
                 <h3>CSS Property Mappings</h3>
+                <h4 id="prop-display-none">`display: none`</h4>
+                <table class="data" aria-labelledby="prop-display-none">
+                    <tbody>
+                        <tr>
+                            <th>CSS Specification</th>
+                            <td>
+                                <a href="https://www.w3.org/TR/css-display-3/#box-generation">`display: none`</a> <!-- fix link -->
+                            </td>
+                        </tr>
+                        <tr>
+                            <th>[[WAI-ARIA-1.2]]</th>
+                            <td>
+                                <a class="core-mapping" href="#ariaHiddenTrue">`aria-hidden="true"`</a>
+                            </td>
+                        </tr>
+                        <tr>
+                            <th>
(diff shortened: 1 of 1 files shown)


jcsteh:
From an editorial standpoint, I'm quite unsure of the right CSS classes, correct cross-referencing, etc. here.

css-aam/index.html
```diff
@@ -156,6 +156,137 @@ <h3>General Rules for Exposing WAI-ARIA Semantics</h3>
...
+                        <tr>
+                            <th>CSS Specification</th>
+                            <td>
+                                <a href="https://www.w3.org/TR/css-display-3/#box-generation">`display: contents`</a> <!-- fix link -->
+                            </td>
+                        </tr>
+                        <tr>
+                            <th>[[WAI-ARIA-1.2]]</th>
+                            <td>
+                                <span class="core-mapping">Element SHOULD be exposed</span>
```
pkra, line 212: If some ARIA spec language is desirable here, two existing options might be 

- not [hidden](https://w3c.github.io/aria/#dfn-hidden)
- [exposed](https://w3c.github.io/aria/#dfn-expose)

The latter doesn't appear to be referenced anywhere in the ARIA repository, so it could also be rephrased to fit here.

Otherwise, a new term might be usfeul

ZoeBijl:
Notes from the css-aam sync today:

- Reviewed PR #2818
- Should there be a definition for “Included” in ARIA?
- ~~TODO: Correctly link to definitions and specs with ReSpec~~
- RESOLUTION: set maximum table of contents level to three.
- RESOLUTION: each property will have a generic “top” level (`<h4>`) with the various values nested inside (`<h5>`)

css-aam/index.html
```diff
@@ -156,6 +156,141 @@ <h3>General Rules for Exposing WAI-ARIA Semantics</h3>
             </section>
             <section id="css-property-mappings">
                 <h3>CSS Property Mappings</h3>
+                <h4 id="prop-display">`display` property</h4>
```
ZoeBijl, line 159: We decided on having a generic top level for each property. Should we add some text here about the property? Maybe link to the relevant CSS spec that defines the property?

RESOLUTION: we decided to _not_ include any text, less is more.

ZoeBijl (approved):
lgtm as discussed during review :)

css-aam/index.html
```diff
@@ -156,6 +156,141 @@ <h3>General Rules for Exposing WAI-ARIA Semantics</h3>
...
+                        </tr>
+                    </tbody>
+                </table>
+                <h5 id="prop-display-contents">`display: contents`</h5>
+                <table class="data" aria-labelledby="prop-display-contents">
+                    <tbody>
+                        <tr>
+                            <th>CSS Specification</th>
+                            <td>
+                                <code>[[[css-display#box-generation|display: none]]]</code>
```
keithamus:
```suggestion
                                <code>[[[css-display#box-generation|display: contents]]]</code>
```

Presumably this should be `display:contents`?

css-aam/index.html
jcsteh: @pkra This is a bit of a tricky case because strictly speaking, display: contents has no impact on accessibility at all. There's no explicit mention anywhere that it should be excluded from the accessibility tree, which implicitly means that it should be included. The problem is that this is commonly misunderstood due to display: none, etc., so I think this should be explicit.

I think the best match in the ARIA spec is the concept of elements that must be [included in the accessibility tree](https://w3c.github.io/aria/#tree_inclusion). However, that could be confusing because that section has a very explicit list of rules which doesn't have any allowance for extension via an AAM spec.

Alternatively, we could just say "not mapped" (since it doesn't map to any property in ARIA) and rely on the comments to make the tree inclusion explicit.

jcsteh: ARIA 7.1 Excluding Elements from the Accessibility Tree explicitly talks about host language semantics which exclude things from the tree. That leads to another possibility here: we could add text to ARIA 7.2 Including Elements in the Accessibility Tree; e.g.

> Elements, including their descendent elements, that have host language semantics specifying that the element has explicit semantic relevance, such as CSS display:contents.

Yet another possibility is that we just link to the [7. Accessibility Tree section of the ARIA spec](https://w3c.github.io/aria/#accessibility_tree), which talks more generally about why something might be included in the accessibility tree (because it can fire events, has a feature which needs to be exposed, etc.).

PR #2820, opened 2026-06-18
editorial: fix synonymous roles

### LINKED ISSUE #2073: Missing role=presentation for "Name from prohibited"
# Describe your concern
While reviewing [5.2.8.6 Roles which cannot be named (Name prohibited)](https://w3c.github.io/aria/#namefromprohibited), I noticed that the role `none` is in the list but not `presentation`. I assumed that since they are synonyms, that listing one or the other is OK however, the [5.3.3 Document Structure Roles](https://w3c.github.io/aria/#document_structure_roles) section lists both `none`/`presentation`.

My concern is that a quick glance of section 5.2.8.6 may yield a false impression that the `presentation` role can be named, and that it's not a true alias for `none`.

I tried to add the `presentation` role in section 5.2.8.6 as a simple `<li>` however, it looks like the list is dynamically generated by `aria/common/script/aria.js` using the characteristics table for each role's definition in `aria/index.html` (and the "Name From: " row and whether it is "prohibited"); this appears to be why `presentation` is missing from 5.2.8.6 because the [`presentation` role definition](https://w3c.github.io/aria/#presentation) has no characteristics table unlike the [`none` role definition](https://w3c.github.io/aria/#none).

Should `presentation`/`none` always be listed together (like in 5.3.3)? And should the `presentation` role definition have the same characteristics table as `none`?

Link to documentation: https://w3c.github.io/aria/

Yes

### LINKED ISSUE #2871: Committing roleinfo updates to main
The current [RoleInfo workflow](https://github.com/w3c/aria/blob/main/.github/workflows/roleInfo.yml) only runs in pushes to the gh-pages branch.

Since #2857, we've moved ariavaluemin and ariavaluemax properties and that creates a failure for the [roleInfo check](https://github.com/w3c/aria/blob/main/.github/workflows/roleInfoCheck.yml) workflow in every PR.

We need to decide how we handle this. IMO,we should manually commit  the differences in roleInfo.js to avoid this false failure.

pkra:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2820--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2820--wai-aria.netlify.app%2Findex.html)

Adds synoymous role markup and updates pre- and post-processing.
- index.html
  - adds new row to characteristics tables for roles none and image
- common/script/
  - aria.js
    - renderIndexEntry()
      - switch check to innerHTML (for performance)
      - add synonymous roles to index entry markup
  - buildRoleInfo.js
    - generateRoleInfoEntry()
      - add synonymRoles array to role objects

Part of #2501
Resolves #2073


pkra:
I intentionally did not commit an update to roleInfo (to test if CI will update it)

Here's the diff from a local generation:

```diff
diff --git a/common/script/roleInfo.js b/common/script/roleInfo.js
index 825ebf20..f8c2f5ee 100644
--- a/common/script/roleInfo.js
+++ b/common/script/roleInfo.js
@@ -6543,6 +6543,7 @@
     name: "image",
     fragID: "image",
     parentRoles: ["section"],
+    synonymRoles: ["img"],
     localprops: [],
     allprops: [
       {
@@ -10444,6 +10445,7 @@
     name: "none",
     fragID: "none",
     parentRoles: ["structure"],
+    synonymRoles: ["presentation"],
     localprops: [
       {
         is: "property",
```

which matches the roleInfo test "failure".

spectranaut:
I'm a bit confused why only `image`/`none` have the synonym row added, and not `img` and `presentational`?

pkra:
@spectranaut 
> I'm a bit confused why only `image`/`none` have the synonym row added, and not `img` and `presentational`?

I'm not sure I understand. https://w3c.github.io/aria/#presentation and https://w3c.github.io/aria/#img have next to no content, in particular no characteristics tables. Are you referring to some other place in the spec?

spectranaut:
> @spectranaut
> 
> > I'm a bit confused why only `image`/`none` have the synonym row added, and not `img` and `presentational`?
> 
> I'm not sure I understand. https://w3c.github.io/aria/#presentation and https://w3c.github.io/aria/#img have next to no content, in particular no characteristics tables. Are you referring to some other place in the spec?

Ahh sorry, I was just trying to take a quick look at this PR before a meeting and I confused myself :sweat_smile: nvm!

spectranaut (approved):
This is great! ty :)

daniel-montalvo (approved):

PR #2838, opened 2026-07-06
Clarify case-insensitive treatment of ARIA roles

### LINKED ISSUE #2548: Are ARIA roles case sensitive or case insensitive?
Are ARIA roles meant to be case sensitive or not? Before #1457, the ARIA spec included this statement:

> 3. Compare the substrings to all the names of the non-abstract WAI-ARIA roles. Case-sensitivity of the comparison inherits from the case-sensitivity of the host language.

The Firefox accessibility team [assumed that meant ARIA roles should be case insensitive in HTML](https://bugzilla.mozilla.org/show_bug.cgi?id=1407167#c2). However, as @zcorpan [pointed out here](https://github.com/w3c/aria/issues/1236#issue-599575993):

> HTML has some attribute values case-sensitive, some ASCII case-insensitive.

I can't find any guidance in the spec either way, but [all three engines are case insensitive as far as I know](https://bugzilla.mozilla.org/show_bug.cgi?id=1407167#c4). Am I missing some guidance or is this currently undefined?

The reason this matters to me right now is that I'm trying to fix [a crash in Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1971642) and was about to add a WPT test which exercises it. However, I can't reasonably add a WPT test which exercises case insensitivity if the spec doesn't say it's case insensitive.

rahimabdi:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2838--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2838--wai-aria.netlify.app%2Findex.html)

Closes #2548 

This PR updates the "WAI-ARIA Roles" section specifying treatment of ARIA roles should be case-insensitive (see WPT tests below).

* [x] "author MUST" tests: n/a
* [x] "user agent MUST" tests: https://github.com/web-platform-tests/wpt/pull/53966/
   * WebKit: n/a
   * Gecko: n/a
   * Blink: n/a
* [x] ACT review? No
* [x] Related APG Issue/PR: n/a
* [x] MDN Issue/PR: n/a

### DIFF
--- a/index.html
+++ b/index.html
@@ -870,7 +870,7 @@
           language semantics, there are no changes in the <abbr title="Document Object Model">DOM</abbr>, only in the <a class="termref">accessibility tree</a>.
         </p>
         <p>
-          User agents MUST use the first token in the sequence of tokens in the <a><code>role</code> attribute</a> value that matches the name of any non-abstract
+          User agents MUST use the first token in the sequence of tokens in the <a><code>role</code> attribute</a> value that is an [=ASCII case-insensitive=] match for the name of any non-abstract
           <abbr title="Accessible Internet Application">WAI-ARIA</abbr> <a>role</a>. Refer to the section on <a><code>role</code> attribute</a> implementation in
           <a href="#host_general_role">Host Languages</a> for further details.
         </p>
--- a/index.html
+++ b/index.html
@@ -16352,7 +16352,7 @@ button.ariaPressed; // null</pre
             attribute value illegal in the host-language syntax; and
           </li>
           <li>
-            The first name literal of a non-abstract <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> role in the list of tokens in the <a><code>role</code> attribute</a> defines
+            The first [=ASCII case-insensitive=] name literal of a non-abstract <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> role in the list of tokens in the <a><code>role</code> attribute</a> defines
             the role according to which the user agent MUST process the element. User Agent processing for roles is defined in the
             <cite><a href="" class="core-mapping">Core Accessibility API Mappings</a></cite> [[CORE-AAM-1.2]].
           </li>


index.html
```diff
@@ -870,7 +870,7 @@ <h2><abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> Roles</h
           language semantics, there are no changes in the <abbr title="Document Object Model">DOM</abbr>, only in the <a class="termref">accessibility tree</a>.
         </p>
         <p>
-          User agents MUST use the first token in the sequence of tokens in the <a><code>role</code> attribute</a> value that matches the name of any non-abstract
+          User agents MUST use the first token in the sequence of tokens in the <a><code>role</code> attribute</a> value that is an [=ASCII case-insensitive=] match for the name of any non-abstract
```
pkra, line 873: I think there needs to be a matching change in the definition of the role attribute (Section 8.1) which has 

>  The first name literal of a non-abstract WAI-ARIA role in the list of tokens in the role attribute defines the role according to which the user agent MUST process the element.

Come to think of it, I think it would be better to remove this sentence here and only clarify this in the definition of the role attribute / Section 8.1.

pkra (changes requested):

index.html
rahimabdi: Agreed @pkra, could it be specified in both places? I've updated the PR.

HaTheo (approved):
LGTM. I am not too familiar with the repercussions of defining it both in 4.1 and 8.1, but I saw pkra seemed to prefer defining it only in 8.1.

index.html
pkra: We just merged https://github.com/w3c/aria/pull/2798 and section 8.1 is now the definitive place.

I strongly prefer not to duplicate normative statements - they get out of sync. Especially here, where there is little reason to look for a duplication all the way in Section 4 should we change something in Section 8.1.

index.html
```diff
@@ -16352,7 +16352,7 @@ <h2><a>Role Attribute</a></h2>
             attribute value illegal in the host-language syntax; and
           </li>
           <li>
-            The first name literal of a non-abstract <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> role in the list of tokens in the <a><code>role</code> attribute</a> defines
+            The first [=ASCII case-insensitive=] name literal of a non-abstract <abbr title="Accessible Rich Internet Applications">WAI-ARIA</abbr> role in the list of tokens in the <a><code>role</code> attribute</a> defines
```
pkra, line 16355: I think this sentence should be replaced with the one above;  "name literal" is just odd here.

front-endian (approved):

HaTheo (approved):

PR #2843, opened 2026-07-14
Fix tag attributes for msqrt and mn

### LINKED ISSUE #40: Tag attributes exposed by <msqrt> and <mn> seem wrong
IIUC, the tag attribute is used to identify the real tag name of the element.
However, the tag attribute exposed by `<msqrt>` is `mroot`, and for `<mn>` is `ms`, both of which seem wrong.
If that's true, I'll submit a PR to fix it.

tannal:
Closes https://github.com/w3c/mathml-aam/issues/40

Correcting the tag attribute for `<mn>` and `<msqrt>`.
The tag name attribute should be the same as the html name.

* [ ] "user agent MUST" tests: https://github.com/web-platform-tests/wpt/pull/61018
   * WebKit: https://searchfox.org/wubkat/rev/d221d71a6a560cb45bb3e9b23e526b6bcfa9e4b3/Source/WebCore/accessibility/atspi/AccessibilityObjectAtspi.cpp#866
   * Gecko: https://searchfox.org/firefox-main/rev/7fa9b602777418732e08b26d1ef6c9945c4bbd72/accessible/generic/LocalAccessible.cpp#1253
   * Blink: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/accessibility/ax_object.cc;l=1558;drc=0976730981877ed30e1cb06c63ad7d60c030c7f4
* [x] ACT review? N/A
* [x] MDN Issue/PR: N/A

### DIFF
--- a/mathml-aam/index.html
+++ b/mathml-aam/index.html
@@ -521,7 +521,7 @@
               <td>
                 <span class="role">Role: <code>ATK_ROLE_STATIC</code></span
                 ><br />
-                <span class="objattr">Object Attribute: <code>tag:ms</code></span>
+                <span class="objattr">Object Attribute: <code>tag:mn</code></span>
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


spectranaut:
Hi Tannal! I just updated your description related to the tests. The tests you are writing/working on in WPT are actually "user agent" tests. The "user agent" means the browser, and WPT tests what the browser should be doing, which is what the MathML-AAM says to do. So any tests for the AAMs do need "user agent" tests, and you are writing them :)

spectranaut (approved):
Thank you for this fix!!

PR #2844, opened 2026-07-14
Accessibility mapping of <a> element

### LINKED ISSUE #39: Accessibility mapping of `<a>` element
cc @lukewarlow @tannal @pepelsbey

The element is currently not mapped in the spec.

Original discussion: https://github.com/mdn/content/pull/44189/changes/6df275f4d0e8267bb50ffd6a313b4117a61942c6#r3267409017

tannal:
Closes https://github.com/w3c/mathml-aam/issues/39

This change map mathml a element to link role in atspi and axapi.
Leave TBD for MSAA and UIA api.
See the mathml-aam issue in detail.

* [ ] "user agent MUST" tests: https://github.com/web-platform-tests/wpt/pull/61018
   * WebKit: will file an issue later
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=2054819
   * Blink: https://issues.chromium.org/issues/510487697
* [x] ACT review? N/A
* [x] Does this need AT implementations? N/A
* [x] MDN Issue/PR: N/A

I got 503 when accessing webkit bug tracker, so for now I can't file the issue. :-)

### DIFF
--- a/mathml-aam/index.html
+++ b/mathml-aam/index.html
@@ -194,7 +194,7 @@
             <tr>
                 <th><abbr title="macOS Accessibility Protocol">AX API</abbr></th>
                 <td>
-                    <div class="general">Use WAI-ARIA mapping, but replace the <code>AXSubrole</code> with <code>AXMathRow</code>.</div>
+                    <div class="general">Use WAI-ARIA mapping, but override the <code>AXSubrole</code> with <code>AXMathRow</code>.</div>
                 </td>
             </tr>
             </tbody>
--- a/mathml-aam/index.html
+++ b/mathml-aam/index.html
@@ -160,6 +160,49 @@
 
       <section>
         <h3>MathML Element Mappings</h3>
+        <h4 id="el-a">`a`</h4>
+                <table class="data" aria-labelledby="el-a">
+                  <tbody>
+                    <tr>
+                      <th>MathML Specification</th>
+                      <td>
+                        <a data-cite="MathML3/chapter5.html#mixing.elements.a">`a`</a>
+                      </td>
+                    </tr>
+                    <tr>
+                      <th>[[wai-aria-1.1]]</th>
+                      <td>`link` role</td>
+                    </tr>
+                    <tr>
+                      <th>MSAA + IAccessible2</th>
+                      <td>
+                        <span class="role">Role: <code>ROLE_SYSTEM_LINK</code></span><br />
+                        <span class="state">IA2 State: <code>IA2_STATE_LINKED</code></span>
+                      </td>
+                    </tr>
+                    <tr>
+                      <th><abbr title="User Interface Automation">UIA</abbr></th>
+                      <td>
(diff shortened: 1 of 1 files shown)


mathml-aam/index.html
```diff
@@ -160,6 +160,43 @@ <h3>Exposing features that do not directly map to accessibility <abbr title="app
...
+            <tbody>
+            <tr>
+                <th>MathML Specification</th>
+                <td>
+                <a data-cite="MathML3/chapter5.html#mixing.elements.a">`a`</a>
+                </td>
+            </tr>
+            <tr>
+                <th>[[wai-aria-1.1]]</th>
+                <td>No corresponding role</td>
```
spectranaut: I'm a little worried that this is so similar to the core-aam definition of an `a` element -- other than the AXSubrole.

For example: The "tag" object attribute is actually always exposed, for all elements, that should be made more clear in the Core-AAM, and the linux mappings in the Core-AAM mention HyperlinkImpl, which is not mentioned here.

I think it might be better to say that there is a corresponding role, and that you should follow it, but that there is an AXSubrole override. So basically, for ATK write "Use the WAI-ARIA mappings" like here https://w3c.github.io/html-aam/#el-a, but for AX AXPI write "Use the WAI-ARIA mappings, but override subrole with..."

spectranaut (changes requested):
I'm a little worried that this is so similar t

mathml-aam/index.html
tannal: Done.

mathml-aam/index.html
```diff
@@ -160,6 +160,45 @@ <h3>Exposing features that do not directly map to accessibility <abbr title="app
...
+            <tbody>
+            <tr>
+                <th>MathML Specification</th>
+                <td>
+                <a data-cite="MathML3/chapter5.html#mixing.elements.a">`a`</a>
+                </td>
+            </tr>
+            <tr>
+                <th>[[wai-aria-1.1]]</th>
+                <td><a class="core-mapping" href="#role-map-link">`link`</a> role if the element has a valid <code>href</code> attribute. For elements that are not links, use the mapping for <a href="#el-mrow"><code>mrow</code></a>.</td>
```
spectranaut:
```suggestion
                <td><a class="core-mapping" href="#role-map-link">`link`</a> role if the element has a valid <code>href</code> attribute. For elements that do not have a valid <code>href</code>, use the mapping for <a href="#el-mrow"><code>mrow</code></a>.</td>
```

When writing specs, it's best to be pedantic, even if that means repeating yourself.

spectranaut (changes requested):
This is looking good, one small change request.

mathml-aam/index.html
tannal: Done. Thanks for the explanation!

spectranaut (approved):
Looks good to me, thanks @tannal !

spectranaut:
I've marked this as "waiting for implementations" because we require two implementations to merge.

PR #2845, opened 2026-07-16
Add aria-actions focus handling wording

### LINKED ISSUE #2691: aria-actions: handling focus when actions are synthetically triggered

smhigley:
🚀 **Netlify Preview**:
- [core-aam preview](https://deploy-preview-2845--wai-aria.netlify.app/core-aam/index.html) &mdash; [core-aam diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fcore-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-2845--wai-aria.netlify.app%2Fcore-aam%2Findex.html)
- [ARIA preview](https://deploy-preview-2845--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2845--wai-aria.netlify.app%2Findex.html)

Closes #2691

This is a PR into the aria-actions branch, not into main. It's split out into its own PR for ease of reviewing this specific update.

I tried to put our agreed-on focus bounce behavior into spec language. In plain language, what happens is:

The browser (not ATs) allow DOM focus to move to the action following the click event, but don't fire API focus events, so screen reader cursors will stay on the referencing element. After a short delay (maybe one tick, maybe a small timeout), browsers will move focus back to the referencing element unless focus has moved somewhere else in the meantime. The API focus event still won't be called, so from a screen reader user's perspective, there have been no focus changes at all. But from a page javascript perspective, the normal events have fired.

I didn't include any wording about what happens if a page author does `preventDefault` on the click or on the first focus event, since that seems too far into the weeds of DOM behavior for ARIA (and more of a JS/eventing concern). Let me know if anyone feels that should be here, though.

### DIFF
--- a/core-aam/index.html
+++ b/core-aam/index.html
@@ -6461,7 +6461,12 @@
             </p>
           </section>
           <h4 id="ariaActions"><code>aria-actions</code></h4>
-          <table aria-labelledby="ariaActions">
+          <p>
+            After an action is invoked, browsers MUST NOT fire the API's <a href="#focus_state_event_table">focus event</a> when the referenced action node receives focus.
+            The browser MUST subsequently return DOM focus to the referencing node also without firing an API <a href="#focus_state_event_table">focus event</a>,
+            unless focus is moved to a different node by a programmatic focus change or a user interaction.
+          </p>
+          <table class="data" aria-labelledby="ariaActions">
             <tbody>
               <tr>
                 <th>ARIA Specification</th>
--- a/core-aam/index.html
+++ b/core-aam/index.html
@@ -6463,7 +6463,7 @@
           <h4 id="ariaActions"><code>aria-actions</code></h4>
           <p>
             After an action is invoked, browsers MUST NOT fire the API's <a href="#focus_state_event_table">focus event</a> when the referenced action node receives focus.
-            The browser MUST subsequently return DOM focus to the referencing node also without firing an API <a href="#focus_state_event_table">focus event</a>,
+            The browser MUST subsequently return DOM focus to the referencing node also without firing an API <a href="#focus_state_event_table">focus event</a>
             unless focus is moved to a different node by a programmatic focus change or a user interaction.
           </p>
           <table class="data" aria-labelledby="ariaActions">
--- a/index.html
+++ b/index.html
@@ -12197,8 +12197,21 @@ button.ariaPressed; // null</pre
               <li>User Agents SHOULD use the accessible names of elements referenced by <code>aria-actions</code> to determine the names of actions that are exposed in a platform accessibility API.</li>
               <li>User Agents MUST NOT expose <code>aria-actions</code> if the Author MUST's are not followed.</li>
             </ul>
+            <p>In order to allow users to invoke actions while remaining on the referencing element, browsers MUST do the following steps to manage focus once an action is invoke via the <code>aria-actions</code> API:</p>
+            <ol>
+              <li>
+                 If the referenced action node is focusable and receives focus as a result of an <code>aria-actions</code> invocation:
+                <ul>
+                  <li>Allow the DOM focus events to fire on the referenced action node as a result of invoking it.</li>
+                  <li>Do not fire the <a>desktop focus event</a>.</li>
+                </ul>
+              </li>
+              <li>If focus is moved either programmatically or via a user interaction, handle that focus change as normal, and do not perform any further <code>aria-actions</code> focus steps.</li>
+              <li>If focus has not moved, the browser MUST move DOM focus back to the referencing element, again without firing a <a>desktop focus event</a>.</li>
+              <li>If the referencing element can no longer be focused, the browser MAY keep focus on the referenced action element and fire a <a>desktop focus event</a>.</li>
+            </ol>
           </div>
-          <table class="property-features">
+          <table class="def">
             <caption>Characteristics:</caption>
             <thead>
               <tr>
--- a/index.html
+++ b/index.html
@@ -12197,17 +12197,18 @@ button.ariaPressed; // null</pre
               <li>User Agents SHOULD use the accessible names of elements referenced by <code>aria-actions</code> to determine the names of actions that are exposed in a platform accessibility API.</li>
               <li>User Agents MUST NOT expose <code>aria-actions</code> if the Author MUST's are not followed.</li>
             </ul>
-            <p>In order to allow users to invoke actions while remaining on the referencing element, browsers MUST do the following steps to manage focus once an action is invoke via the <code>aria-actions</code> API:</p>
+            <p>To prevent an unexpected focus change away from the referencing element after users invoke actions, browsers MUST perform the following steps to manage focus once an action is invoked via <code>aria-actions</code>:</p>
             <ol>
               <li>
-                 If the referenced action node is focusable and receives focus as a result of an <code>aria-actions</code> invocation:
+                 If the referenced action node is focusable and received focus as a result of an <code>aria-actions</code> invocation:
                 <ul>
                   <li>Allow the DOM focus events to fire on the referenced action node as a result of invoking it.</li>
                   <li>Do not fire the <a>desktop focus event</a>.</li>
                 </ul>
               </li>
+              <li>If the referenced action node is not focusable or did not receive focus, do not perform any further <code>aria-actions</code> focus steps.</li>
               <li>If focus is moved either programmatically or via a user interaction, handle that focus change as normal, and do not perform any further <code>aria-actions</code> focus steps.</li>
-              <li>If focus has not moved, the browser MUST move DOM focus back to the referencing element, again without firing a <a>desktop focus event</a>.</li>
+              <li>If focus has remained on the referenced action node, the browser MUST move DOM focus back to the referencing element, again without firing a <a>desktop focus event</a>.</li>
               <li>If the referencing element can no longer be focused, the browser MAY keep focus on the referenced action element and fire a <a>desktop focus event</a>.</li>
             </ol>
           </div>


spectranaut:
@twilco I can't add you to the review list but if you can take a look

core-aam/index.html
```diff
@@ -6461,7 +6461,12 @@ <h4>Not Mapped</h4>
             </p>
           </section>
           <h4 id="ariaActions"><code>aria-actions</code></h4>
-          <table aria-labelledby="ariaActions">
+          <p>
+            After an action is invoked, browsers MUST NOT fire the API's <a href="#focus_state_event_table">focus event</a> when the referenced action node receives focus.
```
cookiecrook, line 6465:
```suggestion
            After an action is invoked, browsers SHOULD NOT fire the system or accessibility-specific <a href="#focus_state_event_table">focus event</a> when the referenced action node receives focus.
```

cookiecrook: Also I think this should be focus "Notification" or "Signals" rather than "Events"? My understanding is that, for most native platform APIs "Event" is referring to an input event like a click (most often user-triggered), but the subsequent broadcasts that goes out over the system would be called "Notifications" on Windows and Apple systems, or "Signals" on GTK.

core-aam/index.html
```diff
@@ -6461,7 +6461,12 @@ <h4>Not Mapped</h4>
...
+            The browser MUST subsequently return DOM focus to the referencing node also without firing an API <a href="#focus_state_event_table">focus event</a>,
```
cookiecrook, line 6466: "subsequently" may be too vague here. does this happen in the same runloop? If not, how long could/should an implementation wait before it's non-compliant?

cookiecrook: Also, same issue as above wrt "event" vs "notification" and "system or accessibility-specific" vs "API"

index.html
```diff
@@ -12197,8 +12197,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
               <li>User Agents SHOULD use the accessible names of elements referenced by <code>aria-actions</code> to determine the names of actions that are exposed in a platform accessibility API.</li>
               <li>User Agents MUST NOT expose <code>aria-actions</code> if the Author MUST's are not followed.</li>
             </ul>
+            <p>In order to allow users to invoke actions while remaining on the referencing element, browsers MUST do the following steps to manage focus once an action is invoke via the <code>aria-actions</code> API:</p>
```
cookiecrook, line 12200:
```suggestion
            <p>In order to allow users to invoke actions while remaining on the referencing element, browsers MUST perform the following steps to manage focus once an action is invoked via <code>aria-actions</code>:</p>
```

cookiecrook: In addition to the first two editorial diffs here, I wasn't sure "the aria-actions API" sounded quite right to me. I'm not set on that final diff through if you want it back.

index.html
```diff
@@ -12197,8 +12197,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+            <ol>
+              <li>
+                 If the referenced action node is focusable and receives focus as a result of an <code>aria-actions</code> invocation:
+                <ul>
+                  <li>Allow the DOM focus events to fire on the referenced action node as a result of invoking it.</li>
+                  <li>Do not fire the <a>desktop focus event</a>.</li>
```
cookiecrook, line 12206:
```suggestion
                  <li>Do not fire the <a>system or accessibility-specific focus notification</a>.</li>
```
Caveat: keep this in line with whatever prose decision is made above on this same issue.

index.html
```diff
@@ -12197,8 +12197,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+                </ul>
+              </li>
+              <li>If focus is moved either programmatically or via a user interaction, handle that focus change as normal, and do not perform any further <code>aria-actions</code> focus steps.</li>
```
cookiecrook, line 12209: > If focus is moved either programmatically or via a user interaction, 

I presume you're not counting an aria-actions invocation here as a "user interaction" so I would recommend a different term that more clearly excludes it.

> handle that focus change as normal

Is there language in the HTML or DOM specs that phrases these more precisely?

index.html
```diff
@@ -12197,8 +12197,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+              <li>If focus has not moved, the browser MUST move DOM focus back to the referencing element, again without firing a <a>desktop focus event</a>.</li>
```
cookiecrook, line 12210: But focus did move to get here, right? So the "if focus has not moved" clause would never be invoked.

ditto prose for "desktop focus event" (these notifications also aren't limited to "desktops")

index.html
```diff
@@ -12197,8 +12197,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+              <li>If the referencing element can no longer be focused, the browser MAY keep focus on the referenced action element and fire a <a>desktop focus event</a>.</li>
```
cookiecrook, line 12211: ditto

cookiecrook (changes requested):

index.html
smhigley: this one was a bit on purpose, since the wording "desktop focus event" is used for this elsewhere in the spec & there's a definition for it as well

index.html
smhigley: I think if a second aria-actions invocation happened here, then presumably this set of focus steps should be abandoned and that aria-actions focus behavior should be kicked off, so I think it still applies 😅

Still open to better or more precise wording in both cases though

index.html
smhigley: This is the spec definition for "desktop focus event", should we change the wording or add an additional term that shares the same def?

> Event from/to the host operating system via the accessibility API, notifying of a change of input focus.

core-aam/index.html
janewman: I don't have a strong preference here but wanted to note that core-aam already uses the term ["Focus event"](https://www.w3.org/TR/core-aam-1.2/#focus_state_event_table) for these platform mappings. If we want to use "Notification" or "Signals" instead, we should change this everywhere for consistency.

core-aam/index.html
mcking65: There should not be a comma before the word 'unless':

```suggestion
            The browser MUST subsequently return DOM focus to the referencing node also without firing an API <a href="#focus_state_event_table">focus event</a>
```

index.html
mcking65: ***
Suggested change:
```
            <p>To prevent unexpected focus change away from the referencing element after users invoke actions, browsers MUST perform the following steps to manage focus once an action is invoked via <code>aria-actions</code>:</p>  
```
***

mcking65: I suggest changing the initial phrase to clarify both:
1) that the user is not on the referencing element, the focus is on the referencing element.
2) The intent is to maintain focus after the action is invoked if doing so is xpected.

index.html
```diff
@@ -12197,8 +12197,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
               <li>User Agents SHOULD use the accessible names of elements referenced by <code>aria-actions</code> to determine the names of actions that are exposed in a platform accessibility API.</li>
               <li>User Agents MUST NOT expose <code>aria-actions</code> if the Author MUST's are not followed.</li>
             </ul>
+            <p>In order to allow users to invoke actions while remaining on the referencing element, browsers MUST do the following steps to manage focus once an action is invoke via the <code>aria-actions</code> API:</p>
+            <ol>
+              <li>
+                 If the referenced action node is focusable and receives focus as a result of an <code>aria-actions</code> invocation:
```
mcking65, line 12203:
```suggestion
                If the referenced action node is focusable and received focus as a result of an <code>aria-actions</code> invocation:
```

index.html
mcking65:
```suggestion
              <li>If the referenced action node is focusable and received focus, move DOM focus back to the referencing element without firing a <a>desktop focus event</a>.</li>
```

mcking65: This list item needs the same 'if' condition as the first list item. It does not need to state that it is the result of the action invocation or that the browser must do the focus change because those things are stated before the start of the list.

mcking65 (approved):
I've made a few suggestions. While they effect meaning, I don't think they change the intended meaning.

core-aam/index.html
smhigley: I agree with what ^ @janewman said, for both specs I made sure to use the same language already present elsewhere for the same concept.

For now, I'd lean towards not creating new wording just for this, since the base concept is already present/described in each spec in other places as well.

core-aam/index.html
smhigley: Comma removed!

The vagueness was intentional in this case b/c when discussed (I think maybe before you joined?) we wanted to let different browsers experiment/make their own decisions about the right amount of time and avoid being too rigid in the spec, especially before we've had wider feedback on it.

Does that sound reasonable? I can see a future where we all realize there's some optimal delay (maybe one run cycle) and update the spec to be specific.

index.html
smhigley: I updated the line to Matt's wording (which also has James' suggestion). I think that reads much better now :D

index.html
smhigley: I updated all the steps slightly, adding a new second item to specifically clarify that if the referenced action node did not get focus, the browser should stop going through the steps.

Also updated the wording of this step, let me know if the whole series now looks better.

core-aam/index.html
jcsteh: Just for what it's worth, all Windows accessibility APIs (MSAA, IA2, UIA) call these "events". Gecko also calls them accessibility events internally. I don't know the Chromium code super well, but a brief look suggests it also calls them accessibility events internally. But as noted above, consistency is the most important goal here.

core-aam/index.html
jcsteh: There's a slight risk of interop issues if we don't align here, but I also totally acknowledge the risks of being too rigid while we're still figuring this out. Personally, I think the best thing to do here is to dispatch the DOM events (mousedown, mouseup, etc.) to invoke the action, then synchronously return the focus. Note that dispatching DOM events happens synchronously, so any click event handler that moves the focus will have done so by the time the dispatch is done. However, this depends somewhat on whether it's feasible for other browsers to do this. For example, if a browser must handle accessibility invocation via an asynchronous input queue, I guess what I'm suggesting might not be possible.

index.html
jcsteh: What if the referencing element is *not* focused when the action is invoked? Do we just allow focus to move to the invoked action target in that case?

My Gecko prototype actually doesn't consider whether the focus is the referencing element. If an action is invoked, it returns focus whence it came, even if the originating focus was some other element. I'm not necessarily saying that is what we should do, but I also don't see a downside.

Either way, perhaps we need to be explicit here about what happens in this case.

smhigley:
@mcking65 / @cookiecrook I think all comments have either been updated in the PR or responded to if you want to take another look!

jcsteh (commented):
Sorry for the century it took me to get to this.

index.html
smhigley: @jcsteh I think that case shouldn't be possible since one of the steps to invoke an action is to first focus the referencing element. The original reason was because authors are not required to have actions in the DOM until the referencing element is focused, so if an AT user requests actions, the AT must focus the referencing element first.

Although this being the web, I can see the use case for specifying a fallback regardless, because chaos and edge cases always happen. I actually like the idea of having focus always return to the referencing element when an action is invoked, even if it was somehow elsewhere to start with, that makes sense to me from a user perspective as well.

core-aam/index.html
smhigley: Yeah my much less informed opinion is the same as yours -- synchronously if possible or in the next run cycle if not seems best. But I'm much less aware of the internal mechanics of browser focus events than you are, so I have no idea to what degree it makes sense to specify timing in the ARIA spec.

IIRC we also discussed keeping this more vague for Apple specifically, since there's often a Windows/macOS divide in how a11y APIs and events are handled under the hood. I"m not sure whether that's actually necessary here though, @cookiecrook probably knows. And also just b/c this is new enough that there should be some leeway for making tweaks out of the gate, I think.

jcsteh: > IIRC we also discussed keeping this more vague for Apple specifically, since there's often a Windows/macOS divide in how a11y APIs and events are handled under the hood.

I don't think that should matter here. Even if the accessibility events are delivered differently to the OS (sync, async, IPC, whatever), the code that *triggers* the delivery of those events would ideally run in a context where it has direct access to the DOM. Otherwise, there's a much greater risk of inconsistent, timing dependent behaviour, which is one of the major reasons we abandoned the idea of AT bouncing the focus.

> And also just b/c this is new enough that there should be some leeway for making tweaks out of the gate, I think.

There's two paths there. We could keep it loose now and tighten it up later... or we can make a decision now and adjust the spec as needed. One thing to consider is that if we define it loosely, there's a significant chance that we just never get around to tightening it, resulting in weird interop and backward compat issues we have to sort out later. In contrast, if we start tight and loosen, we're much likely to update the spec sooner because it's actually blocking implementers. Ideally, I think we'd get some high level feedback from implementers now as to what is feasible within their engines.

That said, I don't feel strongly about this and won't block on it, just flagging the potential interop risk.

core-aam/index.html
jcsteh, line 6465: Should this just refer to the relevant section in the ARIA spec? That section provides a more precise set of steps as to what the browser should do, which could be important if we do decide to introduce timing rules.

index.html
```diff
@@ -12197,8 +12197,22 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+              <li>
+                 If the referenced action node is focusable and received focus as a result of an <code>aria-actions</code> invocation:
+                <ul>
+                  <li>Allow the DOM focus events to fire on the referenced action node as a result of invoking it.</li>
+                  <li>Do not fire the <a>desktop focus event</a>.</li>
+                </ul>
+              </li>
+              <li>If the referenced action node is not focusable or did not receive focus, do not perform any further <code>aria-actions</code> focus steps.</li>
+              <li>If focus is moved either programmatically or via a user interaction, handle that focus change as normal, and do not perform any further <code>aria-actions</code> focus steps.</li>
+              <li>If focus has remained on the referenced action node, the browser MUST move DOM focus back to the referencing element, again without firing a <a>desktop focus event</a>.</li>
```
jcsteh, line 12211: With regard to the [other comment thread where we're discussing timing](https://github.com/w3c/aria/pull/2845#discussion_r3985056043), we could amend the wording like this if other implementers agree:

```suggestion
              <li>Synchronously after the DOM events for the invocation have been dispatched, if focus has remained on the referenced action node, the browser MUST move DOM focus back to the referencing element, again without firing a <a>desktop focus event</a>.</li>
```

core-aam/index.html
smhigley: I'd be happy to be more specific now, if you, James, and someone from Chrome or Edge are in agreement. We could also just update "subsequently" to something like "immediately" and still have a little wiggle room for same run cycle vs. next depending on internal browser constraints.

How would that sound? Is there better wording that browser engine folks can think of?

twilco: > **jcsteh:** Personally, I think the best thing to do here is to dispatch the DOM events (mousedown, mouseup, etc.) to invoke the action, then synchronously return the focus.

> **jcsteh:** Ideally, I think we'd get some high level feedback from implementers now as to what is feasible within their engines.

This is how [WebKit has implemented it](https://github.com/WebKit/WebKit/blob/59da037a4b6636a933a1cf6041f2de4c4b6dd163/Source/WebCore/accessibility/AccessibilityObject.cpp#L1661), and it seems to work well based on my testing of some (admittedly simple) `aria-actions` examples while making said change. So unless @cookiecrook can think of a reason otherwise, this is feasible for (and already landed in) WebKit.

If we do end up aligning on something else though, happy to change things on our side.

index.html
twilco: For what it's worth — WebKit currently returns focus whence it came, regardless of whether the whence is the referencing element. But as with my other comment, if we find reason to do something else, we can definitely change this in WebKit.

index.html
mcking65: The comma before "again", and the word "again" expose this to misinterpretation.

```suggestion
              <li>If focus has remained on the referenced action node, the user agent MUST move DOM focus back to the referencing element without firing a <a>desktop focus event</a>.</li>
```

index.html
```diff
@@ -12197,8 +12197,22 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
               <li>User Agents SHOULD use the accessible names of elements referenced by <code>aria-actions</code> to determine the names of actions that are exposed in a platform accessibility API.</li>
               <li>User Agents MUST NOT expose <code>aria-actions</code> if the Author MUST's are not followed.</li>
             </ul>
+            <p>To prevent an unexpected focus change away from the referencing element after users invoke actions, browsers MUST perform the following steps to manage focus once an action is invoked via <code>aria-actions</code>:</p>
```
mcking65, line 12200: In normative statements, are the terms "browser" and "user agent" interchangeable? In the preceding normative statements in this section, the term "user agent" is used. Should we consider being consistent within a given section of the spec even if we aren't consistent across the entire spec?


```suggestion
            <p>To prevent an unexpected focus change away from the referencing element after users invoke actions, user agents MUST perform the following steps to manage focus once an action is invoked via <code>aria-actions</code>:</p>
```

index.html
mcking65: The comma before the word "again" and the word "again" open this statement to misinterpretation.

Note that it is not necessary to include the words "browser MUST" because they precede the list.

I think steps 4 and 5 need to be combined, otherwise they conflict. Step 4 says the browser MUST move focus to the referencing element, but the browser can't do so if the referencing element isn't focusable. So, step 5 is really an exception that applies to step 4; it is not a separate step.

```suggestion
              <li>If focus has remained on the referenced action element and the referencing element is still focusable, move DOM focus back to the referencing element without firing a <a>desktop focus event</a>. Otherwise, if the referencing element can no longer be focused, optionally leave focus on the referenced action element and fire a <a>desktop focus event</a>.</li>
```

index.html
mcking65: Delete this line if you accept my suggestion for the previous line.

core-aam/index.html
```diff
@@ -6461,7 +6461,12 @@ <h4>Not Mapped</h4>
             </p>
           </section>
           <h4 id="ariaActions"><code>aria-actions</code></h4>
-          <table aria-labelledby="ariaActions">
+          <p>
+            After an action is invoked, browsers MUST NOT fire the API's <a href="#focus_state_event_table">focus event</a> when the referenced action node receives focus.
+            The browser MUST subsequently return DOM focus to the referencing node also without firing an API <a href="#focus_state_event_table">focus event</a>
+            unless focus is moved to a different node by a programmatic focus change or a user interaction.
```
mcking65, line 6467: The case that the referencing element is no longer focusable should be included.

```suggestion
            unless the referencing element is no longer focusable or focus is moved to a different node by a programmatic focus change or a user interaction.
```

mcking65 (changes requested):
Sorry for the tardy review ... super close, but I think there are a couple of places where the wording isn't saying exactly what we mean. Line 12211 in the ARIA file and the related line in core are the places where I made suggestions that I think represent necessary changes.

index.html
smhigley: Updated this with both suggestions 👍

core-aam/index.html
smhigley: I feel like Core AAM must reference specific ARIA spec sections somewhere, but I'm having trouble finding an example. Do you / does anyone else happen to know how to do that?

smhigley: I'm going to resolve this for now, since I think this PR is at least consistent with existing wording in the specs. @cookiecrook feel free to reopen if you think it should stay here!

core-aam/index.html
smhigley: Resolving since it's more specific now!

smhigley:
@mcking65 / @jcsteh updated!

core-aam/index.html
jcsteh: Looking further, the table below already references the aria-actions section in the ARIA spec:

> `<a class="property-reference" href="#aria-actions"><code>aria-actions</code></a>`

I wonder whether it's sufficient to just say something like:

```
After an action is invoked, browsers MUST NOT fire the API's <a href="#focus_state_event_table">focus event</a> when the referenced action node receives focus.
See the <a class="property-reference" href="#aria-actions"><code>aria-actions</code></a> section of the ARIA Specification regarding the steps for managing focus.
```

index.html
```diff
@@ -12197,8 +12197,21 @@ <h2>Definitions of States and Properties (all aria-* attributes)</h2>
...
+              <li>
+                 If the referenced action node is focusable and received focus as a result of an <code>aria-actions</code> invocation:
+                <ul>
+                  <li>Allow the DOM focus events to fire on the referenced action node as a result of invoking it.</li>
+                  <li>Do not fire the <a>desktop focus event</a>.</li>
+                </ul>
+              </li>
+              <li>If the referenced action node is not focusable or did not receive focus, do not perform any further <code>aria-actions</code> focus steps.</li>
+              <li>If focus is moved either programmatically or via a user interaction, handle that focus change as normal, and do not perform any further <code>aria-actions</code> focus steps.</li>
+              <li>After the DOM events for the invocation have been dispatched, if focus has remained on the referenced action element and the referencing element is focusable, move DOM focus back to the referencing element without firing a <a>desktop focus event</a>. If the referencing element cannot be focused, user agents MAY leave focus on the referenced action element and fire a <a>desktop focus event</a>.</li>
```
jcsteh, line 12211: I have a slight preference for being more precise here and I haven't seen any objections, but I don't consider this blocking if there's pushback. This is just adding the word "immediately".

```suggestion
              <li>Immediately after the DOM events for the invocation have been dispatched, if focus has remained on the referenced action element and the referencing element is focusable, move DOM focus back to the referencing element without firing a <a>desktop focus event</a>. If the referencing element cannot be focused, user agents MAY leave focus on the referenced action element and fire a <a>desktop focus event</a>.</li>
```

jcsteh (approved):
Thanks for all your work on this. Two thoughts, but neither of them are blocking if there's pushback or they can't be easily resolved.

core-aam/index.html
smhigley: Nice, I'll use that!

smhigley:
@jcsteh updated both with your suggested wording, I like that better :D

mcking65 (approved):
This is looking really good. Also, thanks for having core-aam point at aria for the trigger response steps.

PR #2849, opened 2026-07-17
editorial: add note in Sec. 6.4 Translatable Attributes

### LINKED ISSUE #2827: Translatable Attributes
# Describe the change

https://www.w3.org/TR/wai-aria-1.3/#translatable-attributes

This section says the language and directionality of each translatable attribute value is the same as the language and directionality of the element. That is useful as a default, but it is not enough when the attribute string has a different language/direction from the element, or contains mixed-language content.

Attribute values cannot contain markup, so authors cannot mark inline language or direction changes inside `aria-label`, `aria-description`, etc.

Adding a note about the limitations and preferred authoring patterns would be useful. For example:

> The element's language and directionality provide the default language and direction for a translatable attribute value. This default is not sufficient when the attribute value is in a different language or requires a different base direction from the element.

> Where language or direction metadata for the text alternative needs to differ from the element, authors SHOULD use a referencing attribute, such as aria-labelledby, aria-describedby, or aria-details, pointing to host-language content on which language and direction can be expressed directly.

> Authors SHOULD avoid using plain string ARIA attributes for text alternatives that contain inline language/direction changes or other text-level markup.

Link to documentation: https://www.w3.org/TR/wai-aria-1.3/#translatable-attributes

Yes.

pkra:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2849--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2849--wai-aria.netlify.app%2Findex.html)

Adds explanatory note why lang&dir currently cannot diverge.

Resolves #2827

### DIFF
--- a/index.html
+++ b/index.html
@@ -12055,6 +12055,9 @@ button.ariaPressed; // null</pre
           <li><pref>aria-rowindextext</pref></li>
           <li><pref>aria-valuetext</pref></li>
         </ul>
+        <p class="note">
+          Due to limitations across the web accessibility stack there is currently no mechanism to specify language and directionality for a translatable attribute separately from the element's language and directionality. This also indirectly applies to <pref>aria-labelledby</pref> and <pref>aria-describedby</pref> because the <a href="" class="accname">Accessible Name and Description Computation</a></cite> [[ACCNAME-1.2]] results in a flat string. Diverging in language or direction can also cause other accessibility problems, see [[WCAG21]]. 
+        </p>
       </section>
       <section id="global_states">
         <h2><dfn>Global</dfn> States and Properties</h2>


index.html
```diff
@@ -12055,6 +12055,9 @@ <h2>Translatable Attributes</h2>
           <li><pref>aria-rowindextext</pref></li>
           <li><pref>aria-valuetext</pref></li>
         </ul>
+        <p class="note">
+          Due to limitations across the web accessibility stack there is currently no mechanism to specify language and directionality for a translatable attribute separately from the element's language and directionality. This also indirectly applies to <pref>aria-labelledby</pref> and <pref>aria-describedby</pref> because the <a href="" class="accname">Accessible Name and Description Computation</a></cite> [[ACCNAME-1.2]] results in a flat string. Diverging in language or direction can also cause other accessibility problems, see [[WCAG21]].
```
xfq, line 12059: About "diverging in language or direction can also cause other accessibility problems", what does this refer to? Are there any examples?

pkra: E.g. WCAG 2.5.3 seems likely to fail. Adjusting one of the examples in [F96](https://www.w3.org/WAI/WCAG22/Techniques/failures/F96): `<button id="sitesearch" aria-label="Find in this site">Finde in dieser Seite</button>`.

xfq: Thanks. I think it might be useful to link to (or add) a concrete example in the note.

cookiecrook: From looking back at the issue, we could consider discussing this in the context of an example like:

```html
<button lang="en" aria-label="This string can't represent multiple languages or directions.">
  There's a certain <span lang="fr">je ne sais quoi</span> about him.
<button>
```
Of note: French and English TTS engines won't have a problem pronouncing either language in that label, so a western+nonwestern language example might be better to use, if you can think of a non-contrived one.

Some other solutions that may be reasonable. 
1. A new note stating that when the computed label of a leaf node changes language or direction, authors should not use aria-label to vend this. Instead uses `nameFrom: contents` so that you can appropriately apply the `lang` and `dir` attributes to substrings where useful.
2. A new note that could mention how NLP parsers work, and that, if multiple languages are used, the different substrings might get shuffled off to different TTS engines and/or difference voices. 


I'm including that sentence again here without the lang markup in case people want to test in various contexts:

> There's a certain je ne sais quoi about him.

```html
<p lang="en">My home town is <span lang="jp">東京</span> but now I live in <span lang="zh">北京</span>.</p> 
```

> My home town is 東京 but now I live in 北京. 

In the second sentence, the same kanji (京 for capital) is used in both Tokyo (東京) and Beijing (北京). 

In my current testing of the second sentence using VoiceOver on macOS, all three languages are pronounced correctly (w/ or w/o the `lang` markup), using three different speech engines, which is why I think the NLP note could work. It should also note however, that sometimes NLP has an ambious result, so the web author should avoid relying on langauge changes in translatable attributes if pronunciation is critical.

cookiecrook: Then again, on-device models might trim that ambiguity rate down to effectively nothing in the near future, so maybe the caveat isn't necessary.

cookiecrook: Does HTML or another spec have a similar note? The HTML `title` attribute, for example, would have this same issue. Instead of writing a new note here, point an ARIA reference to the note in HTML. If it doesn't exist in HTML, this issue could be duped to one on that spec.

xfq: re language detection, see https://www.w3.org/TR/string-meta/#language-detection

re HTML, I couldn't find any specific notes in HTML—only some old bug reports (like [23305](https://www.w3.org/Bugs/Public/show_bug.cgi?id=23305) and [8931](https://www.w3.org/Bugs/Public/show_bug.cgi?id=8931)). I recall there were more in-depth discussions, though I can't seem to locate them right now.

However, this guideline has been part of the i18n specdev for years; perhaps you could refer to it: https://www.w3.org/TR/international-specs/#markup_elements_attributes

daniel-montalvo (changes requested):
Thanks much @pkra for getting this started.

Because the github web interface doesn't make it easy to provide a suggestion directly that spans multiple lines if you're using a screen reader, I'm suggesting my edits here.

<blockquote>
        <p>
          The web accessibility stack does not currently provide a mechanism to specify language and directionality for a translatable attribute other than the element's language and directionality. This also applies to <pref>aria-labelledby</pref> and <pref>aria-describedby</pref> because the <a href="" class="accname">Accessible Name and Description Computation</a> [[ACCNAME-1.2]] results in a flat string.
        </p>
        <p>
          Translatable attribute values that don't match the language and directionality of their element can also cause other accessibility problems. It is best practice for authors, as specified in <a class="informativ" href="https://www.w3.org/TR/international-specs#markup_elements_attributes">Internationalization best practices for web developers</a>, to ensure that the language and directionality of the value of a translatable attribute matches the language and directionality of the element.</p>
</blockquote>

- Made this be part of the prose directly.
- Rephrased the first part where we portray the current situation as a web accessibility stack limitation.
- Specified that what causes accessibility problems is having a translatable attribute value that doesn't match the language and directionality of its element. This hopefully addresses @xfq  's concerns. I think we should make this part self-explanatory so that it does not require a dedicated example.
- Removed the reference to WCAG as it's too broad for this particular context.
- Added the reference to internationalization best practices for web developers as suggested by @xfq.

pkra:
@daniel-montalvo in the interest of time, I'll close this and assign the issue to you.

PR #2852, opened 2026-07-23
editorial: remove elements no longer in SVG 2 Editor's Draft

dd8:
This removes mappings for elements that are no longer in the SVG 2 Editors Draft due to lack of implementations - the elements in question were never implemented by any browser. Also fixes one typo (interpretted -> interpreted)


daniel-montalvo (approved):
Thanks @dd8.

PR #2860, opened 2026-08-03
CORE-AAM: Require ariaNotify announcements to be translated when the page is auto-translated

### LINKED ISSUE #265: Core-AAM: new UA req that ariaNotify strings should be translated when browser view is auto-translated
> I took an action to make a new issue for Core-AAM based on a question from @pkra... The implementation should translate `ariaNotify` strings if the browser/engine is running in a translated rendering via built-in "Translate this page" features. For example, using a variant of @smhigley's example:
> 
> ```html
> <html lang="en">
>     <body>
>         …[chinese content]…
>     </body>
> </html>
> ```
> 
> ```js
> el.ariaNotify('希望你能听懂中文'); // this announcement becomes a translated string when browser view is translated from the Chinese
> ```
> 
> I think this can be a single RFC-2119 req for UAs, and leave the specifics as implementation details. Note: Sarah is doing some additional tests to determine if this is an existing problem with Live Regions. 

 _Originally posted by @cookiecrook in [#2828](https://github.com/w3c/aria/issues/2828#issuecomment-4994847888)_

cookiecrook:
🚀 **Netlify Preview**:
- [core-aam preview](https://deploy-preview-2860--wai-aria.netlify.app/core-aam/index.html) &mdash; [core-aam diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fcore-aam%2F&doc2=https%3A%2F%2Fdeploy-preview-2860--wai-aria.netlify.app%2Fcore-aam%2Findex.html)

## Summary

Adds a new **Translation** step to the `ariaNotify` algorithm mapping table in Core-AAM.

When the user agent presents a document in a machine-translated rendering (e.g. via a built-in "Translate this page" feature), the announcement passed to `ariaNotify` is currently conveyed to assistive technologies in the original content language rather than the language the user sees. This change adds an RFC-2119 requirement that the UA MUST apply the same translation to the announcement before performing the platform-specific notification steps, and a note that the platform notification should not be triggered until translation completes (so the original and translated strings aren't spoken in sequence).

## Details

- New `Translation` row in the `ariaNotify` algorithm mapping table, placed after the existing `Language` row and before the platform-specific rows.
- Note referencing that the same translation step is expected to apply to live region announcements before their platform notifications, per the discussion in the issue.
- Changelog entry added.

Verified in [smhigley's test case](https://codepen.io/editor/smhigley/pen/019f6bf9-12b0-7c95-a290-1642d204c808) that `ariaNotify` strings are not currently auto-translated by browsers.

Closes w3c/core-aam#265

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core-aam/index.html
+++ b/core-aam/index.html
@@ -12558,6 +12558,21 @@
             </td>
           </tr>
           <tr>
+            <th>Translation</th>
+            <td>
+              <p>
+                When the user agent is presenting the document in a machine-translated rendering (for example, via a built-in "Translate this page" feature), the user agent MUST apply the same
+                translation to <var>announcement</var> before performing any of the platform-specific steps below, so that the announcement is conveyed to assistive technologies in the language
+                presented to the user rather than in the language of the original content.
+              </p>
+              <p class="note">
+                The user agent should not trigger the platform accessibility notification until the translation step has completed; otherwise assistive technologies may present the original and
+                translated strings in sequence. The same translation step is expected to apply to <a href="https://www.w3.org/TR/wai-aria-1.2/#aria-live">live region</a> announcements before their
+                corresponding platform notifications are triggered.
+              </p>
+            </td>
+          </tr>
+          <tr>
             <th>MSAA + IAccessible2</th>
             <td>
               <p>No implementation specified (<a href="#arianotify-fallback-note">see fallback note</a>).</p>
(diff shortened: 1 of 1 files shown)


cookiecrook:
@twilco has [WebKit PR 71083](https://github.com/WebKit/WebKit/pull/71083) up already. 🔥

core-aam/index.html
```diff
@@ -12662,6 +12677,7 @@ <h2>Change Log</h2>
         <h2>Substantive changes since the last <a href="https://www.w3.org/TR/2022/CR-core-aam-1.2-20221122/">Candidate Recommendation Snapshot</a></h2>
         <ul>
           <!-- EdNote: After each WD publish, move contents of this list into the next one below. -->
+          <li>30-Jul-2026: Require <code>ariaNotify</code> announcements to be translated when the browser view is machine-translated</li>
```
spectranaut: These change logs are automatically generated when publishing, I believe... but now I'm not sure if that is still the case since we changed to the monorepo. I'll check with Daniel.

spectranaut (commented):
This seems fine to me except for an editorial question I have about when we update the changelog! I'll approve once I understand that.

core-aam/index.html
spectranaut:
```suggestion
```

Discussed with  James, we will update this section before publishing, not in individual PRs.

spectranaut (changes requested):

xfq (approved):

PR #2862, opened 2026-08-04
Editorial: I18n considerations for translatable attributes

### LINKED ISSUE #2827: Translatable Attributes

daniel-montalvo:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2862--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2862--wai-aria.netlify.app%2Findex.html)

Closes #2827


xfq (dismissed):
A full stop needs to be added after the last sentence. Otherwise looks good to me.

index.html
```diff
@@ -12060,6 +12060,18 @@ <h2>Translatable Attributes</h2>
           <li><pref>aria-rowindextext</pref></li>
           <li><pref>aria-valuetext</pref></li>
         </ul>
+        <p>
+          The web accessibility stack does not currently provide a mechanism to specify language and directionality for a translatable attribute other than the language and directionality of its containing element. This includes <pref>aria-labelledby</pref> and <pref>aria-describedby</pref> because the <a href="" class="accname">Accessible Name and Description Computation</a> [[ACCNAME-1.2]] results in a flat string.
```
cookiecrook: Editorial:
```suggestion
          The web platform does not provide a mechanism to specify language and directionality for a translatable attribute separate from the language and directionality of its containing element. This includes <pref>aria-labelledby</pref> and <pref>aria-describedby</pref> because the <a href="" class="accname">Accessible Name and Description Computation</a> [[ACCNAME-1.2]] results in a flat string.
```

index.html
```diff
@@ -12060,6 +12060,18 @@ <h2>Translatable Attributes</h2>
...
+        </p>
+        <p>
+          Translatable attribute values that don't match the language and directionality of their containing element can also cause other accessibility problems. It is best practice for authors to ensure that the language and directionality of the value of a translatable attribute matches the language and directionality of its containing element.
```
cookiecrook: If you call out "other problems," it would be wise to list a known example. 

Alternately, cut the first sentence for brevity, because the second stands on its own.

index.html
cookiecrook: Substantive: Should this note be in AccName instead of ARIA?

index.html
```diff
@@ -12060,6 +12060,18 @@ <h2>Translatable Attributes</h2>
...
+        </p>
+        <p>
+          Where language or direction metadata for the translatable attribute value needs to differ from that of its containing element, it is recommended that authors use a referencing attribute, such as aria-labelledby, aria-describedby, or aria-details, pointing to host-language content on which language and direction can be expressed directly. This would allow translation of these values when the user agent is presenting the document in a machine-translated rendering (for example, via a built-in "Translate this page" feature). Authors are encouraged not to use plain string ARIA attributes for text alternatives that contain inline language/direction changes or other text-level markup.
```
```suggestion
          Where language or direction metadata for the translatable attribute value needs to differ from that of its containing element, authors can use a referencing attribute, such as aria-labelledby, aria-describedby, or aria-details, pointing to an element, on which language and direction attributes can be expressed directly. This allows translation of these values when the user agent is presenting the document in a machine-translated rendering, via a built-in "Translate this page" feature. 
          
          Authors are encouraged to avoid using plain text ARIA attributes values for text alternatives that contain inline language/direction changes or other text-level markup.
```

index.html
```diff
@@ -12060,6 +12060,18 @@ <h2>Translatable Attributes</h2>
...
+        </p>
+        <p>
+          For more details, see  <a class="informativ" href="https://www.w3.org/TR/international-specs#markup_elements_attributes">Internationalization best practices for web developers</a>.
```
jnurthen: typo
```suggestion
          For more details, see <a class="informative" href="https://www.w3.org/TR/international-specs#markup_elements_attributes">Internationalization best practices for web developers</a>.
```

index.html
jnurthen: aria-labelledby, aria-describedby, or aria-details should be wrapped in &lt;code&gt; elements here

index.html
jnurthen: Nit - I think `<a href="" class="accname">Accessible Name and Description Computation</a>` should be wrapped in a `<cite></cite>`

jnurthen: > Substantive: Should this note be in AccName instead of ARIA?

IMO no - although the note could be duplicated in accname potentially

jnurthen (dismissed):

index.html
daniel-montalvo: I also think it belongs here, and should potentially be duplicated in or referenced from accname at some point.

index.html
daniel-montalvo: I'd just cut off the first sentence as you suggest. Listing individual problems here may end up making this a maintenance burden if the guidance changes in the future.

index.html
daniel-montalvo: I prefer to keep "plain string" as it is what the i18n guidance is using.
Also, fixed "attribute<del>s</del> values" typo.
Otherwise, happy to apply:

```suggestion
          Where language or direction metadata for the translatable attribute value needs to differ from that of its containing element, authors can use a referencing attribute, such as <code>aria-labelledby</code>, <code>aria-describedby</code>, or <code>aria-details</code>, pointing to an element, on which language and direction attributes can be expressed directly. This allows translation of these values when the user agent is presenting the document in a machine-translated rendering, via a built-in "Translate this page" feature. 
          
          Authors are encouraged to avoid using plain string ARIA attribute values for text alternatives that contain inline language/direction changes or other text-level markup.
```

daniel-montalvo:
@aphillips @xfq I've updated this based on other ARIA members review as well as on our conversation on [6 august](https://www.w3.org/2026/08/06-i18n-minutes.html).

Let me know if you are happy with this or you would propose further tweaks.

xfq:
I’m a bit confused.

The first paragraph seems to suggest that the `aria-labelledby` and `aria-describedby` attributes shouldn't be used, yet the third paragraph appears to say they can be. It feels a bit contradictory.

index.html
pkra: I think this is an important sentence to keep. I gave an example in https://github.com/w3c/aria/pull/2849/changes#r3614624182, almost straight from a WCAG failure document.

pkra: I'd be very intrigued to see an intentionally authored example where aria-label is in a different language from the element's content without being problematic.

pkra: Someone kindly pointed me to https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes:

> The lang attribute [...] specifies the primary language for the element's contents and for any of the element's attributes that contain text.

Generally speaking, having different (primary) languages in accname and visible content seems to invariably fail WCAG 2.5.3.

daniel-montalvo:
Hi @xfq 

> I’m a bit confused.
> 
> The first paragraph seems to suggest that the `aria-labelledby` and `aria-describedby` attributes shouldn't be used,

It tries to explicitly communicate that `aria-labelledby` and `aria-describedby`, do not allow by themselves to specify language and directionality.

> yet the third paragraph appears to say they can be.

It tries to say they can be used Where language or direction metadata for the translatable attribute value needs to differ from that of its containing element as a "referencing attribute" and in combination with appropriate language markup in the containing element.

Would removing the explicit mentions to these attributes in the first paragraph alleviate the confusion?

xfq:
> Would removing the explicit mentions to these attributes in the first paragraph alleviate the confusion?

Yeah, I think this would be better. Thanks.

index.html
```diff
@@ -12070,6 +12070,20 @@ <h2>Translatable Attributes</h2>
           <li><pref>aria-rowindextext</pref></li>
           <li><pref>aria-valuetext</pref></li>
         </ul>
+        <p>
+          The web platform does not provide a mechanism to specify language and directionality for a translatable attribute separate from the language and directionality of its containing element.
```
pkra: I think a reference to the DOM spec might be clearer than "web platform", see https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes

daniel-montalvo: Done

daniel-montalvo:
@xfq Just checking in with you to confirm the latest updates addressed your point above.

xfq:
Thanks, Daniel. Looks good to me.

index.html
daniel-montalvo: Addressed as per comments here and in https://www.w3.org/2026/09/17-aria-minutes.html#48dc

daniel-montalvo:
@cookiecrook @pkra  -- prose updated as per https://www.w3.org/2026/09/17-aria-minutes.html#48dc, comments marked as resolved. Should be ready for your re-review.

PR #2863, opened 2026-08-04
Update a no href to map to link role

### LINKED ISSUE #608: Map link with direct click listener to link
From this related issue:

> For HTML and SVG, Chromium seems to use Link role for `<a>` element without href, as long as it has mouse event listeners:
> […trimmed reply…]
> (I haven't checked other browsers). 

 _Originally posted by @fred-wang in [#39](https://github.com/w3c/mathml-aam/issues/39#issuecomment-4894047549)_

Later @twilco confirmed this is also aligned with Gecko and WebKit, so the HTML-AAM should update, too.

Note that the implementations are (or should be) limited to direct listeners, not delegated listeners on ancestor elements like the body.

lolaodelola:
Closes [html-aam/608](https://github.com/w3c/html-aam/issues/608#event-28735005457)

Describe Change Here!

* [x] For every spec that this PR edits, please add the appropriate `spec:<spec_name>` label. If you don't have privileges to do this, editors will do it for you.

   * WebKit: https://github.com/WebKit/WebKit/pull/67035
   * Gecko: 
   * Blink:
* [ ] ACT review?
* [ ] Does this need AT implementations?: N/A
* [ ] Related APG Issue/PR:
* [ ] MDN Issue/PR:


html-aam/index.html
```diff
@@ -17060,6 +17110,7 @@ <h3>Change Log</h3>
         <section>
           <h4>Substantive changes since moving to the <a href="https://www.w3.org/WAI/ARIA/">Accessible Rich Internet Applications Working Group</a> (03-Nov-2019)</h4>
           <ul>
+            <li>12-Aug-2026: Add `a with click listener` to have `link` role in keeping with browser implementations. See <a href="https://github.com/w3c/html-aam/issues/608">GitHub Issue 608</a> </li>
```
HaTheo: NIT: missing period/. before `</li>` for consistency.

.gitignore
```diff
@@ -220,3 +220,6 @@ pip-log.txt
 
 # Local Netlify folder
 .netlify
+
+# Jet Brains
```
HaTheo, line 224: NIT: Calling it out as this seems auxiliary like something JetBrains added. Might be good to separate if we feel like this is a helpful addition to '.gitignore', but that might be personal preference.

html-aam/index.html
```diff
@@ -405,6 +405,56 @@ <h4 id="el-a-no-href">`a` <span class="el-context">(no `href` attribute)</span><
...
+            </td>
+          </tr>
+          <tr>
+            <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
+            <td>
+              <div class="general">Use WAI-ARIA mapping</div>
+            </td>
+          </tr>
+          <tr>
+            <th>Comments</th>
```
HaTheo, line 453: Might be good to specify that this applies only to direct onClick and click event listeners in the comments for clarity. Especially as delegated event listeners become more common and James' callout.

html-aam/index.html
HaTheo: NIT: Should explicitly say without click listener for clarity.

HaTheo:
NIT feedback otherwise LGTM. I also looked into Area tags and found that they were treated similarly in Firefox but not in chrome/safari it didn't look like.

lolaodelola:
Thanks @HaTheo! I think we may need more discussion on the area tags.

spectranaut:
@HaTheo can you leave a review? you only left a comment :)

HaTheo (approved):

.gitignore
lolaodelola: I manually added this following the existing pattern. You think this should be taken out and put in a different PR?

html-aam/index.html
lolaodelola: You mean like no href and no click listener?

HaTheo: Yes, since it's broken out into three options: a (represented as a hyperlink); a (without href); and a (with a click listener) there's a bit of overlap between an a without an href and an a with a click listener. I just thought specifying it would clear up any ambiguity. I was thinking this could say "a (without href or click listener)"

lolaodelola: Gotcha!

lolaodelola:
```suggestion
                <span class="el-context">(no <a data-cite="html/links.html#attr-hyperlink-href">`href`</a> attribute, no click listener)</span>
```

lolaodelola:
@HaTheo I've gone ahead and implemented your changes bar the .gitignore one (see comment).

@rahimabdi @shirsha Do you folks mind having a look at this?

Issue #2861, opened 2026-08-03
Security Horizontal Review


simoneonofri:
Dear Group, this is a finding from the [Security Review](https://github.com/w3c/security-request/issues/134).

**Reviewed version:** [4 June 2026 WAI-ARIA 1.3 Working Draft](https://www.w3.org/TR/2026/WD-wai-aria-1.3-20260604/).

**Security and privacy self-review:** [w3c/aria issue #2814](https://github.com/w3c/aria/issues/2814)

## The `aria-notify` policy relaxes the Permission Policy default

**Concern:** WAI-ARIA 1.3 introduces `Document.ariaNotify()`, allowing documents to send author-controlled announcements through the accessibility platform. The security boundary works as follows:

  1. A document calls `ariaNotify()` to generate an announcement.
  2. Access is controlled by the `aria-notify` Permission Policy.
  3. The policy has a default allowlist of `*`.
  4. Consequently, a cross-origin frame does not require explicit delegation from the embedding document.
  5. Repeated calls may produce notification flooding or other forms of user disruption, while the intended filtering behavior remains unresolved.

Even if the [HTML discussion](https://github.com/whatwg/html/issues/11004) presents the `*` default as supporting accessibility by default. However, it also relaxes the usual secure-by-default boundary for cross-origin content, and the current [Security Considerations](https://www.w3.org/TR/2026/WD-wai-aria-1.3-20260604/#security-considerations) do not explain this trade-off.

**Impact:** Untrusted cross-origin content may use the announcement channel without prior delegation. Depending on browser and assistive technology behavior, repeated or misleading announcements could disrupt users or interfere with their understanding of the page.

**Recommendation:** Understand if the permission policy is necessary for this feature and if the default allowlist of `*` is necessary, as it depends on the actual usage of `ariaNotify()`. If necessary, the specification should explain the trade-off of relaxing secure-by-default and which accessibility requirements justify the relaxation; otherwise, updating the default should be fine.

**Affected sections:**
- https://www.w3.org/TR/2026/WD-wai-aria-1.3-20260604/#ARIANotifyMixin
- https://www.w3.org/TR/2026/WD-wai-aria-1.3-20260604/#arianotify-permissions-policy
- https://www.w3.org/TR/2026/WD-wai-aria-1.3-20260604/#security-considerations

Thanks,

Simone

P.S. If you have any suggestions for testing the feature (e.g., assistive technology to put in the lab), I'd be happy to hear them.

[...]

Issue #2837, opened 2026-07-03
Add intro and context to ARIA Notify


daniel-montalvo:
As noted by @matatk in https://github.com/w3c/a11y-request/issues/165, we should have some sort of intro and context for ARIA Notify. I am happy to take this one on by reusing the [explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/Accessibility/AriaNotify/explainer.md)

PR #2869, opened 2026-08-19
Editorial: ARIA Notify intro and i18n example

### LINKED ISSUE #2828: Add explicit language and direction metadata to `AriaNotificationOptions`

https://www.w3.org/TR/wai-aria-1.3/#ARIANotifyMixin

`ariaNotify(announcement, options)` can be used to notify people who are blind or have low vision, identifying dynamic changes in the content, but it creates a natural-language string outside normal DOM text content.

The current `AriaNotificationOptions` only carries `priority` and does not carry language or string direction. The [i18n requirement](https://w3c.github.io/string-meta/) is that natural-language string fields must have determinable language and direction, preferably from metadata.

Please add explicit language and direction metadata to `AriaNotificationOptions`. For example:

```
enum AriaTextDirection { "ltr", "rtl", "auto" };

dictionary AriaNotificationOptions {
  AriaNotifyPriority priority = "normal";
  DOMString lang;
  AriaTextDirection dir = "auto";
};
```

Authors should provide `lang` when the announcement language differs from the language of the node or document. They should also provide `dir` when the string direction is known. And user agents should expose the language and direction metadata to platform accessibility APIs where the platform supports it.

daniel-montalvo:
🚀 **Netlify Preview**:
- [ARIA preview](https://deploy-preview-2869--wai-aria.netlify.app/index.html) &mdash; [ARIA diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Faria%2F&doc2=https%3A%2F%2Fdeploy-preview-2869--wai-aria.netlify.app%2Findex.html)

Addresses:

- #2828 
- #2861 
- #2837

### DIFF
--- a/index.html
+++ b/index.html
@@ -16871,7 +16871,7 @@ button.ariaPressed; // null</pre
           Element includes ARIANotifyMixin;
           Document includes ARIANotifyMixin;
         </pre>
-        <p>Since ariaNotify does not allow to specify the language of the notification, the notification inherits that of its containing element</p>
+        <p>Since ariaNotify does not have a parameter to specify the language of the notification, the notification inherits that of its containing element</p>
         <p>For cases where the language of the notification differs from that of its containing element, authors can use  a separate container element and add language markup to convey the language.</p>
         <aside class="example">
           <p>
--- a/index.html
+++ b/index.html
@@ -16872,7 +16872,7 @@ button.ariaPressed; // null</pre
           Document includes ARIANotifyMixin;
         </pre>
         <p>Since ariaNotify does not allow to specify the language of the notification, the notification inherits that of its containing element</p>
-        <p>For cases where the language of the notification differs from that of its containing element, authors can specify  a courier element and add language markup to convey the language.</p>
+        <p>For cases where the language of the notification differs from that of its containing element, authors can use  a separate container element and add language markup to convey the language.</p>
         <aside class="example">
           <p>
             The original page language is English, but there is a notification in Spanish.
--- a/index.html
+++ b/index.html
@@ -16871,7 +16871,7 @@ button.ariaPressed; // null</pre
           Element includes ARIANotifyMixin;
           Document includes ARIANotifyMixin;
         </pre>
-        <p>Since ariaNotify doesn't allow to specify the language of the notification, the notification inherits that of its containing element</p>
+        <p>Since ariaNotify does not allow to specify the language of the notification, the notification inherits that of its containing element</p>
         <p>For cases where the language of the notification differs from that of its containing element, authors can specify  a courier element and add language markup to convey the language.</p>
         <aside class="example">
           <p>
--- a/index.html
+++ b/index.html
@@ -17281,7 +17281,7 @@ el.getAttribute("aria-label"); // Returns "Publish"</pre
     <section class="informative" id="security-considerations">
       <h2>Security Considerations</h2>
       <p>
-        The <a>"aria-notify"</a> permissions policy has a default allowlist of "*". This relaxes the Permision Policy default because requiring embedders to opt in for each frame would leave announcements coming
+        The <a>"aria-notify"</a> permissions policy has a default allowlist of "*". This relaxes the Permission Policy default because requiring embedders to opt in for each frame would leave announcements coming
         from third-party content inaccessible by default. Authors who want tighter control can still restrict the "aria-notify" policy explicitly, for example with
         <code>&lt;iframe allow="aria-notify 'none'"&gt;</code> or <code>Permissions-Policy: aria-notify=(self)</code>.
       </p>
--- a/index.html
+++ b/index.html
@@ -17282,7 +17282,7 @@ el.getAttribute("aria-label"); // Returns "Publish"</pre
       <h2>Security Considerations</h2>
       <p>
         The <a>"aria-notify"</a> permissions policy has a default allowlist of "*". This relaxes the Permission Policy default because requiring embedders to opt in for each frame would leave announcements coming
-        from third-party content inaccessible by default. Authors who want tighter control can still restrict the "aria-notify" policy explicitly, for example with
+        from third-party content inaccessible by default. Cross-origin embedded content can invoke this user-facing channel without explicit delegation, which can cause uncontrolled disruption for uses of assistive technologies. Authors who want tighter control can still restrict the "aria-notify" policy explicitly, for example with
         <code>&lt;iframe allow="aria-notify 'none'"&gt;</code> or <code>Permissions-Policy: aria-notify=(self)</code>.
       </p>
     </section>
--- a/index.html
+++ b/index.html
@@ -16856,6 +16856,9 @@ button.ariaPressed; // null</pre
       </section>
       <section id="ARIANotifyMixin" class="normative" data-dfn-for="ARIANotifyMixin" data-link-for="ARIANotifyMixin">
         <h2>Interface Mixin <dfn>ARIANotifyMixin</dfn></h2>
+        <p>
+          The ariaNotify API allows authors to provide notifications for accessibility. The <code>ariaNotify</code> method accepts a string as a first parameter and a configuration object as a second (optional) parameter.
+        </p>
         <pre class="idl" data-cite="webidl">
           enum AriaNotifyPriority { "normal", "high" };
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -16854,7 +16857,26 @@ <h2>Interface Mixin <dfn>ARIANotifyMixin</dfn></h2>
           Element includes ARIANotifyMixin;
           Document includes ARIANotifyMixin;
         </pre>
-
+        <p>Since ariaNotify doesn't allow to specify the language of the notification, the notification inherits that of its containing element</p>
```
giacomo-petri:
```suggestion
        <p>Since ariaNotify does not allow to specify the language of the notification, the notification inherits that of its containing element</p>
```

to align with the rest of the spec

index.html
```diff
@@ -17244,7 +17266,11 @@ <h2>Example IDL Attribute Usage</h2>
     </section>
     <section class="informative">
       <h2>Security Considerations</h2>
-      <p>This specification introduces no new security considerations.</p>
+      <p>
+        The <a>"aria-notify"</a> permissions policy has a default allowlist of "*". This relaxes the Permision Policy default because requiring embedders to opt in for each frame would leave announcements coming
```
giacomo-petri:
```suggestion
        The <a>"aria-notify"</a> permissions policy has a default allowlist of "*". This relaxes the Permission Policy default because requiring embedders to opt in for each frame would leave announcements coming
```

index.html
```diff
@@ -16854,7 +16857,26 @@ <h2>Interface Mixin <dfn>ARIANotifyMixin</dfn></h2>
           Element includes ARIANotifyMixin;
           Document includes ARIANotifyMixin;
         </pre>
-
+        <p>Since ariaNotify doesn't allow to specify the language of the notification, the notification inherits that of its containing element</p>
+        <p>For cases where the language of the notification differs from that of its containing element, authors can specify  a courier element and add language markup to convey the language.</p>
```
giacomo-petri:
```suggestion
        <p>For cases where the language of the notification differs from that of its containing element, authors can use  a separate container element and add language markup to convey the language.</p>
```

"courier" is intended to be "container", isn't it?

giacomo-petri (dismissed):

index.html
```diff
@@ -16854,7 +16857,26 @@ <h2>Interface Mixin <dfn>ARIANotifyMixin</dfn></h2>
           Element includes ARIANotifyMixin;
           Document includes ARIANotifyMixin;
         </pre>
-
+        <p>Since ariaNotify does not allow to specify the language of the notification, the notification inherits that of its containing element</p>
```
janewman: >Since ariaNotify does not allow to specify the language of the notification

This sentence seems a little strange from my read, as ariaNotify does allow authors to specify the language of the notification by relying on the document or element's language, which is documented clearly below.

Here's an optional suggestion to clarify:
```suggestion
        <p>Since ariaNotify does not have a parameter to specify the language of the notification, the notification inherits that of its containing element</p>
```

janewman (approved):
lgtm with a small nit.

index.html
```diff
@@ -17250,7 +17272,11 @@ <h2>Example IDL Attribute Usage</h2>
     </section>
     <section class="informative">
       <h2>Security Considerations</h2>
-      <p>This specification introduces no new security considerations.</p>
+      <p>
+        The <a>"aria-notify"</a> permissions policy has a default allowlist of "*". This relaxes the Permission Policy default because requiring embedders to opt in for each frame would leave announcements coming
```
cookiecrook, line 17284: I'm don't recall *why* this decision was made... Is it worth an editorial note pointing to prior discussions, or explaining why? I've never seen Live Regions abused in an ad frame, so maybe that's the reason? If theres a direct link to tjhis detail in the explainer or discussion, should it be listed or linked here?

index.html
```diff
@@ -17250,7 +17272,11 @@ <h2>Example IDL Attribute Usage</h2>
...
+        from third-party content inaccessible by default. Cross-origin embedded content can invoke this user-facing channel without explicit delegation, which can cause uncontrolled disruption for uses of assistive technologies. Authors who want tighter control can still restrict the "aria-notify" policy explicitly, for example with
```
cookiecrook, line 17285: And there is is nothing stopping implementations from restricting this for untrusted sources, right? If not, we could mention that as a "UAs SHOULD/MAY" in the spec.

index.html
```diff
@@ -17250,7 +17272,11 @@ <h2>Example IDL Attribute Usage</h2>
...
+        <code>&lt;iframe allow="aria-notify 'none'"&gt;</code> or <code>Permissions-Policy: aria-notify=(self)</code>.
```
cookiecrook, line 17286: This might be seen as prioritizing author ease over AT users re: Priorities of Constituencies?

cookiecrook: Though that's not really a Security risk... It's more just a "user annoyance" pester risk.

index.html
```diff
@@ -16860,7 +16863,26 @@ <h2>Interface Mixin <dfn>ARIANotifyMixin</dfn></h2>
...
           Document includes ARIANotifyMixin;
         </pre>
-
+        <p>Since ariaNotify does not have a parameter to specify the language of the notification, the notification inherits that of its containing element</p>
+        <p>For cases where the language of the notification differs from that of its containing element, authors can use  a separate container element and add language markup to convey the language.</p>
+        <aside class="example">
+          <p>
+            The original page language is English, but there is a notification in Spanish.
+          </p>
+          <!-- ReSpec needs these examples to be unindented. -->
```
cookiecrook, line 16880: But they are indented.

index.html
```diff
@@ -16860,7 +16863,26 @@ <h2>Interface Mixin <dfn>ARIANotifyMixin</dfn></h2>
...
+          <pre class="example highlight html">
```
cookiecrook, line 16881: This example could be more terse, and might be easier to read if there is a generated element. 

Something like:
```js
// To create a one-off notification that doesn't match the 
// language of the page or any other existing element, authors 
// could append a dynamically created courier element.

var courier = document.createElement('div')
courier.setAttribute('lang', 'es');
document.body.appendElement(courier);
courier.ariaNotify('Este texto debería leerse con la voz en español'); 
```

Note: I did not test the syntax above. Please verify if you decide to use this.

cookiecrook: For clarity, this is a suggestion, but the longer existing example is fine if the WG prefers. (I have no objection to disregarding my suggestion here.)

cookiecrook (commented):
Perhaps the i18n example should have been a different PR than the Sec/Privacy changes... I would have approved this example diff either way, but am marking the PR Review as a "Comment" (not "Approved") because of the questions in the lower Security section.

index.html
cookiecrook: But if ariaNotify is mentioned in this context, should Live Regions be mentioned, too?

index.html
```diff
@@ -16868,7 +16871,26 @@ <h2>Interface Mixin <dfn>ARIANotifyMixin</dfn></h2>
           Element includes ARIANotifyMixin;
           Document includes ARIANotifyMixin;
         </pre>
-
+        <p>Since ariaNotify does not have a parameter to specify the language of the notification, the notification inherits that of its containing element</p>
+        <p>For cases where the language of the notification differs from that of its containing element, authors can use  a separate container element and add language markup to convey the language.</p>
```
smhigley, line 16875: nit: this isn't really a containing/container element, I think. It seems more appropriate to call it the source element or the calling element, since `ariaNotify` is a method and not a child or descendant.

index.html
smhigley: it would be `document.body.appendChild(el)`, but otherwise it looks good to me! In my very subjective view, I also like this approach better / it feels closer to what I'd do in practice, but the example below also 100% works.

index.html
daniel-montalvo: Which shows we have old comments that we may need to clean up at some point :-) There are eight more instances of this in the spec, and all of which related examples seems to carry over some sort of indentation. To my knowledge these examples render correctly, which indicates that Respec doesn't need these unindented anymore.

index.html
daniel-montalvo: Just made this shorter following @cookiecrook and @smhigley  's advice.

I kept "spanishNotification" as there was previously a [comment thread](https://github.com/w3c/aria/pull/2869#discussion_r3821633866) along the lines of not using "courier".

daniel-montalvo:
@giacomo-petri @smhigley @cookiecrook  comments are now addressed, except for those related to the security considerations, wihch I'll put on a separate PR.

janewman:
> @giacomo-petri @smhigley @cookiecrook comments are now addressed, except for those related to the security considerations, wihch I'll put on a separate PR.

This lgtm, but you might want to update the PR description to not claim this handles the security issue.

janewman (approved):

smhigley (approved):
LGTM too!

giacomo-petri (approved):

daniel-montalvo:
HI @xfq The group is ready to merge this in. Are you happy with how it's looking like?

PR #2873, opened 2026-08-24
[graphics-aam] Fix broken links to "Computed Role" in core-aam

tidoust:
The spec still used the previous URL for the "Computed Role" links and the anchors are lost during redirection.

This update fixes the link using the more idiomatic cross-spec section link shorthand that ReSpec supports:
https://respec.org/docs/#cross-spec-section-links

(Note that creates link to the current specification under /TR, which is 1.1, but then the links will auto-update themselves once the group decides to make 1.2 the current specification).

### DIFF
--- a/graphics-aam/index.html
+++ b/graphics-aam/index.html
@@ -329,7 +329,7 @@
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


daniel-montalvo (approved):
Thanks much @tidoust

PR #2895, opened 2026-09-08
Editorial: clarify ARIA 1.2 historical references

### LINKED ISSUE #2893: update language referring to ARIA 1.2
There are a couple of references to ARIA 1.2 specifically which we might want to update for 1.3.

- "WAI-ARIA 1.2 is a member of the WAI-ARIA 1.2 suite" (probably remove both)
- combobox - might need "starting with" e.g.
   - "Major Changes to combobox role in ARIA 1.2"
   -  "has changed significantly in ARIA 1.2"
   - ": Validity changes combobox for ARIA 1.2"
 - "deprecated on this role in ARIA 1.2" - maybe "since"
 - "Deprecated in ARIA 1.2" - maybe "since"
 - "Global use deprecated in ARIA 1.2"
 - "This state is being deprecated as a global state in ARIA 1.2"
 - "This state has been deprecated as a global state in ARIA 1.2. "
 - "This property is being deprecated as a global property in ARIA 1.2."

dvd233:
Closes #2893

Clarifies ARIA 1.2 references that were carried into the ARIA 1.3 draft:

- removes the hard-coded 1.2 version from the suite membership statement;
- uses "starting with" for combobox changes and "since" for ongoing deprecations;
- keeps the source table labels and preprocessing/rendering strings synchronized so deprecated states and properties remain marked correctly.

Validation:

- Prettier 3.6.0 with the repository's 200-column setting;
- `node --check` for all three changed scripts;
- ReSpec 37.3.6 render completed successfully;
- rendered-output assertions confirmed all targeted old forms are absent, the replacements are present, and all eight `data-deprecated` markers are still generated.

ReSpec reported only the existing unused-definition warnings for "Expose" and "Target Element"; neither definition is changed by this PR.

### DIFF
--- a/common/script/aria.js
+++ b/common/script/aria.js
@@ -269,7 +269,7 @@ const buildInheritedStatesProperties = function (item) {
       const suffix = isState ? " (state)" : "";
       const tag = isState ? "sref" : "pref";
       const req = property.required ? " <strong>(required)</strong>" : "";
-      const dep = property.deprecated ? " <strong>(deprecated on this role in ARIA 1.2)</strong>" : "";
+      const dep = property.deprecated ? " <strong>(deprecated on this role since ARIA 1.2)</strong>" : "";
 
       return `<li><${tag}>${property.name}</${tag}>${suffix}${req}${dep}</li>\n`;
     })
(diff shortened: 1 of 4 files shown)


pkra:
Thanks for the PR but it turns out this is not a good first issue after all.

